PR #1, opened 2021-10-18
PYI-312: create terraform networking module

philandstuff:
This contains a terraform module which creates the low-level
networking infrastructure required to deploy the components in this
repo.  In particular, it creates:

- VPC
- subnets, routing tables, internet gateway

### DIFF
--- a/terraform/networking/variables.tf
+++ b/terraform/networking/variables.tf
@@ -1,0 +1,10 @@
+variable "environment" {
+  type = string
+}
+
+locals {
+  default_tags = {
+    Environment = var.environment
+    Source      = "github.com/alphagov/di-ipv-core-back/terraform/networking"
+  }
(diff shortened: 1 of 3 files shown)


terraform/networking/variables.tf
```diff
@@ -0,0 +1,10 @@
+variable "environment" {
+  type = string
+}
+
+locals {
+  default_tags = {
+    Environment = var.environment
+    Source      = "github.com/alphagov/di-ipv-core-back/terraform-networking"
```
merlinc: Should this be?

```suggestion
    Source      = "github.com/alphagov/di-ipv-core-back/terraform/networking"
```

merlinc (dismissed):

merlinc (approved):

PR #17, opened 2021-11-15
Update terraform default tags to work for localstack and change to unique statementID for the lamda permissions

dannyhayes-gds:
### What changed
Update the default tags to work when running terraform in localstack.

Updated the statementID to be unique between the base lambda permission and the alias permission as this was causing issues during deployments to localstack.

### Why did it change
Allows our terraform files to be ran and deployed into a localstack docker instance for local development.

### Issue tracking

- [PYI-300](https://govukverify.atlassian.net/browse/PYI-300)

 - New env var "use_localstack" which is defaulted to "false" here. This value is set to true from the di-ipv-local-startup repo.


terraform/lambda/variables.tf
```diff
@@ -2,14 +2,19 @@ variable "environment" {
...
+
 variable "lambda_zip_file" {
   default     = "../../../../di-ipv-core-back/build/distributions/di-ipv-core-back.zip"
   description = "Location of the Lambda ZIP file"
   type        = string
 }
 
 locals {
-  default_tags = {
+  default_tags = var.use_localstack ? null : {
```
idrop, line 11: just a question: does localstack not work with these tags set?

dannyhayes-gds: It doesn't seem like it does from a bit of a google. And this is also the same thing that the auth team do for their localstack stuff. Without this basically the terraform apply command will just get stuck at various places and end up failing.

idrop (approved):
works for me on localstack

PR #19, opened 2021-11-16
PYI-337: Re-use endpoint module to deploy all of the oauth lambda infrastructure.

dannyhayes-gds:
### What changed
Create new terraform files for the user-info and token-exchange lambda's and parameterise the rest of the endpoint module variables.

### Why did it change
Made the endpoint module re-usable for all future lambdas that we need to deploy.

### Issue tracking

- [PYI-337](https://govukverify.atlassian.net/browse/PYI-337)

### DIFF
--- a/terraform/lambda/token.tf
+++ b/terraform/lambda/token.tf
@@ -1,12 +1,12 @@
-module "authorize" {
+module "token" {
   source      = "../modules/endpoint"
   environment = var.environment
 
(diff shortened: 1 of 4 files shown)


merlinc (approved):
I'm not sure we need `user-identity.tf` at the moment, but I don't think it'll break anything, so let's merge.

kerrr (approved):
LGTM

PR #20, opened 2021-11-16
Move restapi http request type into variable

dannyhayes-gds:
### What changed
Move the rest api http type into a variable.

### Why did it change
So that we can configure each lamda endpoint to be either a GET or POST as needed.

### Issue tracking

- [PYI-337](https://govukverify.atlassian.net/browse/PYI-337)


merlinc (approved):
LGTM

PR #26, opened 2021-11-18
Fix AccessTokenHandler lambda to correctly parse a x-www-form-urlencoded http request body

dannyhayes-gds:
### What changed
Updated the token exchange lambda to correctly parse a x-www-form-urlencoded String body into a TokenRequestDto.

### Why did it change
The POST request body is x-www-form-urlencoded instead of JSON.

### Issue tracking

- [PYI-286](https://govukverify.atlassian.net/browse/PYI-286)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,9 @@ dependencies {
             "com.fasterxml.jackson.core:jackson-core:2.13.0",
             "com.fasterxml.jackson.core:jackson-databind:2.13.0",
             "com.fasterxml.jackson.core:jackson-annotations:2.13.0",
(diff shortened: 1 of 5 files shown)


merlinc (approved):
It looks ok for now while we're trying to unblock things.

Once we're unblocked:
 - I think you've missed a test for `ResponseBodyHelper`
 - `test.com` is a valid domain name - I'd recommended using `example.com` as they're defined for testing in an RFC - [wikipeda](https://en.wikipedia.org/wiki/Example.com)
 - I can't remember what our team consensus was on the name of "helpers" 😁

idrop (approved):
For now, its a copy of what auth did on di-authentication-api 👍

PR #27, opened 2021-11-18
BAU Github action to grade test on push and pull request

idrop:
## Proposed changes

Run a github action to run our tests in this repo. We'd like to have our tests run automatically when we raise a PR, so we can be more confident with the code under review

### What changed

Added a github action through the github actions UI.

## Checklists

- [ ] the `on` triggers make sense
- [ ] - [ ] that `gradlew test` is appropriate

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,4 +21,4 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Build with Gradle
-      run: ./gradlew test
+      run: ./gradlew build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,24 @@
+# This workflow will build a Java project with Gradle and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Java CI with Gradle
+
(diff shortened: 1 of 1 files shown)


kerrr (dismissed):

.github/workflows/test.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+        cache: gradle
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
```
jamietanna: Should this not already be executable, as it's committed into the repo as executable?

.github/workflows/test.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Build with Gradle
+      run: ./gradlew test
```
jamietanna: If we run a `build` it'll make sure we've run `check` which could include static analysis, as well as `assemble` and `javadoc`

```suggestion
      run: ./gradlew build
```

.github/workflows/test.yml
idrop: Hi Jamie, you're right, I just went with the default action that magically appeared in the github UI. I'll yank the line

kerrr (approved):

PR #31, opened 2021-11-19
BAU sonarcloud report on PR

idrop:
## Proposed changes

Add a sonarqube report when PRs are raised. Check out the sonarcloud reports below.

### What changed

Update gradle, github actions.


.github/workflows/test.yml
```diff
@@ -18,5 +19,20 @@ jobs:
...
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
Wynndow: Where do these secrets live - I'm guessing they're set in the GitHub UI somewhere?
Could you add a comment describing what the GitHub token is called and which bot account owns it, just so it's easy to track down if it needs rotating at any point? Also perhaps something similar for the Sonar token?

idrop: I'll add to team manual if that's ok?

Wynndow: I'd be more keen to see it here, or at least a pointer to the team manual? If someone is looking at this in the future who isn't familiar with the page in the team manual they may not think to go looking there.

I'm just remembering the issue with tracking stuff down when the Travis incident happened.

idrop: @Wynndow I've removed the GITHUB_TOKEN with no discernable bad effect, and commented on the SONAR_TOKEN in the script

Wynndow: https://media3.giphy.com/media/d6nvFKK6Dr9YbBDeNQ/giphy.gif

Wynndow (approved):

PR #33, opened 2021-11-24
PYI-356: Set up DynamoDB table for user credentials

Wynndow:
### What changed

**These changes have been manually tested and the result can be seen in the AWS dev account**

Defines a table in DynamoDB for the lambdas to store and retrieve user
info received from credential issuers.

It defines it with a composite primary key based on the users session ID
and the credential issuer. This will have to change in future when we
use longer term IDs for users.

The TF also defines the IAM policy to allow the lambdas to access the
DB, as well as providing the table name to the lambda as an env var.

### Issue tracking

- [PYI-356](https://govukverify.atlassian.net/browse/PYI-356)

### DIFF
--- a/terraform/lambda/dynamodb.tf
+++ b/terraform/lambda/dynamodb.tf
@@ -1,0 +1,46 @@
+resource "aws_dynamodb_table" "user-credentials-table" {
+  name         = "user-credentials-table"
+  hash_key     = "SessionId"
+  range_key    = "CredentialIssuer"
+  billing_mode = "PAY_PER_REQUEST"
+
+  attribute {
+    name = "SessionId"
+    type = "S"
+  }
+
+  attribute {
+    name = "CredentialIssuer"
+    type = "S"
+  }
+
+  tags = local.default_tags
+}
+
+resource "aws_iam_policy" "dynamo-db-user-credentials-table-policy" {
+  name = "dynamo-db-user-credentials-table-policy"
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Sid = "AllowDynamoDbReadWrite"
+        Action = [
+          "dynamodb:PutItem",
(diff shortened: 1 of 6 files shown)


terraform/lambda/dynamodb.tf
```diff
@@ -0,0 +1,46 @@
+resource "aws_dynamodb_table" "user-credentials-table" {
+  name         = "user-credentials-table"
```
philandstuff: 😬  naming is hard

1) don't include the resource type in the name:
```suggestion
  name         = "user-credentials"
```
2) auth already has a [DynamoDB table called user-credentials](https://github.com/alphagov/di-authentication-api/blob/main/ci/terraform/shared/dynamodb.tf#L2) which does something different.

Ideas: `user-verifiable-credentials`, `user-VCs`, `user-issued-credentials`. Dunno

philandstuff: 3. should we include the environment name in the resource name to avoid collisions if/when we have multiple environments in the same account?

Wynndow: Oh good spot. `user-issued-credentials` sounds like a good shout. I'll update to that.

Wynndow: We found previously on Verify that we started with the pattern of using the env as a prefix. We quickly became annoyed by it as it crept everywhere and was difficult to back out. 
It might be easier to add it retrospectively if we find that we're deploying multiple envs to the same account though. Is that likely though?

Wynndow: Although looking at it, we already do it here, so perhaps I should prefix the table.

Wynndow: Have fixed up if you wouldn't mind taking another look.

mack-mcq (approved):

philandstuff (approved):

PR #35, opened 2021-11-26
PYI-362 get credential issuer token

SunilDGDS:
Given an authorization code from front end, exchange it for a token from the credential issuer.

### What changed
Added a new lambda to di-ipv-core-back which will be called by the front end and in turn will call the credential issuer for 
an OAuth token.

A subsequent story uses the token. 

The API for this lambda is documented on the repo `README.md`.

Co-Authored-By: Phillip Miller

### Issue tracking

- [PYI-362](https://govukverify.atlassian.net/browse/PYI-362)

## Checklists

- [ ] Ensure the API docs make sense
- [ ] Check that its ok to return a 400 response in all error conditions

### DIFF
--- a/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
+++ b/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
@@ -5,12 +5,24 @@ import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.nimbusds.oauth2.sdk.AuthorizationCode;
+import com.nimbusds.oauth2.sdk.AuthorizationCodeGrant;
+import com.nimbusds.oauth2.sdk.SerializeException;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
```diff
@@ -4,34 +4,63 @@
...
-        CredentialIssuerRequestDto request = objectMapper.convertValue(body, CredentialIssuerRequestDto.class);
+        CredentialIssuerRequestDto request = RequestHelper.convertRequestBody(input.getBody(), CredentialIssuerRequestDto.class);
 
         var errorResponse = validate(request);
         if (errorResponse.isPresent()) {
             return errorResponse.get();
-        } else {
+        }
+
+        CredentialIssuerConfig credentialIssuerConfig = getCredentialIssuerConfig(request).get();
```
dannyhayes-gds: Potential for a NoSuchElementException here. Its been recognised by sonar as a bug and we should probably be doing a getCredentialIssuerConfig(request).isPresent() or getCredentialIssuerConfig(request).isEmpty() check before calling the .get()

idrop: 👍 the `validate` method above does the check to see if the config is present, so it feels safe to reach for the `.get()` here. I'll take a look at how to make it cleaner

Wynndow (approved):

PR #37, opened 2021-11-29
PYI-395: Store auth code in dynamoDB after its generated in AuthorizationHandler

dannyhayes-gds:
Co-authored-by: Tal Nagra <tal.nagra@digital.cabinet-office.gov.uk>

### What changed
Create generic DataStore that uses the DynamoDbEhancedClient to interact with our DB.
Create a configuration service to load and configure the DataStore with environment varaiables
Persist the generated auth code from within the authorisation handler

### Why did it change
So that when a request to exchange an auth code for a access token comes through, we can validate if their auth code is one that was generated via us and for the correct sessionID.

### Issue tracking

- [PYI-395](https://govukverify.atlassian.net/browse/PYI-395)

IS_LOCAL - used to configure the correct type of DynamoDb client, i.e one pointing to real aws or to localhost
ENVIRONMENT - used in the construction of dynamodb table names i.e "{environment}-auth-codes"

### DIFF
--- a/src/main/java/uk/gov/di/ipv/lambda/AuthorizationHandler.java
+++ b/src/main/java/uk/gov/di/ipv/lambda/AuthorizationHandler.java
@@ -13,10 +13,12 @@ import org.slf4j.LoggerFactory;
 import uk.gov.di.ipv.domain.ErrorResponse;
 import uk.gov.di.ipv.helpers.ApiGatewayResponseGenerator;
 import uk.gov.di.ipv.service.AuthorizationCodeService;
+import uk.gov.di.ipv.service.ConfigurationService;
 
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 public class AuthorizationHandler
@@ -26,12 +28,12 @@ public class AuthorizationHandler
 
     private final AuthorizationCodeService authorizationCodeService;
 
-    public AuthorizationHandler(AuthorizationCodeService authorizationCodeService) {
-        this.authorizationCodeService = authorizationCodeService;
-    }
-
     public AuthorizationHandler() {
         this.authorizationCodeService = new AuthorizationCodeService();
+    }
+
+    public AuthorizationHandler(AuthorizationCodeService authorizationCodeService) {
+        this.authorizationCodeService = authorizationCodeService;
     }
 
     @Override
@@ -50,7 +52,10 @@ public class AuthorizationHandler
             return ApiGatewayResponseGenerator.proxyJsonResponse(400, ErrorResponse.MissingRedirectURI);
         }
 
-        AuthorizationCode authorizationCode = authorizationCodeService.generateAuthorisationCode();
+        AuthorizationCode authorizationCode = authorizationCodeService.generateAuthorizationCode();
+
+        // TODO: Load sessionID value properly
+        authorizationCodeService.persistAuthorizationCode(UUID.randomUUID().toString(), authorizationCode.getValue());
 
         Map<String, Identifier> payload = Map.of("code", authorizationCode);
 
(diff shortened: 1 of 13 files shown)


terraform/lambda/dynamodb.tf
```diff
@@ -44,3 +57,31 @@ resource "aws_iam_policy" "access-user-issued-credentials-table" {
     ]
   })
 }
+
+resource "aws_iam_policy" "access-auth-codes-table" {
```
Wynndow, line 61: I think you're going to need a bit more terraform to attach this policy to the lambda execution role. You can see how I did it for the user-cred table here: https://github.com/alphagov/di-ipv-core-back/blob/main/terraform/modules/endpoint/iam.tf#L27

Using this pattern you can scope the perms to just the lambdas that need access to that table.

dannyhayes-gds: Ah yeah good shout, thanks. I've added the additional terraform resources that I think are required.

src/main/java/uk/gov/di/ipv/lambda/AuthorizationHandler.java
```diff
@@ -13,10 +13,12 @@
 import uk.gov.di.ipv.domain.ErrorResponse;
 import uk.gov.di.ipv.helpers.ApiGatewayResponseGenerator;
 import uk.gov.di.ipv.service.AuthorizationCodeService;
+import uk.gov.di.ipv.service.ConfigurationService;
```
Wynndow, line 16: I don't think this import is being used.

src/main/java/uk/gov/di/ipv/persistence/item/AuthorizationCodeItem.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.di.ipv.persistence.item;
+
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+
+@DynamoDbBean
+public class AuthorizationCodeItem {
```
Wynndow, line 7: I'm wondering if we might change this to be a more generic item (and table) in future. Rather than maintaining a table just for auth codes, we might end up with the access token and maybe the user_id from Auth&Orc in there. Maybe some other bits. I guess we can change this if/when we get there though.

terraform/lambda/dynamodb.tf
```diff
@@ -44,3 +57,31 @@ resource "aws_iam_policy" "access-user-issued-credentials-table" {
...
+        Sid = "AccessAuthCodesTable"
+        Action = [
+          "dynamodb:PutItem",
+          "dynamodb:UpdateItem",
+          "dynamodb:BatchWriteItem",
+          "dynamodb:GetItem",
+          "dynamodb:BatchGetItem",
+          "dynamodb:Scan",
+          "dynamodb:Query",
+          "dynamodb:ConditionCheckItem"
```
Wynndow, line 77: We have the delete method in the new `DataStore` class (which I'm not sure we'll need), but it won't work as we don't have the `dynamodb:DeleteItem` permission in here. I don't think we should add it now, but it's worth noting in case we do need it in future.

Wynndow (approved):
Just a comment about an unused import, but it can always be fixed in another PR.

terraform/lambda/dynamodb.tf
dannyhayes-gds: Ah yeah good catch. Interestingly it did seem to work in localstack however so thats interesting to note that these permissions aren't potentially being used locally? Something to potentially investigate

Wynndow: Yeah it'd be interesting to know. perhaps localstacks modeling of AWS perms isn't bob on...

PR #42, opened 2021-11-30
BAU Java Spotless check

idrop:
## Proposed changes

A java linter that runs before the `gradle compileJava` task, so really up in your grille. It'll also get run when a PR is created or updated.

This could pair well with a pre-push and/or pre-commit git hook.

Fixes can be applied with `./gradlew :spotlessApply` and are the 2nd commit on this PR.

More info on spotless here: https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format


Wynndow (dismissed):
I'm in for this. Although I wouldn't want it on a pre-commit or push hook. Don't want to get blocked from doing things quickly. Part of the PR checks would be good though.

build.gradle
```diff
@@ -57,3 +61,14 @@ sonarqube {
...
+
+spotless {
+    java {
+        target "**/*.java"
+        googleJavaFormat("1.12.0").aosp()
+        importOrder "", "javax", "java", "\\#"
+    }
+}
+
+compileJava.dependsOn "spotlessCheck"
```
jamietanna: This shouldn't be required, as it's bound into the `check` phase

build.gradle
```diff
@@ -57,3 +61,14 @@ sonarqube {
         property "sonar.host.url", "https://sonarcloud.io"
     }
 }
+
+spotless {
+    java {
+        target "**/*.java"
```
jamietanna: Spotless team recommends this as it includes less code unnecessarily:

```suggestion
        target "src/**/*.java"
```

build.gradle
```diff
@@ -1,8 +1,12 @@
+import org.gradle.api.JavaVersion
+import org.gradle.api.tasks.bundling.Zip
+
 plugins {
     id "java"
     id "application"
     id "idea"
     id "org.sonarqube" version "3.3"
+    id "com.diffplug.spotless" version "5.17.1"
```
jamietanna: Can we not bump to a later version?

build.gradle
```diff
@@ -1,8 +1,12 @@
+import org.gradle.api.JavaVersion
+import org.gradle.api.tasks.bundling.Zip
```
jamietanna: were these needed?

build.gradle
```diff
@@ -57,3 +61,14 @@ sonarqube {
         property "sonar.host.url", "https://sonarcloud.io"
     }
 }
+
+spotless {
```
jamietanna, line 71: should we also add `groovyGradle` formatting?

kerrr (approved):

PR #48, opened 2021-12-02
PYI-403: credential issuer config

Danomanic:
### What changed

Adds terraform to set an env variable on the credential issuer lambda that will contain a base64 encoded string of a yaml file defined in the di-ipv-config repo.

Adds SSM call to retrieve the credential issuer config base64 string then decodes and parses into a credential issuers object to be used by the credential issuer handler.

### Issue tracking

- [PYI-403](https://govukverify.atlassian.net/browse/PYI-403)

### DIFF
--- a/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
+++ b/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
@@ -1,0 +1,42 @@
+package uk.gov.di.ipv.dto;
+
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+
+public class CredentialIssuers {
+
+    private final Set<CredentialIssuerConfig> credentialIssuerConfigs;
+
+    public CredentialIssuers() {
+        credentialIssuerConfigs = Collections.emptySet();
+    }
+
+    public CredentialIssuers(Set<CredentialIssuerConfig> credentialIssuerConfigs) {
+        this.credentialIssuerConfigs = credentialIssuerConfigs;
+    }
+
+    public Set<CredentialIssuerConfig> getCredentialIssuerConfigs() {
+        return credentialIssuerConfigs;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CredentialIssuers that = (CredentialIssuers) o;
+        return Objects.equals(credentialIssuerConfigs, that.credentialIssuerConfigs);
+    }
+
+    @Override
+    public String toString() {
+        return "CredentialIssuers{" +
+                "credentialIssuerConfigs=" + credentialIssuerConfigs +
+                '}';
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(credentialIssuerConfigs);
+    }
+}
--- a/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
+++ b/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
@@ -25,18 +25,14 @@ public class CredentialIssuerHandler implements RequestHandler<APIGatewayProxyRe
 
     private final CredentialIssuers credentialIssuers;
 
-    private final ConfigurationService configurationService;
-
     public CredentialIssuerHandler(CredentialIssuerService credentialIssuerService, ConfigurationService configurationService) {
         this.credentialIssuerService = credentialIssuerService;
-        this.configurationService = configurationService;
         this.credentialIssuers = configurationService.getCredentialIssuers();
     }
 
     public CredentialIssuerHandler() {
-        this.configurationService = new ConfigurationService();
         this.credentialIssuerService = new CredentialIssuerService();
-        this.credentialIssuers = configurationService.getCredentialIssuers();
+        this.credentialIssuers = new ConfigurationService().getCredentialIssuers();
     }
 
     @Override
--- a/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
+++ b/src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
@@ -6,43 +6,37 @@ import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
+import java.util.Collections;
+import java.util.Optional;
 import net.minidev.json.JSONObject;
 import uk.gov.di.ipv.domain.CredentialIssuerException;
 import uk.gov.di.ipv.domain.ErrorResponse;
 import uk.gov.di.ipv.dto.CredentialIssuerConfig;
 import uk.gov.di.ipv.dto.CredentialIssuerRequestDto;
+import uk.gov.di.ipv.dto.CredentialIssuers;
 import uk.gov.di.ipv.helpers.ApiGatewayResponseGenerator;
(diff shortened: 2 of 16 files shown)


src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
```diff
@@ -0,0 +1,42 @@
...
+        return credentialIssuerConfigs;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CredentialIssuers that = (CredentialIssuers) o;
+        return Objects.equals(credentialIssuerConfigs, that.credentialIssuerConfigs);
+    }
```
idrop, line 29: are we ever comparing `CredentialIssuers`? If not we shouldn't bother with an equals and hashcode method

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -1,28 +1,42 @@
...
     public String getAuthCodesTableName() {
         return System.getenv("AUTH_CODES_TABLE_NAME");
     }
 
+    public String getCredentialIssuerConfigParameterStoreKey() {
+        return System.getenv("CREDENTIAL_ISSUER_CONFIG_PARAMETER_STORE_KEY");
+    }
+
+    public CredentialIssuers getCredentialIssuers() {
+        String value = ssmProvider.get(getCredentialIssuerConfigParameterStoreKey());
```
idrop: A small thing: could we inline this var on rename it to something other than value?

src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
Danomanic: Only in the test shouldLoadCredentialIssuersFromBase64EncodedString() - Happy for this to stay in, or want me to refactor the test?

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
Danomanic: Done

src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
```diff
@@ -6,43 +6,37 @@
...
-    );
-    protected static final CredentialIssuerConfig FRAUD_ISSUER = new CredentialIssuerConfig(
-            "FraudIssuer",
-            URI.create("http://credential-issuer-stub:8084/token"),
-            URI.create("http://credential-issuer-stub:8084/credential")
-    );
+    private final CredentialIssuers credentialIssuers;
 
-    protected static final Set<CredentialIssuerConfig> CREDENTIAL_ISSUERS = Set.of(PASSPORT_ISSUER, FRAUD_ISSUER);
+    private final ConfigurationService configurationService;
```
Wynndow: I think we can drop this field. Just set the `credentialIssuers` field in the constructors directly (in the no args constructor just call `new ConfigurationService().getCredentialIssuers()`)

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -1,38 +1,54 @@
...
+        return Boolean.parseBoolean(System.getenv("IS_LOCAL"));
     }
 
     public String getAuthCodesTableName() {
         return System.getenv("AUTH_CODES_TABLE_NAME");
     }
 
+    public String getCredentialIssuerConfigParameterStoreKey() {
+        return System.getenv("CREDENTIAL_ISSUER_CONFIG_PARAMETER_STORE_KEY");
+    }
```
Wynndow: I'd be tempted to just inline this in the method below.

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -1,38 +1,54 @@
...
+
+    public CredentialIssuers getCredentialIssuers() {
+        return CredentialIssuerLoader.loadCredentialIssuers(
```
Wynndow, line 32: I thought it might be nice to use the inbuilt deserializing thing in Powertools, seeing as we're using it. I had a quick crack at it [here](https://github.com/alphagov/di-ipv-core-back/blob/bau-deserialize-with-powertools/src/main/java/uk/gov/di/ipv/service/ConfigurationService.java#L32). Would need to change the CredentialssuerLoader into [a transformer](https://github.com/alphagov/di-ipv-core-back/blob/bau-deserialize-with-powertools/src/main/java/uk/gov/di/ipv/helpers/Base64YamlTransformer.java) though.

terraform/modules/endpoint/main.tf
```diff
@@ -21,9 +21,10 @@ resource "aws_lambda_function" "lambda_function" {
 
   environment {
     variables = {
-      USER_ISSUED_CREDENTIALS_TABLE_NAME = var.user_issued_credentials_table_name
-      AUTH_CODES_TABLE_NAME = var.auth_codes_table_name
-      ACCESS_TOKENS_TABLE_NAME = var.access_tokens_table_name
+      USER_ISSUED_CREDENTIALS_TABLE_NAME           = var.user_issued_credentials_table_name
```
Wynndow: I think you can drop these terraform changes - I've taken them over into [my PR](https://github.com/alphagov/di-ipv-core-back/pull/50/files).

terraform/modules/endpoint/variables.tf
```diff
@@ -96,6 +96,12 @@ variable "access_tokens_table_name" {
   description = "Name of the DynamoDB access-tokens table"
 }
 
+variable "credential_issuer_config_parameter_store_key" {
```
Wynndow: As above.

src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
```diff
@@ -86,7 +80,7 @@ public APIGatewayProxyResponseEvent handleRequest(APIGatewayProxyRequestEvent in
     }
 
     private CredentialIssuerConfig getCredentialIssuerConfig(CredentialIssuerRequestDto request) {
-        return CREDENTIAL_ISSUERS.stream()
+        return credentialIssuers.getCredentialIssuerConfigs().stream()
```
Wynndow, line 79: We need to add some logic to refresh the credential issuers after a certain period, say 5 minutes. I'm thinking this might be a reasonable place to do it. We could have a field on the `CredentialIssuers` object that is a timestamp of when it expires (set when the object is created). We could then check here if it's expired and re-assign `credentialIssuers` with a call to `getCredentialIssuers()` if it has.
We could probably do it as another ticket though. Let's have a chat about it.

src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
Danomanic: Good idea, done!

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
Danomanic: I agree, done!

terraform/modules/endpoint/main.tf
Danomanic: Removed!

Danomanic: Done

terraform/modules/endpoint/variables.tf
Danomanic: Done!

src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
idrop: ok 👍 , pls keep it in then 😄

Wynndow (dismissed):

Wynndow (approved):
🎉

PR #50, opened 2021-12-06
PYI-419: Create credential-issuers param

Wynndow:
**To be merged with https://github.com/alphagov/di-ipv-config/pull/55**

### What changed

Adds an SSM parameter that contains the base64 encoded credential
issuers config yaml files from the di-ipv-config repo.

Also handles providing the name of that parameter to the
credential-issuer lambda as an env variable.

### Issue tracking

- [PYI-419](https://govukverify.atlassian.net/browse/PYI-419)

### DIFF
--- a/terraform/lambda/parameter-store.tf
+++ b/terraform/lambda/parameter-store.tf
@@ -1,0 +1,5 @@
+resource "aws_ssm_parameter" "credential-issuers-config" {
+  name  = "/${var.environment}/credential-issuers-config"
+  type  = "String"
+  value = var.credential_issuers_config
+}
--- a/terraform/lambda/credential-issuer.tf
+++ b/terraform/lambda/credential-issuer.tf
@@ -14,4 +14,7 @@ module "credential-issuer" {
   allow_access_to_user_issued_credentials_table = true
   user_issued_credentials_table_policy_arn      = aws_iam_policy.policy-user-issued-credentials-table.arn
   user_issued_credentials_table_name            = aws_dynamodb_table.user-issued-credentials.name
+
+  credential_issuer_config_parameter_store_key = aws_ssm_parameter.credential-issuers-config.name
 }
+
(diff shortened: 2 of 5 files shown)


terraform/lambda/parameter-store.tf
```diff
@@ -0,0 +1,5 @@
+resource "aws_ssm_parameter" "credential-issuers-config" {
+  name  = "/${var.environment}/credential-issuers-config"
```
dannyhayes-gds, line 2: Do we need the / characters here? Should the name of this be something like ${var.environment}-credential-issuers-config

dannyhayes-gds:
Do we need to create a dev.yml file here that will be used by concourse? Or maybe I'm misunderstanding or that will be a different PR?

terraform/lambda/parameter-store.tf
Wynndow: The param store has an idea of using paths to namespace stuff, so you could pull all params under `/dev/` for example. We've said in the ADR that we want to namespace params in case we want to deploy multiple envs to the same account. I figured using the paths syntax will give us that as well as being able to use the hierarchy if needed.

Wynndow:
> Do we need to create a dev.yml file here that will be used by concourse? Or maybe I'm misunderstanding or that will be a different PR?

This is over in the [di-ipv-config](https://github.com/alphagov/di-ipv-config/blob/main/credential-issuers/dev.yaml) repo. I'd managed to delete it by accident in my PR on that repo which is confusing, but I've resolved it now.

terraform/lambda/parameter-store.tf
Wynndow: https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-hierarchies.html

kerrr (approved):

idrop (approved):

PR #51, opened 2021-12-07
PYI-412: Create new ipv-session endpoint

dannyhayes-gds:
Co-authored-by: Tal Nagra <tal.nagra@digital.cabinet-office.gov.uk>

### What changed
Create new IpvSession lambda to initialise an ipvSession.
Add new terraform resources to create the aws api gateway endpoint.
Add new terraform resources to create new DynamoDB table to store the ipv-session

### Why did it change
To be used by core front to initialise a session with core-back.

### Issue tracking

- [PYI-412](https://govukverify.atlassian.net/browse/PYI-412)

### DIFF
--- a/src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
+++ b/src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
@@ -1,0 +1,32 @@
+package uk.gov.di.ipv.service;
+
+import uk.gov.di.ipv.persistence.DataStore;
+import uk.gov.di.ipv.persistence.item.IpvSessionItem;
+
+import java.time.Instant;
+import java.util.UUID;
+
+public class IpvSessionService {
+
+    private final DataStore<IpvSessionItem> dataStore;
+    private final ConfigurationService configurationService;
+
+    public IpvSessionService() {
+        this.configurationService = ConfigurationService.getInstance();
+        dataStore = new DataStore<>(configurationService.getIpvSessionTableName(), IpvSessionItem.class);
+    }
+
+    public IpvSessionService(DataStore<IpvSessionItem> dataStore, ConfigurationService configurationService) {
+        this.dataStore = dataStore;
+        this.configurationService = configurationService;
+    }
+
+    public IpvSessionItem generateIpvSession() {
+        IpvSessionItem ipvSessionItem = new IpvSessionItem();
+        ipvSessionItem.setIpvSessionId(UUID.randomUUID().toString());
+        ipvSessionItem.setCreationDateTime(Instant.now().toString());
+        dataStore.create(ipvSessionItem);
+
+        return ipvSessionItem;
+    }
+}
--- a/src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
+++ b/src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
@@ -21,12 +21,12 @@ public class IpvSessionService {
         this.configurationService = configurationService;
     }
 
-    public IpvSessionItem generateIpvSession() {
+    public String generateIpvSession() {
         IpvSessionItem ipvSessionItem = new IpvSessionItem();
         ipvSessionItem.setIpvSessionId(UUID.randomUUID().toString());
         ipvSessionItem.setCreationDateTime(Instant.now().toString());
         dataStore.create(ipvSessionItem);
 
-        return ipvSessionItem;
+        return ipvSessionItem.getIpvSessionId();
     }
 }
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/di/ipv/lambda/IpvSessionHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
+        String requestBody = "userId=" + testUserId;
+        event.setBody(requestBody);
+
+        APIGatewayProxyResponseEvent response = ipvSessionHandler.handleRequest(event, mockContext);
+
+        Map<String, Object> responseBody = objectMapper.readValue(response.getBody(), Map.class);
+
+        assertEquals(200, response.getStatusCode());
```
tal-nagra: The 200 can be replaced with an appropriate constant from the nimbus jar

src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
```diff
@@ -0,0 +1,32 @@
...
+        this.configurationService = ConfigurationService.getInstance();
+        dataStore = new DataStore<>(configurationService.getIpvSessionTableName(), IpvSessionItem.class);
+    }
+
+    public IpvSessionService(DataStore<IpvSessionItem> dataStore, ConfigurationService configurationService) {
+        this.dataStore = dataStore;
+        this.configurationService = configurationService;
+    }
+
+    public IpvSessionItem generateIpvSession() {
```
tal-nagra: Thus far we haven't exposed the *Item classes beyond the service layer.
At this stage it's probably best to just return a string or UUID containing the ipv session id unless the session creation date time is needed by core-front. 
In which case it would be worth creating a separate class in the domain package to return from the service. So that the *Item classes have a single responsibility of facilitation reading/writing to the database and the *Handler classes don't have knowledge of the *Item classes

src/test/java/uk/gov/di/ipv/service/IpvSessionServiceTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+class IpvSessionServiceTest {
+
+    @Mock
+    DataStore<IpvSessionItem> mockDataStore;
```
tal-nagra: add `private` access modifier

src/test/java/uk/gov/di/ipv/service/IpvSessionServiceTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Mock
+    ConfigurationService mockConfigurationService;
```
tal-nagra: add `private` access modifier

terraform/lambda/ipv-session.tf
```diff
@@ -0,0 +1,17 @@
...
+  source      = "../modules/endpoint"
+  environment = var.environment
+
+  rest_api_id            = aws_api_gateway_rest_api.ipv_internal.id
+  rest_api_execution_arn = aws_api_gateway_rest_api.ipv_internal.execution_arn
+  root_resource_id       = aws_api_gateway_rest_api.ipv_internal.root_resource_id
+  http_method            = "POST"
+  path_part              = "ipv-session"
+  handler                = "uk.gov.di.ipv.lambda.IpvSessionHandler::handleRequest"
+  function_name          = "${var.environment}-create-ipv-session"
```
tal-nagra, line 11: I think this PR also needs the amend we made to the bash script to deploy the new lambda function to localstack ?

tal-nagra (changes requested):

terraform/lambda/ipv-session.tf
dannyhayes-gds: That will be a separate PR within the localstartup repo

src/main/java/uk/gov/di/ipv/service/IpvSessionService.java
dannyhayes-gds: Ah yeah good shout. Not 100% sure if the full session object will be needed for core-front. Most likely not as I think it will just be using the ipvSessionId and passing it around in a header. I'll change to just return the id field.

src/test/java/uk/gov/di/ipv/lambda/IpvSessionHandlerTest.java
dannyhayes-gds: I've found the tests to be a bit clearer by specifying the specific number code, but thats more just a personal preference so I don't mind swapping this around.

tal-nagra (approved):

PR #53, opened 2021-12-07
PYI-403: Use transformer with SSMProvider

Wynndow:
## Proposed changes

The SSM provider from powertools has the ability to do the deserializing
for you. You just need to write a custom transformer.

This converts our current logic to a transformer for base64 encoded
yaml. This should mean we can use it in future if we need to deserialize
other yaml.

### Issue tracking

- [PYI-403](https://govukverify.atlassian.net/browse/PYI-403)


Wynndow:
Closing due to being unable to cache the way we want with an inline transformer.

PR #58, opened 2021-12-07
PYI 421 cache credential issuer config

SunilDGDS:
In the credential issuers , we now store the current value received when retrieving credentials from SSM  .  Initially it is blank but subsequently populated . If the value is different to the one  retrieved, then we stored (serialise) the new credentials

- [PYI-421](https://govukverify.atlassian.net/browse/PYI-421)

### DIFF
--- a/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
+++ b/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
@@ -1,6 +1,7 @@
 package uk.gov.di.ipv.dto;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+
 import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
@@ -9,15 +10,13 @@ public class CredentialIssuers {
 
     private final Set<CredentialIssuerConfig> credentialIssuerConfigs;
 
-    @JsonIgnore
-    private String source;
+    @JsonIgnore private String source;
 
     public CredentialIssuers() {
         credentialIssuerConfigs = Collections.emptySet();
     }
 
-    public CredentialIssuers(
-        Set<CredentialIssuerConfig> credentialIssuerConfigs, String source) {
+    public CredentialIssuers(Set<CredentialIssuerConfig> credentialIssuerConfigs, String source) {
         this.credentialIssuerConfigs = credentialIssuerConfigs;
         this.source = source;
     }
--- a/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
+++ b/src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
@@ -1,5 +1,6 @@
 package uk.gov.di.ipv.dto;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
@@ -8,8 +9,17 @@ public class CredentialIssuers {
 
     private final Set<CredentialIssuerConfig> credentialIssuerConfigs;
 
+    @JsonIgnore
+    private String source;
+
     public CredentialIssuers() {
         credentialIssuerConfigs = Collections.emptySet();
+    }
+
+    public CredentialIssuers(
+        Set<CredentialIssuerConfig> credentialIssuerConfigs, String source) {
+        this.credentialIssuerConfigs = credentialIssuerConfigs;
+        this.source = source;
     }
 
     public CredentialIssuers(Set<CredentialIssuerConfig> credentialIssuerConfigs) {
@@ -21,14 +31,6 @@ public class CredentialIssuers {
     }
 
     @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        CredentialIssuers that = (CredentialIssuers) o;
-        return Objects.equals(credentialIssuerConfigs, that.credentialIssuerConfigs);
-    }
-
-    @Override
     public String toString() {
         return "CredentialIssuers{" + "credentialIssuerConfigs=" + credentialIssuerConfigs + '}';
     }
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
```diff
@@ -24,8 +24,9 @@
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
 
     private final CredentialIssuerService credentialIssuerService;
+    private final ConfigurationService configurationService;
 
-    private final CredentialIssuers credentialIssuers;
+    private CredentialIssuers credentialIssuers = null;
```
Wynndow: I think we don't actually need to explicitly set this to null here. Maybe it's better to though.

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -29,9 +29,20 @@ public String getAuthCodesTableName() {
...
+    public CredentialIssuers getCredentialIssuers(CredentialIssuers credentialIssuers) {
+
+        String credentialIssuerConfigBase64 =
+                ssmProvider.get(System.getenv("CREDENTIAL_ISSUER_CONFIG_PARAMETER_STORE_KEY"));
+
+        if (credentialIssuers == null
+                || !credentialIssuerConfigBase64.equals(credentialIssuers.getSource())) {
+            var credentialIssuersResult =
+                    CredentialIssuerLoader.loadCredentialIssuers(credentialIssuerConfigBase64);
+            credentialIssuersResult.setSource(credentialIssuerConfigBase64);
```
Wynndow: I think if we do the `setSource` in the `CredentialIssuerLoader` we could just do `return credentialIssuerLoader.loadCredentialIssuers(credentialIssuerConfigBase64)` here which might be a bit cleaner.

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -29,9 +29,20 @@ public String getAuthCodesTableName() {
...
-    public CredentialIssuers getCredentialIssuers() {
-        return CredentialIssuerLoader.loadCredentialIssuers(
-                ssmProvider.get(System.getenv("CREDENTIAL_ISSUER_CONFIG_PARAMETER_STORE_KEY")));
+    public CredentialIssuers getCredentialIssuers(CredentialIssuers credentialIssuers) {
+
+        String credentialIssuerConfigBase64 =
+                ssmProvider.get(System.getenv("CREDENTIAL_ISSUER_CONFIG_PARAMETER_STORE_KEY"));
+
+        if (credentialIssuers == null
+                || !credentialIssuerConfigBase64.equals(credentialIssuers.getSource())) {
```
Wynndow: It might be nice to move this source checking logic to a method on the `CredentialIssuers` class. That way we could do something like: 
`if (credentialIssuers == null || credentialIssuers.fromDifferentSource(credentialIssuerConfigBase64))`

src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
```diff
@@ -35,11 +41,35 @@ void getCredentialIssuers() throws URISyntaxException {
...
-        assertEquals(credentialIssuers, underTest.getCredentialIssuers());
+        when(ssmProvider.get(any()))
+                .thenReturn(
+                        CREDENTIAL_ISSUER_CONFIG_BASE64_VERSION_1,
+                        CREDENTIAL_ISSUER_CONFIG_BASE64_VERSION_1);
+
+        CredentialIssuers credentialIssuers1 = underTest.getCredentialIssuers(credentialIssuers);
+        CredentialIssuers credentialIssuers2 = underTest.getCredentialIssuers(credentialIssuers);
+
+        assertEquals(credentialIssuers1, credentialIssuers2);
```
Wynndow: I think we need to add an assertion that `CredentialIssuerLoader.loadCredentialIssuers` has only been called once in this scenario. As it stands it's not really prooving the caching logic we want.

src/main/java/uk/gov/di/ipv/dto/CredentialIssuers.java
```diff
@@ -1,17 +1,25 @@
...
+
 import java.util.Collections;
-import java.util.Objects;
 import java.util.Set;
 
 public class CredentialIssuers {
 
     private final Set<CredentialIssuerConfig> credentialIssuerConfigs;
 
+    @JsonIgnore private String source;
```
Wynndow: I've just tried it and I don't think we need the `@JsonIgnore` annotation here. The object mapper will happily give us a CredentialIssuers object with this field set to null.

src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
kerrr: I agree, don't need to set it explicitly.

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
SunilDGDS: done

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
SunilDGDS: done

src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
SunilDGDS: done

src/main/java/uk/gov/di/ipv/lambda/CredentialIssuerHandler.java
Wynndow: @SunilDGDS If you whip this `null` out I'll give the PR a 👍

Wynndow (approved):
👏

PR #63, opened 2021-12-08
BAU: Test local SSM Client

Wynndow:
### What changed

This adds a test to ensure that the SSM client is sending requests to
localhost when the "IS_LOCAL" env variable is set to true.

### Why did it change

This was mostly done to implement the mocking of environment variables.
This is achieved with the systems stubs library.

We depend on env variables more and more and they make testing
difficult. Having an easy way to mock them should be valuable.

### DIFF
--- a/src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
+++ b/src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
@@ -12,8 +12,10 @@ import java.util.Optional;
 
 public class ConfigurationService {
 
+    public static final int LOCALHOST_PORT = 4567;
+    private static final String LOCALHOST_URI = "http://localhost:" + LOCALHOST_PORT;
     private static final long DEFAULT_BEARER_TOKEN_TTL_IN_SECS = 3600L;
-    private static final String LOCALHOST_URI = "http://localhost:4567";
+    private static final String IS_LOCAL = "IS_LOCAL";
 
     private final SSMProvider ssmProvider;
 
@@ -22,7 +24,7 @@ public class ConfigurationService {
     }
 
     public ConfigurationService() {
-        if (Boolean.parseBoolean(System.getenv("IS_LOCAL"))) {
+        if (isRunningLocally()) {
             this.ssmProvider =
                     ParamManager.getSsmProvider(
                             SsmClient.builder()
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
```diff
@@ -72,4 +92,37 @@ void shouldReturnSameCredentialIssuersWhenBase64EncodingHasNotChanged() {
...
+        Exception e =
+                assertThrows(
+                        Exception.class,
+                        () -> new ConfigurationService().getSsmProvider().get("any-old-thing"));
+
+        System.out.println(e.getClass());
+        System.out.println(e.getMessage());
+        System.out.println(e.getStackTrace());
+        System.out.println(getAllServeEvents());
+        System.out.println(getAllServeEvents().size());
```
dannyhayes-gds: Do we need to keep this println() calls?

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -22,7 +24,9 @@ public ConfigurationService(SSMProvider ssmProvider) {
     }
 
     public ConfigurationService() {
-        if (Boolean.parseBoolean(System.getenv("IS_LOCAL"))) {
+        if (Boolean.parseBoolean(System.getenv(IS_LOCAL))) {
+            System.out.println("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
```
dannyhayes-gds: Same here?

src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
Wynndow: Not at all. This is just me working out why the test wasn't passing when running with GitHub actions. Will be removing them.

Wynndow: OK, should be good to go now

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
```diff
@@ -22,7 +24,7 @@ public ConfigurationService(SSMProvider ssmProvider) {
     }
 
     public ConfigurationService() {
-        if (Boolean.parseBoolean(System.getenv("IS_LOCAL"))) {
+        if (Boolean.parseBoolean(System.getenv(IS_LOCAL))) {
```
Danomanic: Would it be worth just calling the method on this class?
```suggestion
        if (isRunningLocally) {
```

Danomanic (dismissed):
Looks good to me.

src/main/java/uk/gov/di/ipv/service/ConfigurationService.java
Wynndow: ~~I did think that, but the method is an instance method, not static. I tried making it static but there were "problems".~~ 
Oh! I just tried it and it totally works. That's confused me but it's better so it's going in.

Danomanic (approved):
Looks good!

PR #64, opened 2021-12-09
BAU: Increase test coverage

Wynndow:
### What changed

This adds a few tests to cover the lines that are missing according to
sonarcloud.
### Why did it change

A lot of these are just testing the no arg constructors for the handlers
and services used by the lambdas. Whilst there is not much merit in
testing them this way, it does boost our coverage report.

Our quality gate is set to 80% coverage. Without these changes we're
quite close to this boundary. Having code in these constructors drag us
down isn't helpful.

### DIFF
--- a/src/test/java/uk/gov/di/ipv/service/UserIdentityServiceTest.java
+++ b/src/test/java/uk/gov/di/ipv/service/UserIdentityServiceTest.java
@@ -7,17 +7,28 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import uk.gov.di.ipv.persistence.DataStore;
 import uk.gov.di.ipv.persistence.item.UserIssuedCredentialsItem;
+import uk.org.webcompere.systemstubs.environment.EnvironmentVariables;
+import uk.org.webcompere.systemstubs.jupiter.SystemStub;
+import uk.org.webcompere.systemstubs.jupiter.SystemStubsExtension;
+import uk.org.webcompere.systemstubs.properties.SystemProperties;
 
 import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Map;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
+import static uk.gov.di.ipv.service.ConfigurationService.IS_LOCAL;
 
+@ExtendWith(SystemStubsExtension.class)
 @ExtendWith(MockitoExtension.class)
 class UserIdentityServiceTest {
+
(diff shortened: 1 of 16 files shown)


src/test/java/uk/gov/di/ipv/service/UserIdentityServiceTest.java
```diff
@@ -7,18 +7,29 @@
...
 import java.util.List;
 import java.util.Map;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
+import static uk.gov.di.ipv.service.ConfigurationService.IS_LOCAL;
 
+@ExtendWith(SystemStubsExtension.class)
```
jamietanna, line 25: :question: are we not able to refactor our implementations to make it so we don't _need_ to do this? I've had to deal with this in the past and it can make tests much slower, harder to parallelise and harder to maintain

kerrr: Yes, we should have a think about this.

kerrr (approved):

src/test/java/uk/gov/di/ipv/service/UserIdentityServiceTest.java
Wynndow: It doesn't feel ideal, but it does seem a convenient way to test code paths we can't inject into.
The tests are still pretty speedy....

jamietanna: We say we can't inject - is that something we can solve by providing test-only constructors that allow us to inject mocks, then production code uses the current means?

Wynndow: That's exactly what we're doing at the moment. These tests are to try to raise our code coverage by exercising those constructors used by the prod code. Been chatting with Kerr about some alternatives - possibly just ignoring them from the code coverage report.

jamietanna: Ah gotcha, in which case I'd recommend just having a test which calls the constructor and does nothing else? Makes it clear its just for coverage / as you say just exclude it?

Wynndow: i _think_ that's what i'm doing here? Thing is, the code the constructors call need certain env variables set - hence the library use.

jamietanna: Apologies, this is me not reading properly - sorry about that :+1: Yep, you've done exactly what I was proposing

Wynndow:
Closed in favour of https://github.com/alphagov/di-ipv-core-back/pull/65

PR #65, opened 2021-12-09
BAU: Ignore no arg constructors from coverage reports

Wynndow:
### What changed

Turns out that jacoco will ignore code marked with an annotation that
includes the word "Generated".

The first commit adds a new annotation we can apply only to constructors which we
can use to ignore those constructors which are painful to test.

The second commit refactors the DataStore to have one constructor used by the
lambda handlers as well as the tests.

The third commit adds a few tests to pick some low hanging fruit with regards to 
coverage.

### Why did it change

To move us away from out code coverage quality gate threshold

### DIFF
--- a/src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
+++ b/src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
@@ -20,7 +20,8 @@ import uk.org.webcompere.systemstubs.properties.SystemProperties;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.HashMap;
-import java.util.Set;
+import java.util.HashSet;
+import java.util.List;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.*;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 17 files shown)


Danomanic (dismissed):
Looks good!

src/test/java/uk/gov/di/ipv/service/ConfigurationServiceTest.java
```diff
@@ -53,18 +54,28 @@
...
+                                        new CredentialIssuerConfig(
+                                                "FraudIssuer",
+                                                new URI("http://www.example.com"),
+                                                new URI("http://www.example.com/credential")))),
                         CREDENTIAL_ISSUER_CONFIG_BASE64_VERSION_1);
     }
 
+    @Test
+    void shouldReturnCredentialIssuersWhenPassedNull() {
+        ConfigurationService underTest = new ConfigurationService(ssmProvider);
```
Danomanic: Should we move this out to a field now if several tests are using it?

Wynndow: Sounds good. have just pushed the change.

Danomanic (approved):
Good to go!

kerrr (approved):

PR #66, opened 2021-12-17
DRY up endpoint module

philandstuff:
Rather than have a bunch of specific variables for each policy we
might add, we can just pass in a list of policy ARNs to attach and
iterate over the list.

This is untested.

### What changed

See above

### Why did it change

Make the terraform code easier to understand and maintain.

### Issue tracking

### Other considerations

### DIFF
--- a/terraform/modules/endpoint/iam.tf
+++ b/terraform/modules/endpoint/iam.tf
@@ -24,27 +24,8 @@ resource "aws_iam_role" "lambda_iam_role" {
   tags = local.default_tags
 }
 
-resource "aws_iam_role_policy_attachment" "user_issued_credentials_table_policy_to_lambda_iam_role" {
-  count      = var.allow_access_to_user_issued_credentials_table ? 1 : 0
+resource "aws_iam_role_policy_attachment" "additional_policies" {
+  for_each   = toset(var.additional_policies)
   role       = aws_iam_role.lambda_iam_role.name
-  policy_arn = var.user_issued_credentials_table_policy_arn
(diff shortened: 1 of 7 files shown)


dannyhayes-gds (approved):
Looks good, much cleaner! I've tried one of the lambdas locally within localstack and it seemed to still work.

terraform/modules/endpoint/iam.tf
```diff
@@ -24,27 +24,8 @@ resource "aws_iam_role" "lambda_iam_role" {
   tags = local.default_tags
 }
 
-resource "aws_iam_role_policy_attachment" "user_issued_credentials_table_policy_to_lambda_iam_role" {
-  count      = var.allow_access_to_user_issued_credentials_table ? 1 : 0
+resource "aws_iam_role_policy_attachment" "additional_policies" {
+  for_each   = toset(var.additional_policies)
```
dannyhayes-gds, line 28: Just tried running this again within localstack from a clean start and I get an error for this line.

Error: Invalid for_each argument
│
│   on ../../di-ipv-core-back/terraform/modules/endpoint/iam.tf line 28, in resource "aws_iam_role_policy_attachment" "additional_policies":
│   28:   for_each   = toset(var.additional_policies)
│     ├────────────────
│     │ var.additional_policies is list of string with 1 element
│
│ The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends on.

PR #68, opened 2022-01-07
Provide a common library and individual handlers for lambdas

TomTucka:
## Proposed changes

All code expect lambda handler and test will been moved into a java library with it's own gradle build.
All lambdas will be built individually and pull in the shared library.

### What changed

All library code moved into `lib/`
Each lambda has been moved into `lambdas/<lambda_name>/`
Build task now outputs too the root in `/dist`

### Why did it change

We only want to deploy the code that's necessary for a lambda to work.

### Issue tracking

### Other considerations


dannyhayes-gds (approved):
I've tested some of the lambdas locally within localstack and things seem to be working fine there.

PR #77, opened 2022-01-17
PYI-540: Update Configuration Service to load Credential Issuers from…

Danomanic:
### What changed

Update Configuration Service to load Credential Issuers from Parameter Store using new path method rather than YAML.

This change will call parameter store with the specified credentialIssuerId and retried all the parameters from within that path. At the moment, this is tokenUrl and credentialUrl. It then maps the response to a CredentialIssuerConfig object to be used by the CredentialIssuerHandler.

In this ticket I have also resoled an issue with Spotbugs not running in the lambda gradle build files by moving the task to the root build.gradle

### Why did it change

To allow a better way to store the Credential Issuer config in Parameter Store.

### Issue tracking

- [PYI-540](https://govukverify.atlassian.net/browse/PYI-540)

ENVIROMENT = dev/qa etc.

### Other considerations


SunilDGDS (dismissed):

kerrr:
Can we split this into two commits and PRs, one for the gradle build changes and the other for the actual config lambda work. Thanks.

lib/build.gradle
```diff
@@ -1,34 +1,33 @@
 plugins {
-    // Apply the java-library plugin for API and implementation separation.
-    id 'java-library'
+	// Apply the java-library plugin for API and implementation separation.
```
kerrr: Can we stick with spaces please.

Danomanic: Done

Danomanic:
> Can we split this into two commits and PRs, one for the gradle build changes and the other for the actual config lambda work. Thanks.

This has been done! Will open a separate PR for the gradle changes.

SunilDGDS (dismissed):

SunilDGDS (approved):

PR #80, opened 2022-01-18
BAU: Pull spotless to root project

Wynndow:
**Review commit by commit.**
### What changed

Moved the spotless definition to the root gradle project
Added the java plugin to the root project
Ran spotlessApply on all files
Correctly sets the xml.required property for jacoco test reports

### Why did it change

Move the spotless config to the root project so we don't have to manage
it in a bunch of places.

Without the Java plugin the `spotlessApply` task is not automatically run when
the `check` task is run (which is in turn run by the build task). Without it we'd 
need to set up a hook to make sure that spotless is being run. This seems easier 
and less fragile.

Sets the `xml.required` property to true for jacoco reports which will fix the sonarqube
issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,10 @@
 plugins {
 	id "org.sonarqube" version "3.3"
+	id "com.diffplug.spotless" version "6.1.0"
+}
+
+repositories {
+	mavenCentral()
 }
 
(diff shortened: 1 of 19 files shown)


Danomanic (approved):
Looks good, and works!

PR #81, opened 2022-01-18
PYI-537 - Update for_each to get each object

TomTucka:
### What changed

stop used `toset()` and do a `for` over the object variable 

### Issue tracking

- [PYI-537](https://govukverify.atlassian.net/browse/PYI-537)


danworth (dismissed):
🏅 keeping to consistent use of camelCase 👍  I've noticed that the ids in our cri config are not in camel case so we should probably update those so that the full parameter path is camelCase 
https://github.com/alphagov/di-ipv-config/blob/main/credential-issuers/dev.yaml#L2

Danomanic (approved):

PR #84, opened 2022-01-18
WIP

Wynndow:
### Issue tracking

### Other considerations

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ sonarqube {
 spotless {
 	java {
 		target "**/src/**/*.java"
-		googleJavaFormat("1.13.0").aosp()
+		googleJavaFormat("1.13.0")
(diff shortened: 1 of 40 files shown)


kerrr:
Yeah, OK that's why we use AOSP format. Lets stick with 4 char indents.

Wynndow:
> Yeah, OK that's why we use AOSP format. Lets stick with 4 char indents.

@kerrr Out of interest, why do you prefer 4 char indents? 
We should also right this down so that in future people aren't wondering why we're using aosp.

kerrr:
Habit more than anything I think. Everywhere i've ever worked on java and every dev environment i've ever used has defaulted to 4char indents. 🤷🏼‍♂️ 
This is where some tab proponent butts in 🙄

PR #87, opened 2022-01-18
PYI-554: Add lambda to return all credentials

Wynndow:
### What changed

This is very similar to the useridentity lambda except without the token
check. This is because the request is coming from the same (our) system.

For now it returns some stub data, but can easily be updated to actually
chat to the database.

I've tried to import as few dependencies as possible, hence the use of
java.util.logging. Very much up for debate.

This doesn't deploy the thing.

### Issue tracking

- [PYI-554](https://govukverify.atlassian.net/browse/PYI-554)

### DIFF
--- a/lambdas/issuedcredentials/src/main/java/uk/gov/di/ipv/core/issuedcredentials/IssuedCredentialsHandler.java
+++ b/lambdas/issuedcredentials/src/main/java/uk/gov/di/ipv/core/issuedcredentials/IssuedCredentialsHandler.java
@@ -6,6 +6,7 @@ import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.helpers.ApiGatewayResponseGenerator;
+import uk.gov.di.ipv.core.library.helpers.RequestHelper;
 import uk.gov.di.ipv.core.library.service.UserIdentityService;
 
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


kerrr:
I wonder if we should be using the logger from `context.getLogger()`?

lambdas/issuedcredentials/src/main/java/uk/gov/di/ipv/core/issuedcredentials/IssuedCredentialsHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @ExcludeFromGeneratedCoverageReport
+    public IssuedCredentialsHandler() {
+        this.userIdentityService = new UserIdentityService();
+    }
+
+    @Override
+    public APIGatewayProxyResponseEvent handleRequest(
+            APIGatewayProxyRequestEvent input, Context context) {
+        String ipvSessionId = input.getBody();
```
kerrr: The session id should be in a header called `ipv-session-id`

lambdas/issuedcredentials/src/main/java/uk/gov/di/ipv/core/issuedcredentials/IssuedCredentialsHandler.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @ExcludeFromGeneratedCoverageReport
+    public IssuedCredentialsHandler() {
+        this.userIdentityService = new UserIdentityService();
+    }
+
+    @Override
+    public APIGatewayProxyResponseEvent handleRequest(
+            APIGatewayProxyRequestEvent input, Context context) {
+        Map<String, String> headers = input.getHeaders() == null ? Map.of() : input.getHeaders();
```
dannyhayes-gds: We have a RequestHelper.getHeaderByKey method that we can use to get the ipv-session-id header value. It makes sure we ignore the case of the header name when trying to read it.

kerrr: 👍🏼

dannyhayes-gds (approved):

PR #92, opened 2022-01-20
PYI-554: Add static initialiser to set http client

Wynndow:
Without this the lambda will throw horrible errors.

### Why did it change

Without this the lambda will throw horrible errors.


Danomanic (approved):
Looks good.

Side note, I think we can just have one instance of this in the `ConfigurationService` - this should mean we can remove this from all handers. I have yet to test this theory though.

SunilDGDS (approved):

Wynndow:
> Looks good.
> 
> Side note, I think we can just have one instance of this in the `ConfigurationService` - this should mean we can remove this from all handers. I have yet to test this theory though.

Oh, that'd be sweet. Having this everywhere is horrid.

PR #93, opened 2022-01-20
BAU - Temporary fix until we move to SAM

TomTucka:
### What changed

Added an IAM role to allow lambda access to parameters

### Why did it change

Lambda can't access new parameters created by module


terraform/lambda/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "credential_issuer_iam_role_id" {
+  description = "The ID of the IAM role used by the credential issuer lambda"
+  value       = module.credential-issuer.iam_role_id
+}
```
kerrr:
```suggestion
}

```

kerrr (dismissed):

kerrr (approved):

PR #96, opened 2022-01-20
BAU: add missing ssm params for new fields on cri config yaml file

dannyhayes-gds:
### What changed
Add new ssm params for the new fields added to the cri config yaml file.

### Why did it change
We require these new fields for use within the core-front app to be stored in SSM.


TomTucka (approved):
🚀

PR #99, opened 2022-01-20
Added policy arn as variable

TomTucka:


danworth (approved):
Looks ok but without being able to `tf plan` it its hard to be sure ....

PR #100, opened 2022-01-20
added value output

TomTucka:
### What changed
provide an output value

### Why did it change
need to provide output


danworth (approved):
🤞

PR #101, opened 2022-01-21
PYI-488 - Initial commit of SAM template

TomTucka:
### What changed

Created a SAM template to deploy core-back
Contains code to:

- Deploy an API Gateway
- Deploy each lambda
- Deploy each DynamoDB table
- Create and attach all relevant IAM permissions

### Why did it change

Programme decision to adopt [SAM for serverless technologies]()

### Issue tracking

- [PYI-488](https://govukverify.atlassian.net/browse/PYI-488)

### Other considerations

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1,0 +1,234 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: >
+  ipv-core-back
+
+Globals:
+  Function:
+    Timeout: 20
+
+Parameters:
+  Environment:
+    Type: String
+
+Resources:
+
+  IPVCoreAPI:
+    Type: AWS::Serverless::Api
+    Properties:
+      Name: IPV Core API Gateway
+      StageName: !Sub ${Environment}
+
+  IPVAccessTokenFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub "ipv-access-token-${Environment}"
+      Handler: uk.gov.di.ipv.core.accesstoken.AccessTokenHandler::handleRequest
+      Runtime: java11
+      PackageType: Zip
+      CodeUri: ../lambdas/accesstoken
+      Architectures:
+        - x86_64
+      MemorySize: 512
+      Environment:
+        Variables:
+          AUTH_CODES_TABLE_NAME: !Select [1, !Split ['/', !GetAtt AuthCodesTable.Arn]]
+          ACCESS_TOKENS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt AccessTokensTable.Arn]]
+      Policies:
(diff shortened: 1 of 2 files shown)


deploy/template.yaml
```diff
@@ -0,0 +1,290 @@
...
+      Handler: uk.gov.di.ipv.core.accesstoken.AccessTokenHandler::handleRequest
+      Runtime: java11
+      PackageType: Zip
+      CodeUri: ../lambdas/accesstoken
+      Architectures:
+        - x86_64
+      MemorySize: 512
+      Environment:
+        Variables:
+          AUTH_CODES_TABLE_NAME: !Select [1, !Split ['/', !GetAtt AuthCodesTable.Arn]]
```
danworth, line 35: I guess the DynamoDB resource doesn't output the table name so we need these shenanigans. Does SAM have something like Terraform locals where we might be able to specify the table names once and then use them throughout?

deploy/template.yaml
```diff
@@ -0,0 +1,290 @@
...
+          ACCESS_TOKENS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt AccessTokensTable.Arn]]
+      Policies:
+        - DynamoDBCrudPolicy:
+            TableName: !Ref AccessTokensTable
```
danworth, line 39: It looks like this returns the table name? can it be used in place of `!Select [1, !Split ['/', !GetAtt AuthCodesTable.Arn]]`?

deploy/template.yaml
```diff
@@ -0,0 +1,290 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: >
+  ipv-core-back
```
danworth, line 4: We might want to expand the description if only to help new joiners etc

danworth (approved):
As I understand this provides a like-for-like replacement of the same resources managed by the existing Terraform modules. I haven't done a cross reference with those Terraform modules but if that would be useful I can. Further testing of the deployment is planned but this is a great start of getting our infra described and managed by SAM 👍  🏅 

I have limited experience with SAM/CloudFormation but this looks good to me, a couple of questions but they can be addressed in future iterations if necessary.

deploy/template.yaml
TomTucka: From what I understand this is doing some magic in the background as you specify `TableName` so It wouldn't work where try specifying the table name for environment variables

deploy/template.yaml
TomTucka: This is what i was looking for at first, But I couldn't find anything that seemed to do the same thing as Terraform Locals

deploy/template.yaml
TomTucka: Yep, We should probably expand what this repo is in the readme as well with explanations for each lambda

PR #102, opened 2022-01-21
BAU: Move instantiation of AWS HttpClient into ConfigurationService to avoid duplicate code in Handlers.

Danomanic:
### What changed

Setting the HttpClient on the Builder rather than using a static method in each class.

### Why did it change

To avoid duplication

### DIFF
--- a/lambdas/accesstoken/src/main/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandler.java
+++ b/lambdas/accesstoken/src/main/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandler.java
@@ -35,13 +35,6 @@ public class AccessTokenHandler
     private final AuthorizationCodeService authorizationCodeService;
     private final ConfigurationService configurationService;
 
-    static {
-        // Set the default synchronous HTTP client to UrlConnectionHttpClient
-        System.setProperty(
-                "software.amazon.awssdk.http.service.impl",
-                "software.amazon.awssdk.http.urlconnection.UrlConnectionSdkHttpService");
-    }
-
     public AccessTokenHandler(
             AccessTokenService accessTokenService,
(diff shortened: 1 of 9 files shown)


danworth:
I get the desire to remove the duplication. I'm not entirely sure if the right place for it is the `ConfigurationService.class` because this setting is also needed for the `DataStore.class` which creates a dynamoDb client and no longer depends on the `ConfigurationService.class` so they're somewhat independent. I'm thinking that we then have the `DataStore.class` reliant on a setting being set by a class it has no way of knowing whether it is even being instantiated elsewhere in the lambda. An alternative to avoid the duplication at each handler level is to explicitly set the Http client implementation when creating the client e.g. https://github.com/alphagov/di-ipv-cri-uk-passport-back/blob/6fba3a63f788bdc115bb262becc8d022777a962d/src/main/java/uk/gov/di/ipv/cri/passport/persistence/DataStore.java#L36 and then each class requiring a http client implementation is not reliant on settings being set elsewhere and which could change independently.

That said I'm probably overthinking this, the tests pass and every handler is going to instantiate a `ConfigurationService` and so it's reasonably safe to presume that the setting will be set for any other class that requires it - that said I think this will break if we add IT tests that only instantiate the `DataStore.class` and so the settings in the `ConfigurationService.class` are not set as we've done in passport-back https://github.com/alphagov/di-ipv-cri-uk-passport-back/blob/6fba3a63f788bdc115bb262becc8d022777a962d/src/integration-test/java/uk/gov/di/ipv/cri/passport/integrationtest/DataStoreIT.java#L24 ... but then keeping them where they are now would equally not work for such a test.

TLDR; I'm not entirely sure about putting it in ConfigurationService.class but happy to be convinced/chat/someone else to approve... 👍

danworth (approved):
Looks good to me, the client creation no longer depends on settings which may or may not have been set elsewhere

PR #103, opened 2022-01-21
BAU: Refactor Services and Handlers so that ConfigurationService is only instantiated once per handler.

Danomanic:
### What changed

Refactor Services and Handlers so that ConfigurationService is only instantiated once per handler.

### Why did it change

This reduces the number of times we instantiate Configuration Service - in some cases 6 instantiations for one lambda call.


lib/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -7,7 +7,6 @@
 import software.amazon.awssdk.enhanced.dynamodb.model.QueryConditional;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
-import uk.gov.di.ipv.core.library.service.ConfigurationService;
```
danworth, line 10: I like that this is no longer given the entire ConfigurationService when it only needs a single value 👍

danworth (approved):
I think this is a good change and begins to untangle the ever growing configuration service form everything else. It already removes passing the Configuration Service to the DataStore class when it only needed a single value 👍  As you say I think this will help when we come to refactor the Configuration Service 🏅

PR #106, opened 2022-01-25
BAU: Pass the environment variable required by lambda

TomTucka:
### What changed

Pass `ENVIRONMENT` as a variable into the Credential Issuer Config Lambda 

### Why did it change

`ENVIRONMENT` is required


danworth (approved):
👍

PR #107, opened 2022-01-25
Pass the ENVIRONMENT environment var to all lambdas

TomTucka:
### What changed

pass `ENVIRONMENT` to all lambdas

### Why did it change

 `ENVIRONMENT` will be needed by most lambdas


danworth (approved):
👍

PR #109, opened 2022-01-26
Add .sdkmanrc

philandstuff:
sdkman supports a checked-in .sdkmanrc file to ensure we're all using
the same JDK version.

@kerrr informed me that we're using Azul Zulu for Java 11 so this is
what I've put in the file.


TomTucka (dismissed):

Danomanic (dismissed):
Brilliant idea!

kerrr (approved):

PR #111, opened 2022-02-01
PYI-570: Update IPV Core Back to Provide Shared Attribute Set as signed JWT

Danomanic:
### What changed
Update IPV Core Back to Provide Shared Attribute Set as signed JWT

Added a JwtHelper to sign the SharedAttributes payload with the Shared Attributes Signing Key stored in KMS.

### Issue tracking

- [PYI-570](https://govukverify.atlassian.net/browse/PYI-570)

- [x] Documented in the [README](./blob/main/README.md)
- [x] Added to deployment repository

### Other considerations


danworth (approved):
This looks good, we think there's a need for integration tests with KMS but our Integration Test strategy is in development so we can add later on. We have run this against the Dev account and it works as expected.

PR #112, opened 2022-02-01
PYI-612: Add shared attributes parameters

Wynndow:
### What changed

This renames the "credential-issuers" module to "parameters" to reflect
that it now contains the terraform for the credential issuers parameters
as well as other stuff.

It also adds some new parameters for the shared attributes signing key
and cert and adds a policy to the SAM template for the shared attributes
function.

It might the case that management of all ssm parameters will be moving
to SAM. In the interest of keeping things in the same place for now I've
defined these params in the terraform, rather than in SAM. This might
make the transition easier if it does happen. If it doesn't, then at
least we're being consistent.

### Issue tracking

- [PYI-612](https://govukverify.atlassian.net/browse/PYI-612)

### Other considerations


deploy/template.yaml
```diff
@@ -37,6 +37,10 @@ Resources:
       Policies:
         - DynamoDBReadPolicy:
             TableName: !Ref UserIssuedCredentialsTable
+        - SSMParameterReadPolicy:
```
danworth, line 40: Great stuff, thanks 👍

terraform/modules/parameters/shared-attributes.tf
```diff
@@ -0,0 +1,17 @@
+resource "aws_ssm_parameter" "shared_attributes_signing_cert" {
+  name        = "/${var.environment}/ipv/core/sharedAttributes/signingCert"
+  description = "The IPV core's shared attributes signing certificate"
```
danworth, line 3: We could add something like `This is used by CRIs to validate the shared attributes JWT signature`

danworth (approved):
Looks good 👍  think this is co-dependent upon https://github.com/alphagov/di-ipv-config/pull/123#pullrequestreview-869116620 so cannot plan until that's merged but it looks good to me

PR #113, opened 2022-02-01
PYI-612: Use the correct key attribute

Wynndow:
### What changed

We need to use the id of the key, not the key_id we set in the data
resources. Currently it just sets the value of the parameter to the keys
alias. Confusing.

See the TF docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/kms_key#attributes-reference

### Issue tracking

- [PYI-612](https://govukverify.atlassian.net/browse/PYI-612)


danworth (approved):
looks good, they could have chosen some better names for the attributes, how is key_id not the key id :)

PR #117, opened 2022-02-03
PYI-640: Use new parameter store config structure

Wynndow:
**Companion PR to do the restructuring: https://github.com/alphagov/di-ipv-config/pull/130**

### What changed

The layout of the config in the parameter store has changed. This
updates the templates to reflect it.

It also updates the configuration service to format the location of each
individual credential issuers config with the new format.

Some lambdas also had permissions on params they didn't need, which have
been removed.

### Issue tracking

- [PYI-640](https://govukverify.atlassian.net/browse/PYI-640)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -34,14 +34,12 @@ Resources:
         Variables:
           ENVIRONMENT: !Sub "${Environment}"
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsTable.Arn]]
-          SHARED_ATTRIBUTES_SIGNING_KEY_ID_PARAM: !Sub "/${Environment}/ipv/core/sharedAttributes/signingKeyId"
+          SHARED_ATTRIBUTES_SIGNING_KEY_ID_PARAM: !Sub "/${Environment}/core/self/sharedAttributesJwtSigningKeyId"
       Policies:
         - DynamoDBReadPolicy:
             TableName: !Ref UserIssuedCredentialsTable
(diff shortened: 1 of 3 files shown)


deploy/template.yaml
```diff
@@ -34,14 +34,14 @@ Resources:
...
           ENVIRONMENT: !Sub "${Environment}"
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsTable.Arn]]
-          SHARED_ATTRIBUTES_SIGNING_KEY_ID_PARAM: !Sub "/${Environment}/ipv/core/sharedAttributes/signingKeyId"
+          SHARED_ATTRIBUTES_SIGNING_KEY_ID_PARAM: !Sub "/${Environment}/core/self/sharedAttributesJwtSigningKeyId"
       Policies:
         - DynamoDBReadPolicy:
             TableName: !Ref UserIssuedCredentialsTable
         - SSMParameterReadPolicy:
-            ParameterName: !Sub ${Environment}/ipv/core/sharedAttributes/*
+            ParameterName: !Sub ${Environment}/core/self/*
```
danworth: In the spirit of having small lambdas each with limited scope of what it can access etc, then strictly speaking this lambda only needs to access `/${Environment}/core/self/sharedAttributesJwtSigningKeyId`, at present I don't think there's much else under that path but in future secure strings could be added for other purposes and perhaps worth considering tightening the scope?

Wynndow: Yep, that makes sense to me! Will fix it up

danworth (approved):
🏅

PR #118, opened 2022-02-04
PYI-639: Increase core-back lambda timeout size

dannyhayes-gds:
### What changed
Increase lambda timeouts to be 40 seconds

### Why did it change
We are currently reaching the 20 second limit due to cold starts

### Issue tracking

- [PYI-639](https://govukverify.atlassian.net/browse/PYI-639)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -5,7 +5,7 @@ Description: >
 
 Globals:
   Function:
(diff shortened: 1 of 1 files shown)


Danomanic (approved):
Agree this should be increased to 40

PR #119, opened 2022-02-04
PYI-638: Fix shared attributes deserializer by not looking for names …

dannyhayes-gds:
…attribute

### What changed
Updated shared attributes deserialiser to not expect the name properties to be under a "names" field.

### Why did it change
The cri response does not structure the names properties under a "names" field so the current deserialising fails.

### Issue tracking

- [PYI-638](https://govukverify.atlassian.net/browse/PYI-638)


Danomanic (approved):
Looks good.

PR #120, opened 2022-02-07
BAU: Fix cred issuer config env var prefix

Wynndow:
### What changed

The credential issuer config lambda has fallen over due to us trying to
fetch an invalid parameter name. This should fix it up.


danworth (approved):
nice fix 👍

PR #122, opened 2022-02-07
BAU: Improve logging for failed requests

Wynndow:
### What changed

We had an issue with the wrong config being set for the credential
issuers. The 404 being returned in the httpResponse was causing the
error logging to just return "null: null". This adds some more details
to make it easier to work out what had gone wrong.

It's a bit wordy but it'll be helpful if things go wrong again.


SunilDGDS (approved):
looks good

PR #123, opened 2022-02-08
PYI-660 Developer SAM Config

TomTucka:
### What changed

Added config for each dev into `samconfig.toml`
Added basic readme with instructions and reference commands

### Why did it change

We want developers to be able to deploy lambdas into a real AWS account while developing them.

### Issue tracking

- [PYI-660](https://govukverify.atlassian.net/browse/PYI-660)

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,0 +1,35 @@
+# Deploy
+
+## Developer Deployment
+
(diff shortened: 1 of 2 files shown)


danworth (approved):
👍

PR #126, opened 2022-02-10
PYI-680: Validate callback url in token endpoint

Wynndow:
**Supersedes https://github.com/alphagov/di-ipv-core-back/pull/125**

### What changed

* Validates the redirect URL provided by the user in the authorisation endpoint by comparing it to the endpoints registered in the clients configuration.
* Validates the redirect URL provided by the client in the auth grant received by the token endpoint, is the same as that received on the authorisation endpoint.
* Refactors the authorisation request validation to a new class.

### Why did it change

OAuth 2 spec made me do it...

### Issue tracking

- [PYI-680](https://govukverify.atlassian.net/browse/PYI-680)

### DIFF
--- a/lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
+++ b/lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
@@ -19,13 +19,14 @@ import com.nimbusds.oauth2.sdk.token.Tokens;
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.ipv.core.library.persistence.item.AuthorizationCodeItem;
 import uk.gov.di.ipv.core.library.service.AccessTokenService;
 import uk.gov.di.ipv.core.library.service.AuthorizationCodeService;
 import uk.gov.di.ipv.core.library.service.ConfigurationService;
 import uk.gov.di.ipv.core.library.validation.ValidationResult;
 
 import java.util.Map;
-import java.util.UUID;
+import java.util.Optional;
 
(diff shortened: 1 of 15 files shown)


lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
```diff
@@ -166,6 +171,33 @@ void shouldReturn400OWhenInvalidAuthorisationCodeProvided() throws Exception {
         assertEquals(OAuth2Error.INVALID_GRANT.getDescription(), errorResponse.getDescription());
     }
 
+    @Test
+    void shouldReturn400WhenRedirectUrlInAuthGrantDoesNotMatchThatSuppliedToAuthEndpoint()
```
Danomanic, line 181: Would it be worth adding a quick test to see what happens when `redirect_uri` isn't in the request? Not sure if it would cause a null pointer exception in `redirectUrlsDoNotMatch()`

lib/src/main/java/uk/gov/di/ipv/core/library/validation/AuthRequestValidator.java
```diff
@@ -0,0 +1,86 @@
+package uk.gov.di.ipv.core.library.validation;
```
Danomanic, line 1: I like this a lot. Much better way to validate the request. Maybe something we could use on other Handlers.

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
Wynndow: Nice, good spot. I've pushed up a commit to address it.

dannyhayes-gds (approved):

PR #127, opened 2022-02-14
read Jwt from accesstokenhandler

SunilDGDS:
### What changed

The accesstoken handler now receives a claimsetJwt in the response body ; claimset identified by 'cleint_assertion' key with the value as the signed jwt . This jwt is extracted and read only . 

- [PYI-683](https://govukverify.atlassian.net/browse/PYI-683)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
@@ -9,8 +9,6 @@ import com.nimbusds.oauth2.sdk.TokenResponse;
 import com.nimbusds.oauth2.sdk.token.AccessToken;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import com.nimbusds.oauth2.sdk.token.Tokens;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.persistence.DataStore;
 import uk.gov.di.ipv.core.library.persistence.item.AccessTokenItem;
@@ -21,8 +19,6 @@ import java.util.Objects;
 public class AccessTokenService {
     private final DataStore<AccessTokenItem> dataStore;
     private final ConfigurationService configurationService;
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(AccessTokenService.class);
 
     @ExcludeFromGeneratedCoverageReport
     public AccessTokenService(ConfigurationService configurationService) {
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
@@ -68,23 +68,4 @@ public class AccessTokenService {
         accessTokenItem.setIpvSessionId(ipvSessionId);
         dataStore.create(accessTokenItem);
     }
-
-    /*public ValidationResult<ErrorObject> extractJwt(String requestBody) {
-
-        Map<String, String> stringMap = RequestHelper.parseRequestBody(requestBody);
-
-        try {
-            SignedJWT clientJwt = SignedJWT.parse(String.valueOf(stringMap.get("client_assertion")));
-
-            JWTClaimsSet claimsSet = clientJwt.getJWTClaimsSet();
-
-            if (claimsSet != null) {
-                return ValidationResult.createValidResult();
-            }
-        } catch (ParseException e) {
-            LOGGER.error("Unable to parse Claims set {} ", e.getMessage());
-            return new ValidationResult<>(false, OAuth2Error.INVALID_CLIENT);
-        }
-        return ValidationResult.createValidResult();
-    }*/
 }
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
@@ -1,5 +1,7 @@
 package uk.gov.di.ipv.core.library.service;
 
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.AccessTokenResponse;
 import com.nimbusds.oauth2.sdk.ErrorObject;
 import com.nimbusds.oauth2.sdk.GrantType;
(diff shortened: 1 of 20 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
```diff
@@ -64,4 +73,23 @@ public void persistAccessToken(AccessTokenResponse tokenResponse, String ipvSess
         accessTokenItem.setIpvSessionId(ipvSessionId);
         dataStore.create(accessTokenItem);
     }
+
+    public ValidationResult<ErrorObject> extractJwt(String requestBody) {
```
dannyhayes-gds: It might be nicer to move this kind of thing into a Validator class, similar to the AuthRequestValidator that we have now?

dannyhayes-gds: Also might be worth renaming this method, it's starting to do a bit more then just extract the JWT from the query params. Especially if the extra validation needed in ticket PYI-684 is going into here as well

lib/src/test/java/uk/gov/di/ipv/core/library/service/AccessTokenServiceTest.java
```diff
@@ -155,4 +156,38 @@ void shouldReturnNullWhenInvalidAccessTokenProvided() {
...
+                "code=12345&client_assertion=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjM0IiwiYXVkIjoiYWRtaW4iLCJpc3MiOiJtYXNvbi5tZXRhbXVnLm5ldCIsImV4cCI6MTU3NDUxMjc2NSwiaWF0IjoxNTY2NzM2NzY1LCJqdGkiOiJmN2JmZTMzZi03YmY3LTRlYjQtOGU1OS05OTE3OTliNWViOGEifQ==.EVcCaSqrSNVs3cWdLt-qkoqUk7rPHEOsDHS8yejwxMw&redirect_uri=http://test.com&grant_type=authorization_code&client_id=test_client_id";
+
+        ValidationResult<ErrorObject> extractJwt =
+                accessTokenService.extractJwt(tokenRequestBody);
+        assertEquals(extractJwt.isValid(),true);
+    }
+
+    @Test
+    void shouldReadClaimSetProvidedAndError(){
+        // second part (claimset is blank in the string)
```
dannyhayes-gds: Are these comments needed?

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
```diff
@@ -64,4 +73,23 @@ public void persistAccessToken(AccessTokenResponse tokenResponse, String ipvSess
...
+
+        Map<String, String> stringMap = RequestHelper.parseRequestBody(requestBody);
+
+        try {
+            SignedJWT str = SignedJWT.parse(String.valueOf(stringMap.get("client_assertion")));
```
dannyhayes-gds: Not sure what str is supposed to stand for? I'm guessing String maybe? If so then could we rename this to be something like jwt or clientJwt

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
```diff
@@ -64,4 +73,23 @@ public void persistAccessToken(AccessTokenResponse tokenResponse, String ipvSess
...
+
+            JWTClaimsSet claimsSet = str.getJWTClaimsSet();
+
+            if (claimsSet != null) {
+                return ValidationResult.createValidResult();
+            }
+        } catch (ParseException e) {
+            LOGGER.error("Unable to parse Claims set {} ", e.getMessage());
+            return new ValidationResult<>(false, OAuth2Error.INVALID_REQUEST_OBJECT);
```
dannyhayes-gds: There is an OAuth2Error.INVALID_CLIENT error object which might be better suited here?

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
SunilDGDS: move extractJwt method to AuthRequestValidator class

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
SunilDGDS: renamed to clientJwt

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
SunilDGDS: used INVALID_CLIENT

lib/src/test/java/uk/gov/di/ipv/core/library/service/AccessTokenServiceTest.java
SunilDGDS: Removed

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
dannyhayes-gds: Sorry I didn't mean the AuthRequestValidator itself as that one is for validation against our /authorisation endpoint. I meant to create a new validator i.e TokenRequestValidator that can do the validation for the /token endpoint

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
```diff
@@ -64,4 +68,23 @@ public void persistAccessToken(AccessTokenResponse tokenResponse, String ipvSess
         accessTokenItem.setIpvSessionId(ipvSessionId);
         dataStore.create(accessTokenItem);
     }
+
+    /*public ValidationResult<ErrorObject> extractJwt(String requestBody) {
```
dannyhayes-gds: I'm guessing this comment block can be removed now?

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
SunilDGDS: New class TokenRequestValidator added

lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
SunilDGDS: Removed comment

lib/src/test/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+        assertEquals(true, extractJwt.isValid());
+    }
+
+    @Test
+    void shouldReadClaimSetProvidedAndError() {
+        String tokenRequestBody =
+                "code=12345&client_assertion=&redirect_uri=http://test.com&grant_type=authorization_code&client_id=test_client_id";
+
+        ValidationResult<ErrorObject> extractJwt = validator.validateExtractedJwt(tokenRequestBody);
+        assertEquals(true, !extractJwt.isValid());
```
dannyhayes-gds: I think this should be changed. It makes it a bit harder to read if your altering the value of the isValid return here.

It will be neater to change to either of the following

```suggestion
        assertEquals(false, extractJwt.isValid());
```
or
```suggestion
        assertFalse(extractJwt.isValid());
```

dannyhayes-gds: And same thing for the test below

SunilDGDS: applied the change

dannyhayes-gds (approved):

PR #131, opened 2022-02-16
PYI-702: Add new ipvClientJwtExpirationTime variable to the CredentialIssuerConfig class

dannyhayes-gds:
### What changed
Added new ipvClientJwtExpirationTime variable to the CredentialIssuerConfig class.

### Why did it change
New variable will be used when generating the ipv-core client JWT and setting its expiry time claim value.

### Issue tracking

- [PYI-702](https://govukverify.atlassian.net/browse/PYI-702)


danworth (approved):
Would it make it easier to use if the type was `Long` rather than `String`, seems the jwt set expiry accepts a `long`? Out of curiosity I changed it to `Long` and the tests all pass so I presume the mapping from ParameterStore string via the ObjectMapper and the serialisation to Json works without further changes. I guess it depends on how it'll be used which you know best so I'll leave it to you 👍

PR #132, opened 2022-02-17
Pyi 685

SunilDGDS:
### What changed
In access token handler, a jwt claimset is received as part of the request body . This claimset is read and validated against configuration values coming from parameter store .

- [PYI-685](https://govukverify.atlassian.net/browse/PYI-685)

### Other considerations

The clientid from the claimset is used to read the configuration .  It is possible that the client does not need to be authenticated ('ATHENTICATION_METHOD set to none in the config) , therefore not receiving the claimset as part of the response body in this case . How do we know that can we read the configuration in advance to check if the client's authentification method ?

### DIFF
--- a/lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
+++ b/lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
@@ -12,6 +12,7 @@ import com.nimbusds.oauth2.sdk.GrantType;
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.TokenResponse;
+import com.nimbusds.oauth2.sdk.auth.verifier.ClientAuthenticationVerifier;
 import com.nimbusds.oauth2.sdk.http.HTTPResponse;
 import com.nimbusds.oauth2.sdk.token.AccessToken;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
@@ -19,6 +20,8 @@ import com.nimbusds.oauth2.sdk.token.Tokens;
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.exceptions.ClientAuthenticationException;
 import uk.gov.di.ipv.core.library.persistence.item.AuthorizationCodeItem;
 import uk.gov.di.ipv.core.library.service.AccessTokenService;
 import uk.gov.di.ipv.core.library.service.AuthorizationCodeService;
@@ -32,6 +35,7 @@ import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -44,6 +48,7 @@ class AccessTokenHandlerTest {
     private AccessTokenService mockAccessTokenService;
     private AuthorizationCodeService mockAuthorizationCodeService;
     private TokenRequestValidator mockTokenRequestValidator;
+    private ClientAuthenticationVerifier<Object> verifier;
 
     private AccessTokenHandler handler;
     private TokenResponse tokenResponse;
@@ -91,8 +96,6 @@ class AccessTokenHandlerTest {
         when(mockAuthorizationCodeService.getAuthorizationCodeItem("12345"))
                 .thenReturn(Optional.of(authorizationCodeItem));
         when(mockAccessTokenService.validateTokenRequest(any()))
-                .thenReturn(ValidationResult.createValidResult());
-        when(mockTokenRequestValidator.validateExtractedJwt(any()))
                 .thenReturn(ValidationResult.createValidResult());
 
         APIGatewayProxyResponseEvent response = handler.handleRequest(event, context);
@@ -176,8 +179,6 @@ class AccessTokenHandlerTest {
 
         when(mockAccessTokenService.validateTokenRequest(any()))
                 .thenReturn(ValidationResult.createValidResult());
-        when(mockTokenRequestValidator.validateExtractedJwt(any()))
-                .thenReturn(ValidationResult.createValidResult());
         when(mockAuthorizationCodeService.getAuthorizationCodeItem("12345"))
                 .thenReturn(Optional.empty());
 
@@ -204,22 +205,21 @@ class AccessTokenHandlerTest {
         APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
         event.setBody(tokenRequestBody);
 
-        ValidationResult validationResultError =
-                new ValidationResult<>(false, OAuth2Error.UNSUPPORTED_GRANT_TYPE);
+        ValidationResult<ErrorResponse> validationResultError =
+                new ValidationResult<>(false, ErrorResponse.INVALID_REQUEST_PARAM);
 
         when(mockAccessTokenService.validateTokenRequest(any()))
                 .thenReturn(ValidationResult.createValidResult());
 
-        when(mockTokenRequestValidator.validateExtractedJwt(any()))
-                .thenReturn(validationResultError);
+        doThrow(new ClientAuthenticationException("error"))
+                .when(mockTokenRequestValidator)
+                .authenticateClient(any(), any());
 
         APIGatewayProxyResponseEvent response = handler.handleRequest(event, context);
         ErrorObject errorResponse = createErrorObjectFromResponse(response.getBody());
         assertEquals(HttpStatus.SC_BAD_REQUEST, response.getStatusCode());
-        assertEquals(OAuth2Error.UNSUPPORTED_GRANT_TYPE.getCode(), errorResponse.getCode());
-        assertEquals(
-                OAuth2Error.UNSUPPORTED_GRANT_TYPE.getDescription(),
-                errorResponse.getDescription());
+        assertEquals(OAuth2Error.INVALID_GRANT.getCode(), errorResponse.getCode());
+        assertEquals(OAuth2Error.INVALID_GRANT.getDescription(), errorResponse.getDescription());
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
```diff
@@ -204,22 +205,21 @@ void shouldReturn400WhenInvalidJwtProvided() throws Exception {
         APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
         event.setBody(tokenRequestBody);
 
-        ValidationResult validationResultError =
-                new ValidationResult<>(false, OAuth2Error.UNSUPPORTED_GRANT_TYPE);
+        ValidationResult<ErrorResponse> validationResultError =
+                new ValidationResult<>(false, ErrorResponse.INVALID_REQUEST_PARAM);
```
Wynndow: I think this can be dropped - doesn't seem to be used.

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
```diff
@@ -230,22 +230,21 @@ void shouldReturn400WhenJwtMissingFromRequestProvided() throws Exception {
         APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
         event.setBody(tokenRequestBody);
 
-        ValidationResult validationResultError =
-                new ValidationResult<>(false, OAuth2Error.INVALID_REQUEST_OBJECT);
+        ValidationResult<ErrorResponse> validationResultError =
+                new ValidationResult<>(false, ErrorResponse.INVALID_REQUEST_PARAM);
```
Wynndow: Same here, I don't think it's used.

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -1,37 +1,118 @@
...
+
+    private String getClientAuthenticationMethod(String clientId)
+            throws ClientAuthenticationException {
+        String clientAuthenticationMethod =
+                configurationService.getClientAuthenticationMethod(clientId);
+        if (clientAuthenticationMethod == null) {
+            LOGGER.error("Failed to find config for client ID: {}", clientId);
+            throw new ClientAuthenticationException(
+                    String.format("Config for client ID '%s' not found", clientId));
         }
```
Wynndow: The ssmProvider in the config service will throw a runtime error if it can't find a value in the parameter store, so i think this block can be safely dropped.

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -1,37 +1,118 @@
...
-        return ValidationResult.createValidResult();
+        return clientAuthenticationMethod;
+    }
+
+    private void validateExpiration(JWTAuthenticationClaimsSet claimsSet)
```
Wynndow: Could this be called something else to make it obvious that it's validating the expiry isn't too far in the future?

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -1,37 +1,118 @@
...
+        OffsetDateTime offsetDateTime =
+                OffsetDateTime.now().plusSeconds(Long.parseLong(maxAllowedTtl));
+        if (expirationTime.getTime() / 1000L > offsetDateTime.toEpochSecond()) {
+            LOGGER.error("Client JWT expiry date is too far in the future");
+            throw new InvalidClientException(
+                    "The client JWT expiry date has surpassed the maximum allowed ttl value");
+        }
+    }
+
+    public static ClientAuthenticationVerifier<Object> getClientAuthVerifier(
```
Wynndow: As this is being called in the constructor I don't think it needs to be static. I think it was originally intended to be called externally and the return value passed into the constructor to allow injection. As we're not doing that I reckon just make it an instant method.

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
```diff
@@ -184,6 +189,13 @@ public Certificate getClientCertificate(String clientId) throws CertificateExcep
                         System.getenv("ENVIRONMENT"), clientId));
     }
 
+    public String getClientAuthenticationMethod(String clientId) {
+        return getParameterFromStore(
```
Wynndow: Should we just use `ssmProvider.get()` rather than this wrapper method?

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -1,37 +1,118 @@
...
+                    throw new ClientAuthenticationException(
+                            String.format(
+                                    "Missing client_assertion jwt for configured client '%s'",
+                                    clientId));
+                }
             }
-        } catch (ParseException e) {
-            LOGGER.error("Unable to parse Claims set {} ", e.getMessage());
-            return new ValidationResult<>(false, OAuth2Error.INVALID_CLIENT);
+        } else {
```
Wynndow: This is quite hard to follow through all the branches - could it be refactored to a well name private method?

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
dannyhayes-gds: Yeah good catch 👍

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
dannyhayes-gds: Yeah sounds good, I'll change this.

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
dannyhayes-gds: We can do? I think I mostly just followed the pattern of some of the other methods we have doing it like this. But I agree its not really adding much with the wrapper method

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
dannyhayes-gds: Yeah good shout, I'll refactor this slightly

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
SunilDGDS: Removed

lambdas/accesstoken/src/test/java/uk/gov/di/ipv/core/accesstoken/AccessTokenHandlerTest.java
SunilDGDS: Removed

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -40,59 +40,59 @@ public TokenRequestValidator(ConfigurationService configurationService) {
     public void authenticateClient(String requestBody, Map<String, String> queryParams)
             throws ClientAuthenticationException {
         if (!queryParams.containsKey(CLIENT_ASSERTION_PARAM)) {
```
Wynndow: Just to be super picky, could this be:
```
if (queryParams.containsKey(CLIENT_ASSERTION_PARAM)) {
    authenticateClientWithJwt(requestBody);
else {
    authenticateClientWithoutJwt(queryParams);
}
```

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -40,59 +40,59 @@ public TokenRequestValidator(ConfigurationService configurationService) {
...
-
-                clientId = clientJwt.getClientID().getValue();
+            authenticateClientWithJwt(requestBody);
+        }
+    }
 
-                String clientAuthenticationMethod = getClientAuthenticationMethod(clientId);
+    private void authenticateClientWithoutJwt(Map<String, String> queryParams)
+            throws ClientAuthenticationException {
+        if (!queryParams.containsKey(CLIENT_ID_PARAM)) {
```
Wynndow: Again, could we use `if (queryParams.containsKey(CLIENT_ID_PARAM))` and switch the blocks?

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
SunilDGDS: refactor to catch ssmProvider exception so that tests do not fall over

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
SunilDGDS: renamed

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
SunilDGDS: change to concrete

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
SunilDGDS: changed to use ssmProvider

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
SunilDGDS: Refactor into 2 distincts smaller method call

lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
dannyhayes-gds: 👍  yeah agree that would be nicer

Wynndow (approved):

PR #136, opened 2022-02-21
BAU: Remove Optionals from configuration service

Wynndow:
### Why did it change

The ssmProvider doesn't return null if it can't find a parameter.
Instead, it will throw a `ParameterNotFoundException` which is a run
time exception.


Danomanic (approved):
I also noticed this on Friday. Good change and refactor.

PR #143, opened 2022-02-22
OpenAPI specifications for our internal and external backend APIs

kerrr:
### Issue tracking

### Other considerations

### DIFF
--- a/openAPI/core-back-external.yaml
+++ b/openAPI/core-back-external.yaml
@@ -1,0 +1,31 @@
+openapi: 3.0.3
+info:
+  title: IPV Core External
+  description: The external api presented by IPV Core Back for use by non IPV Core services
+  version: 1.0.0
+servers:
+  - url: 'https://example.com'
+paths:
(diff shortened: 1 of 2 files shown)


dannyhayes-gds:
I think the only potential thing missing here is an endpoint on the internal gateway for core-front to use to retrieve a list of the current credentials that have been retrieved for the current user journey in order to display some minimal representation of what has already been retrieved from each cri on the core-front debug page.

Currently this is our "issued-credentials" lambda under the /issued-credentials endpoint which could definitely do with a better name. Perhaps this could fit under a new path of something like /debug/credentials. And anything under the /debug/ path could be for things that are only related to our non-production environments etc?

dannyhayes-gds:
Also I wasn't sure where our shared-attributes lambda fits into this? I guess it would occur during the /event/cri/start endpoint so I imagine it would belong there but with a better lambda name as it might be doing a bit more than just the shared attributes stuff?

PR #147, opened 2022-02-22
BAU: Update lambda permission and refactor param name for max ttl for…

dannyhayes-gds:
… client auth jwts

### What changed

- Refactor param name to be clearer what it is used for.
- Add missing lambda permission to read the param from SSM.


Wynndow (approved):

Danomanic (approved):
Looks good.

PR #148, opened 2022-02-24
SPIKE: Create Lambda handler to remove Configuration Service dependency.

Danomanic:
Create Lambda handler to remove Configuration Service dependency. Also remove `lib` dependency from gradle and have domain objects within the handler module.


lambdas/experimentcriconfig/src/main/java/uk/gov/di/ipv/core/experimentcriconfig/ExperimentCRIConfigHandler.java
```diff
@@ -0,0 +1,139 @@
...
+        implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+
+    public static final int LOCALHOST_PORT = 4567;
+    private static final String LOCALHOST_URI = "http://localhost:" + LOCALHOST_PORT;
+    private static final String IS_LOCAL = "IS_LOCAL";
+    public static final String CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX =
+            "CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX";
+
+    private final SSMProvider ssmProvider;
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
Wynndow, line 39: In the interests of going quick, is it worth trying to use Gson rather than Jackson here? Something something reflection.

Danomanic: Good idea! Will implement :)

Wynndow: Hmm. It might mean pulling in a new dependency though. I wonder what the tradeoff is/which way is faster.

dannyhayes-gds: Could be worth spinning up 2 versions of the lambda? One with jackson and one with Gson to see if there is a big difference or not?

PR #151, opened 2022-02-25
Pyi 735 spike remove config service use gson

SunilDGDS:
### What changed
Experimental lambda using gson instead of object mapper to benchmark performance

- [PYI-735](https://govukverify.atlassian.net/browse/PYI-735)

## Checklists

There are some duplications (sonar ) but this is due to use of some common classes in the 2 experimental self-containedpackages , less likely to happen in a real project .  In the future, only one of the packages will be used


Danomanic:
Closing this.

PR #155, opened 2022-03-02
BAU: Wrap shared attributes vc_http_api claim in claims claim

Wynndow:
### What changed

RFC 0008 actually specifies that the custom vc_http_api claim should be
nested under a top level claim in the JWT called `claims`.

This is to keep in line with the OIDC spec which we're sort of
following. https://openid.net/specs/openid-connect-core-1_0.html#RequestObject

This is quite a naive approach and there are probably better ways to do
it using objects from the OIDC library. This implementation is probably
going to change again though as we need to start encrypting these shared
attributes, so I'm delaying that until then.

### Why did it change

To meet the spec we've laid out.


idrop (approved):
Thanks @Wynndow

PR #157, opened 2022-03-04
PYI-790: New External & Internal AWS Api Gateways via openapi

dannyhayes-gds:
### What changed
Split the existing AWS gateway into 2, "external" and "internal.

For now this will not be changing the structure of the endpoints. These can be changed as we begin on work on updating the lambdas to work with the new EventResponse types.

### Why did it change
The gateways are now modelled using openapi.

### Issue tracking

- [PYI-790](https://govukverify.atlassian.net/browse/PYI-790)

Related PR's:

- https://github.com/alphagov/di-ipv-config/pull/206
- https://github.com/alphagov/di-ipv-core-front/pull/67

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -13,11 +13,22 @@ Parameters:
 
 Resources:
 
-  IPVCoreAPI:
+  IPVCoreInternalAPI:
     Type: AWS::Serverless::Api
     Properties:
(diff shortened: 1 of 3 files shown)


danworth (approved):
This looks good, let's see how using the openAPI spec to define the api gateways works out.

PR #160, opened 2022-03-08
Add old original api gateway output to fix the param store stack

dannyhayes-gds:
### What changed
Add old api gateway id in order to fix the param store stack.

This is a temporary change in order to allow the ipv-core-back-{env} stack to be deployed. This param will be removed after that and the param store stack will be updated.

### Why did it change
Cloudformation does not allow us to change the outputs since they are used by the paramter-store stack.

### Issue tracking

- [PYI-790](https://govukverify.atlassian.net/browse/PYI-790)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -393,6 +393,10 @@ Resources:
           KeyType: "HASH"
 
 Outputs:
+  IPVCoreAPIGatewayID:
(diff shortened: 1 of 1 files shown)


danworth (approved):
Interesting CloudFormation behaviour for us to keep in mind during future refactoring

PR #161, opened 2022-03-08
BAU: Fix sam template syntax

dannyhayes-gds:
### What changed
Fixed SAM template file syntax with missing "Export" property.


danworth (approved):
`sam validate` was ok. 👍

PR #169, opened 2022-03-11
PYI-813: Update /journey/cri/return lambda to return a JourneyResponse

dannyhayes-gds:
### What changed
Updated the /request-evidence "credentialIssuer" lambda to move under the /journey/cri/return endpoint and renamed the lambda.

Moved the JourneyResponse class into the common lib folder to be re-used by this lambda.

Made the CredentialIssuerReturn lambda return a "/journey/next"journey response.

### Why did it change
This lambda will be called by core-front when a user is redirected back onto-core front. The main functionality remains unchanged, but now we return a "/journey/next" journey response to ensure that core-front makes a request to the journey engine lambda in order to determine where to send the user next.

### Issue tracking

- [PYI-813](https://govukverify.atlassian.net/browse/PYI-813)


kerrr (approved):
Looks good

PR #170, opened 2022-03-14
PYI 779

SunilDGDS:
### What changed
Currently the sharedattribute lamdba in core-back generates a data structure for the core attributes (name, address, dob), and returns it in a serialized JWT . The data structure has been modified according to the new spec and included in the Verifiable Credentials . The new data structure will also be used as shared attributes to different CRIs .

- [PYI-779](https://govukverify.atlassian.net/browse/PYI-779)

## Checklists
I have split the main tasks into 3 ( subtasks )  just to define the working steps.  The whole work is deployed under the main task ticket (PYI-779)

### Other considerations

There is likely to be more changes in the data structure ...


.gitignore
```diff
@@ -32,3 +32,6 @@ dist
 
 # Ignore Gradle build output directory
 build
+
+# Ignore AWS SAM folder
+*/**/.aws-sam
```
Wynndow: Would be good add the linebreak back at the end of the file.

lambdas/sharedattributes/src/main/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandler.java
```diff
@@ -77,17 +77,17 @@ private SharedAttributesResponse getSharedAttributes(String ipvSessionId)
         Map<String, String> credentials =
                 userIdentityService.getUserIssuedCredentials(ipvSessionId);
 
-        List<SharedAttributes> sharedAttributes = new ArrayList<>();
+        List<SharedAttributes> sharedAttributesNew = new ArrayList<>();
```
Wynndow: Can we drop the use of `New`? As it loses context once merged and is hard to understand.

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
```diff
@@ -107,10 +141,10 @@ void shouldExtractSessionIdFromHeaderAndReturnSharedAttributesAndStatusOK()
         when(userIdentityService.getUserIssuedCredentials(SESSION_ID))
                 .thenReturn(
                         Map.of(
+                                "CredentialIssuer",
+                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_2),
                                 "CredentialIssuer1",
-                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_1),
-                                "CredentialIssuer2",
-                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_2)));
+                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_1)));
```
Wynndow: It's a bit confusing that `CREDENTIAL_INPUT_2` is first in this map - maybe rename it's key to `CredentialIssuer2` and re-order them?

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
```diff
@@ -123,36 +157,30 @@ void shouldExtractSessionIdFromHeaderAndReturnSharedAttributesAndStatusOK()
...
+                if (jo.get("streetAddress") != null) {
+                    if (jo.get("streetAddress").asText().equals("758 Huel Neck")) {
+                        assertEquals("758 Huel Neck", jo.get("streetAddress").asText());
+                    }
+                    if (jo.get("postalCode").asText().equals("S5 6UN")) {
+                        assertEquals("S5 6UN", jo.get("postalCode").asText());
+                    }
+                }
+            }
+        }
```
Wynndow: I think this can be simplified. You can probably drop the null checks - this is just a test so we should be confident in the data we're working with. 
Also, you've got if statements doing a pre-check on the assertions. So you're tests will silently fail if the assertion is wrong.

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
```diff
@@ -123,36 +157,30 @@ void shouldExtractSessionIdFromHeaderAndReturnSharedAttributesAndStatusOK()
...
+        assertEquals(2, (vcAttributes.get("name")).size());
+        assertEquals(3, (vcAttributes.get("address")).size());
+        assertEquals(3, (vcAttributes.get("birthDate")).size());
 
         RSASSAVerifier rsaVerifier =
                 new RSASSAVerifier((RSAPublicKey) getCertificate().getPublicKey());
         assertTrue(signedJWT.verify(rsaVerifier));
+        System.out.println();
```
Wynndow: I think you can drop this line.

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
```diff
@@ -217,9 +244,9 @@ void shouldReturn500WhenUnableToSignSharedAttributes() throws JsonProcessingExce
                 .thenReturn(
                         Map.of(
                                 "CredentialIssuer1",
-                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_1),
+                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_2),
                                 "CredentialIssuer2",
-                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_2)));
+                                objectMapper.writeValueAsString(CREDENTIAL_INPUT_1)));
```
Wynndow: These keys now don't match with the names of the values? Can they be re-ordered to match?

lib/src/main/java/uk/gov/di/ipv/core/library/domain/SharedAttributes.java
```diff
@@ -1,82 +1,68 @@
...
-    public Optional<Map<String, String>> getAddress() {
+    public Optional<Set<Address>> getAddress() {
         return Optional.ofNullable(address);
     }
 
-    public Optional<List<Map<String, String>>> getAddressHistory() {
-        return Optional.ofNullable(addressHistory);
-    }
-
+    @JsonPropertyOrder({"name", "birthDate", "address"})
```
Wynndow: Do we need this on the builder?

lib/src/main/java/uk/gov/di/ipv/core/library/domain/SharedAttributesDeserializer.java
```diff
@@ -28,43 +29,49 @@ protected SharedAttributesDeserializer(Class<?> vc) {
...
+                    nameParts.forEach(
+                            namePart ->
+                                    namePartsList.add(
+                                            objectMapper.convertValue(namePart, NameParts.class)));
+                }
             }
+
+            Name names = new Name(namePartsList);
+            nameList.add(names);
+            sharedAttributesBuilder.setName(nameList);
```
Wynndow, line 56: A name node is a list of objects. Each object has one key which is `nameParts`.
I feel like this may not handle the situation where the `name` list has more than one object. I think it will add all objects under `nameParts` to one `Name` object, rather than building individual `namePartsList`s for each entry in the `name` list.

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
SunilDGDS: There are 3 addresses blocks , one of them does not have streetaddress (ie set to null & is not shown) . This is why I need to check if the address is not null . Also the address blocks are not always displayed in the same order . I cannot say for ex , assert that block address[1].streetname equala "758.." . To get round , I check if the  streetaddress.exists & therefore assert ..

lib/src/main/java/uk/gov/di/ipv/core/library/domain/SharedAttributesDeserializer.java
SunilDGDS: this is correct . The list of nameparts are added to the Name object (also shown correctly in the test)

lib/src/main/java/uk/gov/di/ipv/core/library/domain/SharedAttributes.java
SunilDGDS: "this" is required by the builder

lambdas/sharedattributes/src/test/java/uk/gov/di/ipv/core/sharedattributes/SharedAttributesHandlerTest.java
```diff
@@ -161,18 +165,24 @@ void shouldExtractSessionIdFromHeaderAndReturnSharedAttributesAndStatusOK()
...
+                        .filter(x -> "35 Idsworth Road".equals(x.getStreetAddress()))
+                        .findAny()
+                        .orElse(null);
+        Address postCode =
+                addressList.stream()
+                        .filter(x -> "38421-3292".equals(x.getPostalCode()))
+                        .findAny()
+                        .orElse(null);
+
+        assertEquals(true, !streetAddress.getStreetAddress().isEmpty());
```
Wynndow: Maybe this could be `assertFalse(streetAddress.getStreetAddress().isEmpty())`

Wynndow (approved):

PR #171, opened 2022-03-14
PYI-782: Create the new CriStart lambda

Danomanic:
### What changed
- Add new Lambda Handler for `/journey/cri/start/{criId}` - This provides the CRI information to the frontend so that it can proceed to the CRI as part of the journey.

### Issue tracking

- [PYI-782](https://govukverify.atlassian.net/browse/PYI-782)

### Other considerations


lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/domain/CriResponse.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CriResponse {
+    @JsonProperty private final String id;
+
+    @JsonProperty private final String authorizeUrl;
+
+    @JsonProperty private final String request;
```
dannyhayes-gds: I don't think we actually need this field anymore. I think the original plan was for this request field to be the "request" property that we send the shared attributes across to the CRI in. But for now we are keeping this as a manual step that core-front will always perform before redirecting to the CRI so it doesn't need to be returned in the CriResponse

lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/domain/CriResponse.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.di.ipv.core.credentialissuer.domain;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CriResponse {
+    @JsonProperty private final String id;
```
dannyhayes-gds, line 7: I think what we do need to add to this response is will be 2 ID values. One for the criId, and then one for the ipvClientId. Both of these params exist within parameter store and will be used by core-front when redirecting onto the cri.

There is one more field that core-front constructs which is "state" but so far this has always just been a hard coded test value that we've never used so I don't think its needed here

lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    public CredentialIssuerStartHandler() {
+        this.configurationService = new ConfigurationService();
+    }
+
+    @Override
+    @Tracing
+    public APIGatewayProxyResponseEvent handleRequest(
+            APIGatewayProxyRequestEvent input, Context context) {
+
+        CredentialIssuerRequestDto request =
```
dannyhayes-gds: I'm don't think this part is quite correct. The CredentialIssuerRequestDto is the type of request that gets passed into the /journey/cri/return endpoint, as it has the auth code provided by the CRI.

In terms of the request coming in here it will just be the criId as a param coming from the AWS gateway endpoint url. You can see an example of how I did the same with the journey engine lambda to get the /journey/{journeyStep} value.

lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                new CriResponse(
+                        credentialIssuerConfig.getId(),
+                        credentialIssuerConfig.getAuthorizeUrl().toString(),
+                        request.getRedirectUri());
+
+        return ApiGatewayResponseGenerator.proxyJsonResponse(200, criResponse);
+    }
+
+    @Tracing
+    private Optional<ErrorResponse> validate(CredentialIssuerRequestDto request) {
```
dannyhayes-gds: Some of this will need updating as we are not expecting an auth code at this point, and I don't think we need the sessionId. It should just be checking if the criId we receive from the /journey/cri/start/{criId} is valid and we can find all the params we need from param store for it etc

openAPI/core-back-internal.yaml
```diff
@@ -111,7 +111,23 @@ paths:
...
+      responses:
+        200:
+          description: "Returns the id, authorizationUrl and request for a CRI."
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/journeyType"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
```
dannyhayes-gds, line 133: It won't let me add a comment to the correct line here, but the line below this one needs updating to use the new AWS function name for your new lambda you've defined in the SAM template file.

```suggestion
        Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${IPVCredentialIssuerStartFunction.Arn}/invocations
```

lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
```diff
@@ -0,0 +1,75 @@
...
+                new CriResponse(
+                        credentialIssuerConfig.getId(),
+                        credentialIssuerConfig.getIpvClientId(),
+                        credentialIssuerConfig.getAuthorizeUrl().toString());
+
+        return ApiGatewayResponseGenerator.proxyJsonResponse(200, criResponse);
+    }
+
+    @Tracing
+    private Optional<ErrorResponse> validate(Map<String, String> pathParameters) {
```
dannyhayes-gds, line 65: Good shout validating the pathParameters are present, I didn't think of doing that for the journeyEngine lambda so I'll add it there as well 👍

openAPI/core-back-internal.yaml
```diff
@@ -118,10 +118,26 @@ paths:
           Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${IPVSharedAttributesFunction.Arn}/invocations
         passthroughBehavior: "when_no_match"
         type: "aws_proxy"
+  /journey/cri/start/{criId}:
+    post:
+      description: Called when the frontend begins the CRI journey.
+      responses:
+        200:
+          description: "Returns the id, authorizationUrl and request for a CRI."
```
dannyhayes-gds: Very minor thing, but could this be updated to be something like

```suggestion
          description: "Returns the id, ipvClientId and authorizationUrl for a CRI."
```

dannyhayes-gds (approved):
Looks good 👍

PR #176, opened 2022-03-15
PYI-521: Get verifiable credentials from CRIs

Wynndow:
### What changed

CRIs will be returning verifiable credentials in the form of signed
JWTs. This updates the credentialissuerreturn lambda to send a post
request for the VC, and receive a signed JWT in response. It then
validates the signature on the response is expected.

We send a POST request to be in line with the Verifiable Credentials API
spec: https://w3c-ccg.github.io/vc-api/#introduction

We send the subject in an unsigned JWT to keep in line with this ADR.
Effectively the plain JWT is a "template" that the CRI can fill in. This
template will need to contain more information for back end only CRIs,
but that will get covered off in a future story. https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0008-identity-ipv-sequence-diagram.md#issue

### Issue tracking

- [PYI-521](https://govukverify.atlassian.net/browse/PYI-521)


lambdas/credentialissuerconfig/build.gradle
```diff
@@ -24,7 +24,8 @@ dependencies {
 			"org.mockito:mockito-core:4.1.0",
 			"org.mockito:mockito-junit-jupiter:4.1.0",
 			"com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.13.0",
-			"com.nimbusds:oauth2-oidc-sdk:9.3.1"
+			"com.nimbusds:oauth2-oidc-sdk:9.3.1",
+			project(":lib").sourceSets.test.output
```
Danomanic, line 28: Just out of interest, what is this doing?

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -84,4 +91,10 @@ public UserIdentityService(
 
         return userIssuedDebugCredentials;
     }
+
+    private boolean isNotPopulatedJsonArray(Object input) {
```
Danomanic, line 95: This was hard to get my head around.

Are we testing that the array is empty, but exists within the vcClaim object?

So an empty `{}`

Would `vcClaim.get(VC_EVIDENCE).length() == 0` work instead? Rather than testing the instances etc.

Danomanic (commented):
Looks good so far, just some tiny questions.

lambdas/credentialissuerconfig/build.gradle
Wynndow: There's a class in the lib projects test directory called `TestFixtures` which has got a bunch of things like private/public keys and pre-signed verifiable credentials. Stuff that you might need that can be a pain to set up for individual tests.
As it's in the lib projects tests it wasn't on the class path for the lambda projects tests. This adds the lib projects test source set to the credenentialissuerconfig lambda projects test dependencies, which makes it accessible.

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Yeah, this is a bit of a mind bender. And the name of it isn't super helpful!

It returns true if:

- The input in null
- The input isn't a JSONArray
- The JSONArray is empty
- The first element in the JSONArray isn't a JSONObject.

It's basically doing a safety check to make sure that value for the evidence attribute in the verifiable credential exists and is an object, before we try and access it.
I hate working with JSON in Java.

lambdas/credentialissuerconfig/build.gradle
Wynndow: So it allows me to do this: https://github.com/alphagov/di-ipv-core-back/pull/176/files#diff-8720e48a70c9d82f7e6ef03d4dd907c0d36c87975f30d1a45b239ed2d6cd56a1R26

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Danomanic: > I hate working with JSON in Java.

I agree - this looks good then!

lambdas/credentialissuerconfig/build.gradle
Danomanic: Makes sense, thanks for explaining the gradle wizardry.

Danomanic (approved):
Looks good to me.

PR #177, opened 2022-03-15
PYI-819: Return a ClientResponse journey type

dannyhayes-gds:
### What changed
Update the authorize lambda to sit under the new /journey/session/end endpoint and update the lambda and package names.

Update the lambda to retrieve the oauth query params for the orc client from the IPV session object instead of the request query params.

Return a ClientResponse object containing all the values that core-front need to redirect the user back onto the orc stub.

### Why did it change
To fit into the new journey event based api structure. And to move the storage and retrieval of the orc clients oauth auth query params from the core-front session to the back-end session.

The newly returned ClientResponse object will be used by core-front to redirect the user back to the orc-client, and will be using values previously stored within the back-end session.

### Issue tracking

- [PYI-819](https://govukverify.atlassian.net/browse/PYI-819)


deploy/template.yaml
```diff
@@ -110,22 +110,22 @@ Resources:
...
             Method: POST
 
-  IPVAuthorizationFunction:
+  IPVSessionEndFunction:
     Type: AWS::Serverless::Function
     Properties:
-      FunctionName: !Sub "ipv-authorization-${Environment}"
-      Handler: uk.gov.di.ipv.core.authorization.AuthorizationHandler::handleRequest
+      FunctionName: !Sub "ipv-session-end-${Environment}"
+      Handler: uk.gov.di.ipv.core.sessionend.SessionEndHandler::handleRequest
```
Danomanic, line 117: I've left the session start under the package `uk.gov.di.ipv.core.session`

Happy to change the StartHandler if you want to standardise on one version.

Danomanic (approved):

PR #179, opened 2022-03-16
PYIC-403 journey engine returns an error page when cri/error status received

SunilDGDS:
### What changed
The journey engine should return an error page when in   cri/error state . This would have been raised by the new lambda that will deal with errors & would have set the state   cri/error state .

- [PYIC-403](https://govukverify.atlassian.net/browse/PYIC-403)


dannyhayes-gds (commented):
Looks good 👍

dannyhayes-gds (dismissed):

dannyhayes-gds (approved):

PR #182, opened 2022-03-17
PYI-743 Import siging key from it's stack

TomTucka:
### What changed

Removed hardcoded ARN value
Imported value from stack

### Why did it change

We need to import the value so it works across multiple environments

### Issue tracking

- [PYI-743](https://govukverify.atlassian.net/browse/PYI-743)


danworth (approved):
Looks good, we need to create kms key stack in dev also.

PR #183, opened 2022-03-17
BAU: Update start session request to receive client auth params in request body

dannyhayes-gds:
### What changed
Updated the start session endpoint to receive the client connection details in the request body rather than as query params,

### Why did it change
Simplifies the request and simplifies the construction of the dto object.

### DIFF
--- a/deploy/samconfig.toml
+++ b/deploy/samconfig.toml
@@ -81,3 +81,19 @@ region = "eu-west-2"
 capabilities = "CAPABILITY_IAM"
 parameter_overrides = "Environment=\"dev-kerrr\""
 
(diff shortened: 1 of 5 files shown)


SunilDGDS (commented):
Makes sense to capture in one Dto . Are we sure that all the values are coming through ? There is no validation now (as previously) ; not too sure if the Json annotations are required in the ClientSessionDetailsDTO

dannyhayes-gds:
The object mapping deserialising will fail if the object being passed in is missing any of the fields etc and will be caught by the JsonProcessingException in the catch block.

And yeah looks your right about the annotations, thanks!

PR #194, opened 2022-03-28
PYIC-844: Update the journey engine to match with the prototype user journey

dannyhayes-gds:
### What changed
Updated the journey engine to match up with the prototype user journey flow

### Why did it change
Makes our user journey a bit more realistic.

### Issue tracking

- [PYIC-844](https://govukverify.atlassian.net/browse/PYIC-844)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserStates.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserStates.java
@@ -2,9 +2,10 @@ package uk.gov.di.ipv.core.library.domain;
 
 public enum UserStates {
     DEBUG_PAGE("page-ipv-debug"),
-    INITIAL_IPV_JOURNEY("page-ipv-start"),
-    TRANSITION_PAGE_1("page-cri-transition"),
-    TRANSITION_PAGE_2("page-cri-transition"),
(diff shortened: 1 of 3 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserStates.java
```diff
@@ -2,9 +2,10 @@
 
 public enum UserStates {
     DEBUG_PAGE("page-ipv-debug"),
-    INITIAL_IPV_JOURNEY("page-ipv-start"),
```
AmritSidhu, line 5: Does this mean we should remove these templates from core-front now?

AmritSidhu (approved):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserStates.java
dannyhayes-gds: I don't think we've used it yet, so yeah might not be needed on core-front either

dannyhayes-gds: I don't think we've used it yet, so yeah might not be needed on core-front either

PR #195, opened 2022-03-29
PYIC-865: Use ES256 to sign JWTs

Wynndow:
### What changed

[PYIC-865: Switch KmsSigner to ES256 and rename](https://github.com/alphagov/di-ipv-core-back/commit/cfa3b345a2f09d159aebf94532d77d90a2cb9e78)[](https://github.com/Wynndow) 

Change the KMS signer to use the ES256 JWS algorithm, rather than RS256.
This is so we stay in step with how we're signing in other parts of the
system. 
  
[PYIC-856: Rename sharedAttributeSigningKey to signingKey](https://github.com/alphagov/di-ipv-core-back/commit/0a078e8f546d615dc617e6873ded724cda7fb206) 

The same key is being used to sign the shared attributes JWT, as well as
the client auth JWT used on the token endpoint.

### Issue tracking

- [PYIC-865](https://govukverify.atlassian.net/browse/PYIC-865)


Danomanic (approved):
LGTM

PR #196, opened 2022-03-30
PYIC-873: Accept ES256 as client auth JWT signature method

Wynndow:
**Config PR to accompany this PR: https://github.com/alphagov/di-ipv-config/pull/273**

### What changed

Updated the public key selector to check the algorithm used to sign the JWT 
and then parse the correct form of public key from either a cert of JWK.

### Why did it change

We should be able to accept ES256 as well as RS256 (which we currently
support).

The auth team will be using ES256.

### Issue tracking

- [PYIC-873](https://govukverify.atlassian.net/browse/PYIC-873)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/ConfigurationServicePublicKeySelector.java
```diff
@@ -40,13 +51,40 @@ public ConfigurationServicePublicKeySelector(ConfigurationService configurationS
...
-        } catch (CertificateException e) {
-            throw new InvalidClientException(e.getMessage());
+
+        JWSAlgorithm algorithm = jwsHeader.getAlgorithm();
+        String clientId = claimedClientID.getValue();
+        String publicKeyMaterial = configurationService.getClientPublicKeyMaterial(clientId);
+
+        switch (algorithm.toString()) {
+            case ES256:
+                try {
```
Danomanic, line 61: For a future refactor:
Something doesn't seem right about having the try catch within the case block. Maybe wrapping the switch within one try catch is safer? However, I'm unsure how you'd provide a specific error message - so ignore for now.

Danomanic (approved):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ConfigurationServicePublicKeySelector.java
Wynndow: Oh interesting. I've not worked with switch's before tbh. I can see why wrapping the whole thing might be a bit cleaner. I've merged this but I'll fix it up to your suggestion.

PR #197, opened 2022-03-30
BAU: Refactor public key selector switch

Wynndow:
### What changed

As per @Danomanic suggestion, wrap the case statement in a try catch, rather than having individual
ones per case block.

### Why did it change

Make it more clean and readable.


lib/src/main/java/uk/gov/di/ipv/core/library/domain/ConfigurationServicePublicKeySelector.java
```diff
@@ -56,35 +56,31 @@ public ConfigurationServicePublicKeySelector(ConfigurationService configurationS
         String clientId = claimedClientID.getValue();
         String publicKeyMaterial = configurationService.getClientPublicKeyMaterial(clientId);
 
-        switch (algorithm.toString()) {
-            case ES256:
-                try {
+        try {
```
Danomanic, line 59: This seems 'nicer'. Thanks for taking the time to do this.

Danomanic (approved):

PR #199, opened 2022-03-31
BAU: Provisioned concurrency for lambdas

Wynndow:
### What changed

This adds the ability to set provisioned concurrency on a per env basis.

With SAM templates you need to set the AutoPublishAlias value to be able
to use ProvisionedConcurrencyConfig.

AutoPublishAlias will set up a version and an alias for the lambda.
ProvisionedConcurrencyConfig is passed directly to the cloudforamtion
Alias resource created by the AutoPublishAlias.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -11,6 +11,15 @@ Parameters:
   Environment:
     Type: String
 
+Mappings:
+  ProvisionedConcurrency:
+    dev:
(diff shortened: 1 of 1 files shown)


kerrr (approved):
Lets give it a try!

PR #212, opened 2022-04-08
BAU: Fix debug page to show evidence.

Danomanic:
### What changed
The debug page currently doesn't render the evidence block due to core-back returning the old 'gpgScore' rather than 'evidence'. This should fix that.

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserIssuedDebugCredential.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserIssuedDebugCredential.java
@@ -12,10 +12,10 @@ import java.util.Map;
 public class UserIssuedDebugCredential {
 
     DebugCredentialAttributes attributes;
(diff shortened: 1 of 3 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserIssuedDebugCredential.java
```diff
@@ -12,10 +12,10 @@
 public class UserIssuedDebugCredential {
 
     DebugCredentialAttributes attributes;
-    private Map<String, Object> gpg45Score;
+    private Map<String, Object> evidence;
```
dannyhayes-gds, line 15: Happy to merge this for now to get things working, but this might be something we should look into for the best name for this. I'm not 100% sure but I think "evidence" is for  only specifically evidence cri types. There are others so having all of the cri gpg45scores under a field called "evidence" might not make sense.

dannyhayes-gds (approved):

PR #214, opened 2022-04-11
PYIC-903 - Add VOT claim to user identity response

dannyhayes-gds:
### What changed
Update the user-identity response to include the VOT (Vector of trust) claim.

The claim will have a value of either "P0" or "P2". P0 is for when the identity check has failed, and P2 is for a successful check. For it to be successful we need to check that the passport validation score is 2, fraud score of 1, and kbv verification score of 2.

### Why did it change
The VOT claim will be used to represent the level of confidence achieved, and will be used by Auth in order to determin wether SPOT needs to be invoked or not.

### Issue tracking

- [PYIC-903](https://govukverify.atlassian.net/browse/PYIC-903)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public Builder setVot(String vot) {
+            this.vot = vot;
+            return this;
+        }
+
+        public UserIdentity build() {
+            return new UserIdentity(vcs, vot);
+        }
```
Wynndow, line 47: What's the benefit of using the builder pattern here, rather than just using the constructor you already have?

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -91,6 +108,74 @@ public Map<String, String> getUserIssuedDebugCredentials(String ipvSessionId) {
...
+        }
+
+        return getVectorOfTrustValue(validPassport, validFraud, validKbv);
+    }
+
+    private Optional<Boolean> isValidScore(
+            JSONArray evidenceArray, String property, int scoreValue) {
+        Long gpg45ScoreValue = ((Map<String, Long>) evidenceArray.get(0)).get(property);
+        if (gpg45ScoreValue != null) {
+            return Optional.of(gpg45ScoreValue.intValue() == scoreValue);
```
Wynndow: I know we won't be receiving it for MVP, but using passport as an example, it's possible to get a validity score of higher than 2. Should this logic be `gpg45ScoreValue.intValue() >= scoreValue`? And perhaps `GPG_45_MAX_VALIDITY_SCORE ` might be called something like `GPG_45_M1A_VALIDITY_SCORE `. Or something. I'm hazy on gpg45 details so I may be way off the mark here.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
dannyhayes-gds: It was more for future proofing than anything else since there are still a fair few more fields to be added to this class in order to match with the spec, and then more again after MVP.

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
dannyhayes-gds: Oh yeah good point, I had this thought this morning and need to double check it. I think for passport specifically though its always either 0 or 2. Not sure about the others though so I'll double check this before merging

Wynndow (approved):

PR #218, opened 2022-04-13
BAU: Add missing kms permissions needed to write onto the audit event queue

dannyhayes-gds:
### What changed
Add missing kms permissions for the kms encryption key that needs to be applied to the lambda for writing to the audit event queue.

### Why did it change
Without these permissions the lambda will fail to write to the SQS queue

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -220,6 +220,13 @@ Resources:
               - 'kms:sign'
             Resource:
               - !ImportValue SigningKeyArn
+          - Sid: kmsAuditEventQueueEncryptionKeyPermission
+            Effect: Allow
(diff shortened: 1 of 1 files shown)


kerrr (dismissed):

deploy/template.yaml
```diff
@@ -68,6 +68,13 @@ Resources:
               - 'kms:sign'
             Resource:
               - !ImportValue SigningKeyArn
+          - Sid: kmsAuditEventQueueEncryptionKeyPermission
```
Wynndow: Is this in the wrong lambda?

Wynndow (approved):

PR #219, opened 2022-04-13
PYIC-859: Add new SqsService to write messages to the audit event queue

dannyhayes-gds:
### What changed

- Create new common service to send messages to our SQSAuditEvent queue.
- Create a enum to hold each of the type events we will send.
- Send event after the CRI VC has been stored.

### Why did it change
For slice 2 we are sending an event to the TxMA SQS queue after a CRI VC has been gathered and validated.

### Issue tracking

- [PYIC-859](https://govukverify.atlassian.net/browse/PYIC-859)

### DIFF
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -21,9 +21,9 @@ import uk.gov.di.ipv.core.library.exceptions.SqsException;
 import uk.gov.di.ipv.core.library.helpers.ApiGatewayResponseGenerator;
 import uk.gov.di.ipv.core.library.helpers.KmsEs256Signer;
 import uk.gov.di.ipv.core.library.helpers.RequestHelper;
+import uk.gov.di.ipv.core.library.service.AuditService;
 import uk.gov.di.ipv.core.library.service.ConfigurationService;
 import uk.gov.di.ipv.core.library.service.CredentialIssuerService;
-import uk.gov.di.ipv.core.library.service.SqsService;
 
 import java.util.Optional;
 
@@ -35,15 +35,15 @@ public class CredentialIssuerReturnHandler
 
     private final CredentialIssuerService credentialIssuerService;
     private final ConfigurationService configurationService;
-    private final SqsService sqsService;
+    private final AuditService auditService;
 
     public CredentialIssuerReturnHandler(
             CredentialIssuerService credentialIssuerService,
             ConfigurationService configurationService,
-            SqsService sqsService) {
+            AuditService auditService) {
         this.credentialIssuerService = credentialIssuerService;
         this.configurationService = configurationService;
-        this.sqsService = sqsService;
+        this.auditService = auditService;
     }
 
     @ExcludeFromGeneratedCoverageReport
@@ -52,7 +52,8 @@ public class CredentialIssuerReturnHandler
         JWSSigner signer = new KmsEs256Signer(configurationService.getSigningKeyId());
 
         this.credentialIssuerService = new CredentialIssuerService(configurationService, signer);
-        this.sqsService = new SqsService(SqsService.getDefaultSqsClient(), configurationService);
+        this.auditService =
+                new AuditService(AuditService.getDefaultSqsClient(), configurationService);
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -66,11 +76,17 @@ public APIGatewayProxyResponseEvent handleRequest(
                             accessToken, credentialIssuerConfig, request.getIpvSessionId());
             credentialIssuerService.persistUserCredentials(verifiableCredential, request);
 
+            sqsService.sendAuditEvent(
+                    AuditEventTypes.IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED);
+
```
Wynndow: Just a thought, the event is "received and signature checked". So this should probably happen before the persistence. If something borks with writing it to Dynamdb we should still have created the audit event.

lambdas/credentialissuerreturn/src/test/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandlerTest.java
```diff
@@ -72,7 +75,8 @@ void setUp() throws URISyntaxException {
     @Test
     void shouldReceive200AndJourneyResponseOnSuccessfulRequest() throws JsonProcessingException {
         CredentialIssuerReturnHandler handler =
-                new CredentialIssuerReturnHandler(credentialIssuerService, configurationService);
+                new CredentialIssuerReturnHandler(
+                        credentialIssuerService, configurationService, sqsService);
```
Wynndow: Might be good to assert in this test that the sqsService has been called correctly?

lib/src/main/java/uk/gov/di/ipv/core/library/service/SqsService.java
```diff
@@ -0,0 +1,47 @@
...
+import com.amazonaws.services.sqs.model.SendMessageRequest;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.di.ipv.core.library.auditing.AuditEvent;
+import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
+import uk.gov.di.ipv.core.library.exceptions.SqsException;
+
+import java.time.Instant;
+
+public class SqsService {
```
kerrr: I think this might be better being `AuditService` or something like that?

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
dannyhayes-gds: Yeah I was a little unsure of wether it should be before or after the persistence. I went with after as I thought it could be hard to follow if we've got an audit trail saying that we "received" a VC but then there being no record of it in the DB if there were DynamoDb issues.

But I see your point as well so I'm happy to change this

lambdas/credentialissuerreturn/src/test/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandlerTest.java
dannyhayes-gds: Yeah good shout 👍

lib/src/main/java/uk/gov/di/ipv/core/library/service/SqsService.java
dannyhayes-gds: Yeah true, I'll change it 👍

lib/src/main/java/uk/gov/di/ipv/core/library/service/SqsService.java
```diff
@@ -0,0 +1,47 @@
...
+
+            sqs.sendMessage(sendMessageRequest);
+        } catch (JsonProcessingException e) {
+            throw new SqsException(e);
+        }
+    }
+
+    private String generateMessageBody(AuditEventTypes eventType) throws JsonProcessingException {
+        AuditEvent auditEvent =
+                new AuditEvent(Instant.now().getEpochSecond(), eventType.toString());
```
Wynndow: If you changed the `eventName` field in the `AuditEvent` class to type `AuditEventTypes`, could you pass the event type in directly? I think that the object mapper would serialize it to the name of the enum which is what we want here? Save calling `toString` here?

dannyhayes-gds: Yeah I think you're right. I'll give it a try, thanks

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
Wynndow: Yeah it's a funny one. I would hope that if something did happen with the DB though that we'd catch an error and log it, so we should be able to work out what happened.

lib/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEvent.java
```diff
@@ -11,12 +11,12 @@ public class AuditEvent {
     @JsonProperty private long timestamp;
 
     @JsonProperty("event_name")
-    private String eventName;
+    private AuditEventTypes eventName;
```
Wynndow: Tiny thing, but maybe this should be called just `event` now?

Wynndow (approved):

kerrr (approved):

PR #232, opened 2022-05-04
PYIC-1065: Lambda and API gateway logs to Splunk

Wynndow:
### What changed

Add CloudFormation resources to start sending our lambda application
logs and API gateway access logs to Splunk.

### Why did it change

For the API gateway logs I've reused the JSON format we have defined for
the core-front Fargate API gateway logs, for consistency. The only
difference is the addition of `responseLatency` which could be helpful
when looking at slow lambdas.

For the lambdas I've explicitly declared a log group resource even
though lambda creates one for you. This is because the lambda created
log groups will store messages indefinitely by default. You can change
this by declaring the log group yourself. As long as the name is the
same as the log group lambda creates and writes to, the function should
write to this new log group. It will be interesting to see if these log
groups deploy cleanly, considering they already exist...

The subscription filters are taking the logs from the newly created log
groups and sending them to Cyber's Centralised Security Logging Service
(CSLS) Kinesis data stream. This is the recommended way of getting our
logs to Splunk. See here for more details:
https://github.com/alphagov/centralised-security-logging-service

### Issue tracking

- [PYIC-1065](https://govukverify.atlassian.net/browse/PYIC-1065)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -30,16 +30,16 @@ Resources:
         DestinationArn: !GetAtt IPVCoreInternalAPILogGroup.Arn
         Format: >-
           {
-          "requestId":      "$context.requestId",
-          "ip":             "$context.identity.sourceIp",
-          "requestTime":    "$context.requestTime",
-          "httpMethod":     "$context.httpMethod",
-          "path":           "$context.path",
-          "routeKey":       "$context.routeKey",
-          "status":         "$context.status",
-          "protocol":       "$context.protocol",
+          "requestId":"$context.requestId",
+          "ip":"$context.identity.sourceIp",
+          "requestTime":"$context.requestTime",
+          "httpMethod":"$context.httpMethod",
+          "path":"$context.path",
+          "routeKey":"$context.routeKey",
+          "status":"$context.status",
+          "protocol":"$context.protocol",
           "responseLatency":"$context.responseLatency",
-          "responseLength": "$context.responseLength"
+          "responseLength":"$context.responseLength"
           }
 
   IPVCoreInternalAPILogGroup:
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -26,6 +26,38 @@ Resources:
...
+    Properties:
+      LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-CoreBackInternal-API-GW-AccessLogs
+      RetentionInDays: 14
+
+  IPVCoreInternalAPILogGroupSubscriptionFilter:
+    Type: AWS::Logs::SubscriptionFilter
+    Properties:
+      DestinationArn: "arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prod"
+      FilterPattern: ""
+      LogGroupName: !Select # Select the string at the 6th index of...
```
danworth, line 56: You should be able to replace this with simply `!Ref IPVCoreInternalAPILogGroup`

danworth: You can replace this with simply `!Ref IPVCoreInternalAPILogGroup` which returns the log group name, here and in the similar places below. Usually `!Ref <logical resource id>` gives you the name of the resource, useful for dynamodb tables etc.

Wynndow: Oh mate. I was looking at the docs and completely missed the Ref part, focussed on the Attributes section. This is going to make things a lot easier...

Wynndow: Fixed up

danworth:
Hey Chris, nice work. I've tried applying this to my dev environment and it fails as you guessed because the log groups already exist. Given we're not live we could delete the stacks (will need to remove the param stack since it imports and export from the core-back stack) and recreate 😢  . I'm guessing you've not deployed this into your dev environment yet? I think is worth doing so just to make sure the logging works as expected before going to the effort of deploying this into build and beyond (I'm also wondering if it will actually work even if the stack doesn't already exist since if we explicitly define the log group and the `AWS::Serverless::Function` expands to also create the same log group, will the stack fail anyway?) If you want a hand to deploy this into your dev environment just let me know, roughly you can use

```
aws-vault exec di-ipv-dev -- sam deploy --config-env dev-danw --config-file samconfig.toml --stack-name ipv-core-back-dev-danw --region eu-west-2
```
but with your values :)

vam-gds (approved):
lgtm

PR #255, opened 2022-05-13
PYIC-1186 Get VTM claim value from config and set on identity response

blakeyp:
### What changed

Get VTM claim value from config and set on identity response

### Why did it change

To ensure we're returning VTM in the identity claims

### Issue tracking
- [PYIC-1186](https://govukverify.atlassian.net/browse/PYIC-1186)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -144,9 +144,9 @@ Resources:
       AutoPublishAlias: live
       ProvisionedConcurrencyConfig:
         !If
-        - AddProvisionedConcurrency
-        - ProvisionedConcurrentExecutions: !Ref ProvisionedConcurrentExecutions
-        - !Ref AWS::NoValue
+          - AddProvisionedConcurrency
+          - ProvisionedConcurrentExecutions: !Ref ProvisionedConcurrentExecutions
+          - !Ref AWS::NoValue
 
   IPVAccessTokenFunctionLogGroup:
     Type: AWS::Logs::LogGroup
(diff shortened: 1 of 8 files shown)


deploy/template.yaml
```diff
@@ -545,6 +545,8 @@ Resources:
                 - 'kms:GenerateDataKey'
               Resource:
                 - !ImportValue AuditEventQueueEncryptionKeyArn
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub ${Environment}/core/coreVtmClaim
```
dannyhayes-gds: I think the param name is slightly wrong here. You are missing the "self" bit in the middle.

```suggestion
            ParameterName: !Sub ${Environment}/core/self/coreVtmClaim
```

deploy/template.yaml
```diff
@@ -787,4 +789,4 @@ Outputs:
     Description: Core Back External API Gateway ID
     Export:
       Name: !Sub "${AWS::StackName}-IPVCoreExternalAPIGatewayID"
-    Value: !Ref IPVCoreExternalAPI
+    Value: !Ref IPVCoreExternalAPI
```
dannyhayes-gds: Could you add an empty end of file line here

dannyhayes-gds (dismissed):

kerrr (dismissed):
Hold this a sec

dannyhayes-gds (approved):

kerrr (approved):
I believe Danny H is following up with a bit of a refactor of the CredientialIssuerStart lambda and test to make some of these changes unnecessary. But good to merge now. 🚀

PR #269, opened 2022-05-16
PYIC-992: Validate verifiable credential JWT received from CRIs

Wynndow:
**This should not be merged before https://github.com/alphagov/di-ipv-core-back/pull/270**

### What changed

Validate verifiable credential JWT received from CRIs

### Why did it change

[PYIC-992: Include enc key and audience in CRI config object](https://github.com/alphagov/di-ipv-core-back/commit/259b6db6c7b881a3122603617381810e383234d0) 

Add a credential issuers public encryption key and audience to the
CredentialIssuerConfig object.

We had these configuration details of CRIs defined separately. This is
most likely as they had been newly added. This refactors the object so
they are deserialized and available on the object with just one call the
parameter store.
  
[PYIC-992: Make ObjectMapper static field on RequestHelper](https://github.com/alphagov/di-ipv-core-back/commit/dadc6b1ba825306788d1dd854e6e7740246b07b9) 

An object mapper is a heavy object. We probably shouldn't be using it at
all in Lambda world. By moving it to a static field we can reduce its
impact but just creating it once, rather than every time the request
helper is called.
  
[PYIC-992: Don't send plain JWT to VC endpoint](https://github.com/alphagov/di-ipv-core-back/commit/7cca6f2eb0fed6b41912624e07af37d538b02773) 

Previously we were sending the subject of the VC to the credential
issues in a plain JWT, when requesting the VC.

We now send the subject in the JAR when we start the journey. We can
remove the plain JWT cruft.
  
[PYIC-922: Validate VC JWTs in new Verifier](https://github.com/alphagov/di-ipv-core-back/commit/3bf0296f04308d8bc73cbbfafbcf989cdda4fa1e) 

Adds a new class specifically for validating JWTs received from CRIs.
This moves the existing signature checking logic from credential issuer
service into this class, and extends it with a check of the claims it
should contain.
  
[PYIC-922: Fetch userId from session to validate VC JWT subject](https://github.com/alphagov/di-ipv-core-back/commit/ad346492f1092cadaa3105ef168223804d14d0d6) 

We are storing the userID received from orch in the JAR they send us in
the session table. This is the value that's used as the subject when
making requests to CRIs. This is then passed back to us as the subject
of the VC JWT.

This pulls that value from the session and validates subject in the JWT
using it.

This also DRYs up the CRI return lambda tests a bunch.

### Issue tracking

- [PYIC-992](https://govukverify.atlassian.net/browse/PYIC-992)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -319,6 +319,7 @@ Resources:
         Variables:
           CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX: !Sub "/${Environment}/core/credentialIssuers"
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsTable.Arn]]
+          IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           SIGNING_KEY_ID_PARAM: !Sub "/${Environment}/core/self/signingKeyId"
           ENVIRONMENT: !Sub "${Environment}"
           POWERTOOLS_SERVICE_NAME: !Sub cri-return-${Environment}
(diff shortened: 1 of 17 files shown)


kerrr (dismissed):
🚀

dannyhayes-gds (approved):

PR #271, opened 2022-05-17
BAU: Define common versions for libs

Wynndow:
### What changed

Define common versions for libs

### Why did it change

We were using different versions of the same libs in different lambdas.

This defines a structure in the root project where we can define the
versions we want to use and then pull them in in the sub projects.

This makes it easy to bump all versions in case of a security issue, as
well as ensuring consistency to help diagnose weird bugs.

This also sorts all the dependencies to scratch an itch.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,3 +30,26 @@ spotless {
 		endWithNewline()
 	}
 }
+
+ext {
+	dependencyVersions = [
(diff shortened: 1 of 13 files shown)


Danomanic (approved):
Looks good, and saves us manually bumping all the gradle modules.

PR #274, opened 2022-05-19
Update passport gpg validity property name

dannyhayes-gds:
### What changed
Update the vot generation to check for the new passport gpd validity property "validityScore".

### Why did it change
We use this value to determine the vot claim. Currently it was always failing to read the validityScore value and always returning P0 as the vot value.


lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -30,7 +30,7 @@ public class UserIdentityService {
     private static final String PASSPORT_CRI_TYPE = "ukPassport";
     private static final String FRAUD_CRI_TYPE = "fraud";
     private static final String KBV_CRI_TYPE = "kbv";
-    private static final String GPG_45_VALIDITY_PROPERTY_NAME = "validity";
+    private static final String GPG_45_VALIDITY_PROPERTY_NAME = "validityScore";
     private static final String GPG_45_FRAUD_PROPERTY_NAME = "fraud";
     private static final String GPG_45_VERIFICATION_PROPERTY_NAME = "verification";
```
blakeyp: Do these also need updating, to identityFraudScore and verificationScore? Had to do similar in https://github.com/alphagov/di-ipv-stubs/pull/99/files

blakeyp (dismissed):

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
dannyhayes-gds: Ah yeah good shout, yeah they will as well if we are changing them on the stubs. I'll change that here then and we will need to get this and your PR merged together

blakeyp (approved):

PR #276, opened 2022-05-20
PYIC-1151: Dedupe sharedClaims in CredentialIssuerStartHandler response JAR.

Danomanic:
### What changed
Deduplicate sharedClaims in the JAR response.

### Issue tracking

- [PYIC-1151](https://govukverify.atlassian.net/browse/PYIC-1151)

### DIFF
--- a/lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
+++ b/lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
@@ -53,7 +53,13 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 import static uk.gov.di.ipv.core.credentialissuer.CredentialIssuerStartHandler.SHARED_CLAIMS;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.*;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.CREDENTIAL_ATTRIBUTES_1;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.CREDENTIAL_ATTRIBUTES_2;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.CREDENTIAL_ATTRIBUTES_3;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.EC_PRIVATE_KEY;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.EC_PUBLIC_JWK;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.RSA_ENCRYPTION_PRIVATE_KEY;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.RSA_ENCRYPTION_PUBLIC_JWK;
 import static uk.gov.di.ipv.core.library.helpers.VerifiableCredentialGenerator.generateVerifiableCredential;
 import static uk.gov.di.ipv.core.library.helpers.VerifiableCredentialGenerator.vcClaim;
 
--- a/lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
+++ b/lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
@@ -233,10 +233,43 @@ class CredentialIssuerStartHandlerTest {
 
         assertEquals(2, (vcAttributes.get("name")).size());
         assertEquals(3, (vcAttributes.get("address")).size());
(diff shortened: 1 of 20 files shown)


lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
```diff
@@ -233,12 +233,45 @@ private void assertSharedClaimsJWTIsValid(String request)
...
+
+        SignedJWT signedJWT = SignedJWT.parse(jweObject.getPayload().toString());
+        JsonNode claimsSet = objectMapper.readTree(signedJWT.getJWTClaimsSet().toString());
+
+        JsonNode sharedClaims = claimsSet.get(SHARED_CLAIMS);
+        assertEquals(1, sharedClaims.get("name").size());
+        assertEquals(2, sharedClaims.get("birthDate").size());
+        assertEquals(2, sharedClaims.get("address").size());
+    }
+
```
blakeyp, line 275: Is it worth adding a test for the edge case in the ticket, where eg some parts of the name are the same but not all?

Danomanic: Patrick has added a test to confirm this edge case is dealt with by using `@EqualsAndHashCode`

lambdas/credentialissuerstart/src/test/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandlerTest.java
```diff
@@ -53,12 +53,7 @@
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.when;
 import static uk.gov.di.ipv.core.credentialissuer.CredentialIssuerStartHandler.SHARED_CLAIMS;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.CREDENTIAL_ATTRIBUTES_1;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.CREDENTIAL_ATTRIBUTES_2;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.EC_PRIVATE_KEY;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.EC_PUBLIC_JWK;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.RSA_ENCRYPTION_PRIVATE_KEY;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.RSA_ENCRYPTION_PUBLIC_JWK;
+import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.*;
```
Wynndow: Could this get switched back to the individual imports?
I'm wondering if we can get spotless to pick this up...

Danomanic: Yep, will sort this. Think Patrick's IntelliJ needs configuring for this.

Danomanic: Fixed.

Wynndow:
~I'm getting a very weird error when trying to run your new tests locally. An aspectJ error, seems to be related to the X-ray tracing annotations. Did you run into anything like this?~
It was IntelliJ being IntelliJ. Ignore me.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/Address.java
```diff
@@ -2,13 +2,16 @@
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.EqualsAndHashCode;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
 @ExcludeFromGeneratedCoverageReport
+@EqualsAndHashCode
```
Wynndow, line 11: This is pretty nifty. 

Why do we need to define them for all of these domain objects though?

Wynndow: Is it the defining of these methods that is actually fixing the bug? So that the entries in the sets can be compared correctly?

Danomanic: @Wynndow - We needed to add this, as when we add the name, address, dob to the set it was adding duplicates.

Java was using the default equals and hashcode method which was not comparing the fields within the objects - just that they matched in memory.

>  Its default implementation simply checks the object references of two objects to verify their equality. By default, two objects are equal if and only if they are refer to the same memory location.

By overriding equals and hashcode, we can ensure just the fields are compared. This had to be done for Name, NameParts, Address, BirthDate domain objects.

Wynndow (approved):

PR #278, opened 2022-05-23
BAU: Remove validFrom/Until dates from name parts

Wynndow:
### What changed

Remove validFrom/Until dates from name parts

### Why did it change

Our `NameParts` class, used to construct and serialize values for the
shared attributes JWT, was including a validFrom and validUntil field
for each name part. We had included an annotation to instruct Jackson to
only include non null values.

When an object is set as the value of a claim in a Nimbus JWT, Nimbus
does not use Jackson to serialize it. This means that the annotations
were being ignore and the validFrom and validUntil were always being
included with null values.

The address CRI was not expecting to receive these fields and this was
causing an error.

Currently we're not expecting to include these fields at this level, so
we should drop them. If we decided to start setting these values in
future we can revisit this and decide on the best way to handle it.


danworth (approved):
I'm not too familiar with this flow but your reasoning makes sense 👍

dannyhayes-gds (approved):

PR #284, opened 2022-05-23
PYIC-1277: Update journey engine to handle cri error responses

dannyhayes-gds:
### What changed
Updated the journey engine to now handle /journey/error journey steps and render the recoverable error page

### Why did it change
This now means we can control how to handle these error scenarios through the journey engine. In the future the big benefit will be that we can now handle these error scenarios by potentially attempting alternative CRI routes where possible.

### Issue tracking

- [PYIC-1277](https://govukverify.atlassian.net/browse/PYIC-1277)

### DIFF
--- a/lambdas/journeyengine/src/main/java/uk/gov/di/ipv/core/journeyengine/JourneyEngineHandler.java
+++ b/lambdas/journeyengine/src/main/java/uk/gov/di/ipv/core/journeyengine/JourneyEngineHandler.java
@@ -150,7 +150,7 @@ public class JourneyEngineHandler
                         updateUserState(CRI_ADDRESS, ipvSessionItem);
                         builder.setJourneyResponse(
                                 new JourneyResponse(criStartUri + ADDRESS_CRI_ID));
-                    } else if (journeyStep.equals(ERROR_STEP)) {
+                    } else {
                         updateUserState(CRI_ERROR, ipvSessionItem);
                         builder.setPageResponse(new PageResponse(PYI_TECHNICAL_ERROR_PAGE.value));
                     }
@@ -159,7 +159,7 @@ public class JourneyEngineHandler
                     if (journeyStep.equals(NEXT_STEP)) {
                         updateUserState(CRI_FRAUD, ipvSessionItem);
                         builder.setJourneyResponse(new JourneyResponse(criStartUri + FRAUD_CRI_ID));
-                    } else if (journeyStep.equals(ERROR_STEP)) {
+                    } else {
                         updateUserState(CRI_ERROR, ipvSessionItem);
                         builder.setPageResponse(new PageResponse(PYI_TECHNICAL_ERROR_PAGE.value));
                     }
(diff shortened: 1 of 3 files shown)


Danomanic (dismissed):
LGTM

lambdas/journeyengine/src/main/java/uk/gov/di/ipv/core/journeyengine/JourneyEngineHandler.java
```diff
@@ -142,34 +146,57 @@ private JourneyEngineResult executeJourneyEvent(
...
+                        builder.setPageResponse(new PageResponse(PYI_TECHNICAL_ERROR_PAGE.value));
+                    }
                     break;
                 case CRI_ADDRESS:
-                    updateUserState(CRI_FRAUD, ipvSessionItem);
-                    builder.setJourneyResponse(new JourneyResponse(criStartUri + FRAUD_CRI_ID));
+                    if (journeyStep.equals(NEXT_STEP)) {
+                        updateUserState(CRI_FRAUD, ipvSessionItem);
+                        builder.setJourneyResponse(new JourneyResponse(criStartUri + FRAUD_CRI_ID));
+                    } else {
```
Wynndow: I feel like this should still be a `else if` block, to make it clear that we're only doing this if the step is an error step. That way if we get an unexpected step for some reason it'll still drop out to the new `default` case.

lambdas/journeyengine/src/main/java/uk/gov/di/ipv/core/journeyengine/JourneyEngineHandler.java
```diff
@@ -142,34 +146,57 @@ private JourneyEngineResult executeJourneyEvent(
...
                     builder.setJourneyResponse(new JourneyResponse(journeyEndUri));
                     break;
                 case DEBUG_PAGE:
                     builder.setPageResponse(new PageResponse(DEBUG_PAGE.value));
                     break;
-                case CRI_ERROR:
-                    builder.setPageResponse(new PageResponse(IPV_ERROR_PAGE.value));
-                    break;
+                default:
+                    updateUserState(CRI_ERROR, ipvSessionItem);
```
Wynndow, line 207: Might be good to have some logging in here to show we've dropped out to the `default` case, which we probably should never actually do.

lambdas/journeyengine/src/main/java/uk/gov/di/ipv/core/journeyengine/JourneyEngineHandler.java
dannyhayes-gds: It won't go to the default case for this kind of scenario. That's only for the actual current user status, this is just checking which journey step we get

Wynndow: Just to capture what we just chatted about IRL - we're going to make the `else` block an `else if` to be specific about the step we've received, and add an `else` block that will throw if we've received an unexpected step.

Wynndow (approved):

PR #288, opened 2022-05-24
PYIC-1270: Add validatecricheck lambda 

Wynndow:
### What changed

Add validatecricheck lambda 

### Why did it change

Add a lambda to validate the check returned by a CRI meets our needs or
reaching an M1A level of assurance.

If it doesn't meet the level required we return a "fail" journey
response. This will lead the user down a session end path (handled by
the journey engine).

### Issue tracking

- [PYIC-1270](https://govukverify.atlassian.net/browse/PYIC-1270)


deploy/template.yaml
```diff
@@ -742,6 +742,54 @@ Resources:
...
+      Environment:
+        Variables:
+          ENVIRONMENT: !Ref Environment
+          POWERTOOLS_SERVICE_NAME: !Sub validate-cri-check-${Environment}
+          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsTable
+      Policies:
+        - DynamoDBReadPolicy:
+            TableName: !Ref UserIssuedCredentialsTable
+      Events:
+        IPVCoreInternalAPI:
```
dannyhayes-gds: I think this has changed now. The internal gateway has changed to the private gateway. You might need to rebase off main to get those changes and see how this is now done

Wynndow: Oh interesting - thanks i'll fix it up

Wynndow: Done

deploy/template.yaml
```diff
@@ -766,6 +766,54 @@ Resources:
...
```
dannyhayes-gds: Think this line needs changing as well?
```suggestion
        IPVCorePrivateAPI:
```

dannyhayes-gds:
Will probs be worth adding details of this lambda to the open-api docs? Although it will be worth double checking with @danworth as I believe there was some issues with using the the open-api stuff with the new private gateways so maybe its not being used anymore

deploy/template.yaml
Wynndow: Bah. Have fixed up. Thanks!
Just checking in with Dan about the open-api gateway spec. I can see the external one being referenced in the template, but not the internal one for the private API.

dannyhayes-gds (approved):

PR #290, opened 2022-05-25
PYIC-1296: Validate fraud check

Wynndow:
### What changed

Validate fraud check

### Why did it change

[PYIC-1296: Add safety checks in Evidence validator](https://github.com/alphagov/di-ipv-core-back/commit/d2117749a8ef626f2d365ba1c358ba7c74eea37a) 

The VCs we receive should always contain an evidence property that is an
array. But it's better to be safe than sorry and do a couple of checks
before trying to process it.

This catches a class cast exception in case the array is not an array,
and adds a null check in case the array does not exist.
  
[PYIC-1296: Validate fraud check](https://github.com/alphagov/di-ipv-core-back/commit/4c3e6c433aa8233a0b3ce5762fa5fc283cd54b57) 

This adds a new evidence validator to encapsulate the rules for
validating a fraud check and hooks it up to the CriCheck validator.

The ticket for this work specifically calls out that the identity fraud
score should be 1 or 2. Looking at GPG45, it's possible for this score
to also be 3: https://www.gov.uk/government/publications/identity-proofing-and-verification-of-an-individual/how-to-prove-and-verify-someones-identity#identity-fraud

I've set the rule to just ensure that the value is 1 or greater. I can't
see that we'd fail someone who received a score of three, even though
we'll never actually receive them.

### Issue tracking

- [PYIC-1296](https://govukverify.atlassian.net/browse/PYIC-1296)


lambdas/validatecricheck/src/main/java/uk/gov/di/ipv/core/validatecricheck/validation/FraudEvidenceValidator.java
```diff
@@ -0,0 +1,22 @@
...
+import uk.gov.di.ipv.core.validatecricheck.domain.FraudEvidence;
+
+public class FraudEvidenceValidator implements CriEvidenceValidator {
+    public static final int ONLY_ELEMENT = 0;
+    public static final int GPG_45_M1A_FRAUD_SCORE = 1;
+    private static final Gson gson = new Gson();
+
+    @Override
+    public boolean validate(JSONArray evidenceArray) {
+        FraudEvidence evidence =
```
dannyhayes-gds, line 14: Just want to double check here, but will GSON be able to correctly convert this if the ci property isn't there?

Wynndow: Yep - there's a test for that scenario here: https://github.com/alphagov/di-ipv-core-back/pull/290/files#diff-ffd5c03bafeb836c40ed264a76fa32747b26f16b90866250e392eccfac2b3c02R75

Although I've realised that the test name is wrong 🙄

dannyhayes-gds: But isn't that test for if the property exists but is an empty list. I think there is the possibility that the ci property could be missing all together?

dannyhayes-gds: As discussed there is another test here that uses a signed JWT without a ci property that proves this will work as expected.

dannyhayes-gds (approved):

PR #292, opened 2022-05-26
PYIC-1247: Load api key from secret manager and include on header for the requests to the passport api

dannyhayes-gds:
### What changed
Load a CRI's private api key from AWS secrets manager if it exists. If it doesn't exist the exception will be caught and null is returned.

The private api key is then sent in the "x-api-key" header to the CRI's /token and /credential requests if it exists.

### Why did it change
The CRI's /token and /credential endpoints are now being secured with private api keys.

### Issue tracking

- [PYIC-1247](https://govukverify.atlassian.net/browse/PYIC-1247)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
@@ -1,10 +1,19 @@
 package uk.gov.di.ipv.core.library.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
 import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
+import software.amazon.awssdk.services.secretsmanager.model.DecryptionFailureException;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
+import software.amazon.awssdk.services.secretsmanager.model.InternalServiceErrorException;
+import software.amazon.awssdk.services.secretsmanager.model.InvalidParameterException;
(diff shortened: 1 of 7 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
```diff
@@ -229,4 +254,55 @@ public String getCoreVtmClaim() {
...
+
+        try {
+            String secretManagerKeyValueJson = getSecretsManagerValue(valueRequest);
+
+            if (secretManagerKeyValueJson != null) {
+                Map<String, String> secret =
+                        objectMapper.readValue(secretManagerKeyValueJson, new TypeReference<>() {});
+                return secret.get(API_KEY);
+            }
+            return null;
```
danworth, line 275: Generally we should prefer to use `Optional` rather than returning `null`s so callers of this function must consider the null case and can do so in a "nicer" way, but I think refactoring to do that now would mean changing quite a bit in this PR so in the interest of time and that this works lets move forward. Something to consider for next time if you like 👍

lib/src/main/java/uk/gov/di/ipv/core/library/service/CredentialIssuerService.java
```diff
@@ -40,6 +40,7 @@
 public class CredentialIssuerService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CredentialIssuerService.class);
+    private static final String API_KEY_HEADER = "x-api-key";
```
danworth, line 43: Are we free to chose a header field or is this restricted to api gateway implementation? Just wondering because KBV we're asking which header field we intend to use so they can be consistent.

danworth (dismissed):
Looks good, a little comment about preferring `Optionals` in places but that can be for another time 👍

lib/src/main/java/uk/gov/di/ipv/core/library/service/CredentialIssuerService.java
dannyhayes-gds: This seemed to be the one that the AWS docs said to use. I'm not aware of another header that we can use

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
dannyhayes-gds: Yeah good shout, I'll do a bit of a refactor of this later today after we can see it working

danworth (approved):
re-approve after rebasing 🏅

PR #295, opened 2022-05-27
PYIC-1279: Return error journey response if Cri error is received

dannyhayes-gds:
### What changed
Return a /journey/error response when we receive an error from a CRI.

### Why did it change
If we receive an Oauth error from a CRI then we need to render the the recoverable error page which will be handled from within the journeyEngine lambda. Returning a /journey/error response ensures the journey engine renders the expected error page.

### Issue tracking

- [PYIC-1279](https://govukverify.atlassian.net/browse/PYIC-1279)


lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -35,6 +35,7 @@
     private static final Logger LOGGER =
             LoggerFactory.getLogger(CredentialIssuerReturnHandler.class);
     private static final String CRI_VALIDATE_ENDPOINT = "/journey/cri/validate/";
+    private static final String JOURNEY_ERROR_ENDPOINT = "/journey/error";
```
Wynndow, line 38: I'm wondering if these endpoints should be in a shared place in the lib somewhere. I think we're using them in a few places now. Not for this PR though.

Wynndow (approved):

PR #296, opened 2022-05-27
PYIC-1313: Update error audit event to include error details

dannyhayes-gds:
### What changed
Updated AuditEvents class to have an optional "extensions" property, and use this new field to store details of the error response we receive from a CRI.

### Why did it change
The extensions property is an optional field for our audit events where we can store any error properties. 

See here for details regarding this property value and how it should be used as described in the TxMA data dictionary: https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0023-crossteam-audit-and-analysis-eventmessage-payload.md

### Issue tracking

- [PYIC-1313](https://govukverify.atlassian.net/browse/PYIC-1313)

### DIFF
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -76,7 +76,7 @@ public class CredentialIssuerReturnHandler
                 RequestHelper.convertRequest(input, CredentialIssuerRequestDto.class);
 
         try {
-            auditService.sendAuditEvent(AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED, null);
+            auditService.sendAuditEvent(AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED);
 
             var errorResponse = validate(request);
 
@@ -105,9 +105,9 @@ public class CredentialIssuerReturnHandler
                             .getUserId());
 
             auditService.sendAuditEvent(
-                    AuditEventTypes.IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED, null);
+                    AuditEventTypes.IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED);
 
-            auditService.sendAuditEvent(AuditEventTypes.IPV_VC_RECEIVED, null);
+            auditService.sendAuditEvent(AuditEventTypes.IPV_VC_RECEIVED);
 
             credentialIssuerService.persistUserCredentials(
                     verifiableCredential.serialize(), request);
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -76,7 +76,7 @@ public class CredentialIssuerReturnHandler
                 RequestHelper.convertRequest(input, CredentialIssuerRequestDto.class);
 
         try {
-            auditService.sendAuditEvent(AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED);
+            auditService.sendAuditEvent(AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED, null);
 
             var errorResponse = validate(request);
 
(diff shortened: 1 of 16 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -24,21 +25,24 @@ public static AmazonSQS getDefaultSqsClient() {
...
     }
 
-    private String generateMessageBody(AuditEventTypes eventType) throws JsonProcessingException {
-        AuditEvent auditEvent = new AuditEvent((int) Instant.now().getEpochSecond(), eventType);
+    private String generateMessageBody(AuditEventTypes eventType, AuditExtensionParams extensions)
+            throws JsonProcessingException {
+        AuditEvent auditEvent =
+                new AuditEvent((int) Instant.now().getEpochSecond(), eventType, extensions);
 
         ObjectMapper objectMapper = new ObjectMapper();
```
Wynndow: I know it's not your change, but can this objectMapper be made static? Newing one up for every audit event is going to be a bit of a drain I think.

lib/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -24,21 +25,24 @@ public static AmazonSQS getDefaultSqsClient() {
         return AmazonSQSClientBuilder.defaultClient();
     }
 
-    public void sendAuditEvent(AuditEventTypes eventType) throws SqsException {
+    public void sendAuditEvent(AuditEventTypes eventType, AuditExtensionParams extensions)
```
Wynndow, line 34: Could you create a new method signature that replicates the old one, that just calls this with `null` for extensions? That way you don't have to call this method with `null` in places you're not passing in the extensions?

Wynndow (commented):
What's the motivation for including these error details?

lib/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
dannyhayes-gds: Yeah good shout, I'll update it

lib/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
dannyhayes-gds: Yep will do 👍

dannyhayes-gds:
Adding the error details to the audit event originally came from here - https://govukverify.atlassian.net/browse/PYIC-1141 and its also an expected thing from the TxMA side of things. Here in their data dictionary it describes the purpose of the "extensions" field which includes error responses - https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0023-crossteam-audit-and-analysis-eventmessage-payload.md

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -105,9 +105,9 @@ public APIGatewayProxyResponseEvent handleRequest(
                             .getUserId());
 
             auditService.sendAuditEvent(
-                    AuditEventTypes.IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED, null);
+                    AuditEventTypes.IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED);
 
-            auditService.sendAuditEvent(AuditEventTypes.IPV_VC_RECEIVED, null);
+            auditService.sendAuditEvent(AuditEventTypes.IPV_VC_RECEIVED);
```
Wynndow: It seems odd we have these two events...

lib/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -25,6 +27,19 @@ public static AmazonSQS getDefaultSqsClient() {
...
+            SendMessageRequest sendMessageRequest =
+                    new SendMessageRequest()
+                            .withQueueUrl(queueUrl)
+                            .withMessageBody(generateMessageBody(eventType, null));
+
+            sqs.sendMessage(sendMessageRequest);
+        } catch (JsonProcessingException e) {
+            throw new SqsException(e);
+        }
+    }
```
Wynndow: I think I was thinking of something more like:
```
    public void sendAuditEvent(AuditEventTypes eventType) throws SqsException {
        sendAuditEvent(eventType, null)
    }
```

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
dannyhayes-gds: Yeah good point. Not sure why we'd need both here they seem like the same thing. Plus I can't find anything about the IPV_CREDENTIAL_RECEIVED_AND_SIGNATURE_CHECKED one here in the list of events - https://docs.google.com/spreadsheets/d/1cLjAdRcpw94uYLnrt7FTeFX5XM_7MLRTDWdf-Zbobj8/edit?pli=1#gid=485429246

dannyhayes-gds: Might be a bit of a cleanup thing to do after to double check that we don't need the 1st one

Wynndow (approved):

PR #297, opened 2022-05-28
PYIC-1337: Update to start returning journey errors

dannyhayes-gds:
### What changed
Update CriStart and ValidateCri lambda's to start returning journey errors instead of 500 error responses.

### Why did it change
These are recoverable error scenario's, so returning an error journey will mean we still attempt to display the recoverable error page and return the user back to orc.

### Issue tracking

- [PYIC-1337](https://govukverify.atlassian.net/browse/PYIC-1337)

### DIFF
--- a/lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
+++ b/lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
@@ -22,6 +22,7 @@ import uk.gov.di.ipv.core.credentialissuer.domain.CriResponse;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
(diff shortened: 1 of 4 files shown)


lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
```diff
@@ -60,6 +61,7 @@
     public static final int BAD_REQUEST = 400;
     public static final int OK = 200;
     public static final String SHARED_CLAIMS = "shared_claims";
+    public static final String ERROR_JOURNEY_STEP_URI = "/journey/error";
```
Wynndow: I'm wondering it it's worth making this:
`public static final JourneyResponse errorJourneyResponse = new JourneyResponse(ERROR_JOURNEY_STEP_URI)`, and then just referencing it in the error returns below...

Wynndow (dismissed):

lambdas/credentialissuerstart/src/main/java/uk/gov/di/ipv/core/credentialissuer/CredentialIssuerStartHandler.java
dannyhayes-gds: Yeah good shout, I'll make the change 👍

Wynndow (approved):

PR #301, opened 2022-06-04
PYIC-1338: Add recoverable error handling for SessionStart lambda

dannyhayes-gds:
### What changed
Updated the session start lambda JAR validation to identify the difference between a recoverable and non recoverable JAR validation error. A recoverable scenario is after the JAR has been decrypted, signature validated and the redirect_uri param is validated.

If a recoverable error occurs then we set the user starting state to be FAILED_CLIENT_JAR with the "error" and "error_description" params being stored on the session object.

Journey engine has been updated to handle the FAILED_CLIENT_JAR state and will ensure the SessionEnd lambda is called where the "error" and "error_description" params are added to the redirect_uri back to orchestrator.

### Why did it change
In these scenarios we want to redirect back to the orchestrator with details of the error. We do this via the Journey engine by moving the user into a FAILED_CLIENT_JAR state which goes straight into the SessionEnd lambda.

### Issue tracking

- [PYIC-1338](https://govukverify.atlassian.net/browse/PYIC-1338)


lambdas/sessionstart/src/main/java/uk/gov/di/ipv/core/session/IpvSessionStartHandler.java
```diff
@@ -112,6 +114,21 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            ClientSessionDetailsDto clientSessionDetailsDto =
+                    generateErrorClientSessionDetails(e.getRedirectUri(), e.getClientId());
+
+            String ipvSessionId =
+                    ipvSessionService.generateIpvSession(
+                            clientSessionDetailsDto, e.getErrorObject());
+
+            Map<String, String> response = Map.of(IPV_SESSION_ID_KEY, ipvSessionId);
+
+            return ApiGatewayResponseGenerator.proxyJsonResponse(HttpStatus.SC_OK, response);
```
Wynndow: Small thing, could this catch go above the `ParseExeption` catch?  That way the `HttpStatus.SC_OK` responses are grouped at the top, and the bad ones at the bottom. Feel free to completely ignore this, it just made sense it my head.

lambdas/sessionstart/src/main/java/uk/gov/di/ipv/core/session/IpvSessionStartHandler.java
```diff
@@ -159,6 +176,12 @@ private ClientSessionDetailsDto generateClientSessionDetails(
                 isDebugJourney);
     }
 
+    @Tracing
+    private ClientSessionDetailsDto generateErrorClientSessionDetails(
+            String redirectUri, String clientId) {
+        return new ClientSessionDetailsDto(null, clientId, redirectUri, null, null, false);
```
Wynndow: I think we need to [include the state value](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) in this if one exists, so we can pass it back with the error and error_description params.

dannyhayes-gds: Is that a thing to return the state in a Oauth error response? I couldn't see anything about it in the Oauth RFC - https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

lambdas/sessionstart/src/main/java/uk/gov/di/ipv/core/session/IpvSessionStartHandler.java
dannyhayes-gds: Sounds good I'll update it 👍

lambdas/sessionstart/src/main/java/uk/gov/di/ipv/core/session/IpvSessionStartHandler.java
dannyhayes-gds: As discussed, spotted that it does list state on the next page of the RFC. I'll update this PR to return a state param if it exists during an error response

Wynndow (approved):

PR #302, opened 2022-06-05
PYIC-1368: generate the identity claim for the UserIdentity response

dannyhayes-gds:
### What changed
Updated the UserIdentity endpoint to return the "identity" claim under the `https://vocab.account.gov.uk/v1/coreIdentity` property name. The claim will only be returned of a there is a vot claim value of P2.

### Why did it change
The identity claim will be used by SPOT and is only required on a successful identity check i.e vot value of P2.

### Issue tracking

- [PYIC-1368](https://govukverify.atlassian.net/browse/PYIC-1368)


lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -83,19 +93,22 @@ public UserIdentity generateUserIdentity(String ipvSessionId, String sub) {
...
-                .build();
+        UserIdentity.Builder userIdentityBuilder =
+                new UserIdentity.Builder().setVcs(vcJwts).setSub(sub).setVot(vot).setVtm(vtm);
+
+        if (vot.equals(VectorOfTrust.P2.toString())) {
+            Optional<IdentityClaim> identityClaim = generateIdentityClaim(credentialIssuerItems);
+            identityClaim.ifPresent(userIdentityBuilder::setIdentityClaim);
+        }
+
+        return userIdentityBuilder.build();
```
Wynndow: This logic seems to imply that we can send a response that has a VOT of P2, but doesn't have an identity claim. Is that possible?

dannyhayes-gds: Ah yeah I was a little unsure on how to handle this to be honest. It's very unlikely to happen. It should only be missing if the generateIdentityClaim method fails to parse the JSON or the "name" or "birthDate" params are missing from the credential we've retrieved.

I think this scenario shouldn't ever really happen and this seemed the most sensible way to handle it as this lambda is called directly by the orchestration so we can't use our normal error handling stuff here. Maybe in this case we would need to change the vot to be P0 instead if something has gone wrong?

I'll double check with @kerrr and see if he has a preference on how we should handle this.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class IdentityClaim {
+    private final List<Name> name;
+
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
```
Wynndow: Does this still work when the field is a list?

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
kerrr: We shouldn't return a `vot:P2` without an identity claim. If there was some sort of error creating this we should probably return a 500 http error.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
dannyhayes-gds: Hmmm not sure actually. I grabbed this from the passport code base. I'll double check as this might not be needed here

Wynndow (approved):

PR #303, opened 2022-06-06
PYIC-1369: Set backend session expiry time

Wynndow:
**See this config PR for adding the timeout parameter: https://github.com/alphagov/di-ipv-config/pull/628**

### What changed

Set backend session expiry time

### Why did it change

We're going to start checking if a users session has expired. To do this
we need to create an expiry time when their session is created. This timeout
should be pulled from config.

### Issue tracking

- [PYIC-1369](https://govukverify.atlassian.net/browse/PYIC-1369)


dannyhayes-gds:
The sessionStart lambda will need read permission for the new SSM param

dannyhayes-gds (approved):

PR #307, opened 2022-06-07
PYIC-1380: Add the ttl value to our dynamoDB items

dannyhayes-gds:
### What changed
Enable the TTL setting on our DynamoDB tables with the "ttl" attribute name.

Add the "ttl" property to all our DynamoDB items that retrieves the SSM backendSessionTtl param and sets an epoch seconds value on the ttl property. This value should be set to 24 hours

### Why did it change
This property will be used to automatically delete our dynamoDB items based of the ttl property on each item.

### Issue tracking

- [PYIC-1380](https://govukverify.atlassian.net/browse/PYIC-1380)


lib/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -45,6 +51,10 @@ public static DynamoDbEnhancedClient getClient(boolean isRunningLocally) {
     }
 
     public void create(T item) {
+        item.setTtl(
```
Wynndow, line 54: nice...

Wynndow (approved):

PR #308, opened 2022-06-08
PYIC-858: Return address attribute to orch

Wynndow:
### Why did it change

[PYIC-858: Update Address domain model](https://github.com/alphagov/di-ipv-core-back/commit/2053d0e52987aa886d9a206b97edd02b33b58e59) 

The domain model we had for an address was out of date. The address
structure is defined in RFC20: https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0020-address-structure.md
  
[PYIC-858: Return address attribute to orch](https://github.com/alphagov/di-ipv-core-back/commit/02d7a9b5651b5b799dd7b40625ed2730f5521988) 

This extracts the address from the VC returned by the address CRI, and
adds it as a standalone claim in the JSON returned to the orchestrator
from the core. The spec for this can be found in RFC14: https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0014-identity-ipv-core-oauth-profile.md

It was decided that we would only return this data if we're at P2, and
that scoring would be added further down the line.

I've deliberately avoided deserializing the JSON contained in the
address VC into an Address domain object. Partly because it doesn't feel
like the core's responsibility to validate the structure at all at this
point in the journey, and partly because it means that things won't
break if the structure is ever updated in the address CRI. We just pass
along what we've been given. If people feel we'd be better using the
Address domain object we already have (updated in previous commit) I'm
happy to update it.

### Issue tracking

- [PYIC-858](https://govukverify.atlassian.net/browse/PYIC-858)

Screenshot 2022-06-08 at 14 26 55


lib/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
```diff
@@ -1,9 +1,11 @@
 package uk.gov.di.ipv.core.library.domain;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.EqualsAndHashCode;
 
 import java.util.List;
 
+@EqualsAndHashCode
```
dannyhayes-gds, line 8: Why do we need this?

dannyhayes-gds (approved):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
Wynndow: It's for the test I extended in [UserIdentityHandlerTest.java](https://github.com/alphagov/di-ipv-core-back/pull/308/files#diff-f99dd793b7e45d90a250ac427fb8fa617e8f8e2ffa6fca67435ce9b2d8f649d9R148). Without adding this it was comparing the objects identities, rather than their equalities.

PR #311, opened 2022-06-09
Fix new env vars in template

dannyhayes-gds:
### What changed
Fix the new env vars within the template file


danworth (approved):
Looks good and cfn-lint is happy now 👍

PR #315, opened 2022-06-10
Bau use ssm for journey engine cri ids

dannyhayes-gds:
### What changed
Load the journey CRI id values from SSM instead of env vars.


Wynndow (approved):

Wynndow (commented):
Do you need to give AWS perms in the template to the lambda?

Wynndow (approved):

PR #322, opened 2022-06-16
BAU: Refactor Configuration Service to be minimal

Danomanic:
### What changed
- Move all Environment Variables into Enum class.
- Replace configuration get methods with single reusable method. This is to reduce the size of the ConfigurationService ready for further refactoring.
- Added Environment and Configuration Enum classes


lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
```diff
@@ -86,40 +90,30 @@ public SSMProvider getSsmProvider() {
...
-
-    public String getIpvSessionTableName() {
-        return System.getenv("IPV_SESSIONS_TABLE_NAME");
+    public String getEnvironmentVariable(EnvironmentVariable environmentVariable) {
+        return System.getenv(environmentVariable.name());
     }
 
-    public String getIpvJourneyCriStartUri() {
-        return System.getenv("IPV_JOURNEY_CRI_START_URI");
+    public String get(ConfigurationVariable configurationVariable) {
```
dannyhayes-gds: Minor thing here but could we maybe rename this method and the one below to something like
`getSsmParameter`

lib/src/main/java/uk/gov/di/ipv/core/library/config/EnvironmentVariable.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.di.ipv.core.library.config;
+
+public enum EnvironmentVariable {
+    ACCESS_TOKENS_TABLE_NAME,
```
dannyhayes-gds, line 4: Not something for this PR, but another nice improvement we could look into is seeing which of these env vars are actually needed. I have a feeling we have a few where maybe its not necessary and is just a bit too "over-configured", or could be moved over to SSM perhaps.

lib/src/main/java/uk/gov/di/ipv/core/library/validation/JarValidator.java
```diff
@@ -104,11 +110,13 @@ private void validateSignature(SignedJWT signedJWT, String clientId)
             } else {
                 concatSignatureJwt = signedJWT;
             }
-            boolean valid =
+            boolean valid;
```
dannyhayes-gds: Does the setting of the boolean value here need to be split off onto a new line? Might be neater to just have it set all in one go as we did before?

Danomanic: Good spot, not sure why I did that :)

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
Danomanic: Good idea, will do!

lib/src/main/java/uk/gov/di/ipv/core/library/config/EnvironmentVariable.java
Danomanic: Yeah, I think this work has shown how many Environment Variables we have, and can then plan to remove those that are not needed.

dannyhayes-gds (dismissed):

lib/src/test/java/uk/gov/di/ipv/core/library/helpers/AuthorizationRequestHelperTest.java
```diff
@@ -217,9 +221,10 @@ private void setupCredentialIssuerConfigMock() {
     }
 
     private void setupConfigurationServiceMock() {
-        when(configurationService.getCoreFrontCallbackUrl()).thenReturn(CORE_FRONT_CALLBACK_URL);
-        when(configurationService.getIpvTokenTtl()).thenReturn(IPV_TOKEN_TTL);
-        when(configurationService.getAudienceForClients()).thenReturn(IPV_ISSUER);
+        when(configurationService.getSsmParameter(ConfigurationVariable.CORE_FRONT_CALLBACK_URL))
+                .thenReturn(CORE_FRONT_CALLBACK_URL);
```
blakeyp: Minor, but to prevent the naming clash/make it clearer here could rename the static class vars with `MOCK_` prefix or something? (ignore if this goes against convention 😄)

Danomanic: Nice spot. Changed!

blakeyp (approved):

PR #326, opened 2022-06-22
PYIC-1485: Log in JSON format

Wynndow:
### What changed

Log in JSON format
Another ticket is to actually make good use of the new capabilities introduced here

### Why did it change

This leverages the AWS docs found here:
https://docs.aws.amazon.com/lambda/latest/dg/java-logging.html#java-logging-log4j2
as well as RFC0030:
https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0030-cross-programme-logging-format.md
and the auth teams previous good work:
https://github.com/alphagov/di-authentication-api/blob/main/ipv-api/src/main/resources/log4j2.xml

This uses the `aws-lambda-java-log4j2` library to allow is to include
the AWS request ID in each log line.

It defines a couple of new gradle configurations in the root project
which are then pulled in by the sub projects to reduce repetition.

It adds a YAML configuration file to each sub project to configure the
logging. Currently we're only adding a couple of custom fields,
session-id and aws-request-id. We can add more as we identify what they
are. The log4j2 docs are good for understanding what's happening:
https://logging.apache.org/log4j/2.x/manual/configuration.html#YAML
https://logging.apache.org/log4j/2.x/manual/json-template-layout.html

As we're already using slf4j to create our loggers we don't need to
update our code o get hold of a logger which is nice.

### Issue tracking

- [PYIC-1485](https://govukverify.atlassian.net/browse/PYIC-1485)

Screenshot 2022-06-22 at 16 10 24

### DIFF
--- a/lambdas/accesstoken/src/main/resources/log4j2.yaml
+++ b/lambdas/accesstoken/src/main/resources/log4j2.yaml
@@ -1,0 +1,20 @@
+Configuration:
+  status: warn
+  appenders:
+    Lambda:
+      name: Lambda
+      JsonTemplateLayout:
+        eventTemplateUri: "classpath:JsonLayout.json"
+        eventTemplateAdditionalField:
+          - key: session-id
+            value: "$${ctx:sessionId:-unknown}"
+          - key: request-id
+            format: JSON
+            value: '{"$resolver": "pattern", "pattern": "%X{AWSRequestId}"}'
+  Loggers:
+    Root:
+      level: info
+      AppenderRef:
+        ref: Lambda
+
+
--- a/build.gradle
+++ b/build.gradle
@@ -38,10 +38,12 @@ ext {
 		awsJavaSdkSqs:'1.12.197',
 		awsLambdaJavaCore:'1.2.1',
 		awsLambdaJavaEvents:'3.11.0',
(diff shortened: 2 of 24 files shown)


kerrr (dismissed):

lambdas/accesstoken/src/main/resources/log4j2.yaml
```diff
@@ -0,0 +1,20 @@
...
+        eventTemplateUri: "classpath:JsonLayout.json"
+        eventTemplateAdditionalField:
+          - key: session-id
+            value: "$${ctx:sessionId:-unknown}"
+          - key: request-id
+            format: JSON
+            value: '{"$resolver": "pattern", "pattern": "%X{AWSRequestId}"}'
+  Loggers:
+    Root:
+      level: info
```
dannyhayes-gds, line 16: Does this mean we only this INFO level logs in this format? Or will it work for WARN/ERROR levels as well?

Wynndow: Yeah, it'll log from info up. If you set it to trace for example, you'd get all the logs. Set it to error and you'd just get error.

dannyhayes-gds (approved):

PR #327, opened 2022-06-23
PYIC-1485: Add identifiers to log lines

Wynndow:
### What changed

Add identifiers to log lines

### Why did it change
This updates each lambda to add the ipv session ID, the credential
issuer ID or the client ID when any are available to the logs produced.
These values will help us run queries over our logging to tie things together.
We might decide to add more in the future.

It uses a new helper to abstract some of this away. The helper adds the
values to the MDC (ThreadContext in log4j2): https://logging.apache.org/log4j/2.x/manual/thread-context.html

The MDC is thread specific but I've had a test and it looks like it gets
reused between Lambda invocations. Because of this we need to make sure
we're clearing the MDC at the end of every lambda. This has resulted in
each lambda getting wrapped in a try block with a `finally` to make sure
it's never missed.

This change has also altered the responses of some of the lambdas in
some situations. Rather than returning a journey error response, I've
changed it to just returning a 400. This happens where we the
ipv-session-id has not been sent in the header. If this ever happens
something is seriously wrong and trying to continue the journey is
probably a bad idea and we should just throw. Opinions encouraged.

### Issue tracking

- [PYIC-1485](https://govukverify.atlassian.net/browse/PYIC-1485)


Wynndow:
Closing in favour of https://github.com/alphagov/di-ipv-core-back/pull/333

PR #328, opened 2022-06-23
BAU: Replace UUIDs with a secure string

Wynndow:
### What changed

Replace UUIDs with a secure string

### Why did it change

I read this article and got overly paranoid about our use or UUIDs for
tokens. https://neilmadden.blog/2018/08/30/moving-away-from-uuids/

This updates our tokens to use a secure random string with 256 bits of
entropy rather than the 122 offered by UUIDs. This means we meet the
OAuth spec of at least 128 bits and preferably 160 bits.


danworth (approved):
Interesting article and seems to make sense 👍

PR #333, opened 2022-06-28
PYIC-1487: Add IDs to JSON logs

Wynndow:
### What changed

Add IDs to JSON logs

### Why did it change

This updates each lambda to add the ipv session ID, the credential
issuer ID, the client ID and a component ID to the logs as soon as
they're available.

The component ID is a constant thing we add to help with filtering logs.
The same value will be added in the frontend logs. I've intentionally
left it environment agnostic (so not a URL) so it's easy to handle
across all deployments.

The logging makes use of the AWS Powertools logging lib. This has
simplified our dependencies, and adds some extra fields to our logs for
free. It also handles clean up of the MDC for us.

Instantiation of the loggers no longer uses slf4j. We're not going to
be changing the logging framework anytime soon and it makes things a
little simpler.

This change has also altered the responses of some of the lambdas in
some situations. Rather than returning a journey error response, I've
changed it to just returning a 400. This happens where we the
ipv-session-id has not been sent in the header. If this ever happens
something is seriously wrong and trying to continue the journey is
probably a bad idea and we should just throw. Opinions encouraged.

### Issue tracking

- [PYIC-1487](https://govukverify.atlassian.net/browse/PYIC-1487)


integration-test/build.gradle
```diff
@@ -23,7 +23,8 @@ dependencies {
 			"com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$rootProject.ext.dependencyVersions.jackson",
 			"com.google.code.gson:gson:$rootProject.ext.dependencyVersions.gson",
 			"com.nimbusds:oauth2-oidc-sdk:$rootProject.ext.dependencyVersions.nimbusdsOauth2OidcSdk",
-			"org.slf4j:slf4j-simple:$rootProject.ext.dependencyVersions.slf4j",
+			"org.apache.logging.log4j:log4j-api:$rootProject.ext.dependencyVersions.log4j",
+			"org.apache.logging.log4j:log4j-core:$rootProject.ext.dependencyVersions.log4j",
```
kerrr, line 27: I think these come in transitively from powertools-logging?

Wynndow: This build.gradle is for the integration tests. I've not added powertools to them as they don't get run as lambdas. Have moved to log4j to be consistent with how we're creating loggers elsewhere, rather than sticking with slf4j.

kerrr (approved):

PR #335, opened 2022-06-30
BAU: Temporary logging to diagnose OAuth state issue

Wynndow:
### What changed

Temporary logging to diagnose OAuth state issue

### Why did it change

We saw a lot of issues with OAuth states not matching during the first
go live period. To help diagnose what's happening we can log them.

The state values are already in the logs, so this shouldn't be an issue.

### DIFF
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -14,6 +14,7 @@ import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import software.amazon.lambda.powertools.logging.Logging;
+import software.amazon.lambda.powertools.logging.LoggingUtils;
 import software.amazon.lambda.powertools.tracing.Tracing;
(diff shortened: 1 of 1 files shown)


lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -190,7 +191,11 @@ private Optional<ErrorResponse> validate(CredentialIssuerRequestDto request) {
             return Optional.of(ErrorResponse.MISSING_OAUTH_STATE);
         }
 
-        if (!request.getState().equals(getPersistedOauthState(request))) {
+        String persistedOauthState = getPersistedOauthState(request);
+        if (!request.getState().equals(persistedOauthState)) {
+            LoggingUtils.appendKey("expected_state", persistedOauthState);
+            LoggingUtils.appendKey("request_state", request.getState());
```
kerrr, line 197: Is this a place where a structured messsage would be better than attaching the value to the loggger?

kerrr (approved):
Happy to approve anyway for now

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
Wynndow: I think so - that's why I've appended the state values as fields with `appendKey()`

PR #337, opened 2022-06-30
BAU: Don't create table schemas more than required

kerrr:
### What changed

Only call `TableSchema.fromBean` once for every table we are using.

### Why did it change

`TableSchema.fromBean` is a heavy opperation and we were calling it every time we wnted to access the table. We don't need to do that.


Wynndow (commented):
Got a couple of unused fields now - dynamoDbEnhancedClient and tableName

Wynndow (approved):

PR #342, opened 2022-07-04
PYIC-1517: Validate Access Token has not expired.

Danomanic:
### What changed
Validate that the Access Token has not expired.

### Why did it change
"We set the lifetime of access tokens returned to the client to one hour. We don’t however validate this in the protected resource endpoint. We should reject access tokens that have expired."

### Issue tracking

- [PYIC-1517](https://govukverify.atlassian.net/browse/PYIC-1517)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/persistence/item/AccessTokenItem.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/persistence/item/AccessTokenItem.java
@@ -11,6 +11,7 @@ public class AccessTokenItem implements DynamodbItem {
     private String ipvSessionId;
     private String revokedAtDateTime;
     private long ttl;
+    private String expiryDateTime;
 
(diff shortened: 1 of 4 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/persistence/item/AccessTokenItem.java
```diff
@@ -11,6 +11,7 @@ public class AccessTokenItem implements DynamodbItem {
     private String ipvSessionId;
     private String revokedAtDateTime;
     private long ttl;
+    private String expiryDateTime;
```
blakeyp, line 14: Not a big deal, but I called this `accessTokenExpiryDateTime` in passport-back just to avoid confusing it with the `ttl` which is the expiry of the dynamodb record

blakeyp (approved):

PR #350, opened 2022-07-05
PYIC-1533: Normalise error responses

Wynndow:
### What changed

Normalise error responses

### Why did it change

This started as normalising the error responses from the user identity
endpoint to match the OAuth error response pattern of a json object with
`error` and `error_description` fields.

In a few places we throw an `HttpResponseExceptionWithErrorBody` error,
which includes one of our own `ErrorResponse` enums to describe what
happens. We were serializing them with `code` and `message` fields. This
meant that the client would receive different style error responses
depending on the error. This updates the `getErrorBody` method on the
`HttpResponseExceptionWithErrorBody` to serialize them in the same way
as OAuth errors. This has had knock on effects into other lambdas.

This may be a TERRIBLE idea.

But I'm hoping it will start a discussion about our use of our
`ErrorResponse` enum and why we do it that way, as I think it's
confusing.

This also refactors the validation in the CRI return lambda from using
optionals and to just throwing, which I think is a bit cleaner.

### Issue tracking

- [PYIC-1533](https://govukverify.atlassian.net/browse/PYIC-1533)


Danomanic (approved):
Brilliant refactor Chris.

PR #354, opened 2022-07-08
PYIC-1538: Revoke access tokens after being used in User Identity Handler.

Danomanic:
### What changed
- Revoke access tokens after being used in User Identity Handler.

### Why did it change
- We have the ability to revoke our access tokens. We only do this if an auth core is used multiple times. We should use the same mechanism to revoke the access token once it’s been used in the user identity endpoint.

### Issue tracking

- [PYIC-1538](https://govukverify.atlassian.net/browse/PYIC-1538)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
@@ -73,14 +73,18 @@ public class AccessTokenService {
         dataStore.create(accessTokenItem);
     }
 
+    public void revokeAccessToken(AccessTokenItem accessTokenItem) throws IllegalArgumentException {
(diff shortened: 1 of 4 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/AccessTokenService.java
```diff
@@ -73,14 +73,18 @@ public void persistAccessToken(AccessTokenResponse tokenResponse, String ipvSess
         dataStore.create(accessTokenItem);
     }
 
+    public void revokeAccessToken(AccessTokenItem accessTokenItem) throws IllegalArgumentException {
```
Wynndow, line 76: 🫶

Wynndow (approved):
Nice.

PR #355, opened 2022-07-08
PYIC-1513: Enrich Audit events

kerrr:
### What changed

Update the audit events and processing to add userId, sessionId and componentId.

### Why did it change

All audit events should have userId, sessionId and componentId as standard.

### Issue tracking

- [PYIC-1513](https://govukverify.atlassian.net/browse/PYIC-XXX)

### DIFF
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -100,13 +100,11 @@ public class CredentialIssuerReturnHandler
                     configurationService.getSsmParameter(
                             ConfigurationVariable.AUDIENCE_FOR_CLIENTS);
 
-            var ae =
+            auditService.sendAuditEvent(
                     new AuditEvent(
                             AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED,
                             componentId,
-                            auditEventUser);
-
-            auditService.sendAuditEvent(ae);
+                            auditEventUser));
 
             CredentialIssuerRequestDto request =
                     RequestHelper.convertRequest(input, CredentialIssuerRequestDto.class);
--- a/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
+++ b/lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
@@ -6,7 +6,6 @@ import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.nimbusds.jose.shaded.json.JSONObject;
-import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
@@ -57,6 +56,8 @@ public class CredentialIssuerReturnHandler
     private final VerifiableCredentialJwtValidator verifiableCredentialJwtValidator;
     private final IpvSessionService ipvSessionService;
 
+    private String componentId;
+
     public CredentialIssuerReturnHandler(
             CredentialIssuerService credentialIssuerService,
             ConfigurationService configurationService,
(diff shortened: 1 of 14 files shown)


lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -90,8 +91,22 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                    configurationService.getSsmParameter(
+                            ConfigurationVariable.AUDIENCE_FOR_CLIENTS);
+
+            var ae =
+                    new AuditEvent(
+                            AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED,
+                            componentId,
+                            auditEventUser);
+
+            auditService.sendAuditEvent(ae);
```
Wynndow: Could inline this - might scan a bit more nicely

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -90,8 +91,22 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-            auditService.sendAuditEvent(AuditEventTypes.IPV_CRI_AUTH_RESPONSE_RECEIVED);
-            RequestHelper.getIpvSessionId(input);
+
+            String ipvSessionId = RequestHelper.getIpvSessionId(input);
+            String userId = ipvSessionService.getUserId(ipvSessionId);
+
+            AuditEventUser auditEventUser = new AuditEventUser(userId, ipvSessionId);
+            this.componentId =
+                    configurationService.getSsmParameter(
+                            ConfigurationVariable.AUDIENCE_FOR_CLIENTS);
```
Wynndow, line 101: In the log helper we're setting the componentID to a hard coded value of `core`. That way it's the same between all envs, and the same Splunk query will work between envs. If you use the audience value, you're gonna get different values per env. Either way I think the two things should match.

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
```diff
@@ -90,8 +91,22 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+
+            var ae =
+                    new AuditEvent(
```
Wynndow, line 104: Why doesn't this Audit event not have extensions like the one produced from `sendIpvVcReceivedAuditEvent()`?

Wynndow (commented):
I'm wondering if the audit service should have access to the configuration service and the session service. That was it can be responsible for creating the `AuditEventUser`s and `AuditEvent`s. So you could just call the audit service with the audit type and the session ID. Would keep the handlers a bit cleaner.

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
kerrr: Yup, think i was trying to debug something here.

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
kerrr: There's been a bit of discussion about what the componentId should be. It's primarily the name for the id we use to represent the core when talking to other services, and used in aud and iss in jwts. Really parameter in config should be `self/componentId` and we should use that as appropriate where when creating and validating jwts.

kerrr: The log helper should reflect that too.

lambdas/credentialissuerreturn/src/main/java/uk/gov/di/ipv/core/credentialissuerreturn/CredentialIssuerReturnHandler.java
kerrr: Not all audit events have anything to put in the exentions block.

kerrr:
> I'm wondering if the audit service should have access to the configuration service and the session service. That was it can be responsible for creating the AuditEventUsers and AuditEvents. So you could just call the audit service with the audit type and the session ID. Would keep the handlers a bit cleaner.

I think there's a bit of thinking to do here. We'll have to see how bespoke each audit message is. This PR is setting a bunch of standard values, but I think as we get into putting more detail in each one, constructing the correct audit message will be more complex and having the Audit service knowhow to construct each one will not be the best option.

Danomanic (approved):

PR #369, opened 2022-07-27
PYIC-1689: Implement evaluate-gpg45-scores lambda

Wynndow:
### What changed

Implement evaluate-gpg45-scores lambda

### Why did it change

[PYIC-1689: Determine if credentials meet M1A](https://github.com/alphagov/di-ipv-core-back/commit/c337aef3ba641a200a6e06fe3da77b3fb3a882d2) 

This adds logic to the evaluate-gpg45-scores lambda to see if the
credentials gathered already are sufficient to meet the M1A profile. It
uses code already written, pulled in from here, for defining the
profiles and scores: https://github.com/alphagov/di-ipv-core-back/tree/wip/gpg45-lib/src/main/java/uk/gov/di/ipv/gpg45

When building the scores from the gathered credentials, for gpg45 scores
that only have one value (unlike evidence which can have multiple), this
will pick the score that has the highest value with the lowest number of
contra indicators. This will be useful in the future, but currently we
only ever expect to have score for each type of score so there is
actually nothing to sort.

If the credentials are determined to have met the M1A profile we end the
users session. If not, we continue them on the journey.
  
[PYIC-1689: Fail journey in gpg45 evaluator if failing credential](https://github.com/alphagov/di-ipv-core-back/commit/02c26516bd1b72967c1580cb87634e816e41a0fd) 

We need to fail a journey if we receive a VC from a CRI that won't allow
us to meet M1A. Currently this is done by the validatecricheck lambda.

This steals some of the logic from there to do an upfront check on all
the credentials gathered so far in the evaluate-gpg45-scores lambda. It
will fail the journey if any VCs don't meet M1A.

This logic will be moved to the CRI selector lambda in future - this
lambda should just be responsible for deciding if we've met a profile or
if we need more evidence. The CRI selector should be responsible for
failing a journey if it decides there are no more paths for a user to
take.

Adding this in makes some of the logic added in the previous commit
redundant - specifically the handling of CIs. This logic will be needed
when the fail check is moved to the CRI selector lambda so it should
stay.

### Issue tracking

- [PYIC-1689](https://govukverify.atlassian.net/browse/PYIC-1689)

### DIFF
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -26,11 +26,11 @@ import java.util.List;
 public class EvaluateGpg45ScoresHandler
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
 
+    private static final Logger LOGGER = LogManager.getLogger();
     public static final String JOURNEY_SESSION_END = "/journey/session/end";
     public static final String JOURNEY_NEXT = "/journey/next";
     public static final String JOURNEY_ERROR = "/journey/error";
-
-    private static final Logger LOGGER = LogManager.getLogger();
+    public static final String JOURNEY_FAIL = "/journey/fail";
     private final UserIdentityService userIdentityService;
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
 
@@ -55,12 +55,21 @@ public class EvaluateGpg45ScoresHandler
             String ipvSessionId = RequestHelper.getIpvSessionId(event);
             List<String> credentials = userIdentityService.getUserIssuedCredentials(ipvSessionId);
 
-            return ApiGatewayResponseGenerator.proxyJsonResponse(
-                    HttpStatus.SC_OK,
-                    gpg45ProfileEvaluator.credentialsSatisfyProfile(credentials, Gpg45Profile.M1A)
-                            ? new JourneyResponse(JOURNEY_SESSION_END)
-                            : new JourneyResponse(JOURNEY_NEXT));
+            JourneyResponse journeyResponse;
+            if (gpg45ProfileEvaluator.anyCredentialsGatheredDoNotMeetM1A(credentials)) {
+                // This will eventually be handled by the CRI select lambda. We are only
+                // failing the journey here for temporary convenience. This lambda should
+                // only have responsibility for ending the journey if we have met a profile.
+                journeyResponse = new JourneyResponse(JOURNEY_FAIL);
+            } else {
+                journeyResponse =
+                        gpg45ProfileEvaluator.credentialsSatisfyProfile(
+                                        credentials, Gpg45Profile.M1A)
+                                ? new JourneyResponse(JOURNEY_SESSION_END)
+                                : new JourneyResponse(JOURNEY_NEXT);
+            }
 
+            return ApiGatewayResponseGenerator.proxyJsonResponse(HttpStatus.SC_OK, journeyResponse);
         } catch (HttpResponseExceptionWithErrorBody e) {
             return ApiGatewayResponseGenerator.proxyJsonResponse(
                     e.getResponseCode(), e.getErrorBody());
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -4,12 +4,74 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import org.apache.http.HttpStatus;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import software.amazon.lambda.powertools.logging.Logging;
+import software.amazon.lambda.powertools.tracing.Tracing;
+import uk.gov.di.ipv.core.evaluategpg45scores.exception.UnknownEvidenceTypeException;
+import uk.gov.di.ipv.core.evaluategpg45scores.gpg45.Gpg45Profile;
+import uk.gov.di.ipv.core.evaluategpg45scores.gpg45.Gpg45ProfileEvaluator;
+import uk.gov.di.ipv.core.library.domain.JourneyResponse;
+import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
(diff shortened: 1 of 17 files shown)


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -4,12 +4,83 @@
...
+    @Tracing
+    @Logging(clearState = true)
     public APIGatewayProxyResponseEvent handleRequest(
-            APIGatewayProxyRequestEvent input, Context context) {
-        return null;
+            APIGatewayProxyRequestEvent event, Context context) {
+        LogHelper.attachComponentIdToLogs();
+        try {
+            String ipvSessionId = RequestHelper.getIpvSessionId(event);
+            List<String> credentials = userIdentityService.getUserIssuedCredentials(ipvSessionId);
```
blakeyp, line 56: Assuming this PR will go in first, but just noting we'll need to update this to get the VCs by userId in [PYIC-1693](https://govukverify.atlassian.net/browse/PYIC-1693)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45Profile.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Checks if that the provided {@code Gpg45Scores} satisfy the {@code Gpg45Scores} for this
+     * profile.
+     *
+     * @param target The {@code Gpg45Scores} to compare.
+     * @return {@code true} if the elements in the provided {@code Gpg45Scores} are all equal or
+     *     greater than the profile's `{@code Gpg45Scores}.
+     */
+    public boolean satisfiedBy(Gpg45Scores target) {
+        return satisfactoryEvidence(target)
```
blakeyp, line 79: So it reads a bit clearer maybe `hasSatisfactoryEvidence` ?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45Scores.java
```diff
@@ -0,0 +1,273 @@
...
+    public int hashCode() {
+        return Objects.hash(evidences, activity, fraud, verification);
+    }
+
+    /**
+     * Compares the score to another. A negative value indicates that there is some negative value
+     * in the {@code difference} between this score and the other.
+     *
+     * @param other
+     * @return
```
blakeyp: Missing the rest of the comment here?

I like these doc comments I think we should do these more 😃

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -0,0 +1,147 @@
...
+
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CLAIM;
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_EVIDENCE;
+
+public class Gpg45ProfileEvaluator {
+
+    private static final Gson gson = new Gson();
+    private static final int NO_SCORE = 0;
+
+    public boolean anyCredentialsGatheredDoNotMeetM1A(List<String> credentials)
```
blakeyp, line 30: Just trying to understand, what's the difference between this and `!credentialsSatisfyProfile(credentials, Gpg45Profile.M1A)`? One is checking the credentials overall meet the profile whereas the other is checking that each credential alone meets the profile?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
kerrr: Yes this will go in first.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45Profile.java
kerrr: Maybe if further clarity is required, `isSatisfiedBy()`
Would then read, `if (foo.isSatisfiedBy(bar))` which would be quite nice.

Wynndow: Yeah, it was the readability I was going for, and your suggestion is nicer than mine Kerr.
The word `Evidence` is very overloaded in this area, so I'm inclined to not use it in a method name.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45ProfileEvaluator.java
kerrr: This is a shortcut that is relevant right now, but will be taken out later. This is what Chris was talking about , starting "We need to fail a journey if we receive a VC from a CRI that won't allow us to meet M1A. ..."
If you look where this is called from you can see an inline comment there too. But this is good thinking about the problem. 👍🏼

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45Scores.java
Wynndow: You can thank @kerrr for these! He wrote this code and I copied it. Will flesh this comment out a bit though.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/gpg45/Gpg45ProfileEvaluator.java
Wynndow: So this is a temporary thing to replace the validatecricheck lambda. If `anyCredentialsGatheredDoNotMeetM1A`  returns true in the handler we fail the journey immediately. So it's effectively checking if the last credential gathered was below score or had CI's and halting the journey at that point like we do at the moment. We can remove the validatecricheck lambda once this is being used.

The `credentialsSatisfyProfile` method will return true if we have enough pieces of evidence to meet the desired profile. It doesn't care if there is evidence that has been gathered that _doesn't _ meet the scores, or has CIs, because this lambda isn't responsible for knowing if there are other CRIs available to go get more evidence from. That's the job of the CRI selector lambda.

So in future, the `anyCredentialsGatheredDoNotMeetM1A` check will get effectively get done by the CRI selector. If we haven't met the gpg45profile yet, this new lambda will pass on to the cri selector lambda. If that lambda decides there's no other paths for a user to take (other CRIs) to try to gather evidence to meet the profile, the user will fail. It allows for more complex journeys in future.

kerrr (dismissed):

blakeyp (approved):

PR #374, opened 2022-07-28
PYIC-1693 Refactor storage: read VCs from new table by userId and delete old one

blakeyp:
### What changed

This PR completes the migration to the 'v2' user-issued-credentials table which now stores the user's VCs by user id instead of session id. All the reads to the old table are updated to read from the new table by user id, and the old table has been removed.

### Why did it change

So that we are reading VCs from the new table, by user id instead of session id. This has been done over 2 PRs to stage out the table changes to production.

### Issue tracking
- [PYIC-1693](https://govukverify.atlassian.net/browse/PYIC-1693)


deploy/template.yaml
```diff
@@ -410,16 +410,13 @@ Resources:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
           CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX: !Sub "/${Environment}/core/credentialIssuers"
-          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsTable.Arn]]
-          USER_ISSUED_CREDENTIALS_V2_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsV2Table.Arn]]
+          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsV2Table.Arn]]
```
dannyhayes-gds, line 413: Does the UserIssedCredentials table name still need "V2" in it or will it be changed in a different PR?

dannyhayes-gds (approved):

deploy/template.yaml
blakeyp: So I realised the table can't just be renamed, as cloudformation sees it as a new table so tries to create a new one. So we'll need to keep the V2 in the name (this also meant we could stage out the change keeping both tables in tandem initially and then dropping the old one)

dannyhayes-gds: Ah okay fair enough 👍

PR #378, opened 2022-08-01
PYIC-1728 Refactor storage: read auth code from session

blakeyp:
### What changed

Read authorization code and metadata from sessions table and delete AuthCodesTable

### Why did it change

We're streamlining the storage of the authorization codes and access tokens so they now sit in the sessions table rather than their own separate tables. The associated metadata for each is also stored in its own object in the session. This PR continues the migration to read the auth codes and metadata from the session (via secondary index) and remove the old auth codes table. Due to only being able to add one secondary index per update, the next PR will handle the access token reading.

### Issue tracking
- [PYIC-1728](https://govukverify.atlassian.net/browse/PYIC-1728)


deploy/template.yaml
```diff
@@ -1087,9 +1059,18 @@ Resources:
       AttributeDefinitions:
         - AttributeName: "ipvSessionId"
           AttributeType: "S"
+        - AttributeName: "authorizationCode"
+          AttributeType: "S"
       KeySchema:
         - AttributeName: "ipvSessionId"
           KeyType: "HASH"
+      GlobalSecondaryIndexes:
+        - IndexName: "authorizationCode-index"
```
kerrr: Do we need to namespace this with `index`. In context is it obvious that this is an index? Does it clash with the attribute name somewhere elsewhere?

lib/src/main/java/uk/gov/di/ipv/core/library/service/AuthorizationCodeService.java
```diff
@@ -1,72 +1,24 @@
...
-        AuthorizationCodeItem authorizationCodeItem = dataStore.getItem(authorizationCode);
-        authorizationCodeItem.setIssuedAccessToken(DigestUtils.sha256Hex(accessToken));
-        authorizationCodeItem.setExchangeDateTime(Instant.now().toString());
-
-        dataStore.update(authorizationCodeItem);
-    }
-
-    public boolean isExpired(AuthorizationCodeItem authCodeItem) {
-        return Instant.parse(authCodeItem.getCreationDateTime())
+    public boolean isExpired(AuthorizationCodeMetadata authorizationCodeMetadata) {
```
kerrr: I think this should just be a method in the `AuthorizationCodeMetadata` object. We can then remove this whole service class.

blakeyp: Done this

deploy/template.yaml
blakeyp: ~Doesn't seem to be anything against this in the documentation, but I've tried it and it doesn't seem to want to create the index. Also noticed if you try to create an index manually in the console it does add `-index` on the end~

It was just being slow creating the index, looks like it does work with just `authorizationCode` 👍

Wynndow (approved):

PR #380, opened 2022-08-01
PYIC-1729 Refactor storage: read access token from session

blakeyp:
### What changed

Read access token and metadata from sessions table and delete AccessTokensTable

### Why did it change

We're streamlining the storage of the authorization codes and access tokens so they now sit in the sessions table rather than their own separate tables. The associated metadata for each is also stored in its own object in the session. This PR completes the migration, reading the access tokens and metadata from the session (via secondary index) and removing the old access tokens table.

### Issue tracking
- [PYIC-1729](https://govukverify.atlassian.net/browse/PYIC-1729)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -185,7 +185,6 @@ Resources:
         Variables:
           ENVIRONMENT: !Sub "${Environment}"
           POWERTOOLS_SERVICE_NAME: !Sub access-token-${Environment}
-          ACCESS_TOKENS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt AccessTokensTable.Arn]]
           CLIENT_AUTH_JWT_IDS_TABLE_NAME: !Ref ClientAuthJwtIdsTable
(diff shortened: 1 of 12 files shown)


deploy/template.yaml
```diff
@@ -1071,6 +1049,12 @@ Resources:
               KeyType: "HASH"
           Projection:
             ProjectionType: "ALL"
+        - IndexName: "accessToken-index"
```
kerrr: Namespace on index. Do we need it?

kerrr (dismissed):

deploy/template.yaml
blakeyp: Doesn't seem to be anything against this in the documentation, but I've tried it and it doesn't see to want to create the index. Also noticed if you try to create an index manually in the console it does add -index on the end

blakeyp: Confirmed the namespace isn't needed, I've renamed to `accessToken`

dannyhayes-gds (approved):

PR #386, opened 2022-08-02
PYIC-1642: Embed `process-journey-step` lambda in Express Step Function

Wynndow:
### What changed

Embed process-journey-step lambda in Express Step Function

### Why did it change

[PYIC-1642: Update openAPI spec for private API](https://github.com/alphagov/di-ipv-core-back/pull/386/commits/bf2164e7909dcc601ba31aec271242b95b04009d) 

This updates the openAPI spec for the internal private API. This
includes endpoints that are shortly to be removed so will change very
soon. The important bit is the configuration for the
/journey/{journeyStep} endpoint. It sets up an integration with the new
step functions StartSyncExecution action.

It's necessary to do this as using SAM on its own to manage the
integration with the API gateway results in only the StartExecution
action being configured. With no way to change it. It also doesn't
appear to allow any form of mapping to be configured.

The mappings included here take an incoming HTTP request and parse out
the body, headers, path parameters and query parameters. It puts all
these values into a JSON object and sets it as the input for the state
machine. It also includes the ARN of the state machine we want to
trigger.

The response mapping just uses a default with a status code of 200. We
have to specify a status code here, but over ride it in the mapping
depending on what we get back from the state machine. The "status" being
checked here is the result of running the state machine. We can have a
"successful" execution that results in a 400, say, from the lambda. It's
only unsuccessful if there is an unexpected error. If the state machine
executed successfully, we pass back the output of the lambda and the
status code. Otherwise we return a 500.
  
[PYIC-1642: Add StateMachine to the template](https://github.com/alphagov/di-ipv-core-back/pull/386/commits/0a1f537f397cbfa0423078f1c054c3b1afe7161b) 

This adds the actual state machine resource to the deploy template. It
uses a very simple state machine definition that just passes all input
to the process-journey-step lambda and returns all of the output.

This also updates the internal API to use the updated openAPI spec.
There were some issues with `sam validate` not being able to handle the
use of the Auth block as well as the `DefinitionBody` block. There are
PRs for SAM to fix this, but in the mean time this includes a workaround
to allow the command to work.
  
[PYIC-1642: Update process-journey-step lambda to use streams](https://github.com/alphagov/di-ipv-core-back/pull/386/commits/5fa82a1d7a11fbff5f444ec4c908187ff24b4d33) 

Previously we were using a proxy integration with the lambdas and AWS
gateway. This meant that the incoming HTTP request was passed through to
the lambda. This could be modelled using the events library.

This no longer works as the input is very different. This updates the
lambda to use received its input from an input stream, and write it to
an output stream. There may well be better ways of doing this, but this
works for now.

### Issue tracking

- [PYIC-1642](https://govukverify.atlassian.net/browse/PYIC-1642)


openAPI/core-back-internal.yaml
```diff
@@ -113,11 +177,43 @@ paths:
...
         uri:
-          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${IPVJourneyEngineFunction.Arn}:live/invocations
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:states:action/StartSyncExecution
         passthroughBehavior: "when_no_match"
-        type: "aws_proxy"
+        requestTemplates:
+          application/x-www-form-urlencoded:
+            Fn::Sub: |
+              {
+                "input": "{\"body\": $util.escapeJavaScript($input.json('$')), \"headers\": {#foreach($header in $input.params().header.keySet())\"$header\": \"$util.escapeJavaScript($input.params().header.get($header))\"#if($foreach.hasNext),#end#end},\"params\": {#foreach($param in $input.params().path.keySet())\"$param\": \"$util.escapeJavaScript($input.params().path.get($param))\"#if($foreach.hasNext),#end#end},\"query\": {#foreach($queryParam in $input.params().querystring.keySet())\"$queryParam\": \"$util.escapeJavaScript($input.params().querystring.get($queryParam))\"#if($foreach.hasNext),#end#end}}",
```
dannyhayes-gds, line 127: Is this what controls how the request headers and path params etc are passing into the step function lambda?

openAPI/core-back-internal.yaml
```diff
@@ -113,11 +177,43 @@ paths:
...
+                "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStateMachine.Name}"
+              }
+        responses:
+          default:
+            statusCode: 200
+            responseTemplates:
+              application/json: |
+                #set ($bodyObj = $util.parseJson($input.body))
```
dannyhayes-gds, line 135: I'm not that familiar with open-api, so are these just comments? Or is this the syntax to show the different types of responses that could be returned on each scenario?

deploy/template.yaml
```diff
@@ -1213,6 +1222,83 @@ Resources:
...
+            Statement:
+              - Effect: Allow
+                Action: 'states:StartSyncExecution'
+                Resource: !Ref JourneyEngineStateMachine
+
+  JourneyEngineStateMachineLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      RetentionInDays: 14
+      LogGroupName: !Sub "/aws/states/journey-engine-state-machine-${Environment}"
```
dannyhayes-gds: Could we maybe rename this a bit. I think it could end up getting confused with our process-journey-engine "statemachine"? Might be clearer to just refer to this as things like JourneyEngineStepFunction?

openAPI/core-back-internal.yaml
Wynndow: Yep - the `StartSyncExecution` action of the step function needs a JSON with an `input` and `stateMachineArn` as a minimum. Our lambdas are set up to expect to receive headers, path params etc, so we need to parse them from the request (`$input` in the above), and pass them on to the step function. The value of the `input` key is the json that the step function receives as input.

openAPI/core-back-internal.yaml
Wynndow: No, this is the mapping to take the response from the step function (`$input`) and map it to the http response returned by the API gateway. The language is [Apache Velocity](https://velocity.apache.org/).

Wynndow: I think the # makes GitHub render it as a comment, but it's actually part of the language.

deploy/template.yaml
Wynndow: Yep, works for me

dannyhayes-gds (dismissed):

dannyhayes-gds (approved):

PR #387, opened 2022-08-02
Fix missing env var and Dynamo permission on new version of initialis…

dannyhayes-gds:
…e session lambda

### What changed
Add missing ENV var and DynamoDB permission to the new version of the initialise session lambda.

### Why did it change
These vars were added to the old "ToBeRemoved" version of the lambda but not the new one. Now that core-front is using the new lambda we are seeing a Dynamo permission denied error due to the missing var and policy.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -475,12 +475,15 @@ Resources:
           ENVIRONMENT: !Sub "${Environment}"
           POWERTOOLS_SERVICE_NAME: !Sub initialise-ipv-session-${Environment}
           IPV_SESSIONS_TABLE_NAME: !Select [ 1, !Split [ '/', !GetAtt SessionsTable.Arn ] ]
+          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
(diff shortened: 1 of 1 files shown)


AmritSidhu (approved):

deploy/template.yaml
```diff
@@ -475,12 +475,15 @@ Resources:
...
           IPV_SESSIONS_TABLE_NAME: !Select [ 1, !Split [ '/', !GetAtt SessionsTable.Arn ] ]
+          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
           SQS_AUDIT_EVENT_QUEUE_URL: !ImportValue AuditEventQueueUrl
       Policies:
         - DynamoDBReadPolicy:
             TableName: !Ref SessionsTable
         - DynamoDBWritePolicy:
             TableName: !Ref SessionsTable
+        - DynamoDBCrudPolicy:
+            TableName: !Ref UserIssuedCredentialsV2Table
```
Wynndow, line 486: My bad, sorry. I think you can remove the two policies above this one now, as the CRUD covers them

dannyhayes-gds: They are for a different table, so I think they are needed still? They are also still there on the other lambda

PR #401, opened 2022-08-10
BAU: Allow A01 in gpg45 evaluator

Wynndow:
### What changed

Allow A01 in gpg45 evaluator

### Why did it change

This fix had been applied to the validator lambda, but we're going to be
using the logic in the gpg45 evaluator soon instead, so also need to
handle this edge case.


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/domain/CredentialEvidenceItem.java
```diff
@@ -37,6 +39,9 @@ public Gpg45Scores.Evidence getEvidenceScore() {
     }
 
     public boolean hasContraIndicators() {
+        if (isIdentityFraud()) {
+            return ci != null && !ci.isEmpty() && !(ci.size() == 1 && ci.get(0).equals(A01));
```
Danomanic, line 43: This is fine for now - but `.get(o)` raises a my eyebrows a little. I will have a think about how we can make this "safer".

Danomanic (approved):

PR #404, opened 2022-08-16
PYIC-1819: Add new JWKS endpoint to external api

dannyhayes-gds:
### What changed
Add new JWKS api endpoint to serve our public keys from.

The api endpoint will pull the value from SSM. This is because we currently store the certs there. This is just temporary until we decide to implement a proper key rotation method.

### Why did it change
As part of the app system integration work they requested the keys be made available from a JWKS endpoint.

### Issue tracking

- [PYIC-1819](https://govukverify.atlassian.net/browse/PYIC-1819)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -77,9 +77,9 @@ Resources:
             Statement:
               - Effect: Allow
                 Action:
-                  - "ssm:GetParameter"
+                  - "ssm:GetParametersByPath"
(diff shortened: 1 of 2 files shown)


Wynndow (dismissed):
Have deployed to my dev stack and it looks good!

Wynndow (approved):
Nice!

PR #409, opened 2022-08-22
PYIC-1834: Contra-indicator storage core stub

Wynndow:
### What changed

Contra-indicator storage core stub

### Why did it change

Adds a lambda that will be deployed to dev and build accounts that will
allow interaction with the contra-indicator storage system for dev and
testing purposes.

It provides a web page that can be used to submit CIs and a user ID to
the system, wrapped in a signed JWT. It will also retrieve the CIs for a
given user ID.

This makes an assumption of the interface for the getContraindicator
lambda, which has not been built yet, and so may need adjusting in
future.

### Issue tracking

- [PYIC-1834](https://govukverify.atlassian.net/browse/PYIC-1834)

### Screenshots
Screenshot 2022-08-22 at 13 30 27

Screenshot 2022-08-22 at 13 30 42


lambdas/contra-indicator-storage-core-stub/src/main/java/uk/gov/di/ipv/core/contraindicatorstoragecorestub/domain/GetCiRequest.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.di.ipv.core.contraindicatorstoragecorestub.domain;
+
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@Getter
+@SuppressWarnings("java:S116") // Field names should comply with a naming convention
```
dannyhayes-gds: We previously used the Jackson @JsonProperty annotation here to handle the snake-case var names. I guess the downside is introducing another gradle dependency but would avoid having to suppress the warning which might be nicer?


But not too fussed about this so happy for you to decide what seems better.

Wynndow: Yeah, I was trying to not use Jackson. The auth team found a lot of speed up in their lambdas by switching from Jackson to Gson.
I think I've just found something that can do a similar thing to `@JsonProperty` though so I might try adding it in.

dannyhayes-gds: Ah okay yeah thats fair enough, happy to leave out jackson then 👍

dannyhayes-gds (dismissed):

dannyhayes-gds (approved):

PR #420, opened 2022-08-26
PYIC-1869: Update core-back to send and validate against the new redirectUri structure for the app

dannyhayes-gds:
### What changed
For the app system set the redirectUri param to use URL path params instead of query params for the criId.

### Why did it change
The app system was incorrectly constructing the redirect uri due to us adding a query param. We realised that a better alternative would be to use a URL path parameter instead of a query parameter.

### Issue tracking

- [PYIC-1869](https://govukverify.atlassian.net/browse/PYIC-1869)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/helpers/AuthorizationRequestHelper.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/helpers/AuthorizationRequestHelper.java
@@ -26,7 +26,6 @@ import uk.gov.di.ipv.core.library.service.ConfigurationService;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Date;
-import java.util.List;
 import java.util.Objects;
 
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.AUDIENCE_FOR_CLIENTS;
@@ -35,8 +34,6 @@ import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.JWT_TTL_SE
 public class AuthorizationRequestHelper {
 
     public static final String SHARED_CLAIMS = "shared_claims";
-    public static final String PARAM_ID = "id";
-    public static final List<String> DCMAW_CRI_IDS = List.of("dcmaw", "stubDcmaw");
 
     private AuthorizationRequestHelper() {}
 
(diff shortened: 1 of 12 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/helpers/AuthorizationRequestHelper.java
```diff
@@ -55,9 +57,16 @@ public static SignedJWT createSignedJWT(
 
         String criId = credentialIssuerConfig.getId();
 
-        URI redirectionURI =
-                getRedirectionURI(
-                        criId, configurationService.getSsmParameter(CORE_FRONT_CALLBACK_URL));
+        URI redirectionURI;
+        if (DCMAW_CRI_IDS.contains(criId)) {
```
dannyhayes-gds: After looking back on this part, I might change this up a bit. It might be nicer to store the full redirect URI value in SSM and just use that rather than construct it here.

dannyhayes-gds: I've updated the PR to do this. Plus this new PR to add the new params to ssm - https://github.com/alphagov/di-ipv-config/pull/876/files#diff-07674faffb5f12730b95c927da7ae98beda7fad03267639510f0d2eadc56e7e2

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/AuthorizationRequestHelper.java
```diff
@@ -17,28 +17,26 @@
...
 
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.AUDIENCE_FOR_CLIENTS;
-import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CORE_FRONT_CALLBACK_URL;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.JWT_TTL_SECONDS;
 
 public class AuthorizationRequestHelper {
 
     public static final String SHARED_CLAIMS = "shared_claims";
     public static final String PARAM_ID = "id";
+    public static final List<String> DCMAW_CRI_IDS = List.of("dcmaw", "stubDcmaw");
```
Wynndow: I don't think either of these are being used.

Wynndow (dismissed):
Just a comment about a couple of unused constants.
I like this - makes things a bit cleaner.

Wynndow (approved):

PR #425, opened 2022-08-31
PYIC-1860: Send VCs to storage system

Wynndow:
### What changed

Send VCs to storage system

### Why did it change

This adds the logic and permissions to start sending VCs to the CI
storage system in all envs other than prod.

We don't want this to actually block anything, so any unexpected errors
that might be thrown are caught and logged.

### Issue tracking

- [PYIC-1860](https://govukverify.atlassian.net/browse/PYIC-1860)


lib/src/main/java/uk/gov/di/ipv/core/library/helpers/LogHelper.java
```diff
@@ -88,6 +92,22 @@ public static void logInfoMessageWithFieldAndValue(
...
+            String logFieldValue2,
+            LogField logField3,
+            String logFieldValue3) {
+        LoggingUtils.appendKey(logField1.getFieldName(), logFieldValue1);
+        LoggingUtils.appendKey(logField2.getFieldName(), logFieldValue2);
+        LoggingUtils.appendKey(logField3.getFieldName(), logFieldValue3);
+        LOGGER.log(logLevel, message);
+        LoggingUtils.removeKeys(logField1.getFieldName(), logField2.getFieldName(), logFieldValue3);
+    }
+
```
kerrr: I think we should be using a structured message rather than this mechanism.

lib/src/main/java/uk/gov/di/ipv/core/library/service/CIStorageService.java
```diff
@@ -0,0 +1,74 @@
...
+import uk.gov.di.ipv.core.library.domain.PutCiRequest;
+import uk.gov.di.ipv.core.library.helpers.LogHelper;
+import uk.gov.di.ipv.core.library.helpers.LogHelper.LogField;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.CI_STORAGE_PUT_LAMBDA_ARN;
+
+public class CIStorageService {
```
kerrr: Super picky, but I think the convention is CiStorageService

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/LogHelper.java
Wynndow: Yeah, i thought that. But the structured message blows up if you give it a value of `null`. The place this is being used is triggered by two scenarios and it means the response may not always have all the fields. I could add a bunch of null checks, but it might leave some slightly weird values in the logs. I think this way null values just don't get included.

lib/src/main/java/uk/gov/di/ipv/core/library/service/CIStorageService.java
Wynndow: That makes my skin crawl though.
Will update and live with it.

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/LogHelper.java
```diff
@@ -88,6 +92,23 @@ public static void logInfoMessageWithFieldAndValue(
...
+            String logFieldValue1,
+            LogField logField2,
+            String logFieldValue2,
+            LogField logField3,
+            String logFieldValue3) {
+        LoggingUtils.appendKey(logField1.getFieldName(), logFieldValue1);
+        LoggingUtils.appendKey(logField2.getFieldName(), logFieldValue2);
+        LoggingUtils.appendKey(logField3.getFieldName(), logFieldValue3);
+        LOGGER.log(logLevel, message);
+        LoggingUtils.removeKeys(logField1.getFieldName(), logField2.getFieldName(), logFieldValue3);
```
blakeyp: Should this be `logField3.getFieldName()` ?

Wynndow:
Closing this PR due to the GitHub build being squiffy. Seeing if a new PR resolves.

PR #430, opened 2022-09-01
PYIC-1881: Store in the session details of when a user returns from a CRI

dannyhayes-gds:
### What changed
Store details of a visited CRI in the core-back session when a user returns. The criId, a boolean of wether a VC was returned, and then the type of OauthError that was returned if relavant are the fields we store in the list of visited CRI's.

### Why did it change
This session data will be used by the cri-selector lambda to help it decide where to send the user next. We can't rely on VC's being present to tell us if a CRI has been visited so these details will be used instead.

### Issue tracking

- [PYIC-1881](https://govukverify.atlassian.net/browse/PYIC-1881)

### DIFF
--- a/lambdas/credentialissuererror/src/main/java/uk/gov/di/ipv/core/credentialissuererror/CredentialIssuerErrorHandler.java
+++ b/lambdas/credentialissuererror/src/main/java/uk/gov/di/ipv/core/credentialissuererror/CredentialIssuerErrorHandler.java
@@ -100,7 +100,7 @@ public class CredentialIssuerErrorHandler
 
             sendAuditEvent(credentialIssuerErrorDto, ipvSessionItem);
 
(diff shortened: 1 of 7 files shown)


lambdas/credentialissuererror/src/main/java/uk/gov/di/ipv/core/credentialissuererror/CredentialIssuerErrorHandler.java
```diff
@@ -99,6 +100,14 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             sendAuditEvent(credentialIssuerErrorDto, ipvSessionItem);
 
+            ipvSessionItem.addVisistedCredentialIssuerDetails(
```
blakeyp: Typo *addVisited

blakeyp (approved):

PR #432, opened 2022-09-02
PYIC-1882

dannyhayes-gds:
### What changed
Moved the GPG45 classes into the common lib to be used between both the evaluate-gpg45-scores & select-cri lambda.

Expanded the app journey cri selection logic. The select CRI lambda now uses the visited CRI data in the session table to track which CRI's have been visited. It then also checks the gathered VC from the app to see if it was successful or not. If successful then the select-cri lambda will send the users to only the address and fraud CRI's, otherwise it will revert back to the original web journey of Passport > address > fraud > kbv.

### Why did it change
The select cri lambda needs to be able to choose the correct next CRI based on where the user has already been in the journey. Depending on wether the user successfully makes it through the app journey or not, then the select-cri lambda will need to respond in the correct way.

### Issue tracking

- [PYIC-1882](https://govukverify.atlassian.net/browse/PYIC-1882)


lambdas/select-cri/src/test/java/uk/gov/di/ipv/core/credentialissuer/SelectCriHandlerTest.java
```diff
@@ -78,14 +87,18 @@ void shouldReturnPassportCriJourneyResponse() throws JsonProcessingException {
     private void mockIpvSessionService() {
         when(mockIpvSessionItem.getClientSessionDetails()).thenReturn(mockClientSessionDetailsDto);
         when(mockIpvSessionService.getIpvSession(TEST_SESSION_ID)).thenReturn(mockIpvSessionItem);
-        when(mockIpvSessionService.getUserId(TEST_SESSION_ID)).thenReturn(TEST_USER_ID);
+        //        when(mockIpvSessionService.getUserId(TEST_SESSION_ID)).thenReturn(TEST_USER_ID);
```
blakeyp: Should this be removed?

blakeyp (dismissed):

blakeyp (approved):

PR #436, opened 2022-09-05
PYIC-1886: Update select-cri lambda to handle failed VC visit checking

dannyhayes-gds:
### What changed
Updated the select-cri lambda to handle the GPG45 score validation of the gathered VC's in order to decide wether to end the journey or not.

Update the error handling within the evaluate-gpg45-scores lambda to only be checking for the presence of CI's. The score checking has been moved into the select-cri lambda.

The select-cri lambda now checks the following when deciding where to send the user:

- Gets the current visited CRI's from the session storage to see if the next planned CRI has previously been visited.
- If it has then it checks if it was returned with a VC or not.
- If it was returned with a VC, it will retrieve the VC and get the evidence scores and check if it was a successful VC or not.

### Why did it change
The select-cri lambda is now responsible for deciding wether there is an alternative CRI available when a failed VC has been gathered or if a user has visited a CRI unsuccessfully etc. For now this is only the app journey that can provide an alternative with the web based passport journey if either the user visited but was returned without a VC, or returned with a failing VC that does not contain any CI's.

### Issue tracking

- [PYIC-1886](https://govukverify.atlassian.net/browse/PYIC-1886)


lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -217,8 +246,94 @@ private APIGatewayProxyResponseEvent getJourneyResponse(String criId) {
...
+            if (criVisitDetails.get().isReturnedWithVc()) {
+                if (!isSuccessfulVc(userId, criId, validator, gpg45Profile)) {
+                    LOGGER.info(
+                            "User has a previous failed visit to {} cri due to a failed identity check. Routing user to failed journey path",
+                            criId);
+                    return Optional.of(getJourneyPyiNoMatchResponse());
+                }
+            } else {
+                LOGGER.info(
+                        "User has a previous failed visit to {} cri due to: {}. Routing user to web journey instead.",
```
dannyhayes-gds: Just realised this log isn't quite right anymore. I'll move this one and the one above slightly so that it can reflect correctly what actually happens

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -8,17 +8,16 @@
 import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 import uk.gov.di.ipv.core.library.domain.gpg45.domain.CredentialEvidenceItem;
 import uk.gov.di.ipv.core.library.domain.gpg45.exception.UnknownEvidenceTypeException;
-import uk.gov.di.ipv.core.library.domain.gpg45.validation.FraudEvidenceValidator;
-import uk.gov.di.ipv.core.library.domain.gpg45.validation.KbvEvidenceValidator;
-import uk.gov.di.ipv.core.library.domain.gpg45.validation.PassportEvidenceValidator;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 
+import javax.swing.*;
```
blakeyp: Star import - is this intended?

dannyhayes-gds: Ooops no it wasn't. I'll remove that

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -169,46 +151,14 @@ private boolean contraIndicatorsPresent(
...
-    private Optional<JourneyResponse> getJourneyResponse(
-            List<CredentialEvidenceItem> credentialEvidenceItems,
-            Predicate<CredentialEvidenceItem> validator,
-            String journeyResponseValue) {
-        for (CredentialEvidenceItem item : credentialEvidenceItems) {
-            if (!validator.test(item)) {
-                return Optional.of(new JourneyResponse(journeyResponseValue));
+                if (evidenceType
+                        .name()
+                        .equals(CredentialEvidenceItem.EvidenceType.VERIFICATION.name())) {
```
Wynndow:
```suggestion
                if (evidenceType.equals(CredentialEvidenceItem.EvidenceType.VERIFICATION)) {
```

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -217,8 +246,98 @@ private APIGatewayProxyResponseEvent getJourneyResponse(String criId) {
...
+            List<VisitedCredentialIssuerDetailsDto> visitedCredentialIssuers,
+            BiPredicate<CredentialEvidenceItem, Gpg45Profile> validator,
+            Gpg45Profile gpg45Profile,
+            String criId,
+            String userId)
+            throws ParseException {
+        Optional<VisitedCredentialIssuerDetailsDto> criVisitDetails =
+                visitedCredentialIssuers.stream()
+                        .filter(cri -> cri.getCriId().equals(criId))
+                        .findFirst();
```
Wynndow: Due to calling `userHasNotVisited` on line 266 this should always find a `VisitedCredentialIssuerDetailsDto`. So I think you could use `orElseThrow()` and remove the need for an Optional here.
```suggestion
        VisitedCredentialIssuerDetailsDto criVisitDetails =
                visitedCredentialIssuers.stream()
                        .filter(cri -> cri.getCriId().equals(criId))
                        .findFirst().orElseThrow();
```

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -217,8 +246,98 @@ private APIGatewayProxyResponseEvent getJourneyResponse(String criId) {
...
+                            visitedCredentialIssuers, gpg45Validator, gpg45Profile, criId, userId);
+
+            if (failedJourneyResponse.isPresent()) {
+                if (criId.equals(dcmawCriId)) {
+                    LOGGER.info("Routing user to web journey");
+                    return Optional.of(getNextWebJourneyCri(visitedCredentialIssuers, userId));
+                }
+                LOGGER.info("Routing user to failed journey path");
+            }
+            return failedJourneyResponse;
```
Wynndow:
```suggestion
            return Optional.empty();
```
I think this might make the intention a little clearer.

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -125,21 +132,55 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            return passportResponse.get();
         }
 
         if (userHasNotVisited(visitedCredentialIssuers, addressCriId)) {
             return getJourneyResponse(addressCriId);
+        } else {
+            Optional<VisitedCredentialIssuerDetailsDto> addressVisitDetails =
+                    visitedCredentialIssuers.stream()
+                            .filter(cri -> cri.getCriId().equals(addressCriId))
+                            .findFirst();
```
Wynndow:
```suggestion
            VisitedCredentialIssuerDetailsDto addressVisitDetails =
                    visitedCredentialIssuers.stream()
                            .filter(cri -> cri.getCriId().equals(addressCriId))
                            .findFirst().orElseThrow();
```
I think this should be safe due to the check above.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/DcmawEvidenceValidator.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.di.ipv.core.library.domain.gpg45.validation;
```
Wynndow: I think this class is identical to `PassportEvidenceValidator`.
When I created these I should probably have written them with the GPG45 elements in mind, rather than the CRIs. They could probably be abstracted to just refer to GPG45 rather than CRIs.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/domain/CredentialEvidenceItem.java
```diff
@@ -21,19 +21,21 @@ public class CredentialEvidenceItem {
...
-    public CredentialEvidenceItem(EvidenceType evidenceType, int score) {
+    public CredentialEvidenceItem(EvidenceType evidenceType, int score, List<String> ci) {
         if (EvidenceType.ACTIVITY.equals(evidenceType)) {
             this.activityHistoryScore = score;
         } else if (EvidenceType.IDENTITY_FRAUD.equals(evidenceType)) {
             this.identityFraudScore = score;
         } else if (EvidenceType.VERIFICATION.equals(evidenceType)) {
             this.verificationScore = score;
         }
+        this.ci = ci;
```
Wynndow, line 33:
```suggestion
    public CredentialEvidenceItem(EvidenceType evidenceType, int score) {
        if (EvidenceType.ACTIVITY.equals(evidenceType)) {
            this.activityHistoryScore = score;
        } else if (EvidenceType.IDENTITY_FRAUD.equals(evidenceType)) {
            this.identityFraudScore = score;
        } else if (EvidenceType.VERIFICATION.equals(evidenceType)) {
            this.verificationScore = score;
        }
        this.ci = Collections.emptyList();
```
I think this constructor is only ever being used to pass in an empty list

Wynndow (commented):
A few comments, but as mentioned in Slack, i'm wondering if the scoring logic can be left in the `evaluate-gpg45-scores` lambda and the success/fail of a check communicated to the `cri-select` lambda in some other way.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/DcmawEvidenceValidator.java
dannyhayes-gds: Ah yeah true. I'll remove this one and rename them to refelect GPG45 types rather than CRI's

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/domain/CredentialEvidenceItem.java
dannyhayes-gds: I kept this in more for potential future use in case we ever need to set this value for future CRI's. Plus I do use it at the moment during tests to setup test items with CI's

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -131,4 +164,80 @@ private List<ContraIndicatorItem> getCIsAndSwallowErrors(
...
+            SignedJWT signedJWT = SignedJWT.parse(credential);
+            JSONObject vcClaim = (JSONObject) signedJWT.getJWTClaimsSet().getClaim(VC_CLAIM);
+            JSONArray evidenceArray = (JSONArray) vcClaim.get(VC_EVIDENCE);
+            if (evidenceArray == null) {
+                CredentialIssuerConfig addressCriConfig =
+                        configurationService.getCredentialIssuer(addressCriId);
+                String vcIss = signedJWT.getJWTClaimsSet().getIssuer();
+                if (vcIss.equals(addressCriConfig.getAudienceForClients())) {
+                    vcStatuses.add(new VcStatusDto(vcIss, true));
+                }
```
Wynndow, line 200: Is it worth logging something here if they don't match? Feels like a situation we should never be in - having a VC with no evidence from a CRI other than the address CRI?

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoreHandlerTest.java
```diff
@@ -48,14 +54,63 @@ class EvaluateGpg45ScoreHandlerTest {
...
-    public static final List<String> CREDENTIALS = List.of("Some", "gathered", "credentials");
+    private static final String M1A_PASSPORT_VC =
+            "eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NiJ9.eyJzdWIiOiJ1cm46dXVpZDplNmUyZTMyNC01YjY2LTRhZDYtODMzOC04M2Y5ZjgzN2UzNDUiLCJhdWQiOiJodHRwczpcL1wvaWRlbnRpdHkuaW50ZWdyYXRpb24uYWNjb3VudC5nb3YudWsiLCJuYmYiOjE2NTg4Mjk2NDcsImlzcyI6Imh0dHBzOlwvXC9yZXZpZXctcC5pbnRlZ3JhdGlvbi5hY2NvdW50Lmdvdi51ayIsImV4cCI6MTY1ODgzNjg0NywidmMiOnsiZXZpZGVuY2UiOlt7InZhbGlkaXR5U2NvcmUiOjIsInN0cmVuZ3RoU2NvcmUiOjQsImNpIjpudWxsLCJ0eG4iOiIxMjNhYjkzZC0zYTQzLTQ2ZWYtYTJjMS0zYzY0NDQyMDY0MDgiLCJ0eXBlIjoiSWRlbnRpdHlDaGVjayJ9XSwiY3JlZGVudGlhbFN1YmplY3QiOnsicGFzc3BvcnQiOlt7ImV4cGlyeURhdGUiOiIyMDMwLTAxLTAxIiwiZG9jdW1lbnROdW1iZXIiOiIzMjE2NTQ5ODcifV0sIm5hbWUiOlt7Im5hbWVQYXJ0cyI6W3sidHlwZSI6IkdpdmVuTmFtZSIsInZhbHVlIjoiS0VOTkVUSCJ9LHsidHlwZSI6IkZhbWlseU5hbWUiLCJ2YWx1ZSI6IkRFQ0VSUVVFSVJBIn1dfV0sImJpcnRoRGF0ZSI6W3sidmFsdWUiOiIxOTU5LTA4LTIzIn1dfSwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIklkZW50aXR5Q2hlY2tDcmVkZW50aWFsIl19fQ.MEYCIQC-2fwJVvFLM8SnCKk_5EHX_ZPdTN2-kaOxNjXky86LUgIhAIMZUuTztxyyqa3ZkyaqnkMl1vPl1HQ2FbQ9LxPQChn";
```
Wynndow: Might be worth moving these VC strings to the TestFixtures class in the lib, so they can be reused elsewhere if needed. No biggy though.

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -125,21 +116,43 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+        } else {
+            VisitedCredentialIssuerDetailsDto addressVisitDetails =
+                    visitedCredentialIssuers.stream()
+                            .filter(cri -> cri.getCriId().equals(addressCriId))
+                            .findFirst()
+                            .orElseThrow();
+
+            if (!addressVisitDetails.isReturnedWithVc()) {
+                LOGGER.info(
+                        "User has a previous failed visit to address cri due to: {}. Routing user to web journey instead.",
```
Wynndow: I think this log line might be a duplicate of the one from the app journey below.

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -217,6 +210,64 @@ private APIGatewayProxyResponseEvent getJourneyResponse(String criId) {
...
+        Optional<VcStatusDto> vc = getVc(currentVcStatuses, criConfig.getAudienceForClients());
+        if (vc.isEmpty()) {
+            if (userHasNotVisited(visitedCredentialIssuers, criId)) {
+                return Optional.of(getJourneyResponse(criId));
+            }
+
+            if (criId.equals(dcmawCriId)) {
+                LOGGER.info("Routing user to web journey");
+                return Optional.of(
+                        getNextWebJourneyCri(visitedCredentialIssuers, currentVcStatuses, userId));
```
Wynndow, line 242: What's the scenario where a user doesn't has a VC status (returned by getVc()), but _has_ visited the CRI?

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45EvidenceValidator.java
```diff
@@ -0,0 +1,27 @@
...
+    public static boolean validate(CredentialEvidenceItem item, Gpg45Profile gpg45Profile) {
+
+        if (item.getStrengthScore() < gpg45Profile.scores.evidences().get(0).strength()) {
+            return false;
+        }
+        if (item.getValidityScore() < gpg45Profile.scores.evidences().get(0).validity()) {
+            return false;
+        }
+        return item.getCi() == null || item.getCi().isEmpty();
+    }
```
Wynndow: It looks like the `validate` method on these validators is no longer being used. The CIs are being checked separately, and the scores are being compared directly to the profiles. They could probably be removed, and a bunch of tests with them.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
dannyhayes-gds: Yeah good shout 👍  I'll add a warn log level here

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
dannyhayes-gds: Ah yeah looks like it was the wrong message as well. I've updated the message now for both to say its routing the users to the failed journey route

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
dannyhayes-gds: So its if a user gets an Oauth error from a CRI. In that scenario they have visited the CRI but won't have a VC. 

Specifically this kind of thing will happen in the app. If the user decides to not continune on the app journey after we send them there, the app will send the user back with an `access_denied` oauth error. The KBV cri is also going to be doing some stuff with oauth errors as well

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45EvidenceValidator.java
dannyhayes-gds: Ah yeah good shout. Thanks I'll remove these and the tests

Wynndow (approved):
🎉

PR #445, opened 2022-09-08
PYIC-1905 Drop VCs if storage for CIs fails

blakeyp:
### What changed

If submitting to CI storage fails for whatever reason, throw an error and don't store the VC - treat it the same way as a VC storage failure and return journey error response.

### Why did it change

If storing a CI in the CI storage system fails for any reason we shouldn't use the VC that was issued in to attempt to meet an identity profile.

### Issue tracking
- [PYIC-1905](https://govukverify.atlassian.net/browse/PYIC-1905)

### DIFF
--- a/lambdas/retrieve-cri-oauth-access-token/src/main/java/uk/gov/di/ipv/core/retrievecrioauthaccesstoken/RetrieveCriOauthAccessTokenHandler.java
+++ b/lambdas/retrieve-cri-oauth-access-token/src/main/java/uk/gov/di/ipv/core/retrievecrioauthaccesstoken/RetrieveCriOauthAccessTokenHandler.java
@@ -151,12 +151,7 @@ public class RetrieveCriOauthAccessTokenHandler
 
                 if (configurationService.isNotRunningInProd()) {
                     LOGGER.info("Submitting VC to CI storage system");
-                    try {
-                        ciStorageService.submitVC(
-                                vc, clientSessionDetailsDto.getGovukSigninJourneyId());
-                    } catch (Exception e) {
-                        LOGGER.error("Exception thrown when calling CI storage system", e);
-                    }
+                    submitVCAndSwallowErrors(vc, clientSessionDetailsDto.getGovukSigninJourneyId());
                 }
 
                 credentialIssuerService.persistUserCredentials(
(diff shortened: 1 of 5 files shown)


lambdas/retrieve-cri-oauth-access-token/src/main/java/uk/gov/di/ipv/core/retrievecrioauthaccesstoken/RetrieveCriOauthAccessTokenHandler.java
```diff
@@ -151,7 +152,17 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        ciStorageService.submitVC(
+                                vc, clientSessionDetailsDto.getGovukSigninJourneyId());
+                    } catch (Exception e) {
+                        LOGGER.error("Exception thrown when calling CI storage system", e);
+                        LOGGER.info(
+                                "Dropping VC due to exception thrown when calling CI storage system");
+                        throw new CredentialIssuerException(
+                                HTTPResponse.SC_SERVER_ERROR,
+                                ErrorResponse.FAILED_TO_SAVE_CREDENTIAL);
+                    }
```
Wynndow: I think you're getting a codesmell cos this is nested try/catch. You might need to refactor it to a private method. This is why the `submitVCAndSwallowErrors` method existed.

Wynndow: Should `ErrorResponse.FAILED_TO_SAVE_CREDENTIAL` be `ErrorResponse.ERROR_CALLING_CI_STORAGE`?

Wynndow (commented):
I think this is the right thing to do eventually. But while we're still working on the integration I think we need to allow VCs to be stored if CI storage fails. So catching that CredentialIssuerException and letting it continue would be useful. Then when the time comes we can remove that outer catch and allow the CredentialIssuerException to be thrown.

It might be that we're just tying ourselves in knots though... Ultimately I want to exercise all the logic without actually acting on it. Which might end up being too much to ask.

blakeyp:
@Wynndow I've rejigged things a bit so we're now swallowing any errors as before. The idea being eventually we can remove that try-catch and catch `CiPutException` further up to return the journey error response (as we now throw this in CiStorageService on lambda failure).

PR #453, opened 2022-09-14
PYIC-1889: Process CIs from storage service

Wynndow:
### What changed

Process CIs from storage service

### Why did it change

This adds logic to start using the CIs stored in the storage service to generate a journey response. The result is currently just compared to the existing logic and logged, rather than being acted upon. This can stay this way until we're ready to start using the system in prod.

This approach uses the issuer of the last known CI to determine which error page to show to a user. This should be consistent with current behaviour, but may not be sustainable long term.

This new logic is also light on tests. When we actually come to start using it, we can add more tests to give us better confidence in it.

### Issue tracking

- [PYIC-1889](https://govukverify.atlassian.net/browse/PYIC-1889)

### DIFF
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -14,7 +14,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import software.amazon.lambda.powertools.logging.Logging;
 import software.amazon.lambda.powertools.tracing.Tracing;
-import uk.gov.di.ipv.core.library.domain.ContraIndicatorItem;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 import uk.gov.di.ipv.core.library.domain.gpg45.Gpg45Profile;
@@ -58,7 +57,6 @@ public class EvaluateGpg45ScoresHandler
     private final UserIdentityService userIdentityService;
     private final IpvSessionService ipvSessionService;
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
-    private final CiStorageService ciStorageService;
     private final ConfigurationService configurationService;
     private final String addressCriId;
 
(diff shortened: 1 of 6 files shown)


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -81,8 +77,9 @@ public EvaluateGpg45ScoresHandler() {
         this.configurationService = new ConfigurationService();
         this.userIdentityService = new UserIdentityService(configurationService);
         this.ipvSessionService = new IpvSessionService(configurationService);
-        this.gpg45ProfileEvaluator = new Gpg45ProfileEvaluator();
-        this.ciStorageService = new CiStorageService(configurationService);
+        this.gpg45ProfileEvaluator =
+                new Gpg45ProfileEvaluator(
+                        new CiStorageService(configurationService), configurationService);
```
Danomanic, line 82: Something about this draws my attention. Should CiStorageService be initialised inside the Gpg45ProfileEvaluator if it's only being used there? Instead of within the Handler?

lib/src/test/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileEvaluatorTest.java
```diff
@@ -588,4 +612,64 @@ void parseGpg45ScoresFromCredentialsShouldReturnCredentialItemsMapForADcmawCrede
         assertEquals(2, dcmawItems.get(0).getEvidenceScore().validity());
         assertEquals(1, dcmawItems.get(0).getActivityHistoryScore());
     }
+
+    @Test
+    void shouldCallCIStorageSystemToGetCIs() throws Exception {
+        Map<CredentialEvidenceItem.EvidenceType, List<CredentialEvidenceItem>> evidenceMap =
```
Danomanic: Could this be moved to the setup? If it's the same across both tests? Mainly for readability.

Danomanic (dismissed):
Tiny comments (more observations than anything else).

Looks good to me.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
Wynndow: I know what you mean about drawing your attention. But I don't want to closely couple the `CiStorageService` class to the `Gpgp45ProfileEvaluator` by having it initialised inside. By passing an instance of it in it keeps them loosley(sp?) coupled and easier to test.

Danomanic: Makes sense Chris! Happy to keep it as it is 👍

dannyhayes-gds (approved):

PR #460, opened 2022-09-20
PYIC-1975: Update Retrieve CRI Credential Lambda to get access_token from SessionItem rather than Request Body

Danomanic:
### What changed
- Update Retrieve CRI Credential Lambda to get access_token from SessionItem rather than Request Body

### Issue tracking

- [PYIC-1975](https://govukverify.atlassian.net/browse/PYIC-1975)


lambdas/process-journey-step/src/main/resources/statemachine/build-statemachine-config.yaml
```diff
@@ -237,6 +237,17 @@ CORE_SESSION_TIMEOUT:
   name: CORE_SESSION_TIMEOUT
   parent: END_JOURNEY
 
+#retrieve cri credential state
+RETRIEVE_CRI_CREDENTIAL:
```
dannyhayes-gds, line 241: Can't remember if its needed exactly. But just to follow the pattern of the other states, could we also add the name and parent field in here

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -96,14 +95,24 @@ public RetrieveCriCredentialHandler() {
...
-        RetrieveCriRequestDto request =
-                RequestHelper.convertRequest(input, RetrieveCriRequestDto.class);
 
+        IpvSessionItem ipvSessionItem;
         try {
             String ipvSessionId = RequestHelper.getIpvSessionId(input);
             ipvSessionItem = ipvSessionService.getIpvSession(ipvSessionId);
+        } catch (HttpResponseExceptionWithErrorBody e) {
+            ErrorResponse errorResponse = e.getErrorResponse();
+            LogHelper.logOauthError(
```
dannyhayes-gds: This wouldn't be an oauth error here. Not sure what the result of the logs would be using the logOauthError if its going to make it look like an oauth error and could be a bit confusing in the logs etc? Might be better to swap to a normal error log here

lambdas/process-journey-step/src/main/resources/statemachine/production-statemachine-config.yaml
```diff
@@ -6,10 +6,10 @@ CRI_STATE:
     next:
       type: basic
       name: next
-      targetState: EVALUATE_GPG45_SCORES
+      targetState: RETRIEVE_CRI_CREDENTIAL
       response:
         type: journey
-        journeyStepId: /journey/evaluate-gpg45-scores
+        journeyStepId: /journey/cri/credential
```
Wynndow, line 12: Should we not be going from a CRI_STATE to the `/journey/cri/access-token` endpoint, and then from there to the `/journey/cri/credential` endpoint? Without collecting the access token we won't be able to get the credential? Or have I misunderstood how it works?

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
Danomanic: Done

lambdas/process-journey-step/src/main/resources/statemachine/build-statemachine-config.yaml
Danomanic: Done

lambdas/process-journey-step/src/main/resources/statemachine/production-statemachine-config.yaml
Danomanic: @dannyhayes-gds - Can correct me if I'm wrong, but core-front calls the `/journey/cri/access-token` before it makes the call to the Journey Lambda. So the`/journey/cri/access-token` isn't actually part of the Journey States.

dannyhayes-gds: Yep at the moment the front-end still calls the access-token lambda same way that used. The user lands back in core-front with an auth code so it just passes that into the access-token lambda to be used, then reverts back here for the next step.

lambdas/retrieve-cri-oauth-access-token/src/main/java/uk/gov/di/ipv/core/retrievecrioauthaccesstoken/RetrieveCriOauthAccessTokenHandler.java
```diff
@@ -140,24 +135,8 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                if (configurationService.isNotRunningInProd()) {
-                    LOGGER.info("Submitting VC to CI storage system");
-                    submitVCAndSwallowErrors(vc, clientSessionDetailsDto.getGovukSigninJourneyId());
-                }
-
-                credentialIssuerService.persistUserCredentials(
-                        vc.serialize(), request.getCredentialIssuerId(), userId);
-            }
 
+            updateIpvSessionItemAccessToken(ipvSessionItem, accessToken);
```
dannyhayes-gds: Could we refactor these 2 methods here slightly. I think it could be a bit confusing at first glance to see when the actual ipvSessionItem is being updated. As this method just adds the access token to the item, then the next method is where we actually update it in Dynamo.

I can imagine the scenario that further down the line we swap the order or these methods but we could miss that fact that it would stop the access token being updated into Dynamo. If that makes sense?

Maybe having the 2 methods just update the ipvSessionItem object. Then a separate statement below here to do the actual `ipvSessionService.updateIpvSession(ipvSessionItem);` call? Also maybe updating the method names to setIpvSessionItemAccessToken and setIpvSessionVisitedCredentials might be clearer to say they aren't doing an update request but only setting a value on the object.

```suggestion
            setIpvSessionItemAccessToken(ipvSessionItem, accessToken);
            setIpvSessionItemVisitedCredentials(ipvSessionItem, request.getCredentialIssuerId(), true, null);
            
            ipvSessionService.updateIpvSession(ipvSessionItem);
```

Danomanic: Valid points - I've updated the method names, and shifted `ipvSessionService.updateIpvSession(ipvSessionItem);` out of the second method, to below the two `set` methods.

kerrr (approved):

PR #467, opened 2022-09-22
BAU: Handle VCs with multiple names

Wynndow:
### What changed

Handle VCs with multiple names

### Why did it change

The way we were generating shared claims wasn't working when a VC contained multiple names. The code was taking all the nameParts from all the names and combining them into one name. This was then being included as an extra name in the final shared claims.

This updates the logic to keep each name received in the VC as an individual. All names from all VCs are then combined into a final set to do the deduplication.

This shouldn't really be a problem with real CRIs, but it did expose an issue with our stub, and we should properly handle this edge case.

Here is an example of the shared claims generated when a VC contains multiple names:
```
"shared_claims": {
    "address": [
      {
        "departmentName": null,
        "addressCountry": null,
        "uprn": null,
        "postalCode": "EH1 9GP",
        "validFrom": null,
        "subBuildingName": null,
        "buildingName": null,
        "organisationName": null,
        "streetName": null,
        "dependentStreetName": null,
        "buildingNumber": null,
        "validUntil": null,
        "dependentAddressLocality": null,
        "addressLocality": null,
        "doubleDependentAddressLocality": null
      },
      {
        "departmentName": null,
        "addressCountry": null,
        "uprn": null,
        "postalCode": "NW1 6XE",
        "validFrom": "1887-01-01",
        "subBuildingName": null,
        "buildingName": "221B",
        "organisationName": null,
        "streetName": "BAKER STREET",
        "dependentStreetName": null,
        "buildingNumber": null,
        "validUntil": null,
        "dependentAddressLocality": null,
        "addressLocality": "LONDON",
        "doubleDependentAddressLocality": null
      }
    ],
    "name": [
      {
        "nameParts": [
          {
            "type": "GivenName",
            "value": "Mary"
          },
          {
            "type": "FamilyName",
            "value": "Watson"
          }
        ]
      },
      {
        "nameParts": [
          {
            "type": "GivenName",
            "value": "Joe Shmoe"
          },
          {
            "type": "FamilyName",
            "value": "Doe The Ball"
          }
        ]
      },
      {
        "nameParts": [
          {
            "type": "GivenName",
            "value": "Mary"
          },
          {
            "type": "FamilyName",
            "value": "Watson"
          },
          {
            "type": "GivenName",
            "value": "Joe Shmoe"
          },
          {
            "type": "FamilyName",
            "value": "Doe The Ball"
          }
        ]
      }
    ],
    "birthDate": [
      {
        "value": "1932-02-25"
      },
      {
        "value": "1985-02-08"
      }
    ]
  },
```

### DIFF
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -231,7 +231,7 @@ public class BuildCriOauthRequestHandler
             throws HttpResponseExceptionWithErrorBody {
         List<String> credentials = userIdentityService.getUserIssuedCredentials(userId);
 
-        Set<SharedClaims> sharedAttributes = new HashSet<>();
+        Set<SharedClaims> sharedClaimsSet = new HashSet<>();
         for (String credential : credentials) {
             try {
                 JsonNode credentialSubject =
@@ -243,7 +243,7 @@ public class BuildCriOauthRequestHandler
                     throw new HttpResponseExceptionWithErrorBody(
                             500, ErrorResponse.CREDENTIAL_SUBJECT_MISSING);
                 }
-                sharedAttributes.add(
+                sharedClaimsSet.add(
                         mapper.readValue(credentialSubject.toString(), SharedClaims.class));
             } catch (JsonProcessingException e) {
                 LOGGER.error("Failed to get Shared Attributes: {}", e.getMessage());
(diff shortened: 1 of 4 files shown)


dannyhayes-gds (approved):

galund:
This doesn't look like a healthy VC to me - multiple DOB with no indication of which one is correct is plain wrong. Actual multiple names is expected in future but not at the moment - why are we trying to handle this now, there's a problem here we should be fixing instead maybe?

kerrr:
I've discussed with George, but for anyone else following along. This Is a fix to a bug in the shared_claims generation, that was exposed by a bug in the stub CRI. The bug was that if there was multiple names in a VC, then the shared_claims generated would contain all the names on the VC (correct) and an additional name that was a collection of all the name parts of the multiple name (the bug).

PR #468, opened 2022-09-22
PYIC-1994

dannyhayes-gds:
### What changed
Update the shared_claims generation to remove extra address claims once we have a VC from the address CRI.

Also ensures that the contents of any failed VC are not included in the shared_claims.

### Why did it change
The fraud CRI will not be able to handle receiving 2 address details in the shared_claims during a mobile journey. So IPV-core needs to only include the address details from the address VC in its shared claims.

Also now it is possible for a user to fail the app CRI but still receive a VC. So for this scenario we don't want to include the contents of this VC in the shared_claims.

### Issue tracking

- [PYIC-1994](https://govukverify.atlassian.net/browse/PYIC-1994)


vam-gds (approved):

vam-gds (dismissed):
ignore my tick - accidentally ticked wrong pr

Danomanic (approved):

PR #474, opened 2022-09-23
PYIC-2009: Fix vot claim for app journey and add dcmaw id's to list of CRIs to try build the identity claim from

dannyhayes-gds:
### What changed
Set the VOT value during the evaluate-gpg45-scores lambda when a profile has been met. The vot value is stored in the session and loaded from there when needed. The value is initialised to P0 when a new session is created.

Fixed the generation of the additional identity claims. Now if we fail to generate one of the claims, it will just log that it couldn't generate the claim. These claims are now returned as Optional's and only added to the overall identity if present.

Also added the app ID's to the list of CRI's that will be looked into when generating the Identity claim. NOTE - There is a small scenario here that could result in the app's failed VC being used to get the name/DOB value's for this claim. But I will address that in a separate PR.

### Why did it change
The VOT and user identity claim generation had previously been hardcoded to expect a passport journey so was failing for an app journey.

### Issue tracking

- [PYIC-2009](https://govukverify.atlassian.net/browse/PYIC-2009)

### DIFF
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -54,6 +54,7 @@ public class EvaluateGpg45ScoresHandler
             List.of(Gpg45Profile.M1A, Gpg45Profile.M1B);
     public static final JourneyResponse JOURNEY_END = new JourneyResponse("/journey/end");
     public static final JourneyResponse JOURNEY_NEXT = new JourneyResponse("/journey/next");
+    public static final String VOT_P2 = "P2";
     private static final Logger LOGGER = LogManager.getLogger();
     private static final Gson gson = new Gson();
     private final UserIdentityService userIdentityService;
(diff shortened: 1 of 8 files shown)


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -111,13 +112,19 @@ public APIGatewayProxyResponseEvent handleRequest(
                     gpg45ProfileEvaluator.contraIndicatorsPresent(
                             evidenceMap, clientSessionDetailsDto);
             if (contraIndicatorErrorJourneyResponse.isEmpty()) {
-                updateSuccessfulVcStatuses(ipvSessionItem, credentials);
```
Wynndow, line 114: We've lost the call to `updateSuccessfulVcStatuses`. Does that matter?

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoreHandlerTest.java
```diff
@@ -284,7 +284,7 @@ void shouldReturn500IfFailedToParseCredentials() throws Exception {
 
     @Test
     void shouldReturn500IfCredentialOfUnknownType() throws Exception {
-        when(configurationService.getCredentialIssuer(any())).thenReturn(addressConfig);
+        //        when(configurationService.getCredentialIssuer(any())).thenReturn(addressConfig);
```
Wynndow: Can this go?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
dannyhayes-gds: Its just moved down below now

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoreHandlerTest.java
dannyhayes-gds: Oh yep 👍 it was just an unnecessary stubbing

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
Wynndow: ha, oh yeah! sorry

Wynndow (approved):

kerrr (approved):

PR #488, opened 2022-09-27
PYIC-2012: Don't include failed VC's in core identity claim

dannyhayes-gds:
### What changed
Only use a VC to populate the CoreIdentity claim if it is a successful VC.

### Why did it change
It is now possible for us to gather a failed App vc but to continue the journey with the old MVP passport journey. For this scenario, this PR will ensure that the successful Passport VC is the one used to populate the CoreIdentity claim.

### Issue tracking

- [PYIC-2012](https://govukverify.atlassian.net/browse/PYIC-2012)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -865,6 +865,7 @@ Resources:
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Select [1, !Split ['/', !GetAtt UserIssuedCredentialsV2Table.Arn]]
           IPV_SESSIONS_TABLE_NAME: !Select [ 1, !Split [ '/', !GetAtt SessionsTable.Arn ] ]
           SQS_AUDIT_EVENT_QUEUE_URL: !ImportValue AuditEventQueueUrl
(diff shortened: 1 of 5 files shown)


deploy/template.yaml
```diff
@@ -878,6 +879,10 @@ Resources:
             TableName: !Ref SessionsTable
         - SSMParameterReadPolicy:
             ParameterName: !Sub ${Environment}/core/self/*
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub ${Environment}/core/credentialIssuers/*
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub ${Environment}/core/credentialIssuers
```
blakeyp: Are both these policies needed, the first doesn't cover the second?

blakeyp (dismissed):

deploy/template.yaml
dannyhayes-gds: Ah yeah true. I just did a copy past of where we have it else where but its not needed here.

blakeyp (approved):

PR #490, opened 2022-09-28
PYIC-2043 Set up `validate-oauth-callback` lambda

Danomanic:
### What changed

Set up a new lambda `validate-oauth-callback`. This will be responsible for validating the returned callback. As a starting point this PR lays the groundwork, setting up the lambda and handling oauth success callbacks only. The existing cri-error handler is still used in the error scenario.

The new journey flow is as follows:
1) validate-cri-callback `/journey/cri/validate-callback`
2) retrieve-cri-oauth-access-token `/journey/cri/access-token`
3) retrieve-cri-credential `/journey/cri/credential`

Note to stage out the change, in the RetrieveCriAccessToken handler we fetch the auth code from the request (old method) if it's not found in the session and we manually set the user state to keep the journey on track. We will tidy this up in a follow-up PR

### Issue tracking
- [PYIC-2043](https://govukverify.atlassian.net/browse/PYIC-2043)


deploy/template.yaml
```diff
@@ -1341,6 +1341,74 @@ Resources:
...
+          USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
+      Policies:
+        - KMSDecryptPolicy:
+            KeyId: !Ref DynamoDBKmsKey
+        - DynamoDBReadPolicy:
+            TableName: !Ref SessionsTable
+        - DynamoDBWritePolicy:
+            TableName: !Ref SessionsTable
+        - DynamoDBReadPolicy:
+            TableName: !Ref UserIssuedCredentialsV2Table
```
blakeyp: Not sure we need this one, we're not using this table here?

deploy/template.yaml
```diff
@@ -1341,6 +1341,74 @@ Resources:
...
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub ${Environment}/core/self/*
```
blakeyp: Again looks like we only use the credential issuer config here so may not need this

blakeyp (dismissed):

lambdas/validate-oauth-callback/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+
+	implementation "com.amazonaws:aws-java-sdk-dynamodb:$rootProject.ext.dependencyVersions.awsJavaSdkDynamodb",
+			"com.amazonaws:aws-java-sdk-sqs:$rootProject.ext.dependencyVersions.awsJavaSdkSqs",
+			"com.amazonaws:aws-lambda-java-core:$rootProject.ext.dependencyVersions.awsLambdaJavaCore",
+			"com.amazonaws:aws-lambda-java-events:$rootProject.ext.dependencyVersions.awsLambdaJavaEvents",
+			"com.google.code.gson:gson:$rootProject.ext.dependencyVersions.gson",
+			"com.fasterxml.jackson.core:jackson-annotations:$rootProject.ext.dependencyVersions.jackson",
+			"com.fasterxml.jackson.core:jackson-core:$rootProject.ext.dependencyVersions.jackson",
+			"com.fasterxml.jackson.core:jackson-databind:$rootProject.ext.dependencyVersions.jackson",
+			"com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:$rootProject.ext.dependencyVersions.jackson",
```
Wynndow: I suspect you can drop all these Jackson dependencies. I recently removed them from all the other Lambdas as they're brought in by the AWS SDK.

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -0,0 +1,203 @@
...
+            setIpvSessionCRIAuthorizationCode(ipvSessionItem, authorizationCode);
+
+            ipvSessionService.updateIpvSession(ipvSessionItem);
+
+            return ApiGatewayResponseGenerator.proxyJsonResponse(
+                    HttpStatus.SC_OK, JOURNEY_NEXT_RESPONSE);
+        } catch (HttpResponseExceptionWithErrorBody e) {
+            ErrorResponse errorResponse = e.getErrorResponse();
+            LogHelper.logOauthError(
+                    "Error in credential issuer return lambda",
```
Wynndow:
```suggestion
                    "Error in validate oauth callback lambda",
```

lambdas/validate-oauth-callback/src/main/resources/log4j2.yaml
```diff
@@ -0,0 +1,19 @@
+Configuration:
```
Wynndow: This could do with converting to JSON format. That way, you can drop all the Jackson dependencies mentioned above (this yaml format breaks without `jackson-dataformat-yaml`).

```
{
  "Configuration": {
    "status": "warn",
    "appenders": {
      "Console": {
        "name": "JsonAppender",
        "target": "SYSTEM_OUT",
        "JsonTemplateLayout": {
          "eventTemplateUri": "classpath:IpvLambdaJsonLayout.json"
        }
      }
    },
    "Loggers": {
      "logger": [
        {
          "name": "JsonLogger",
          "level": "info",
          "additivity": false,
          "AppenderRef": {
            "ref": "JsonAppender"
          }
        }
      ],
      "Root": {
        "level": "info",
        "AppenderRef": {
          "ref": "JsonAppender"
        }
      }
    }
  }
}
```

Wynndow (commented):
Looks good! Just some small comments

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -0,0 +1,203 @@
...
+            APIGatewayProxyRequestEvent input, Context context) {
+        LogHelper.attachComponentIdToLogs();
+
+        IpvSessionItem ipvSessionItem = null;
+        String credentialIssuerId = null;
+
+        try {
+            String ipvSessionId = RequestHelper.getIpvSessionId(input);
+
+            CredentialIssuerRequestDto request =
```
dannyhayes-gds, line 79: If the intention is for this lambda to be called when the user is redirected with an Oauth error as well then I'm not sure if this part will work. Plus we are missing some of the useful features of the criError lambda. We would log what the oauth error was and set the visited credential issuer list etc.

Is this kind of thing going to be handled in another PR or needed within this one?

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -0,0 +1,203 @@
...
+
+            return ApiGatewayResponseGenerator.proxyJsonResponse(
+                    HttpStatus.SC_OK, JOURNEY_NEXT_RESPONSE);
+        } catch (HttpResponseExceptionWithErrorBody e) {
+            ErrorResponse errorResponse = e.getErrorResponse();
+            LogHelper.logOauthError(
+                    "Error in credential issuer return lambda",
+                    errorResponse.getCode(),
+                    errorResponse.getMessage());
+            return ApiGatewayResponseGenerator.proxyJsonResponse(
```
dannyhayes-gds, line 121: Do we need to set the visitedCredentials here as well?

lib/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEventTypes.java
```diff
@@ -9,5 +9,7 @@ public enum AuditEventTypes {
     IPV_VC_RECEIVED,
     IPV_JOURNEY_END,
     IPV_IDENTITY_ISSUED,
-    IPV_CRI_AUTH_RESPONSE_RECEIVED
+    IPV_CRI_AUTH_RESPONSE_RECEIVED,
+
```
dannyhayes-gds: Don't need the space here

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
blakeyp: Good point, for now we are still using the cri-error handler (called from the frontend).
Need to clarify whether we should be handling errors in here instead, if so will raise a separate subtask

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
blakeyp: I think the rationale for not doing that here was because we're not doing so for HttpResponse errors in the retrieve-cri-oauth-access-token handler, because this error is not recoverable?

lambdas/validate-oauth-callback/build.gradle
blakeyp: Removed

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
kerrr: Yes, the idea is for this lambda to be the only one that is called when the user gets redirected to the callback url. So the frontend should not need to make any decisions as to whether the callback was an error or not.

kerrr (approved):

PR #507, opened 2022-10-06
Add additional summary logs to lambdas to help debug journeys

dannyhayes-gds:
### What changed
Add some additional logs to each lambda summarising what happened.

### Why did it change
Should ensure that each lambda will at least log something about what has happened and help in the future with debugging and following user journeys through the system.


kerrr:
Can we make these structured rather than injecting values into the message.

kerrr:
Sorry to be a pain, can we avoid using this appendKey, removeKey pattern. Instead we should be structuring the message itself, with some thing like:
https://github.com/alphagov/di-ipv-core-back/blob/b15eb45bcfcf6dce68f5f825d9ce0a834b0599cf/lib/src/main/java/uk/gov/di/ipv/core/library/domain/SharedClaimsResponse.java#L50-L56

dannyhayes-gds:
Ah okay I see what you mean now. Yeah I can update that

kerrr (approved):
I'm not sure about having `message.message` in some of these, but let's put this in now and we can think about how we can make that better at another point

PR #519, opened 2022-10-13
PYIC-2059 Delete all of user's VCs if any expire within session timeout

blakeyp:
### What changed

On initialising a user's IPV session, check if they have any stored VCs that are due to expire within the session timeout. If there are, clear all of their stored VCs (not just expired ones). Otherwise keep them all.

### Why did it change

For Save & Return we want to keep the user's stored VCs so they can continue their journey. However, if there are stored VCs that have expired or may expire in the course of the user's journey, we should delete them. Because the VCs are all linked (eg. address feeds into fraud) and may have different expiration times, it's safer to delete all of them if any one has expired.

### Issue tracking
- [PYIC-2059](https://govukverify.atlassian.net/browse/PYIC-2059)

### DIFF
--- a/lib/src/test/java/uk/gov/di/ipv/core/library/persistance/DataStoreTest.java
+++ b/lib/src/test/java/uk/gov/di/ipv/core/library/persistance/DataStoreTest.java
@@ -173,6 +173,21 @@ class DataStoreTest {
     }
 
     @Test
+    void shouldGetItemsFromDynamoDbTableViaLessThanOrEqualFilterExpression() {
+        when(mockDynamoDbTable.query(any(QueryEnhancedRequest.class))).thenReturn(mockPageIterable);
+        when(mockPageIterable.stream()).thenReturn(Stream.empty());
+
+        dataStore.getItemsWithAttributeLessThanOrEqualValue(
+                "partition-key-12345", "an-attribute", "a-value");
+
+        verify(mockDynamoDbEnhancedClient)
+                .table(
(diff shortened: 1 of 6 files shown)


dannyhayes-gds (dismissed):

lib/src/test/java/uk/gov/di/ipv/core/library/persistance/DataStoreTest.java
```diff
@@ -172,6 +172,21 @@ void shouldGetItemsFromDynamoDbTableViaPartitionKeyQueryRequest() {
...
+        when(mockPageIterable.stream()).thenReturn(Stream.empty());
+
+        dataStore.getItemsWithAttributeLessThanOrEqualValue(
+                "partition-key-12345", "an-attribute", "a-value");
+
+        verify(mockDynamoDbEnhancedClient)
+                .table(
+                        eq(TEST_TABLE_NAME),
+                        ArgumentMatchers.<TableSchema<AuthorizationCodeItem>>any());
+        verify(mockDynamoDbTable).query(any(QueryEnhancedRequest.class));
```
Wynndow: Is there any way to make this more specific to the input? Maybe with an argument captor on the `query` call? Just to make sure the query is constructed correctly?

blakeyp: Good shout, I've added a captor to check the expression and attribute name+value

Wynndow (approved):
🧨

PR #524, opened 2022-10-13
BAU: Use structured logging

kerrr:
### What changed

Update to use structured logging in the `TokenRequestValidator`

### Why did it change

We prefer structured logging over attaching fields for specific messages.

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
@@ -11,7 +11,8 @@ import com.nimbusds.oauth2.sdk.id.JWTID;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import software.amazon.lambda.powertools.logging.LoggingUtils;
+import org.apache.logging.log4j.message.MapMessage;
+import org.apache.logging.log4j.message.StringMapMessage;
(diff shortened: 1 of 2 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/validation/TokenRequestValidator.java
```diff
@@ -94,14 +106,11 @@ private ClientAuthenticationVerifier<Object> getClientAuthVerifier(
...
-        LoggingUtils.appendKey(
-                LogHelper.LogField.JTI_USED_AT_LOG_FIELD.getFieldName(),
-                clientAuthJwtIdItem.getUsedAtDateTime());
-        LOGGER.warn("The client auth JWT id (jti) has already been used");
-        LoggingUtils.removeKeys(
-                LogHelper.LogField.JTI_LOG_FIELD.getFieldName(),
-                LogHelper.LogField.JTI_USED_AT_LOG_FIELD.getFieldName());
+        var message =
+                new MapMessage()
+                        .with("jwtValidation", "The client auth JWT id (jti) has already been used")
```
dannyhayes-gds, line 112: Either the name of these map fields can use the enum values from the LogHelper.LogField enum, or those values can be removed as they aren't being used now.

kerrr: Yeah, I'll clear them up.

kerrr: I removed the LogHelper enums

PR #534, opened 2022-10-18
BAU: Call lambdas directly in CRI return step function

Wynndow:
### What changed

Call lambdas directly in CRI return step function

### Why did it change

This updates the lambdas in the CRI return step function to call each other directly, rather than by calling the process journey step lambda.

This does mean that we'll lose a couple of states, namely `RETRIEVE_CRI_OAUTH_ACCESS_TOKEN` and `RETRIEVE_CRI_CREDENTIAL`, but I'm not sure that it matters much. If there are errors we need to investigate we'll be able to see from the lambda logs where the journey broke down.

### DIFF
--- a/deploy/criReturnStepFunction.asl.json
+++ b/deploy/criReturnStepFunction.asl.json
@@ -7,16 +7,6 @@
       "ResultPath": "$.lambdaResult",
       "Next": "RouteJourneyResponse"
     },
-    "ProcessJourneyStep": {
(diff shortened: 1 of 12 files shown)


kerrr (commented):
🤔 We should probably have a think about how we could share some of the journey engine state machine config between environments.

kerrr (approved):

PR #535, opened 2022-10-18
PYIC-2067: add successful property to IPV_VC_RECEIVED audit event

dannyhayes-gds:
### What changed
Moved the "successful" vc check logic into a common helper class.

Added a successful boolean property to the `IPV+VC_RECEIVED` audit event.

Also updated the successful VC check logic to look at the presence of CI's when making its decision. For the Fraud CRI the A01 CI is allowed when determining the VC's success status for the journey, but for the audit event it will be marked as not successful.

### Why did it change
This extra property will help with tracking the success rate of each CRI.

### Issue tracking

- [PYIC-2067](https://govukverify.atlassian.net/browse/PYIC-2067)


kerrr (approved):
We're going to have to look at our proliferation of helpers and what we have in the mono-lib at some point. But that's perhaps not for today.

PR #537, opened 2022-10-19
PYIC-2075: Send audit event when GPG45 profile matched

Wynndow:
### What changed

Send audit event when GPG45 profile matched

### Why did it change
We currently don't send an audit message when a user has matched against a GPG45 profile.

This adds that audit event, with details of the profile, the scores achieved, and the txn ID's from the VC's used.

We send the full evidence block of the VCs in audit events when we fetch them from CRIs, so the txn's here can be used to find those audit events and get more details.

### Issue tracking

- [PYIC-2075](https://govukverify.atlassian.net/browse/PYIC-2075)

### DIFF
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -5,6 +5,8 @@ import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.google.gson.Gson;
+import com.nimbusds.jose.shaded.json.JSONArray;
+import com.nimbusds.jose.shaded.json.JSONObject;
 import com.nimbusds.jwt.SignedJWT;
 import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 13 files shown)


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -96,19 +120,27 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                if (matchedProfile.isPresent()) {
+                    auditService.sendAuditEvent(
+                            buildProfileMatchedAuditEvent(
+                                    ipvSessionItem,
+                                    matchedProfile.get(),
+                                    gpg45Scores,
+                                    credentials));
                     ipvSessionItem.setVot(VOT_P2);
                     journeyResponse = JOURNEY_END;
                     message.with("lambdaResult", "A GPG45 profile has been met")
```
dannyhayes-gds, line 146: Minor thing, but is it worth adding the profile that we've matched into this log message?

Wynndow: We already do this in the `Gpg45ProfileEvaluator` - https://github.com/alphagov/di-ipv-core-back/pull/537/files#diff-f61afdbba01ff0a1ecb6f163d3a5e438472e54882b62b50988e02fcd12006d44R95

dannyhayes-gds: Ah yeah true 👍

dannyhayes-gds (approved):

PR #550, opened 2022-10-26
Remove old csls subscription filters

dannyhayes-gds:
### What changed
We have been dual publishing to 2 csls endpoints, however the old one is no longer needed and can be removed.

### Why did it change
New csls endpoint is the only one required, so the old subscription filter can be removed.


TobySaundersGDS (approved):
looks good to me

PR #560, opened 2022-10-31
PYIC-2080 Use CI scoring config to calculate user's overall CI score

blakeyp:
### What changed

Read the CI scoring values stored as json in a secret. For each of the user's stored CIs add up the "detected" scores to give an overall score. If this is greater than the allowed threshold stop the journey.

Also no longer check for the presence of CIs when recording a VC as "successful" or not - the profile score is sufficient and we don't want to fail the journey in the case that CIs are present but within the allowed threshold. The only exception is we still check for the presence of CIs when we send the "success" audit event on retrieving a credential from a CRI.

### Why did it change

Previously we were simply checking if there were any stored CIs and failing the journey on that basis (except for only-A01). Now we take into account the "detected" scores and only fail the journey if the total score exceeds the configured threshold (which won't be the case for only-A01). Eventually we will also handle the "checked" scores for doing mitigations. 

### Issue tracking
- [PYIC-2080](https://govukverify.atlassian.net/browse/PYIC-2080)

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1483,7 +1483,6 @@ Resources:
           IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
           CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX: !Sub "/${Environment}/core/credentialIssuers"
-          CI_SCORING_THRESHOLD: !Sub "/${Environment}/core/self/ciScoringThreshold"
           CI_STORAGE_GET_LAMBDA_ARN: !Sub
             - "arn:aws:lambda:eu-west-2:${contra_indicator_storage_account_id}:function:getContraIndicators-${env}"
             - contra_indicator_storage_account_id: !FindInMap
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1483,6 +1483,7 @@ Resources:
           IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
           CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX: !Sub "/${Environment}/core/credentialIssuers"
+          CI_SCORING_THRESHOLD: !Sub "/${Environment}/core/self/ciScoringThreshold"
           CI_STORAGE_GET_LAMBDA_ARN: !Sub
             - "arn:aws:lambda:eu-west-2:${contra_indicator_storage_account_id}:function:getContraIndicators-${env}"
             - contra_indicator_storage_account_id: !FindInMap
@@ -1528,6 +1529,8 @@ Resources:
             ParameterName: !Sub ${Environment}/core/self/*
         - SQSSendMessagePolicy:
             QueueName: !ImportValue AuditEventQueueName
+        - AWSSecretsManagerGetSecretValuePolicy:
+            SecretArn: !Sub arn:aws:secretsmanager:eu-west-2:*:secret:${Environment}/contra-indicator-scoring-values-*
         - Statement:
             - Sid: kmsAuditEventQueueEncryptionKeyPermission
               Effect: Allow
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1530,7 +1530,7 @@ Resources:
         - SQSSendMessagePolicy:
             QueueName: !ImportValue AuditEventQueueName
         - AWSSecretsManagerGetSecretValuePolicy:
-            SecretArn: !Sub arn:aws:secretsmanager:eu-west-2:*:secret:${Environment}/contra-indicator-scoring-values-*
+            SecretArn: !Sub arn:aws:secretsmanager:eu-west-2:*:secret:/${Environment}/core/self/ci-scoring-config-*
         - Statement:
             - Sid: kmsAuditEventQueueEncryptionKeyPermission
               Effect: Allow
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -210,7 +210,7 @@ public class EvaluateGpg45ScoresHandler
 
             CredentialIssuerConfig addressCriConfig =
                     configurationService.getCredentialIssuer(addressCriId);
-            boolean isSuccessful = VcHelper.isSuccessfulVc(signedJWT, addressCriConfig, true);
+            boolean isSuccessful = VcHelper.isSuccessfulVcIgnoringCi(signedJWT, addressCriConfig);
 
             vcStatuses.add(new VcStatusDto(signedJWT.getJWTClaimsSet().getIssuer(), isSuccessful));
         }
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -18,6 +18,7 @@ import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
 import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
 import uk.gov.di.ipv.core.library.auditing.AuditExtensionGpg45ProfileMatched;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
+import uk.gov.di.ipv.core.library.domain.ContraIndicatorScores;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 import uk.gov.di.ipv.core.library.domain.gpg45.Gpg45Profile;
@@ -44,9 +45,11 @@ import uk.gov.di.ipv.core.library.service.UserIdentityService;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.ADDRESS_CRI_ID;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CI_SCORING_THRESHOLD;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CLAIM;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_EVIDENCE;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_EVIDENCE_TXN;
@@ -210,7 +213,13 @@ public class EvaluateGpg45ScoresHandler
 
             CredentialIssuerConfig addressCriConfig =
                     configurationService.getCredentialIssuer(addressCriId);
-            boolean isSuccessful = VcHelper.isSuccessfulVc(signedJWT, addressCriConfig, true);
+            Map<String, ContraIndicatorScores> ciScoresMap =
+                    configurationService.getContraIndicatorScoresMap();
+            int ciScoreThreshold =
+                    Integer.parseInt(configurationService.getSsmParameter(CI_SCORING_THRESHOLD));
+            boolean isSuccessful =
+                    VcHelper.isSuccessfulVc(
+                            signedJWT, addressCriConfig, ciScoresMap, ciScoreThreshold);
 
             vcStatuses.add(new VcStatusDto(signedJWT.getJWTClaimsSet().getIssuer(), isSuccessful));
         }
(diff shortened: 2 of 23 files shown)


Danomanic (dismissed):

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/VcHelper.java
```diff
@@ -46,23 +53,37 @@ public static boolean isSuccessfulVc(
...
     private static boolean isValidEvidence(
-            List<CredentialEvidenceItem> credentialEvidenceList, boolean isFraudAllowedA01) {
+            List<CredentialEvidenceItem> credentialEvidenceList,
+            Map<String, ContraIndicatorScores> ciScoresMap,
+            int ciScoreThreshold) {
         try {
             for (CredentialEvidenceItem item : credentialEvidenceList) {
                 if (item.getType().equals(CredentialEvidenceItem.EvidenceType.EVIDENCE)) {
-                    return Gpg45EvidenceValidator.isSuccessful(item);
+                    return Gpg45EvidenceValidator.isSuccessful(item, ciScoresMap, ciScoreThreshold);
```
dannyhayes-gds: I think for all these successful checks we can remove any checking of CI's. I don't think it really determines if a VC is successfull anymore, as if the CI is mitigated then it is successful.

I think these "success" checks should just be based on the actual gpg45 score value, i.e if it is 0 or not.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45DcmawValidator.java
```diff
@@ -1,21 +1,31 @@
...
-                return false;
+    public static boolean isSuccessful(
+            CredentialEvidenceItem item,
+            Map<String, ContraIndicatorScores> ciScoresMap,
+            int ciScoreThreshold) {
+        if (item.getFailedCheckDetails() != null) {
+            return false;
+        }
+        if (item.getValidityScore() != 0) {
+            if (item.getCi() == null || item.getCi().isEmpty()) {
```
dannyhayes-gds: I think we can get rid of the ci scoring checks here now.

Should double check with @kerrr how we want to handle the "success" status of the audit events. So when we retrieve a new VC, we sent an audit event that contains that VC's success status that uses this code. For the audit message is it fine to exclude CI checks in that status as well, or for the audit message do we want to mark it as not successful if it contains a CI?

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45EvidenceValidator.java
```diff
@@ -1,17 +1,28 @@
...
 
-    public static boolean isSuccessful(CredentialEvidenceItem item) {
-        if (item.getCi() == null || item.getCi().isEmpty()) {
-            return item.getValidityScore() != 0;
+    public static boolean isSuccessful(
+            CredentialEvidenceItem item,
+            Map<String, ContraIndicatorScores> ciScoresMap,
+            int ciScoreThreshold) {
+        if (item.getValidityScore() != 0) {
+            if (item.getCi() == null || item.getCi().isEmpty()) {
```
dannyhayes-gds: Same here

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45FraudValidator.java
```diff
@@ -11,11 +15,16 @@ private Gpg45FraudValidator() {
...
-            return item.getIdentityFraudScore() != 0;
-        } else if (item.getCi().size() == 1) {
-            return isFraudAllowedA01 && item.getCi().get(0).equals(A01);
+    public static boolean isSuccessful(
+            CredentialEvidenceItem item,
+            Map<String, ContraIndicatorScores> ciScoresMap,
+            int ciScoreThreshold) {
+        if (item.getIdentityFraudScore() != 0) {
+            if (item.getCi() == null || item.getCi().isEmpty()) {
+                return true;
```
dannyhayes-gds: Same here

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45VerificationValidator.java
```diff
@@ -1,17 +1,28 @@
...
 
-    public static boolean isSuccessful(CredentialEvidenceItem item) {
-        if (item.getCi() == null || item.getCi().isEmpty()) {
-            return item.getVerificationScore() != 0;
+    public static boolean isSuccessful(
+            CredentialEvidenceItem item,
+            Map<String, ContraIndicatorScores> ciScoresMap,
+            int ciScoreThreshold) {
+        if (item.getVerificationScore() != 0) {
+            if (item.getCi() == null || item.getCi().isEmpty()) {
```
dannyhayes-gds: Same here

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -210,7 +213,13 @@ private List<VcStatusDto> generateVcSuccessStatuses(List<SignedJWT> credentials)
...
             CredentialIssuerConfig addressCriConfig =
                     configurationService.getCredentialIssuer(addressCriId);
-            boolean isSuccessful = VcHelper.isSuccessfulVc(signedJWT, addressCriConfig, true);
+            Map<String, ContraIndicatorScores> ciScoresMap =
+                    configurationService.getContraIndicatorScoresMap();
+            int ciScoreThreshold =
+                    Integer.parseInt(configurationService.getSsmParameter(CI_SCORING_THRESHOLD));
+            boolean isSuccessful =
+                    VcHelper.isSuccessfulVc(
+                            signedJWT, addressCriConfig, ciScoresMap, ciScoreThreshold);
```
dannyhayes-gds: I don't think we should do the CI scoring here. It should be at the top level across all VC's gathered. Not against an individual VC

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -63,9 +65,20 @@ public Optional<JourneyResponse> getJourneyResponseForStoredCis(
...
-        if (foundContraIndicators) {
+        Map<String, ContraIndicatorScores> contraIndicatorScoresMap =
+                configurationService.getContraIndicatorScoresMap();
+
+        int ciScore = 0;
+        for (String ci : ciSet) {
+            ContraIndicatorScores scoresConfig = contraIndicatorScoresMap.get(ci);
+            ciScore += scoresConfig.getDetectedScore();
+        }
+        LOGGER.info("User's CI score: {}", ciScore);
```
kerrr: We should always use structured logging with separate fields for values, rather than interpolating values into strings.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorScores.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.di.ipv.core.library.domain;
+
+import lombok.Getter;
+
+@Getter
+public class ContraIndicatorScores {
```
kerrr: Bike shedding, but this seems to represent a single score, so should be `ContraindicatorScore` singular.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorScores.java
```diff
@@ -0,0 +1,20 @@
...
+    private String ci;
```
kerrr: I'd be tempted to call this ciCode.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
kerrr: Yes, agreed, the scoring is not required here.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45DcmawValidator.java
kerrr: 🤔 We shouldn't do CI scoring here. Even if we think that having a CI or not has some basing on success it's not the same issue as the scoring of CIs. As such it doesn't belong in this PR. So just leave this change out.

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/VcHelper.java
kerrr: I think the reality is that apart from Fraud, VCs won't have a non zero profile score if they contain a CI. Lets just leave this for now.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45DcmawValidator.java
dannyhayes-gds: I think a change might still need to be made potentially. This code is being called to trigger when a VC is considered successful or not during the select-cri lambda. Which probably doesn't need to count CI's into that fact anymore. We previously had it return false if a CI was present unless it was the Fraud CRI specifically with A01

blakeyp: I've updated this now so we only check for presence of CIs for the VC success check in the retrieve-cri-credential lambda which uses it for the audit log. When we record VC success in the select-cri lambda we now ignore the CI check so the journey can continue.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileEvaluator.java
blakeyp: Updated to use structured log

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorScores.java
blakeyp: Updated

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorScores.java
blakeyp: I've left this as this is mapping the JSON object we're using in the config which uses `ci` as the field

deploy/template.yaml
```diff
@@ -1483,6 +1483,7 @@ Resources:
           IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           USER_ISSUED_CREDENTIALS_TABLE_NAME: !Ref UserIssuedCredentialsV2Table
           CREDENTIAL_ISSUERS_CONFIG_PARAM_PREFIX: !Sub "/${Environment}/core/credentialIssuers"
+          CI_SCORING_THRESHOLD: !Sub "/${Environment}/core/self/ciScoringThreshold"
```
dannyhayes-gds: Do we need this ENV var? I think you've got it in the ConfigurationVariable class as well so I think that's the only one we need

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/domain/CredentialEvidenceItem.java
```diff
@@ -77,7 +77,7 @@ public Gpg45Scores.Evidence getEvidenceScore() {
         return new Gpg45Scores.Evidence(getStrengthScore(), getValidityScore());
     }
 
-    public boolean hasContraIndicators() {
+    public boolean hasContraIndicatorsExcludingFraudA01() {
```
dannyhayes-gds: I think this should be refactored slightly now. We no longer really need to be hard coding the fraud A01 scenario anymore.

I think this method should go back to its original name, and the logic inside that does the check of it being fraud etc can be removed. This method should just return true if any item has any CI at all

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
```diff
@@ -230,6 +232,30 @@ public String getCriPrivateApiKey(String criId) {
         }
     }
 
+    public Map<String, ContraIndicatorScore> getContraIndicatorScoresMap() {
+        GetSecretValueRequest valueRequest =
+                GetSecretValueRequest.builder()
+                        .secretId(
+                                String.format(
+                                        "/%s/core/self/ci-scoring-config",
```
dannyhayes-gds: We should maybe move this string secrets manager var name into the ConfigurationVariable class

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/domain/CredentialEvidenceItem.java
blakeyp: So this is used for the VC success check for the audit event in retrieve-cri-credential - I was thinking it should still be considered "success" if it's fraud with only the A01 CI?

deploy/template.yaml
blakeyp: Good spot I've removed

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigurationService.java
blakeyp: Done

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/domain/CredentialEvidenceItem.java
dannyhayes-gds: I think that would be a change in the behaviour from what we currently have. Currently for the audit event, if it's fraud with A01 that is marked as not successful. Its only marked as successful for the select-cri bit

dannyhayes-gds: I think for the audit event the logic we went with originally, was that if any CI is on the VC then it would be marked as not successful.

blakeyp: Ah I see what you mean now yeah, I forgot about the flag being set to false there. Will update this then

dannyhayes-gds (approved):

PR #566, opened 2022-11-14
PYIC-2143: Add CI's to the session to be mitigated

dannyhayes-gds:
### What changed
Store in the user session when a CI should be mitigated. This new feature will be hidden behind a SSM param feature flag to ensure this isn't used in any environments before the rest of the required work is complete.

When a CI is returned from the CI storage system it is added to the session to be mitigated. If no new CI's are received but there is still a CI in the session still marked as mitigatable then it skips all the CI scoring and gpg45 profile checking.

If the CI in the session is no longer mitigatable or there aren't any CI's at all then the normal CI scoring and gpg45 profile checking is done.

### Why did it change
This new data in the session will be used by the `select-cri` lambda to decide when to send a user on a mitigation journey. When the user is in a mitigation journey, most of the normal logic of the evaluate gpg45 lambda can be skipped. The select-cri lambda will be responsible for updating this new session field to be not "mitigatable" once all the mitigation journeys have been completed.

### Issue tracking

- [PYIC-2143](https://govukverify.atlassian.net/browse/PYIC-2143)

### DIFF
--- a/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
+++ b/lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
@@ -99,9 +99,7 @@ public class EvaluateGpg45ScoresHandler
         this.configurationService = new ConfigurationService();
         this.userIdentityService = new UserIdentityService(configurationService);
         this.ipvSessionService = new IpvSessionService(configurationService);
-        this.gpg45ProfileEvaluator =
-                new Gpg45ProfileEvaluator(
-                        new CiStorageService(configurationService), configurationService);
+        this.gpg45ProfileEvaluator = new Gpg45ProfileEvaluator(configurationService);
         this.ciStorageService = new CiStorageService(configurationService);
         this.auditService =
                 new AuditService(AuditService.getDefaultSqsClient(), configurationService);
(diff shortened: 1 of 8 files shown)


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -123,46 +133,43 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                    journeyResponse = JOURNEY_END;
-                    message.with("lambdaResult", "A GPG45 profile has been met")
-                            .with("journeyResponse", JOURNEY_END);
-                } else {
-                    journeyResponse = JOURNEY_NEXT;
-                    message.with("lambdaResult", "No GPG45 profiles have been met")
-                            .with("journeyResponse", JOURNEY_NEXT);
-                }
+            List<ContraIndicatorItem> ciItems;
+            ciItems =
```
blakeyp, line 135: Now we're fetching the CIs here is it worth just passing `ciItems` into the  `getJourneyResponseForStoredCis` call, rather than making a second request to the CI storage system inside Gpg45ProfileEvaluator?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -123,46 +133,43 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                LOGGER.info(message);
-
-                return ApiGatewayResponseGenerator.proxyJsonResponse(
-                        HttpStatus.SC_OK, journeyResponse);
+            boolean ciMitigationJourneysEnabled =
+                    Boolean.parseBoolean(
+                            configurationService.getSsmParameter(CI_MITIGATION_JOURNEYS_ENABLED));
+            Optional<JourneyResponse> mitigationJourneyResponse =
+                    getNextMitigationJourneyResponse(
+                            ciMitigationJourneysEnabled, ipvSessionItem, ciItems);
```
blakeyp: Not a big deal but might be clearer to just pull the `if (ciMitigationsJourneyEnabled)` check outside of the method rather than passing a boolean flag in which slightly conceals what it's doing

blakeyp (dismissed):

Danomanic (dismissed):

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
dannyhayes-gds: Yeah good catch. I'll update it

blakeyp (approved):

PR #568, opened 2022-11-17
Pyic 2144

dannyhayes-gds:
### What changed
Update end-mitigation-journey lambda to collect and send the mitigating VC's to the CI storage system for the MJ01 mitigation journey. This will look for a new fraud VC that does not contain the same matching CI.

The lambda will also end the mitigation journey for ipv-core by updating in the user sessions to say that the journey has completed.

### Why did it change
This is the 1st mitigation journey that IPV core needs to be able to handle. This lambda will be called as the final step to any mitigation journey.

### Issue tracking

- [PYIC-2144](https://govukverify.atlassian.net/browse/PYIC-2144)


lambdas/end-mitigation-journey/src/main/java/uk/gov/di/ipv/core/endmitigationjourney/EndMitigationJourneyHandler.java
```diff
@@ -5,21 +5,215 @@
...
+        return ApiGatewayResponseGenerator.proxyJsonResponse(HttpStatus.SC_OK, JOURNEY_NEXT);
+    }
+
+    private Optional<List<String>> validateMitigationJourney(
+            List<String> credentials,
+            ContraIndicatorItem ciItem,
+            String mitigationId,
+            ConfigurationService configurationService)
+            throws UnknownMitigationJourneyException {
+        // check mitigation id and call vorrect validation method
```
blakeyp: Typo *correct

lambdas/end-mitigation-journey/src/main/java/uk/gov/di/ipv/core/endmitigationjourney/EndMitigationJourneyHandler.java
```diff
@@ -5,21 +5,215 @@
...
+                    ipvSessionItem.getClientSessionDetails().getGovukSigninJourneyId(),
+                    ipAddress);
+        } catch (CiPostMitigationsException e) {
+            StringMapMessage mapMessage =
+                    new StringMapMessage()
+                            .with(
+                                    "message",
+                                    "Failed to send mitigation request to the CI storage system")
+                            .with("reason", e.getMessage());
+            LOGGER.error(mapMessage);
```
blakeyp, line 215: Shouldn't we be failing the journey here because CI storage has probably failed to mark the CIs as mitigated? (or is the intention to just swallow the error for now?)

lib/src/main/java/uk/gov/di/ipv/core/library/domain/PostCiMitigationRequest.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Data
+public class PostCiMitigationRequest {
+    @SerializedName(value = "govuk_signin_journey_id")
+    private final String govukSigninJourneyId;
+
+    @SerializedName(value = "ip_address")
+    private final String ipAddress;
+
+    @SerializedName(value = "signed_jwt_list")
```
blakeyp: We've named this `signed_jwts` in the CI storage request, but can easily update it if we want to

dannyhayes-gds: 👍  I'll update it here

lambdas/end-mitigation-journey/src/main/java/uk/gov/di/ipv/core/endmitigationjourney/EndMitigationJourneyHandler.java
dannyhayes-gds: Yeah for now I thought it would be better to just log the error and move on. This way it means that the lambda will still mark this particular mitigation journey as complete so that we don't get stuck in a loop retrying the same mitigation journey. And if the CI Storage failed for whatever reason then we can still attempt an identity verification with CI scoring etc

blakeyp (approved):

PR #573, opened 2022-11-23
PYIC-2202: Update userIdentityService to handle new passport version of DCMAW credential

dannyhayes-gds:
### What changed
Start looking at the DCMAW VC when trying to generate the passport claim.

Check that the VC is successful before using it to generate the passport claim.

Stop throwing exceptions if the passport or drivingPermit property cannot be found when generating passport/drivingPermit claim. Instead just logs a warning and returns an empty optional.

### Why did it change
DCMAW will soon release a new passport journey meaning that we could receive slightly different VC's. These changes ensure that the various user identity claims are still generated when we encounter the different possible DCMAW VC's.

### Issue tracking

- [PYIC-2202](https://govukverify.atlassian.net/browse/PYIC-2202)

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -14,6 +14,7 @@ import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.MapMessage;
+import org.apache.logging.log4j.message.StringMapMessage;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.domain.BirthDate;
 import uk.gov.di.ipv.core.library.domain.DebugCredentialAttributes;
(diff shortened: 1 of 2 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -54,7 +55,8 @@ public class UserIdentityService {
     private static final String DRIVING_PERMIT_PROPERTY_NAME = "drivingPermit";
     private static final List<String> ADDRESS_CRI_TYPES =
             List.of(ADDRESS_PROPERTY_NAME, "stubAddress");
-    private static final List<String> PASSPORT_CRI_TYPES = List.of("ukPassport", "stubUkPassport");
+    private static final List<String> PASSPORT_CRI_TYPES =
```
galund, line 58: At some point it might be worth turning this around and having it in configuration. For example, against each issuer URI we could have a list of properties we can get from it, eg "credentialSubject.passport", "evidence", "credentialSubject.drivingPermit".

There's something of a smell from this "type" system ending up with these things getting hard-coded here. It's not urgent to fix, but it's only going to get more messy until we do :-)

kerrr: Yeah, Danny and I talked about this needing an overhaul at some stage. I've been thinking about how much Core should just introspect VCs and go with what is there vs validate what we expect from a particular CRI . Your suggestion sounds like a good balance.

kerrr: This update is intentionally the minimum to get it working.

kerrr (approved):
Looks good to me 👍🏼 🚀

PR #574, opened 2022-11-24
BAU: User StringMapMessage

kerrr:
### What changed

Use StringMapMessage in place of MapMessage.

### Why did it change

When creating structured logs we were using a raw MapMessage. Since we always use strings for key and values, we can use StringMapMessage and avoid using a raw generic type or dealing with the complex generic type of MapMessage.


dannyhayes-gds (approved):
Thanks for sorting this! I've been meaning to do this as well but kept forgetting 😄 🚀

PR #582, opened 2022-11-29
PYIC-2213: Update internal API Gateway to provide endpoint for new lambda

Danomanic:
### What changed
- Update internal API Gateway to provide endpoint for new lambda

### Issue tracking

- [PYIC-2213](https://govukverify.atlassian.net/browse/PYIC-2213)

### DIFF
--- a/openAPI/core-back-internal.yaml
+++ b/openAPI/core-back-internal.yaml
@@ -263,6 +263,23 @@ paths:
         passthroughBehavior: "when_no_match"
         type: "aws_proxy"
 
+  /journey/check-existing-identity:
+    get:
+      description: "Check Existing Identity"
+      responses:
+        200:
+          description: "Check Existing Identity response"
+          content:
+            application/json:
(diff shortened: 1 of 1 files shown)


openAPI/core-back-internal.yaml
```diff
@@ -263,6 +263,23 @@ paths:
...
+  /journey/check-existing-identity:
+    get:
+      description: "Check Existing Identity"
+      responses:
+        200:
+          description: "Check Existing Identity response"
+          content:
+            application/json:
+              schema:
+                type: "object"
```
dannyhayes-gds: I think this should be returning a journeyResponse each time so this can be the following
```suggestion
                $ref: "#/components/schemas/journeyType"
```

Danomanic: Nice spot - done

dannyhayes-gds (approved):

PR #583, opened 2022-11-29
PYIC-2214: Update journey engine state machine config for new check existing identity journey.

Danomanic:
### What changed

Update journey engine state machine config for new check existing identity journey.

### Issue tracking

- [PYIC-2214](https://govukverify.atlassian.net/browse/PYIC-2214)

### DIFF
--- a/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
@@ -62,6 +62,17 @@ INITIAL_IPV_JOURNEY:
     next:
       type: basic
       name: next
+      targetState: IPV_IDENTITY_START_PAGE
+      response:
+        type: page
+        pageId: page-ipv-identity-start
+IPV_IDENTITY_START_PAGE:
+  name: IPV_IDENTITY_START_PAGE
+  parent: null
+  events:
+    next:
+      type: basic
+      name: next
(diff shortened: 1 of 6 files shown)


lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
```diff
@@ -58,6 +58,17 @@ MITIGATION_JOURNEY:
 INITIAL_IPV_JOURNEY:
   name: INIT
   parent: null
+  events:
+    next:
+      type: basic
+      name: next
+      targetState: CHECK_EXISTING_IDENTITY
```
dannyhayes-gds, line 76: I think this event is being applied to the wrong state. Currently this means we are checking the for an existing identity as the very first thing in IPV Core. But from the figma journey diagram, I think we first need to display the IPV_IDENTITY_START_PAGE then do the check after that

kerrr: Yeah, sorry, I think my description was wrong in the ticket was wrong.

kerrr: I've updated the ticket with the correct description (I think).

Danomanic: Corrected this!

dannyhayes-gds (approved):

PR #586, opened 2022-11-30
BAU: Set vot value to P2 on a successful reuse journey

dannyhayes-gds:
### What changed
Set the vot value in the users session to be P2 if the check-existing-identity lambda matches with a gpg45 profile.

### Why did it change
This value will be loaded from the users session when we build the final identity and will be returned to orchestration.


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -112,6 +113,10 @@ public APIGatewayProxyResponseEvent handleRequest(
                                             "Matched profile and within CI threshold so returning reuse journey")
                                     .with("profile", matchedProfile.get().getLabel());
                     LOGGER.info(message);
+
+                    ipvSessionItem.setVot(VOT_P2);
+                    ipvSessionService.updateIpvSession(ipvSessionItem);
```
Danomanic, line 118: I'll update the tests for this post-merge.

But looks good.

Danomanic (approved):

PR #596, opened 2022-12-06
BAU: Rename UserIssuedCredential to VcStoreItem

kerrr:
### What changed

Rename `UserIssuedCredentialItem` to `VcStoreItem`.

This is a mostly mechanical IDE driven rename refactor.   

### Why did it change

The UserIssuedCredentialsItem is the DynamoDB item for the "Vc Store". We will be refactoring how we handle various VCs in the system and we want to make sure that we are dealing with the wrapping DynamoDB item here.

### Reviewing

It will be easiest to review by commits.
 1. The first commit is mostly mechanical IDE driven rename of the class and variables that the IDE detected.
 2. Next is removal of the method that was not required that had `UserIssuedCredentials` in the name.
 3. Next is moving debug credentials handling out of user identity service 
 4. Next is a spotless fixup
 5. Next is some additional variable renaming.


lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -73,76 +64,71 @@ public UserIdentityService(ConfigurationService configurationService) {
...
 
     public UserIdentityService(
-            ConfigurationService configurationService,
-            DataStore<UserIssuedCredentialsItem> dataStore) {
+            ConfigurationService configurationService, DataStore<VcStoreItem> dataStore) {
         this.configurationService = configurationService;
         this.dataStore = dataStore;
     }
 
     public List<String> getUserIssuedCredentials(String userId) {
```
dannyhayes-gds, line 79: We should probably rename this method, maybe
```suggestion
    public List<String> getVcStoreItemCredentials(String userId) {
```

lambdas/build-debug-credential-data/src/main/java/uk/gov/di/ipv/core/builddebugcredentialdata/BuildDebugCredentialDataHandler.java
```diff
@@ -56,11 +71,54 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                    userIdentityService.getUserIssuedDebugCredentials(
-                            clientSessionDetailsDto.getUserId()));
+                    getUserIssuedDebugCredentials(clientSessionDetailsDto.getUserId()));
         } catch (HttpResponseExceptionWithErrorBody e) {
             return ApiGatewayResponseGenerator.proxyJsonResponse(
                     e.getResponseCode(), e.getErrorBody());
         }
     }
+
+    public Map<String, String> getUserIssuedDebugCredentials(String userId) {
```
dannyhayes-gds, line 81:
```suggestion
    public Map<String, String> getVcStoreItemDebugCredentials(String userId) {
```

lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
kerrr: I was intentionally leaving this one. It is badly named, but it returns the underlying raw VC jwt, not the VcStoreItem. I think we want to have a richer object that wraps the raw jwt, but that's not this PR.

lambdas/build-debug-credential-data/src/main/java/uk/gov/di/ipv/core/builddebugcredentialdata/BuildDebugCredentialDataHandler.java
kerrr: Again, I think this is a different thing. I was looking at renaming and thought just pulling it out into the only place we used it would be a good first step. Then we can figure out if this is even a thing we need, or if there is a better way of doing it.

dannyhayes-gds (approved):

PR #601, opened 2022-12-08
PYIC-2241 Decoupled CoreEncryptionKeyArn

vam-gds:
### What changed
PYIC-2241 Decoupled CoreEncryptionKeyArn
This is imported from encryption-key-${environment}, but the stack needs to be renamed so that it can be migrated to secure pipelines. This PR hardcodes the value and removes the import which should make it possible to rename the stacks safely.

### Why did it change
PYIC-2241 Decoupled CoreEncryptionKeyArn
This is imported from encryption-key-${environment}, but the stack needs to be renamed so that it can be migrated to secure pipelines. This PR hardcodes the value and removes the import which should make it possible to rename the stacks safely.

### Issue tracking

- [PYIC-2241](https://govukverify.atlassian.net/browse/PYIC-2241)

### Other considerations

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -71,6 +71,7 @@ Mappings:
       criReturnStepFunctionLogLevel: ALL
       pgw500ErrorLimit: 2
       egw500ErrorLimit: 2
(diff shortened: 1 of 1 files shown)


TobySaundersGDS (approved):
👍

PR #611, opened 2023-01-05
BAU Remove duplicated addition to visited CRIs in session

blakeyp:
### What changed

Remove duplicated addition to visited CRIs stored in session. Only add once, on credential retrieval.

### Why did it change

We were adding the visited CRIs twice, once in the CRI access token handler and again in the CRI credential retrieval handler, causing the list to contain duplicates in the session.


kerrr (commented):
👍🏼

kerrr (approved):
Remember to select approve 🙄

PR #614, opened 2023-01-06
PYIC-2298 Clear oauth session on process journey step

blakeyp:
### What changed

Update process-journey-step to clear the oauth session on any state change. Also add a new check in validate-oauth-callback as the oauth session may now be null.

### Why did it change

We need to make sure that we are managing the state of OAuth connections with CRIs in a more robust way. We should track this “OAuth Session” so that after a user has returned form a CRI we close the session, so that it is no longer valid. In particular this fixes a 'bug' spotted where you can go back from the pre-kbv page to fraud CRI and get a new fraud VC issued.

### Issue tracking
- [PYIC-2298](https://govukverify.atlassian.net/browse/PYIC-2298)

[PYIC-2298]: https://govukverify.atlassian.net/browse/PYIC-2298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/ProcessJourneyStepHandler.java
+++ b/lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/ProcessJourneyStepHandler.java
@@ -12,6 +12,7 @@ import software.amazon.lambda.powertools.tracing.Tracing;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.config.EnvironmentVariable;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.dto.CredentialIssuerSessionDetailsDto;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.helpers.StepFunctionHelpers;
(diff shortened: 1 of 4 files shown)


lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/ProcessJourneyStepHandler.java
```diff
@@ -121,6 +122,10 @@ private Map<String, Object> executeJourneyEvent(
                     journeyStep,
                     ipvSessionItem);
 
+            clearOauthSessionIfExists(ipvSessionItem);
+
+            ipvSessionService.updateIpvSession(ipvSessionItem);
```
dannyhayes-gds, line 127: ~~Might be worth moving this update call inside the `clearOauthSessionIfExists` method, so that we are only making the dynamoDB update when we need to?~~

dannyhayes-gds: Oh wait never mind, I see now its because we are making the update no matter anyways for other reasons. Ignore the above

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -112,6 +113,10 @@ public Map<String, Object> handleRequest(CredentialIssuerRequestDto request, Con
 
             ipvSessionService.updateIpvSession(ipvSessionItem);
 
+            var mapMessage =
+                    new StringMapMessage().with("message", "Successfully validated oauth callback");
```
dannyhayes-gds: Might be useful to add the criId of the request into this log message as another property?

blakeyp: Good idea I've added that now

dannyhayes-gds (approved):

PR #618, opened 2023-01-09
PYIC-2300: Add attempt-recovery event to all states

dannyhayes-gds:
### What changed
Add new ATTEMPT_RECOVERY_STATE to the state machine.

Update all other states to use this as a parent in order to inherit the attempt-recovery event which sends users to the evaluategpg45Scores lambda.

### Why did it change
This event could be used potentially from any point within the system so we want all states to be able to handle this event.

### Issue tracking

- [PYIC-2300](https://govukverify.atlassian.net/browse/PYIC-2300)

[PYIC-2300]: https://govukverify.atlassian.net/browse/PYIC-2300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Danomanic (approved):
LGTM

PR #663, opened 2023-01-31
PYIC-2413 Change SNS event to SQS event

blakeyp:
### What changed

Read the message from the body of the incoming SQS event not SNS event

### Why did it change

As we have an SQS queue sitting in between the SNS topic and the lambda, the lambda actually needs to read the SNS message from an SQS event, inside the body of the record

### Issue tracking
- [PYIC-2413](https://govukverify.atlassian.net/browse/PYIC-2413)

[PYIC-2413]: https://govukverify.atlassian.net/browse/PYIC-2413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Danomanic (approved):
LGTM

PR #667, opened 2023-02-02
BAU Set up logger and log not throw when no VCs found

blakeyp:
### What changed

Set up logger and add info logs with context. Also in the case when no VCs are found for the given user id, log rather than throwing an error - because this is expected behaviour for users who haven't been through IPV

### Why did it change

So we can do structured logging from this lambda, and not throw errors for expected behaviour


Danomanic (approved):
LGTM

PR #673, opened 2023-02-06
PYIC-2419 Add audit event delete

blakeyp:
### What changed

Send audit event `IPV_DELETE_USER_DATA` when lambda executes successfully and has deleted some data. Include `source-topic-arn` which identifies the SNS topic the message came from.

### Why did it change

So we have audit logs for deleted data

### Issue tracking
- [PYIC-2419](https://govukverify.atlassian.net/browse/PYIC-2419)

[PYIC-2419]: https://govukverify.atlassian.net/browse/PYIC-2419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/delete-user-data/src/config.ts
+++ b/lambdas/delete-user-data/src/config.ts
@@ -1,5 +1,5 @@
 export const config = {
-  componentId: "ipv-core-delete-user-data",
+  environment: process.env.ENVIRONMENT || "",
   userIssuedCredentialsTableName: process.env.USER_ISSUED_CREDENTIALS_TABLE_NAME || "",
   sqsAuditEventQueueUrl: process.env.SQS_AUDIT_EVENT_QUEUE_URL || "",
 };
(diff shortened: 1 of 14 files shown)


lambdas/delete-user-data/src/config.ts
```diff
@@ -1,3 +1,5 @@
 export const config = {
+  componentId: "ipv-core-delete-user-data",
```
kerrr: `componentId` should be the URI representing core. In production this is https://identity.account.gov.uk and is the config param at `core/self/componentId`

kerrr:
Have you contacted TxMA about adding a new audit event?

blakeyp:
> Have you contacted TxMA about adding a new audit event?

Haven't yet, will do

lambdas/delete-user-data/src/config.ts
blakeyp: Changed this to now get the componentId from SSM

kerrr (approved):

PR #695, opened 2023-03-01
BAU: Refactor getCriReponse method in SelectCriHandler

AmritSidhu:
### What changed

Extracted the logic of checking whether the user has a passport VC into a separate method 
Removed the unnecessary else statements after the return statements 
Used ternary operators for simple if-else statements Used string interpolation to make the log messages more readable Reordered the conditions in the if statements to prioritise the simpler ones first

### Why did it change

The method is convoluted and difficult to read. This should hopefully make it that bit easier

### DIFF
--- a/lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
+++ b/lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
@@ -241,17 +241,14 @@ public class SelectCriHandler
             String journeyId,
             String userId)
             throws ParseException {
-        CredentialIssuerConfig criConfig = configService.getCredentialIssuer(criId);
 
+        CredentialIssuerConfig criConfig = configService.getCredentialIssuer(criId);
         Optional<VcStatusDto> vc = getVc(currentVcStatuses, criConfig.getAudienceForClients());
+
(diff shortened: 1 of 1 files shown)


lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -241,47 +241,34 @@ private Optional<JourneyResponse> getCriResponse(
...
-        } else if (Boolean.FALSE.equals(vc.get().getIsSuccessfulVc())) {
-            var message =
-                    new StringMapMessage()
-                            .with(
-                                    "message",
-                                    "User has a previous failed visit to a cri due to a failed identity check")
-                            .with("criId", criId);
-            LOGGER.info(message);
+            LOGGER.info(
+                    "User has a previous failed visit to a CRI due to an OAuth error for CRI ID {}",
```
kerrr: We explicitly don't what to have values interpolated into log messages. We want to use structured logging,  separating out the values into their own fields.

message.message is not ideal though. Where a better field name does not exist we should use `message.description`

kerrr:
Other than the logging, looks good.

kerrr (commented):
👍🏼 🚀

kerrr (approved):
🙄

PR #699, opened 2023-03-01
BAU: Rename CredentialIssuerRequestDto to CriCallbackRequest

kerrr:
### What changed

Rename `CredentialIssuerRequestDto` to `CriCallbackRequest`

### Why did it change

This object represents the data we receive in the OAuth callback from the Credential Issuer. Since we also make requests to Credential Issuers this rename is helpful in scoping the data to the callback.

Also rename variables and methods to better reflect this change.

### DIFF
--- a/lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
+++ b/lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
@@ -86,13 +86,13 @@ public class ValidateOAuthCallbackHandler
     @Override
     @Tracing
     @Logging(clearState = true)
-    public Map<String, Object> handleRequest(CriCallbackRequest request, Context context) {
+    public Map<String, Object> handleRequest(CriCallbackRequest callbackRequest, Context context) {
         LogHelper.attachComponentIdToLogs();
 
         IpvSessionItem ipvSessionItem = null;
 
         try {
-            String ipvSessionId = request.getIpvSessionId();
+            String ipvSessionId = callbackRequest.getIpvSessionId();
             if (ipvSessionId == null) {
                 throw new HttpResponseExceptionWithErrorBody(
                         HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_IPV_SESSION_ID);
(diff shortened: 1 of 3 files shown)


lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -86,37 +86,37 @@ public ValidateOAuthCallbackHandler() {
     @Override
     @Tracing
     @Logging(clearState = true)
-    public Map<String, Object> handleRequest(CredentialIssuerRequestDto request, Context context) {
```
rpayal, line 89: Param name can be criCallbackRequest instead of callbackRequest.

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
```diff
@@ -242,7 +242,7 @@ private void setIpvSessionCRIAuthorizationCode(
     }
 
     @Tracing
-    private String getPersistedOauthState(CredentialIssuerRequestDto request) {
```
rpayal, line 245: Param name to criCallbackRequest.

rpayal (dismissed):

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
kerrr: I think I'm happy with `callbackRequest` as the var name. It's a reasonable balance between verbosity and conciseness in this context.

lambdas/validate-oauth-callback/src/main/java/uk/gov/di/ipv/core/validateoauthcallback/ValidateOAuthCallbackHandler.java
kerrr: Oops. Good spot that this one wasn't renamed. Using `callbackRequest` as used elsewhere.

Danomanic (approved):

PR #705, opened 2023-03-09
PYIC-2536 Update select-cri for visited driving licence CRI

blakeyp:
### What changed

Update select-cri to handle visited driving licence CRI - only return 'passport' i.e. doc check page response if neither passport nor driving licence have been visited. Also handle the case where a returning DCMAW user already has a web journey driving licence VC.

### Why did it change

To handle the onward journey from the new driving licence CRI so it knows where to go next

### Issue tracking
- [PYIC-2536](https://govukverify.atlassian.net/browse/PYIC-2536)

[PYIC-2536]: https://govukverify.atlassian.net/browse/PYIC-2536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
+++ b/lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
@@ -144,7 +144,14 @@ public class SelectCriHandler
                         passportCriId,
                         passportCriId,
                         userId);
-        if (passportResponse.isPresent()) {
+        Optional<JourneyResponse> drivingLicenceResponse =
+                getCriResponse(
+                        visitedCredentialIssuers,
+                        currentVcStatuses,
+                        drivingLicenceCriId,
+                        drivingLicenceCriId,
+                        userId);
+        if (passportResponse.isPresent() && drivingLicenceResponse.isPresent()) {
             return passportResponse.get();
         }
 
@@ -255,8 +262,11 @@ public class SelectCriHandler
 
         if (vc.isEmpty()) {
             if (userHasNotVisited(visitedCredentialIssuers, criId)) {
-                if (criId.equals(dcmawCriId) && hasPassportVc(currentVcStatuses)) {
-                    LOGGER.info("User already has a passport VC, continuing a web journey");
+                if (criId.equals(dcmawCriId)
+                        && (hasPassportVc(currentVcStatuses)
+                                || hasDrivingLicenceVc(currentVcStatuses))) {
+                    LOGGER.info(
+                            "User already has a passport or driving licence VC, continuing a web journey");
                     return Optional.of(
                             getNextWebJourneyCri(
                                     visitedCredentialIssuers, currentVcStatuses, userId));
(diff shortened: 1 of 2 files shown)


lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
```diff
@@ -144,7 +144,14 @@ private JourneyResponse getNextWebJourneyCri(
...
                         userId);
-        if (passportResponse.isPresent()) {
+        Optional<JourneyResponse> drivingLicenceResponse =
+                getCriResponse(
+                        visitedCredentialIssuers,
+                        currentVcStatuses,
+                        drivingLicenceCriId,
+                        drivingLicenceCriId,
+                        userId);
+        if (passportResponse.isPresent() && drivingLicenceResponse.isPresent()) {
```
Danomanic, line 154: Just out of interest, does this still work if the logic is:

``` if (passportResponse.isPresent() || drivingLicenceResponse.isPresent())```

File: lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java [hunk 2]
```diff
@@ -144,7 +144,14 @@ private JourneyResponse getNextWebJourneyCri(
...
             return passportResponse.get();
```
rpayal, line 155: Returning only passportResponse. Do we need to include drivingLicenceResponse in return too ?

kerrr (dismissed):
Looks good to me 🚀

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
kerrr: 🤔

blakeyp: It wouldn’t as response.isPresent() is true if that CRI hasn’t been visited yet - so we need both to not be present

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
blakeyp: In this case they’d actually be the same, both send you to the doc check page which drives which of the two the user actually goes to. But correct me if I’m wrong @Danomanic

Danomanic: It doesn't use the response (apart from checking it is present), so passing back either the passport or driving licence is fine here.

lambdas/select-cri/src/main/java/uk/gov/di/ipv/core/selectcri/SelectCriHandler.java
Danomanic: Fair enough. Thank you.

Danomanic (approved):
Looks good, but I think @kerrr may been looking at a different approach.

kerrr (approved):
Yes, this does what we think it should, but it's confusing to read. We will need to refactor for greater clarity.

🚀

PR #715, opened 2023-03-17
PYIC-2422 Return user to multi-doc-page on access-denied

blakeyp:
### What changed

On an access denied error from passport or driving licence CRI (and when a user cancels out of a CRI) return them back to the multiple doc selection page - if both CRIs are enabled. Otherwise go to pyi-no-match as before.

### Why did it change

The user has the option of using their passport or driving licence instead in this scenario.

### Issue tracking
- [PYIC-2422](https://govukverify.atlassian.net/browse/PYIC-2422)

[PYIC-2422]: https://govukverify.atlassian.net/browse/PYIC-2422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
@@ -255,9 +255,25 @@ SELECT_CRI:
 CRI_UK_PASSPORT:
   name: CRI_UK_PASSPORT
   parent: CRI_STATE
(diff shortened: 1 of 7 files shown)


rpayal (dismissed):

TalatAydinGDS (dismissed):

TalatAydinGDS (approved):
Looks good to me

PR #771, opened 2023-04-21
PYIC-2667 Use new CRI config api key flags

blakeyp:
### What changed

Use new `requiresApiKey` flags in the CRI config to check whether the CRI requires an api key or not, and only try to fetch the api key if it does require one.

### Why did it change

In the retrieve CRI access token + credential lambdas we always try to retrieve the CRI's api key from secrets manager. In cases where the CRI or env doesn't need an api key, this causes an unnecessary call to secrets provider and results in a warning log to say the api key wasn't found. 

This prevents this by using a new flag in the CRI config to check whether the CRI requires an api key or not. Also makes the existing log error level, so it's flagged as an error when the api key is missing when we actually expect it to be there.

### Issue tracking
- [PYIC-2667](https://govukverify.atlassian.net/browse/PYIC-2667)

[PYIC-2667]: https://govukverify.atlassian.net/browse/PYIC-2667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandlerTest.java
+++ b/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandlerTest.java
@@ -141,7 +141,8 @@ class BuildCriOauthRequestHandlerTest {
                         "{}",
                         RSA_ENCRYPTION_PUBLIC_JWK,
                         "http://www.example.com/audience",
-                        URI.create("http://www.example.com/callback/criId"));
+                        URI.create("http://www.example.com/callback/criId"),
+                        true);
 
         addressCredentialIssuerConfig =
                 new CredentialIssuerConfig(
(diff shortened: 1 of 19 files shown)


AmritSidhu (approved):
LGTM so will approve. However you may want to check how we do isUnavailable flag for example, so that way you wont have to add to the constructor every time and you can call the param directly in its own method. May help with consistency

blakeyp:
> LGTM so will approve. However you may want to check how we do isUnavailable flag for example, so that way you wont have to add to the constructor every time and you can call the param directly in its own method. May help with consistency

Thanks @AmritSidhu good point but I think as this flag is part of the nested CRI config under `connections` it makes more sense to keep it in the class, unlike `unavailable` which is a level above

kerrr:
I think this is fine. There will definitely be a trade off in different bits of code it's not worth while grabbing all CRI config when you just need a single value. So we should watch for that, but I think this change is fine.

Looking forward there are some things in here that we will want to look at. `CredentialIssuerService` is not particularly well factored ATM. Two separate lambdas call this shared class, but they each use non overlapping sections of code. I don't think we gain much from this, when we could have the code sections discretely in each lambda.

Also there are a bunch of logging in here that use parameterised strings. It would be good to clean that up when we find them. In separate commits :)

🚀

PR #776, opened 2023-04-27
The correlation should be against any VCs containing a name or a birth date

TalatAydinGDS:
…hdate

### What changed

The correlation should be against any VCs containing a name or birthdate. If so, should the correlation check only be performed if the field is present?  

### Why did it change

 If we don't see a name or birth date properties in credentialIssuer's VC.  We don't throw any exception from correlation check for name and birth date.

### Issue tracking

- [PYIC-2688](https://govukverify.atlassian.net/browse/PYIC-2688)

### Other considerations

[PYIC-2688]: https://govukverify.atlassian.net/browse/PYIC-2688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -30,6 +30,7 @@ import java.text.ParseException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -188,13 +189,16 @@ private <T> T getJsonProperty(
             JsonNode jsonNode,
             String propertyName,
             String credentialIssuer,
-            CollectionType valueType)
+            CollectionType valueType,
+            boolean validateCorrelation)
```
blakeyp, line 193: Would it be clearer to make this more generic, something like `shouldThrowErrorIfMissing`?

TalatAydinGDS: I used the validateCorrelation that we decided together on in today's our call. I think this is more appropriate as it not only throws an exception, but also returns an empty list. There is business according this param in method not only error message. Therefore, I think validateCorrelation is more suitable.

blakeyp (approved):

PR #785, opened 2023-05-02
BAU: Removed unused lines from pre-merge-checks.yml which are now car…

chris-dco:
### What changed

Removes unused code from build pipeline which previous triggered the integration test to run.

### Why did it change

Integration tests was previously commented out, however the code to log into AWS was still present. This caused Dependabot to fail as it did not have permissions to access secrets. 

This PR removes all of this unused code and resolve the dependabot issues.

### Issue tracking
N/A

### Other considerations

N/A


TobySaundersGDS (approved):
Looks good to  me

PR #788, opened 2023-05-04
[PYIC-2707] Add featureSet property to configuration services

therealslimsladey:
### What changed

Added featureSet property to the configuration services and constructors to accept the featureSet.

### Why did it change

Configuration returned by the configuration services will depend on whether there is a featureSet in play or not.

### Issue tracking

- [PYIC-2707](https://govukverify.atlassian.net/browse/PYIC-2707)

- No environment variables or secrets were added or changed

### Other considerations

[PYIC-2707]: https://govukverify.atlassian.net/browse/PYIC-2707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -53,12 +53,20 @@ public class ConfigService {
     private final SecretsProvider secretsProvider;
     private final ObjectMapper objectMapper = new ObjectMapper();
 
-    public ConfigService(SSMProvider ssmProvider, SecretsProvider secretsProvider) {
+    private final String featureSet;
+
+    public ConfigService(
+            SSMProvider ssmProvider, SecretsProvider secretsProvider, String featureSet) {
         this.ssmProvider = ssmProvider;
         this.secretsProvider = secretsProvider;
+        this.featureSet = featureSet;
     }
 
-    public ConfigService() {
+    public ConfigService(SSMProvider ssmProvider, SecretsProvider secretsProvider) {
+        this(ssmProvider, secretsProvider, null);
+    }
+
+    public ConfigService(String featureSet) {
         if (isRunningLocally()) {
             this.ssmProvider =
                     ParamManager.getSsmProvider(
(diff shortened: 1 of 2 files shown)


chris-dco (approved):
Looks good.

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -53,12 +53,20 @@ public class ConfigService {
     private final SecretsProvider secretsProvider;
     private final ObjectMapper objectMapper = new ObjectMapper();
 
-    public ConfigService(SSMProvider ssmProvider, SecretsProvider secretsProvider) {
+    private final String featureSet;
```
AmritSidhu, line 56: I think this becomes unused as you are setting `String featureSet` in the constructor on line 59.

therealslimsladey:
Sorry, I'm not sure I follow your comment. The instance variable `featureSet` still has to be declared, the constructor is simply initialising it to its immutable value.

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
therealslimsladey: Instance variable `featureSet` still needs to be declared though. Maybe I'm misunderstanding your comment?

AmritSidhu: Apologies my java is rusty, I forgot you had to declare the types when passing it in eventhough already defining it. However I think my comment still somewhat stands. The `private final String featureSet` can be removed as its unused / not declared here. Does it not need to be set before being passed into the constructor if you are defining it as a final string like that

AmritSidhu: Ah i've just noticed what im referring to is also showing as a code smell if you take a look at the SonarCloud report

therealslimsladey: It is currently unused, but we're getting the instance variable in now so the constructors with featureSet are available for the backend services.

AmritSidhu (approved):

PR #810, opened 2023-05-15
Update the Lambda to support both API Gateway and JSON requests

TalatAydinGDS:
### What changed

Update the build-cri-oauth-request lambda so that it can be called directly using JSON

### Why did it change

Update the Lambda to support both API Gateway and JSON requests.

### Issue tracking

- [PYIC-2732](https://govukverify.atlassian.net/browse/PYIC-2732)

### Other considerations

[PYIC-2732]: https://govukverify.atlassian.net/browse/PYIC-2732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -127,14 +127,15 @@ public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
             String ipAddress = getIpAddress(input);
             String featureSet = getFeatureSet(input);
             credentialIssuerConfigService.setFeatureSet(featureSet);
-            String criId = getJourney(input);
+            String journey = getJourney(input);
 
-            var errorResponse = validate(criId);
+            var errorResponse = validate(journey);
             if (errorResponse.isPresent()) {
                 return new JourneyErrorResponse(
                         JOURNEY_ERROR_PATH, HttpStatus.SC_BAD_REQUEST, errorResponse.get());
             }
 
+            String criId = journey;
             CredentialIssuerConfig credentialIssuerConfig = getCredentialIssuerConfig(criId);
 
             if (credentialIssuerConfig == null) {
@@ -263,11 +264,11 @@ public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
     }
 
     @Tracing
-    private Optional<ErrorResponse> validate(String criId) {
-        if (criId == null || StringUtils.isBlank(criId)) {
+    private Optional<ErrorResponse> validate(String journey) {
+        if (journey == null || StringUtils.isBlank(journey)) {
             return Optional.of(ErrorResponse.MISSING_CREDENTIAL_ISSUER_ID);
         }
-        LogHelper.attachCriIdToLogs(criId);
+        LogHelper.attachCriIdToLogs(journey);
         return Optional.empty();
     }
 
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -51,6 +51,8 @@ import uk.gov.di.ipv.core.library.statemachine.BaseJourneyLambda;
 import java.net.URISyntaxException;
 import java.text.ParseException;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static uk.gov.di.ipv.core.library.domain.CriConstants.ADDRESS_CRI;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CLAIM;
@@ -64,13 +66,13 @@ import static uk.gov.di.ipv.core.library.helpers.RequestHelper.getJourney;
 
 public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
     private static final Logger LOGGER = LogManager.getLogger();
-    private static final String CRI_ID = "criId";
     private static final String DCMAW_CRI_ID = "dcmaw";
     public static final String SHARED_CLAIM_ATTR_NAME = "name";
     public static final String SHARED_CLAIM_ATTR_BIRTH_DATE = "birthDate";
     public static final String SHARED_CLAIM_ATTR_ADDRESS = "address";
     public static final String DEFAULT_ALLOWED_SHARED_ATTR = "name,birthDate,address";
     public static final String REGEX_COMMA_SEPARATION = "\\s*,\\s*";
+    public static final Pattern LAST_SEGMENT_PATTERN = Pattern.compile("/([^/]+)$");
 
     private final ObjectMapper mapper = new ObjectMapper();
     private final CredentialIssuerConfigService credentialIssuerConfigService;
@@ -135,7 +137,7 @@ public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
                         JOURNEY_ERROR_PATH, HttpStatus.SC_BAD_REQUEST, errorResponse.get());
             }
 
-            String criId = journey;
+            String criId = getCriIdFromJourney(journey);
             CredentialIssuerConfig credentialIssuerConfig = getCredentialIssuerConfig(criId);
 
             if (credentialIssuerConfig == null) {
@@ -221,6 +223,14 @@ public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
                     ErrorResponse.FAILED_TO_CONSTRUCT_REDIRECT_URI,
                     e.getMessage());
         }
+    }
+
+    private String getCriIdFromJourney(String journey) {
+        Matcher matcher = LAST_SEGMENT_PATTERN.matcher(journey);
+        if (matcher.find()) {
+            return matcher.group(1);
+        }
+        return journey;
     }
 
     private CriResponse getCriResponse(
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -265,7 +265,7 @@ public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
 
     @Tracing
     private Optional<ErrorResponse> validate(String journey) {
-        if (journey == null || StringUtils.isBlank(journey)) {
+        if (StringUtils.isBlank(journey)) {
             return Optional.of(ErrorResponse.MISSING_CREDENTIAL_ISSUER_ID);
         }
         LogHelper.attachCriIdToLogs(journey);
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -1,9 +1,6 @@
 package uk.gov.di.ipv.core.buildcrioauthrequest;
 
 import com.amazonaws.services.lambda.runtime.Context;
-import com.amazonaws.services.lambda.runtime.RequestHandler;
-import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
-import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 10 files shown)


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CriResponse.java
```diff
@@ -2,14 +2,20 @@
...
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 
 @Getter
-public class CriResponse {
+@EqualsAndHashCode(callSuper = true)
+public class CriResponse extends JourneyResponse {
```
chris-dco: I think if you don't extend this from JourneyResponse and update BaseJourneyLambda giving it the return type would be better.

lib/src/main/java/uk/gov/di/ipv/core/library/statemachine/BaseJourneyLambda.java
```diff
@@ -20,8 +20,9 @@
 public abstract class BaseJourneyLambda
```
chris-dco: So that CriResponse.java doesn't need to extend from JourneyResponse instead make this `BaseJourneyLambda<T>` and update `handleRequest` to:
```java
protected abstract T handleRequest(JourneyRequest request, Context context);
```

This will enable you to update BuildCriOauthRequestHandler to:
```java
public class BuildCriOauthRequestHandler extends BaseJourneyLambda<CriResponse>
```

lib/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
```diff
@@ -1,13 +1,18 @@
...
+import java.util.Map;
+
 @AllArgsConstructor
+@NoArgsConstructor
 @Data
+@Builder
 public class JourneyRequest {
     private String ipvSessionId;
     private String ipAddress;
+    private Map<String, String> pathParameters;
```
chris-dco: The only path parameter which would be required across all of the lambdas is `journey` which would contain something like `/journey/cri/build-oauth-request/dcmaw` and would be passed in via the Step Function or API Gateway. So his could be just:
```java
private String journey;
private String ipvSessionId;
private String ipAddress;
...
```

lib/src/test/java/uk/gov/di/ipv/core/library/helpers/RequestHelperTest.java
```diff
@@ -210,4 +212,24 @@ void getFeatureSetShouldReturnDefault() {
 
         assertEquals("default", RequestHelper.getFeatureSet(event));
     }
+
+    @Test
+    void getPathParametersShouldReturnPathParameters() {
+        var event = new APIGatewayProxyRequestEvent();
+        event.setPathParameters(Map.of("criId", DCMAW_CRI));
```
chris-dco: See comment in JourneyRequest in which case you would test by setting the `journey` field to `/journey/cri/build-oauth-request/dcmaw`

chris-dco (changes requested):

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CriResponse.java
TalatAydinGDS: I agree that a generic approach would be better. I will pick up where remove the inheritance of 'BaseJourneyLambda' from the subtask 'Update the Lambda to support both API Gateway and JSON request'. (PYIC-2750) In some other Lambdas, we are returning object then 'JourneyResponse,' and they are also extended from 'JourneyResponse'. (For example: build-client-auth-response) If we are not going to follow that approach, I will consider making it generic while extending from 'RequestHandler' in all Lambdas, keeping this in mind. However, It would be more appropriate to fix this in other Lambdas as well and follow same approach.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
TalatAydinGDS: Chris, I am bit confused about your this comment.As you can seed this API gateway definition, /journey/cri/build-oauth-request/{criId}, the lonk comes as a path variable. I am getting this just like before from the path variable. While implementing your request in other subtask to update the lambda and remove support for the API Gateway request, are you expecting me to take this{criId} path variable and set it to the journey field? For that: "aws"
        requestTemplates:
          application/x-www-form-urlencoded:
            Fn::Sub: |
              {
               "journey": "$input.path('$.criId)", 
                "ipvSessionId": "$input.params('ipv-session-id')", 
                "ipAddress": "$input.params('ip-address')"
              }

  If you confirm, I will implement this change within in other subtask. If this is a requirement for this task. I can add it here as well. You know, other option, the front end could send us the journey instead of criId. We can discuss this at your convenience.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CriResponse.java
chris-dco: Yes this'll need to be fixed in all the lambdas.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
chris-dco: Hi Talat. The Process Journey Step Function is called with the following information, which we'd also use to call this Lambda with:
```json
{
  "ipvSessionId": "xbMDE2w3BNYQThqMEO4IXGQPZh_PEqKj-r7DkiNnf3Q",
  "ipAddress": "51.149.8.27",
  "journey": "/journey/error"
}
```

If you can update the Lambda to take the path from the `journey` field and for API Gateway to supply the path to the `journey` field then calling the Lambda from the Process Journey Step Function will be a piece of cake.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CriResponse.java
TalatAydinGDS: implemented the interface as we have decided.

lib/src/main/java/uk/gov/di/ipv/core/library/statemachine/BaseJourneyLambda.java
TalatAydinGDS: response to the BasResponse interface as we have decided.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
TalatAydinGDS: Done

lib/src/test/java/uk/gov/di/ipv/core/library/helpers/RequestHelperTest.java
TalatAydinGDS: This test was added for the 'criId' path variable

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -252,11 +264,11 @@ private JWEObject signEncryptJar(
     }
 
     @Tracing
-    private Optional<ErrorResponse> validate(Map<String, String> pathParameters) {
-        if (pathParameters == null || StringUtils.isBlank(pathParameters.get(CRI_ID))) {
+    private Optional<ErrorResponse> validate(String journey) {
+        if (journey == null || StringUtils.isBlank(journey)) {
```
rpayal: .isBlank will take care of null value also.

lib/src/main/java/uk/gov/di/ipv/core/library/statemachine/BaseJourneyLambda.java
```diff
@@ -52,12 +54,15 @@ public Map<String, Object> handleRequest(Map<String, Object> event, Context cont
...
             var featureSet = getFeatureSet(request);
             var journeyRequest =
-                    new JourneyRequest(ipvSessionId, ipAddress, clientOAuthSessionId, featureSet);
-
+                    JourneyRequest.builder()
+                            .ipvSessionId(getIpvSessionId(request))
+                            .ipAddress(getIpAddress(request))
+                            .journey(getJourney(request))
+                            .clientOAuthSessionId(clientOAuthSessionId)
+                            .featureSet(featureSet)
```
rpayal: can remove variable clientOAuthSessionId & featureSet.......and directly call the req. helper 
method.

openAPI/core-back-internal.yaml
```diff
@@ -327,10 +327,8 @@ paths:
           application/x-www-form-urlencoded:
             Fn::Sub: |
               {
-                "ipvSessionId": "$input.params('ipv-session-id')", 
-                "ipAddress": "$input.params('ip-address')",
-                "clientOAuthSessionId": "$input.params('client-session-id')",
-                "featureSet": "$input.params('feature-set')"
+                "ipvSessionId": "$input.params('ipv-session-id')",
+                "ipAddress": "$input.params('ip-address')"
               }
```
rpayal: removed req param's "clientOAuthSessionId" & "featureSet" ?

rpayal (dismissed):

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -135,6 +138,25 @@ public static String getFeatureSet(APIGatewayProxyRequestEvent event) {
         return getFeatureSet(event.getHeaders());
     }
 
+    public static String getJourney(JourneyRequest request) {
+        return request.getJourney();
+    }
+
+    public static Optional<String> getJourney(APIGatewayProxyRequestEvent event) {
+        Optional<String> criId = getPathVariableFromEvent(event, CRI_ID);
```
chris-dco: `CRI_ID` is very specific to this lambda, can you make this more generic as not all Lambdas which use this method will be looking for CRI_ID.

lib/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -135,6 +138,25 @@ public static String getFeatureSet(APIGatewayProxyRequestEvent event) {
         return getFeatureSet(event.getHeaders());
     }
 
+    public static String getJourney(JourneyRequest request) {
+        return request.getJourney();
+    }
+
+    public static Optional<String> getJourney(APIGatewayProxyRequestEvent event) {
```
chris-dco: Although I can see why the Optional has been used it's the only place in this file where it's done this way. All others return `null`. So that we can be consistent with the reset of the file can you return `null` instead? Same with `getPathVariableFromEvent`.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -59,14 +57,15 @@
...
+import static uk.gov.di.ipv.core.library.helpers.RequestHelper.getIpvSessionId;
+import static uk.gov.di.ipv.core.library.helpers.RequestHelper.getJourney;
 
-public class BuildCriOauthRequestHandler
-        implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+public class BuildCriOauthRequestHandler extends BaseJourneyLambda {
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String CRI_ID = "criId";
-    private static final int OK = 200;
     private static final String DCMAW_CRI_ID = "dcmaw";
```
chris-dco, line 69: This is no longer used.
```java
private static final String DCMAW_CRI_ID = "dcmaw";
```

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -66,11 +67,15 @@ class CheckExistingIdentityHandlerTest {
 
     private static final String TEST_FEATURE_SET = "test-feature-set";
     private static final String TEST_CLIENT_OAUTH_SESSION_ID = SecureTokenHelper.generate();
+
+    private static final String TEST_JOURNEY = DCMAW_CRI;
```
chris-dco, line 71: As it's a requirement that the Step Function passes the full URL e.g. `/journey/cri/build-oauth-request/dcmaw` in the `journey` parameter on the Lambda when running in either API Gateway or Step Function the TEST_JOURNEY should be changed to `/journey/cri/build-oauth-request/dcmaw` in order to reflect that and the code should be able to handle this.

chris-dco (changes requested):

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
TalatAydinGDS: We are obtaining the 'criId' from path variable  /journey/cri/build-oauth-request/{criId}, within the API Gateway and setting it in the journey field. We retrieve the configurations from the system manager based on thi criId. I'am not sure how to obtain the full path. If we do get the full path, how do we extract criId from it? I truly don't understand this detail in the task description and task requirements. Please can discuss about that.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
TalatAydinGDS: This is using in the line 234.

blakeyp (approved):

chris-dco (approved):

PR #827, opened 2023-05-18
Pyic 2760 add the check existing identity lambda to the journey engine step function

chris-dco:
### What changed

* Added the basic structure of the different paths which can be taken. New Lambdas can now be added to the ProcessJourneyStepResult
* Added the CheckExistingIdentityLambda Lambda to the Journey Engine Step Function
  * Also passes through clientOAuthSessionId and featureSet from request.
* Added null checking to RequestHelper for featureSet and clientOAuthSessionId as changes were causing null pointer exceptions.
* Updated `/journey/{journeyStep}` so that the clientOAuthSessionId and featureSet were passed in.
* Added `lambdas/*/bin` to .gitignore so that it's easier to see for uncommited changed when developing.

### Why did it change

Added CheckExistingIdentityLambda to Process Journey State Machine

### Issue tracking

- [PYIC-2760](https://govukverify.atlassian.net/browse/PYIC-2760)

### Other considerations

None

[PYIC-2760]: https://govukverify.atlassian.net/browse/PYIC-2760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
@@ -1,11 +1,8 @@
 package uk.gov.di.ipv.core.library.helpers;
 
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
 import org.apache.http.HttpStatus;
-import org.apache.http.NameValuePair;
-import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.StringMapMessage;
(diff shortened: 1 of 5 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -124,7 +108,7 @@ public static String getClientOAuthSessionId(JourneyRequest event) {
                                 LOG_MESSAGE_DESCRIPTION.getFieldName(),
                                 "Client session id missing in header.");
         validateClientOAuthSessionId(clientSessionId, message);
-        return clientSessionId;
+        return StringUtils.isBlank(clientSessionId) ? null : clientSessionId;
```
rpayal, line 111: Is in request i.e JourneyRequest clientSessionId coming as empty/blank instead of null ?
If yes then also update null check in validateClientOAuthSessionId() to isBlank check.

openAPI/core-back-internal.yaml
```diff
@@ -311,7 +311,9 @@ paths:
             Fn::Sub: |
               {
                 "ipvSessionId": "$input.params('ipv-session-id')", 
-                "ipAddress": "$input.params('ip-address')"
+                "ipAddress": "$input.params('ip-address')",
+                "clientOAuthSessionId": "$input.params('client-session-id')",
```
rpayal, line 315: clientOAuthSessionId is not needed for check-existing-identity ....so do we need to still add it here ?
As its only needed for lambda build-client-oauth-response.

blakeyp (approved):

PR #830, opened 2023-05-18
Reconfigure API Gateway so that it calls the Lambda using JSON

TalatAydinGDS:
### What changed

Reconfigure API Gateway so it calls the build-cri-oauth-request lambda using JSON

### Why did it change

We'll be calling these journey lambdas within the journey engine step function so will be removing API Gateway

### Issue tracking

- [PYIC-2733](https://govukverify.atlassian.net/browse/PYIC-2733)

[PYIC-2733]: https://govukverify.atlassian.net/browse/PYIC-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openAPI/core-back-internal.yaml
+++ b/openAPI/core-back-internal.yaml
@@ -158,6 +158,12 @@ paths:
   /journey/cri/build-oauth-request/{criId}:
     post:
       description: Called when the frontend begins the CRI journey.
+      parameters:
+        - name: criId
+          in: path
+          required: true
+          schema:
+            type: string
       responses:
         200:
           description: "Returns the id, clientId and authorizationUrl for a CRI."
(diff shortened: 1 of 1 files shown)


openAPI/core-back-internal.yaml
```diff
@@ -170,7 +176,27 @@ paths:
         uri:
           Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${BuildCriOauthRequestFunction.Arn}:live/invocations
         passthroughBehavior: "when_no_match"
-        type: "aws_proxy"
+        type: "aws"
+        requestTemplates:
+          application/x-www-form-urlencoded:
+            Fn::Sub: |
+              {
+                "journey": "$input.params('criId')",
```
chris-dco: As the Step Function will send the full URL e.g. `/journey/cri/build-oauth-request/dcmaw` to the Lambda the API Gateway should do the same. In this case I believe the code you need is.

```json
"journey": "/journey/cri/build-oauth-request/$input.params('criId')"
```

chris-dco (changes requested):

openAPI/core-back-internal.yaml
TalatAydinGDS: updated and tested on the dev

blakeyp (approved):

chris-dco (approved):

PR #886, opened 2023-06-06
PYIC-2915: Correct F2F handoff state naming (dev and build)

therealslimsladey:
### What changed
Correct F2F handoff state naming (dev and build)

### Why did it change
Name change suggested by Tech Lead

### Issue tracking

- [PYIC-2915](https://govukverify.atlassian.net/browse/PYIC-2915)

- No environment variables or secrets were added or changed

### Other considerations

[PYIC-2915]: https://govukverify.atlassian.net/browse/PYIC-2915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
@@ -318,13 +318,12 @@ CRI_F2F:
     pending:
       type: basic
       name: pending
-      targetState: CRI_F2F_HANDOFF
+      targetState: F2F_HANDOFF_PAGE
       response:
         type: page
         pageId: page-face-to-face-handoff
-CRI_F2F_HANDOFF:
-  name: CRI_F2F_HANDOFF
-  parent: CRI_F2F
+F2F_HANDOFF_PAGE:
+  name: F2F_HANDOFF_PAGE
 CRI_KBV:
   name: CRI_KBV
   parent: CRI_STATE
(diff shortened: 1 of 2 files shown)


chris-dco (dismissed):

lambdas/process-journey-step/src/main/resources/statemachine/build/ipv-core-main-journey.yaml
```diff
@@ -287,6 +287,20 @@ CRI_ADDRESS:
...
+    pending:
+      type: basic
+      name: pending
+      targetState: CRI_F2F_HANDOFF
+      response:
+        type: page
+        pageId: page-face-to-face-handoff
+CRI_F2F_HANDOFF:
+  name: CRI_F2F_HANDOFF
+  parent: CRI_F2F
```
kerrr: The convention is that states that start CRI_ are for when the user is at a CRI. I think this state is supposed to be when the user is at the "handoff page", so should probably be something like F2F_HANDOFF_PAGE.
Also I don't know why it would have CRI_F2F as a parent. It probably doesn't need to have any events.

therealslimsladey: Amended, and ditto for dev journey engine config

AmritSidhu:
You may want to check this isnt covered by Chris's work: https://github.com/alphagov/di-ipv-core-back/pull/885

chris-dco (approved):

PR #900, opened 2023-06-08
BAU: Add f2f check validator

Wynndow:
We have logic to check if VC's are deemed "successful". This is used when deciding where to send a user, but also for deciding which VCs we should extract data from when constructing an identity.

We need this for the new f2f VC as it will be the only VC that contains the name and date of birth data.

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45F2fValidator.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45F2fValidator.java
@@ -1,0 +1,17 @@
+package uk.gov.di.ipv.core.library.domain.gpg45.validation;
+
+import uk.gov.di.ipv.core.library.domain.gpg45.domain.CredentialEvidenceItem;
+
+public class Gpg45F2fValidator {
+
+    private Gpg45F2fValidator() {
+        throw new IllegalStateException("Utility class");
+    }
+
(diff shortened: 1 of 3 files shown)


shivanshuit914 (dismissed):

shivanshuit914 (approved):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/validation/Gpg45F2fValidator.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.di.ipv.core.library.domain.gpg45.validation;
+
+import uk.gov.di.ipv.core.library.domain.gpg45.domain.CredentialEvidenceItem;
+
+public class Gpg45F2fValidator {
+
+    private Gpg45F2fValidator() {
+        throw new IllegalStateException("Utility class");
```
chris-dco, line 8: An empty constructor would be fine and would matche up with other static classes.

Wynndow: This follows the same pattern as the other validators we have for the other "evidence types". I seem to recall we initially did it like this to appease sonarcloud. If you have a class with only one (or maybe more) static methods it shouts at you that it's a utility class and you should take steps to prevent anyone actually instantiating an instance.

PR #912, opened 2023-06-15
BAU: Explicitly use X-Ray env vars

Wynndow:
### What changed

Explicitly use X-Ray env vars

### Why did it change

The powertools x-ray lib is configurable with env vars to control what information in sends to AWS in its data. You can prevent it from sending method responses and errors. The docs just mention setting the env vars which we've done, however we're seeing lots of stack traces X-Ray. This may be causing an intermittent error we're seeing where subsegments can't be sent on some lambda in build and staging: `SEVERE: Exception while sending segment (SubsegmentImpl) over UDP for entity ## handleRequest"`. This is because the message is too large.

Looking at the powertools lib code, it looks like you might actually need to tell the tracing annotations to look for the env vars for config. This is rolling that out on one lambda to see if it helps. If it does we can roll this out further.


Wynndow:
Closing as I noticed the default capture mode is env var anyway

PR #932, opened 2023-06-22
BAU: Apply singleton pattern to ConfigService

Wynndow:
### What changed

Apply singleton pattern to ConfigService

### Why did it change

It might be useful for certain classes to have direct access to a ConfigService without it having to be passed down through a chain of method calls, and without having to instantiate a whole new instance of it for themselves.

Applying the singleton pattern like this would allow some class in the depths of the system to use an already existing config service. This would hopefully help with performance of the lambda, and also benefit from the parameter caching we're already doing.

There can be issues with this pattern when in a multithreaded environment, but I don't think this is an issue when running in the context of a lambda.

### DIFF
--- a/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
+++ b/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
@@ -55,7 +55,7 @@ public class DataStoreIpvSessionIT {
                     "The environment variable 'IPV_SESSIONS_TABLE_NAME' must be provided to run this test");
         }
 
-        ConfigService configService = new ConfigService();
+        ConfigService configService = ConfigService.getInstance();
 
         ipvSessionItemDataStore =
                 new DataStore<>(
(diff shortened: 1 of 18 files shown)


kerrr:
Lets have a chat about this one. Gut feeling is that I'm not sure this is a good idea, nut I'm open to being convinced.

therealslimsladey:
My two-pennorth:
- singleton makes sense to me
- not sure it'll have any significant impact as the underlying cache is a static anyway (if it weren't we'd see no evidence of caching and subjectively, I've seen the caching in play when I've amended parameters)
- my money's on aggregating the parameters into a single SSM parameter blob and then accessing that to extract the params

Wynndow:
@therealslimsladey The caching thing I mentioned is more of an added side benefit - my main motivator for this was having the ability in future to new up a config service without feeling guilty. We might be doing some work that needs some small helper classes to access the config service, but I don't want to have to plum in passing it all the way down from the handler. And creating a whole new one (with the associated SSM clients) in these tiny classes just felt a bit much. By having the singleton, we could just access the existing resource. Which would come with the cache it's already got. But that's just a nice side effect and as you say, probably not going to make any real difference.

PR #933, opened 2023-06-22
PYIC-2991

TalatAydinGDS:
to exclude the email property from ClaimSet when it is null and rename the property to 'emailAddress'

### What changed

to exclude the email property from ClaimSet when it is null and rename the property to 'emailAddress'

### Why did it change

Address CRI doesn't expectation email and null value.

### Issue tracking

- [PYIC-2991](https://govukverify.atlassian.net/browse/PYIC-2991)

[PYIC-2991]: https://govukverify.atlassian.net/browse/PYIC-2991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TalatAydinGDS:
We had to create a new DTO because of the serialisation issue. Despite ignoring the emailAddress field in Jackson, it was still added as emailAddress=null.
Also, tried to the jackson serialisation. But format changed. As a solution, we create a new DTO for this purpose.

shivanshuit914 (approved):

PR #952, opened 2023-06-28
PYIC-3009: Set journeyType in session based on config value

lwebb-iig:
The users session is configured to run an existing or a new journey depending on if a featureSet is active.

### What changed

- IpvSessionService updated to consider parameter value `"self/journey/type"` to dictate the set journey type

### Why did it change

A users journey type is set in their session by the initialise-ipv-session lambda. We want that lambda to check the new parameter for the journey type, which may or may not have been overridden by a featureSet, and set the appropriate value.

### Issue tracking

- [PYIC-3009](https://govukverify.atlassian.net/browse/PYIC-3009)

### Other considerations

[PYIC-3009]: https://govukverify.atlassian.net/browse/PYIC-3009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/src/main/java/uk/gov/di/ipv/core/library/domain/IpvJourneyTypes.java
```diff
@@ -13,4 +13,8 @@ public enum IpvJourneyTypes {
     public String getValue() {
         return value;
     }
+
+    public static IpvJourneyTypes get(String journeyType) {
+        return IpvJourneyTypes.valueOf(journeyType.toUpperCase().replace("-", "_"));
+    }
```
Wynndow: I'm wondering if rather than adding this method, we just set the value of the parameter to the expected value (`IPV_CORE_MAIN_JOURNEY` for example), then we can just call `IpvJourneyTypes.valueOf(journeyType)`

Wynndow (commented):
Just the comment about the get method, otherwise all good.

lib/src/test/java/uk/gov/di/ipv/core/library/service/IpvSessionServiceTest.java
```diff
@@ -152,8 +158,33 @@ void shouldCreateSessionItemForDebugJourney() {
...
+                .create(ipvSessionItemArgumentCaptor.capture(), eq(BACKEND_SESSION_TTL));
+        assertNotNull(ipvSessionItemArgumentCaptor.getValue().getIpvSessionId());
+        assertNotNull(ipvSessionItemArgumentCaptor.getValue().getCreationDateTime());
+
+        assertEquals(
+                ipvSessionItemArgumentCaptor.getValue().getIpvSessionId(),
+                ipvSessionItem.getIpvSessionId());
+        assertEquals(
+                INITIAL_IPV_JOURNEY_STATE, ipvSessionItemArgumentCaptor.getValue().getUserState());
+        assertEquals(ipvSessionItem.getJourneyType(), IpvJourneyTypes.IPV_CORE_REFACTOR_JOURNEY);
```
Wynndow: sonar wants you to swap these two args, so they match with `expected` and `actual`.

Wynndow (approved):

PR #955, opened 2023-06-28
PYIC-3046 Handle F2F restart event to clear VCs and return user to start

blakeyp:
### What changed

Add 'restart' event for the F2F pending page which points to the new reset-identity lambda, which clears the user's existing VCs and F2F CRI response item and returns next which takes them to the start of the journey (app eligibility/doc select page for now)

### Why did it change

When a returning user hits the F2F pending page they have the option to 'restart' and prove their identity another way. This handles the backend logic for that scenario

### Issue tracking
- [PYIC-3046](https://govukverify.atlassian.net/browse/PYIC-3046)

[PYIC-3046]: https://govukverify.atlassian.net/browse/PYIC-3046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
+++ b/lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
@@ -53,7 +53,6 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.DCMAW_CRI;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.F2F_CRI;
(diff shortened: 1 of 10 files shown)


shivanshuit914 (dismissed):

kerrr:
🤔  I'm wondering if we should have separate lambdas here? A reset identity lambda that just does that and can be wired in to various places. Or rethink how the check existing identity lambda is factored so that it makes more sense to call it in different parameters to do different collections of things. We need to have a relook at check existing identity anyway since it's doing a lot of the same stuff that `evaluate-gpg45-scores` is already doing. 🤔

blakeyp:
@kerrr I'm thinking about something like evaluate-gpg45-scores just spitting out a result (matches profile or not) and have that feed into check-existing-identity so the scoring/profile check doesn't have to be duplicated. But I think that would be a lot of work to untie the journey response/CI stuff wrapped up in evaluate-gpg45-scores. But yeah it would make more sense to have a dedicated reset identity lambda for this anyway so we can do that for now.

shivanshuit914 (approved):

PR #961, opened 2023-07-03
PYIC-3013: Test statemachine classes

Wynndow:
### What changed

Test statemachine classes

### Why did it change

We've previously taken an approach of not testing directly the classes we use to build and operate the statemachine, but instead testing them implicitly by testing all the steps in the statemachine file. You can see these tests in ProcessJourneyStepHandlerTest. This file is currently 795 lines long.

Maintaining this testing strategy is not sustainable. There are too many tests, and they're prone to not being updated. Rather than trying to test the statemachine file here, we should instead unit test the "building blocks" of the statemachine system, and then allow the journeys themselves to be tested with the functional test suite. This will result in more maintainable tests that are easier to reason about and hopefully more useful.

Once we've moved over to using the refactored journey statemachine file, I propose we get rid of the majority of tests in
ProcessJourneyStepHandlerTest.

### Issue tracking

- [PYIC-3013](https://govukverify.atlassian.net/browse/PYIC-3013)

[PYIC-3013]: https://govukverify.atlassian.net/browse/PYIC-3013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/statemachine/State.java
+++ b/lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/statemachine/State.java
@@ -3,7 +3,6 @@ package uk.gov.di.ipv.core.processjourneystep.statemachine;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
-import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.processjourneystep.statemachine.events.Event;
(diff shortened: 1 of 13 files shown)


lwebb-iig (approved):
LGTM - at a glance, I think I prefer the idea of splitting out [ProcessJourneyStepHandler](https://github.com/alphagov/di-ipv-core-back/blob/main/lambdas/process-journey-step/src/test/java/uk/gov/di/ipv/core/processjourneystep/ProcessJourneyStepHandlerTest.java ) into a more manageable per-journey/use case files perhaps rather than throwing away entirely? Although I agree the current maintainability of these tests is quite daunting

PR #967, opened 2023-07-03
PYIC-3032: Add CI scoring to ContraIndications domain object

TalatAydinGDS:
### What changed

Add CI scoring to ContraIndications domain object

### Why did it change

Add a new getContraIndicatorScore method which will

Fetch/cache? the contra indicators score map

Return the sum over the CI detected scores

Add getLatestContraIndicator method to return most recent ContraIndicator.

### Issue tracking

- [PYIC-3032](https://govukverify.atlassian.net/browse/PYIC-3032)

[PYIC-3032]: https://govukverify.atlassian.net/browse/PYIC-3032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
@@ -12,6 +12,15 @@ import java.util.Optional;
 public class ContraIndications {
     private final Map<String, ContraIndicator> contraIndicators;
 
+    public Integer getContraIndicatorScores(
+            final Map<String, ContraIndicatorScore> contraIndicatorScores) {
+        return contraIndicators.keySet().stream()
+                .map(
+                        contraIndicatorCode ->
+                                contraIndicatorScores.get(contraIndicatorCode).getDetectedScore())
+                .reduce(0, Integer::sum);
+    }
+
     public Optional<ContraIndicator> getLatestContraIndicator() {
         return contraIndicators.values().stream()
                 .max(Comparator.comparing(ContraIndicator::getIssuanceDate));
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
@@ -2,13 +2,26 @@ package uk.gov.di.ipv.core.library.domain;
 
 import lombok.Builder;
 import lombok.Getter;
-import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
-import java.util.Map;
+import java.util.*;
 
 @Getter
-@Builder
-@ExcludeFromGeneratedCoverageReport
+@Builder(toBuilder = true)
 public class ContraIndications {
     private final Map<String, ContraIndicator> contraIndicators;
+
+    private final Map<String, ContraIndicatorScore> contraIndicatorScores;
+
+    public Integer getContraIndicatorScores() {
+        return contraIndicators.keySet().stream()
+                .map(
+                        contraIndicatorCode ->
+                                contraIndicatorScores.get(contraIndicatorCode).getDetectedScore())
+                .reduce(0, Integer::sum);
+    }
+
+    public Optional<ContraIndicator> getLatestContraIndicator() {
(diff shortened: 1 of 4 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
```diff
@@ -2,13 +2,26 @@
 
 import lombok.Builder;
 import lombok.Getter;
-import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
-import java.util.Map;
+import java.util.*;
```
therealslimsladey: Please check IDE settings - looks like it's set to wildcard import if more than n explicit imports.
GDS standard is to avoid wildcard imports.

rpayal (dismissed):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
```diff
@@ -2,13 +2,26 @@
...
 
 @Getter
-@Builder
-@ExcludeFromGeneratedCoverageReport
+@Builder(toBuilder = true)
 public class ContraIndications {
     private final Map<String, ContraIndicator> contraIndicators;
+
+    private final Map<String, ContraIndicatorScore> contraIndicatorScores;
```
therealslimsladey: I reconsidered this. I don't think the scores should be part of the domain object but rather something passed into the scoring method.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicator.java
```diff
@@ -8,13 +8,18 @@
 import java.util.List;
 
 @Getter
-@Builder
+@Builder(toBuilder = true)
 @ExcludeFromGeneratedCoverageReport
```
therealslimsladey: We should probably remove the exclusion if there are tests for this class?

therealslimsladey (changes requested):

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
TalatAydinGDS: Yes, it was overlooked. Thank you

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
TalatAydinGDS: removed property, method and unit tests

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicator.java
TalatAydinGDS: Added it with a different issuance date but same other fields for testing purposes.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndications.java
therealslimsladey: Hi. I meant keep the scoring method but pass the score data as an argument.

lib/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicator.java
```diff
@@ -1,20 +1,25 @@
...
 
 import java.time.Instant;
 import java.util.List;
 
 @Getter
-@Builder
-@ExcludeFromGeneratedCoverageReport
-public class ContraIndicator {
+@Builder(toBuilder = true)
+@EqualsAndHashCode
```
therealslimsladey: We don't need the equals, hashcode, and compareTo - nor does ContraIndicator need to implement Comparable since `ContraIndications.getLatestContraIndicator` supplies a Comparator on issuanceDate directly, hence ContraIndicator doesn't need the automatic ordering that we were giving it. Simplifies/eliminates the ContraIndicator tests as well.

therealslimsladey (changes requested):

therealslimsladey (approved):

PR #976, opened 2023-07-04
BAU: Simplify statemachine definition

Wynndow:
### What changed

Simplify statemachine definition

### Why did it change
This commit is an attempt to make the statemachine definition less verbose and easier to understand. It does a few things.

This commit is an attempt to make the statemachine definition less
verbose and easier to understand. It does a few things.

* It removes the "name" property from states and events. These are now
  interpreted from the key used to define the body of them in the yaml.
  This used to set the name value on the objects as we still need it.
* It moves the "response" definition (effectively how to produce the
  desired state), from the event to the state itself. This makes things
  easier to reason about and prevents a bunch of duplication.
* It removes some duplicated states in favour of a shared one.
* Removes the StateMachineResult class as it's no longer needed. We can
  just use the state itself and the response it contains.
* Removes the criEvent type. It's not really needed now as States can
  have a response of type criResponse. This does push some logic into
  the basic event, but it means the definition is simpler.

This has been deployed and it works, but only with the new statemachine
files. If this change was to be applied, it would need to go in after
we've switched.

This has been deployed and it works, but only with the new statemachine files. If this change was to be applied, it would need to go in after we've switched.


lwebb-iig (approved):
I generally think a lot of the changes here are well worth merging as it significantly reduces redundant & duplicated information in the statemachine files. If we consider the original goal of the refactor epic and the long term goals of having the statemachine files more readable and representative of the UX/User journey, then these changes certainly take a large step towards that goal.

lwebb-iig (changes requested):
Needs a spotless check

Wynndow:
Closing in favour of splitting this out into separate tickets to be done in a sprint

PR #990, opened 2023-07-06
BAU: Quote log level for step functions

Wynndow:
### What changed

Quote log level for step functions

### Why did it change

This reverts an early commit to remove the logging config completely when set to OFF. We've discovered that if you don't quote the OFF value it gets transformed to false, which breaks stuff.

This adds the quotes, and also adds the joureyEngine step function logging level to the env map to be consistent.


TobySaundersGDS (approved):
Nice1

PR #1015, opened 2023-07-12
BAU: Fix compile warnings

Wynndow:
### What changed

Fix compile warnings

### Why did it change

When we compile the lambdas with gradle (`./gradlew build`), we see warnings generated for a couple of the sub-projects.

```
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

This makes a few small modifications to remove those warnings.


lwebb-iig (approved):

lwebb-iig:
Approved - sonarqube is saying 50% coverage on new code which you might want to look at but not a huge deal @Wynndow

Wynndow:
Yeah. I think it's because of a catch block that I needed to add. It's actually not possible to reach it as the input that might throw the error (JsonProcessingException) is hardcoded rather than coming from an unvetted source. So I'm not too worried.

PR #1018, opened 2023-07-12
BAU Add Journey Engine docs

kerrr:
Add some outline docs for journey engine concepts and ways of specifying states.

### DIFF
--- a/docs/journey-engine-concepts.md
+++ b/docs/journey-engine-concepts.md
@@ -1,0 +1,138 @@
+# Journey Engine
+
+## Terms
+
+### Journey Engine
+
+The Journey Engine is what tells IPV Core Front what to show the user next. IPV Core Front calls the journey engine with `/journey/<journey-event>` requests to the internal API Gateway and the Journey Engine returns a Journey Response, that tells IPV Core front to show a page in IPV Core, redirect to a CRI, or a redirect to the originating OAuth client. 
+
+(The Journey Engine can also tell IPV Core Front to fire a new Journey Event back to the Journey Engine, though this is being phased out.)
+
+### Journey Engine Step Function
+
+This is the AWS Step Function that orchestrates the running of the lambdas that make up the Journey Engine.
+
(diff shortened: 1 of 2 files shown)


lwebb-iig (approved):

docs/journey-engine-concepts.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Journey State
+
+A Journey Sate represents a point in the user's journey. The definition includes the events that can be accepted when the user is in this state and what happens when that event is received.
+
+### Journey Event
+
+A Journey Event is a named event that has occurred in the users journey. IPV Core Front is responsible for interpreting the action a user has taken and turning it into an event that can be sent to the Journey Engine. The event is represented as a string in requests to the internal API Gateway `/journey/<journey-event>`
+
+### Journey Response
```
Wynndow, line 35: There's something else here too that overlaps quite a lot and I think can be the source of some confusion.

When we transition states in the state machine we define a "response" which is what the process-journey-step lambda returns. They are `JourneyStepResponse`s and you can see the types (page, journey, error, cri), defined  [here](https://github.com/alphagov/di-ipv-core-back/tree/main/lambdas/process-journey-step/src/main/java/uk/gov/di/ipv/core/processjourneystep/statemachine/responses). The journey responses from here are initially used by the step function to do some routing. OR (I think), returned to the frontend. The CRI response defined here is actually a journey response that is constructed slightly differently and then interpreted by the step function to run the build-oauth-request lambda. Confusing. The page and error responses get sent to the frontend.

We also define a different set of responses that have similar types and are derived from a [`BaseResponse`](https://github.com/alphagov/di-ipv-core-back/blob/main/lib/src/main/java/uk/gov/di/ipv/core/library/domain/BaseResponse.java). The types are JourneyResponse, ProvenUserIdentityDetails, CriResponse, ClientResponse. These tend to be the returns of lambdas themselves.

We should differentiate these things and make it obvious what they are.

PR #1026, opened 2023-07-17
PYIC-3196: Move responses to state from events

Wynndow:
### What changed

Move responses to state from events

### Why did it change

This changes the way we define what the response for a state transition should be. Previously we defined it on the event. This moves it to being defined on the state.

This makes it easier to understand what's happening in the state machine. Previously you had to understand which event was called to arrive at a particular state to know what response was output by the statemachine. Now you just need to look at the state a user might be in.

Doing it this way removes a lot of duplication.

We don't have any scenarios where we want a different response for a state depending on the event that took us there, so this is fine.

We can also remove the criEvent class as we no longer need it. It was generating the cri response for us, but that's now done in the state.

### Issue tracking

- [PYIC-3196](https://govukverify.atlassian.net/browse/PYIC-3196)

[PYIC-3196]: https://govukverify.atlassian.net/browse/PYIC-3196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blakeyp (approved):
Nice 🎉

PR #1052, opened 2023-07-27
BAU: Handle unmatched M1B journeys

Wynndow:
### What changed

Return no match for failed DCMAW journey

### Why did it change

We have two journeys that are targeting the M1B profile - the app
journey and the driving license journey. Both of these journeys require a
user to get a score of 2 in the fraud CRI.

If the user gets a 0, they'll be shown a no-match page and their journey
will end. If they get a one however, they will be allowed to continue as
it's not a failed VC.

In the case of the app journey the fraud CRI is the last CRI in that
journey. The only event we were handling from there was `end`. We were
expecting a user to have either got a 0 or a 2 from fraud. A 0 would
have ended their journey as it would have contained a CI. A 2 would have
resulted in a matched profile and we'd have ended the journey
successfully.
If they got a 1 we were calling the `next` event, which
wasn't defined. This led to an error and the user saw a technical error
page. They couldn't get back to the service.

In the case of the driving license journey the user completes the fraud
CRI as the penultimate CRI. If they scored a 1 in fraud we send them to
KBV's, even though there's no chance of them matching M1B. If they
completed KBVs successfully they still wouldn't match M1B and we'd try
to call the next event, leading to an error.

These changes handle that `next` event in both scenarios, and shows the
user a no-match page.

For the driving license journey we should ideally be failing the user
earlier, after the fraud CRI. That will need more thinking about and can
be handled in another PR.

We could, and maybe should in future, route them differently to further
CRIs to allow them to improve their evidence score (a passport
successful passport check would get them to M1A). But that would require
more thinking.

### DIFF
--- a/lambdas/process-journey-step/src/main/resources/statemachine/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-step/src/main/resources/statemachine/ipv-core-main-journey.yaml
@@ -227,6 +227,8 @@ CRI_FRAUD_J1:
   events:
     end:
       targetState: IPV_SUCCESS_PAGE
+    next:
+      targetState: PYI_NO_MATCH
 
 # Passport journey (J2)
 CRI_UK_PASSPORT_J2:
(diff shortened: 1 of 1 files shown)


therealslimsladey (approved):
lgtm
by the power of journey engine
:-)

PR #1053, opened 2023-07-27
BAU: Update log message to be more accurate

AmritSidhu:
### What changed

Update log message to be more accurate


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -213,7 +213,7 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
                         new StringMapMessage()
                                 .with(
                                         LOG_MESSAGE_DESCRIPTION.getFieldName(),
-                                        "F2F VC unsuccessful.");
+                                        "Failed to match a profile");
```
blakeyp: Maybe "F2F return - failed to match a profile" ?

blakeyp (approved):

PR #1058, opened 2023-07-31
BAU Add criId to oauth error log

blakeyp:
### What changed

Add criId to oauth error log

### Why did it change

Makes debugging easier if we can immediately see which CRI the error came from


TalatAydinGDS (approved):
LGTM

PR #1079, opened 2023-08-03
Spike logic for GPG45 scores min required given fraud value

blakeyp:
Method to find the minimum remaining scores required to reach a profile given the current fraud score and a list of profiles. If the fraud score for a given profile is already reached, we know the profile can be met and so the remaining 'diff' to reach that profile is added to the set of 'compatible' scores. Taking the min of all the compatible scores gives us the minimum remaining scores required to reach one of the profiles

https://govukverify.atlassian.net/browse/PYIC-3293

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Scores.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Scores.java
@@ -2,6 +2,7 @@ package uk.gov.di.ipv.core.library.domain.gpg45;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Scores.java
```diff
@@ -142,6 +143,30 @@ private List<Gpg45Scores.Evidence> diffEvidence(Gpg45Scores target) {
         return evidenceDiff;
     }
 
+    public Gpg45Scores minimumRequiredToMeetProfileGivenCurrentFraudScore(
```
lwebb-iig, line 146: My only suggestion would be to maybe replace `RequiredToMeetProfile` with something shorter as you can infer some of this from the fact we have a `profiles` argument

lwebb-iig (commented):
One optional suggestion with the naming, otherwise LGTM

PR #1103, opened 2023-08-09
PYIC-3206: Use CiScoring lambda as part of check-existing-identity flow

lwebb-iig:
Include the CiScoring lambda as part of the check-existing-identity flow

## Proposed changes

Run the new CiScoring lambda as part of the check-existing-identity flow, before the check-existing-identity lambda itself. CI scoring should be done twice, once by the new lambda and once by the check-existing-identity lambda.

If a user has breaching CIs CIScoring should redirect from check-existing-identity and end the flow. If the user doesn’t it should allow the flow to move to the check-existing-identity.

### What changed

- JourneyEngineStepFunction updated to include CiScoring lambda
- Permissions for JourneyEngineStepFunction to invoke CiScoring lambda
- Updated CiScoring to return `JOURNEY_NEXT` on no breaching CIs rather than `JOURNEY_EVALUATE` to allow use in both step functions
- Updated CriReturnStepFunction to handle `JOURNEY_NEXT` responses from CiScoring (new CiScoringResponse choice)

Example flow when breaching CIs present:
image

Example flow when no breaching CIs present:
image

Updated CriReturnStepFunction routing
image

### Why did it change

Prior work required to only enter the check-existing-identity flow when required

### Issue tracking

- [PYIC-3206](https://govukverify.atlassian.net/browse/PYIC-3206)

### Other considerations

[PYIC-3206]: https://govukverify.atlassian.net/browse/PYIC-3206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TalatAydinGDS (dismissed):
LGTM

deploy/criReturnStepFunction.asl.json
```diff
@@ -89,10 +89,23 @@
...
+          "Not": {
+            "Variable": "$.lambdaResult.journey",
+            "IsPresent": true
+          },
+          "Next": "ReturnToFrontend"
         },
         {
           "Variable": "$.lambdaResult.journey",
-          "StringEquals": "/journey/evaluate",
+          "StringEquals": "/journey/next",
```
Wynndow: Just wondering if we can do without this new choice task and keep the template a bit dryer. We're not using a `next` string in the `RouteJourneyResponse` task so could use it in there?

lwebb-iig: Good point - I missed that. essentially can just add below to the `RouteJourneyResponse`, definitely cleaner & fits in more with existing patterns
```json
{
        "Variable": "$.lambdaResult.journey",
        "StringEquals": "/journey/next",
        "Next": "EvaluateGpg45Scores"
}
```

Wynndow (dismissed):

Wynndow (approved):

PR #1104, opened 2023-08-09
PYIC-3333: implement logic to get minimum strength score from session item

shivanshuit914:
## Proposed changes

- Use strength score from IPV sesssion and send it to CI for F2F cri.

### Issue tracking

- [PYIC-3333](https://govukverify.atlassian.net/browse/PYIC-3333)

### Other considerations

[PYIC-3333]: https://govukverify.atlassian.net/browse/PYIC-3333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -270,10 +270,32 @@ private JWEObject signEncryptJar(
...
+                Gpg45ScoresDto requiredScores = gpg45Score.getRequiredScores();
+                if (requiredScores.getFraud() > 0) {
+                    continue;
+                }
+                List<EvidenceDto> evidences = requiredScores.getEvidences();
+                int minRequiredStrengthScore =
+                        evidences.stream()
+                                .mapToInt(EvidenceDto::getStrength)
+                                .summaryStatistics()
+                                .getMin();
```
blakeyp: I think this is not quite right. When there's a list of evidence eg. [EV_32, EV_42] it means 2 pieces of evidence are required, one strength 3 the other 4. However for F2F and the profiles we're dealing with there should only ever be 1 piece of evidence in the list - so I think we can just assume and just get the strength score of the first piece of evidence. So something like `int minRequiredStrength = evidences.get(0).getStrength()`

blakeyp (approved):

PR #1138, opened 2023-08-22
PYIC-3397: Route fail-with-no-ci to new escape page

AmritSidhu:
### What changed

Route fail-with-no-ci to new escape page

### Why did it change

If the user has a thin file or abandons KBV (selects ‘cannot answer the questions’) KBV CRI returns to us a fail VC with no CI. Currently we redirect the user to the pyi-kbv-thin-file page. We need to send them to a new ‘escape’ page which will offer DCMAW as an alternative route to prove their identity. 

### Issue tracking

- [PYIC-3397](https://govukverify.atlassian.net/browse/PYIC-3397)

[PYIC-3397]: https://govukverify.atlassian.net/browse/PYIC-3397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -216,6 +216,18 @@ PYI_ESCAPE:
     end:
       targetState: END
 
+PYI_CRI_ESCAPE:
+  response:
+    type: page
+    pageId: pyi-cri-escape
+  events:
+    dcmaw:
+      targetState: CRI_DCMAW_J5
+    f2f:
+      targetState: CRI_F2F_J5
+    end:
+      targetState: END
+
 ERROR:
   response:
     type: error
@@ -296,6 +308,9 @@ CRI_KBV_J2:
       targetState: IPV_SUCCESS_PAGE
     fail-with-no-ci:
       targetState: PYI_KBV_THIN_FILE
+      checkFeatureFlag:
+        criEscapeFlag:
+          targetState: PYI_CRI_ESCAPE
 
 # Driving licence journey (J3)
 CRI_DRIVING_LICENCE_J3:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -216,6 +216,18 @@ PYI_ESCAPE:
...
 
+PYI_CRI_ESCAPE:
+  response:
+    type: page
+    pageId: pyi-cri-escape
+  events:
+    dcmaw:
+      targetState: CRI_DCMAW
+    f2f:
+      targetState: CRI_F2F_J4
```
blakeyp: I'm thinking here we might want to define separate target states for the 2 options rather than pointing to existing journeys CRI_DCMAW and CRI_F2F_J4? Particularly for CRI_DCMAW all its fail states route to the doc check page which would be very confusing for the user/they could get stuck in a loop. I'm not sure product/UCD have considered this so will raise it with them

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -216,6 +216,18 @@ PYI_ESCAPE:
     end:
       targetState: END
 
+PYI_CRI_ESCAPE:
+  response:
+    type: page
+    pageId: pyi-cri-escape
+  events:
+    dcmaw:
+      targetState: CRI_DCMAW
```
blakeyp: Also think we'll want to define a new state for this DCMAW journey so that it skips going to the address and fraud CRIs as the user will have already been through them

blakeyp (approved):

PR #1143, opened 2023-08-23
BAU: Refactor CiMit VC parsing

Wynndow:
### What changed

Refactor CiMit VC parsing

### Why did it change

Rather than building an object by mapping separate parts of the VC into it, we can just define the object models to match the VC and let the deserializer work it's magic.

This solves an issue with using the CiMit service to fetch the VC from our stub. The VC received from our stub doesn't include some information that the proper VC will, such as the mitigating credential, included alongside the mitigation code. Without the mitigating credential we were getting a NPE when trying to map the object in the original way. Letting the deserializer handle it all means we don't encounter this problem - the missing data will just be set to null. We include all data from the stub that we actually need, so this is fine.


TalatAydinGDS (dismissed):
LGTM

TalatAydinGDS (approved):
LGTM

PR #1146, opened 2023-08-24
BAU: Reduce config service cache time in dev

Wynndow:
### What changed

Reduce config service cache time in dev

### Why did it change

I was bored of waiting three minutes for param changes in my dev deploy to take effect. This change removes the cache in dev and leaves it at 3 minutes in all other envs.


TalatAydinGDS (approved):
LGTM

lwebb-iig (approved):

rpayal (approved):

PR #1157, opened 2023-08-30
SPIKE: Enable snapstart in dev

Wynndow:
### What changed

Enable snapstart in dev

### Why did it change

This is a spike to see what happens if we use snapstart. https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html

* It only applies changes in the dev env
* Snapstart only works with x86_64 architecture
* Our step functions weren't invoking our lambdas with a published version (via an alias). Instead they were just invoking whatever the latest version was. You need to specify the version for snapstart to work, hence the changes to the step function resources in the template.
* There needed to be a change to the TokenRequestValidator. The audience value to validate the request against had been baked in to the snapstart image and was set to the wrong thing. This ensures that the audience gets set at first runtime.

These changes made the dev experience much...snappier. It might be a good thing to implement, and if we like it we could maybe even use it in prod and turn off provisioned concurrency. Do we *need* to run arm64 in prod though? We'd also need to do a bunch of checking to ensure we weren't shooting ourselves in the foot like with the audience issue mentioned above.

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -94,7 +94,6 @@ public class CheckExistingIdentityHandler
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
     private final AuditService auditService;
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
-    private final String componentId;
 
     @SuppressWarnings("unused") // Used by AWS
     public CheckExistingIdentityHandler(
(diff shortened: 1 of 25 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -340,7 +337,7 @@ private AuditEvent buildProfileMatchedAuditEvent(
                         ipAddress);
         return new AuditEvent(
                 AuditEventTypes.IPV_GPG45_PROFILE_MATCHED,
-                componentId,
+                configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
```
blakeyp, line 340: Is there a cost to these repeated calls to SSM? I guess not because it's cached?

blakeyp (dismissed):

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Yep - it'll be slightly more than what we were doing before (storing the component ID on init), but a drop in the ocean in terms of SSM calls.

Wynndow: But yeah, we have the three minute cache so it's not like there's going to be loads more.

blakeyp (approved):

PR #1169, opened 2023-09-04
BAU: Update sdkman config to use java 17

kerrr:
### What changed

Update sdkman config to use java 17

### Why did it change

We now use java 17 for our lambdas. This change will use the correct java version for people using sdkman.


therealslimsladey (approved):
lgtm

PR #1170, opened 2023-09-04
PYIC-3490: Fix GPG45 profile matching bug

DanCorderIPV:
### What changed

When comparing a GPG45 score to a profile try all permutations of evidence until one matches or we have tried them all.

### Why did it change

Previously the code ordered evidence by strength and then validity in the score and the profile and then compared the evidences in each position.
However, both strength and validity are equally important in the evidence so the ordering doesn't always work meaning that a profile needing 33 and 22 would fail to match against a score with evidence 42 and 33 because the check would try the 42 against the 33 required and fail it.

### Issue tracking

- [PYIC-3490](https://govukverify.atlassian.net/browse/PYIC-3490)

[PYIC-3490]: https://govukverify.atlassian.net/browse/PYIC-3490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
@@ -1,9 +1,9 @@
 package uk.gov.di.ipv.core.library.domain.gpg45;
 
-import java.util.List;
-
 import uk.gov.di.ipv.core.library.domain.gpg45.Gpg45Scores.Evidence;
 import uk.gov.di.ipv.core.library.helpers.ListHelper;
+
+import java.util.List;
 
 /** Enumeration of all GPG 45 profiles along with their reference Gpg45Scores value. */
 public enum Gpg45Profile {
@@ -87,7 +87,8 @@ public enum Gpg45Profile {
             return false;
         }
 
-        // No profile needs more than 3 pieces of evidence so calling getPermutations should be safe.
+        // No profile needs more than 3 pieces of evidence so calling getPermutations should be
+        // safe.
         var achievedEvidencePermutations = ListHelper.getPermutations(achievedEvidences);
 
         for (List<Evidence> achievedEvidencePermutation : achievedEvidencePermutations) {
@@ -99,7 +100,8 @@ public enum Gpg45Profile {
         return false;
     }
 
-    private boolean evidencePermutationIsSatisfactory(List<Evidence> requiredEvidences, List<Evidence> achievedEvidences) {
+    private boolean evidencePermutationIsSatisfactory(
+            List<Evidence> requiredEvidences, List<Evidence> achievedEvidences) {
         for (int i = 0; i < requiredEvidences.size(); i++) {
             var requiredEvidence = requiredEvidences.get(i);
             var achievedEvidence = achievedEvidences.get(i);
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
@@ -1,5 +1,10 @@
 package uk.gov.di.ipv.core.library.domain.gpg45;
 
+import java.util.List;
+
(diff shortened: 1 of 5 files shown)


lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
```diff
@@ -63,28 +68,46 @@ public Gpg45Scores difference(Gpg45Scores target) {
...
+
+        if (requiredEvidences.size() > achievedEvidences.size()) {
             return false;
         }
 
-        for (int i = 0; i < scores.getEvidences().size(); i++) {
-            var sourceEvidence = scores.getEvidence(i);
-            var targetEvidence = target.getEvidence(i);
+        // No profile needs more than 3 pieces of evidence so calling getPermutations should be
+        // safe.
```
Wynndow, line 91: How bad does `getPermutations` get if a user had four pieces of evidence? This isn't a world we're in right now, but I don't think the number of evidence scores they've achieved so far is limited by the number of pieces of evidence any one profile requires. In a hypothetical journey in the future where they'd tried different pieces of evidence and failed, then run mitigation journeys to counteract those CIs, they could start building up more than 3. An edge case but worth considering maybe.

lib/src/test/java/uk/gov/di/ipv/core/library/helpers/ListHelperTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private <T> boolean listsOfListsAreEquivalent(List<List<T>> a, List<List<T>> b) {
+        if (a.size() != b.size()) {
+            return false;
+        }
+
+        return a.stream()
+                .allMatch(al -> b.stream().anyMatch(bl -> listsContainSameElements(al, bl)));
+    }
```
Wynndow: I'm not sure this is quite right, for example this test passes:
```
    @Test
    void test() {
        List<List<Integer>> a = List.of(List.of(1, 2, 3), List.of(1, 2, 3), List.of(1, 2, 3));
        List<List<Integer>> b = List.of(List.of(1, 2, 3), List.of(2, 3, 4), List.of(3, 4, 5));

        assertTrue(listsOfListsAreEquivalent(a, b));

    }
```

You could implement a comparator for the lists of lists, then sort the two lists and assert equality? Something like this? https://stackoverflow.com/questions/35761864/java-sort-list-of-lists

lib/src/test/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45ProfileTest.java
```diff
@@ -28,19 +36,96 @@ void shouldMatchExact() {
...
+                Arguments.of(3, 3, 2, 2), // Exact match
+                Arguments.of(2, 2, 3, 3), // Exact match reversed
+                Arguments.of(4, 2, 3, 3), // Stronger evidence has weaker validity
+                Arguments.of(2, 4, 3, 3), // Weaker evidence has stronger validity
+                Arguments.of(4, 4, 4, 4) // Higher strength and validity still matches
+                );
+    }
+
+    @ParameterizedTest
+    @MethodSource("ShouldMatchDoubleEvidenceScoresTestCases")
```
Wynndow, line 51: Not used `@MethodSource` before - it's nice!

Wynndow (commented):
Looks good, just a thing about the testing of the ListHelper

lib/src/main/java/uk/gov/di/ipv/core/library/domain/gpg45/Gpg45Profile.java
DanCorderIPV: 4 items have 24 permutations, and 5 have 120, so they should also be fine.

lib/src/test/java/uk/gov/di/ipv/core/library/helpers/ListHelperTest.java
DanCorderIPV: Good point! I'll fix this

Wynndow (approved):
Nice!

PR #1187, opened 2023-09-11
BAU: Use process step response instead of journey

Wynndow:
### What changed

Use process step response instead of journey

### Why did it change

Now we have the process step response, we can use it in the places that were using the journey step response.

This also refactors the ProcessStepResponse to not use the `Map.of()` method. This doesn't allow nulls. If we're not passing any values to the lambda this will blow up. Constructing it with a new HashMap works.


therealslimsladey (approved):
lgtm

lwebb-iig (approved):

PR #1192, opened 2023-09-13
PYIC-3539: Track CRI connection

Wynndow:
### What changed

Track CRI connection

### Why did it change

[PYIC-3539: Maintain users connection with a CRI](https://github.com/alphagov/di-ipv-core-back/commit/ea996d8eb89de2085131a582e3a75b328a97216b) 

If a user has been sent to a CRI they'll need to use the same
connections when they are returned to core. That is, they'll need to do
the OAuth dance with the same CRI they were sent to.
This is an important distinction in the case that we have rotated the
config for a CRI - perhaps new endpoint, new keys etc. If this change
happens after a user has initiated an OAuth session with a CRI but
before they return, they need to use the config at the point they
initiated it to complete the journey.

This stores the connection used in their CRI Oauth session details. That
connection is then re-used when coming back to core.

There are some places where we fetch CRI config independently of having
a users active CRI Oauth session. In the async journey for example, and
when building an identity. We do this by looking up the issuer of a VC
and then fetching config based on this. This implied that a CRI
shouldn't change their component ID.

[PYIC-3539: Tests for previous commit](https://github.com/alphagov/di-ipv-core-back/commit/9117ba30ec8f48a2e53754ae5f22807deca544f4) 

This updates the tests for the changes made in the previous commit. I'm
commiting them separately to make reviewing easier - there's a fair
amount of boilerplate change in these changes.

[PYIC-3539: Use active connection directly less](https://github.com/alphagov/di-ipv-core-back/commit/b623ce8b1f97d6de752ea3aabcd7c76e70309a8f) 

In a few places we were fetching a CRIs config using the method in the
config service for fetching the active connection.
In these places we were just trying to determine the CRIs component ID.
We have a method for doing that without pulling back the entire config.

It also means that we're using the "active" connection less directly.
Sure, the component ID method is using it under the hood, but we can
refactor that at some point in the future if we needed to.

### Issue tracking

- [PYIC-3539](https://govukverify.atlassian.net/browse/PYIC-3539)

[PYIC-3539]: https://govukverify.atlassian.net/browse/PYIC-3539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Wynndow:
Closing in favour of two separate PRs

PR #1193, opened 2023-09-13
PYIC-3539: Store CRI connection

Wynndow:
### What changed

Store CRI connection. The interesting bit is happening here: https://github.com/alphagov/di-ipv-core-back/pull/1193/files#diff-274862ae6bce0c95efeb08dd49c38d2a2fdfb49f2bc32e8847ccbb3e190abeabR152

### Why did it change

We need to store the connection the users started a journey with a CRI with, so that they can pick this up when they return.

This needs merging before we start using the stored connection in other lambdas to ensure that the value is there to be read.

### Issue tracking

- [PYIC-3539](https://govukverify.atlassian.net/browse/PYIC-3539)

[PYIC-3539]: https://govukverify.atlassian.net/browse/PYIC-3539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -25,6 +25,8 @@ import uk.gov.di.ipv.core.library.domain.ContraIndicatorMitigation;
 import uk.gov.di.ipv.core.library.domain.ContraIndicatorScore;
 import uk.gov.di.ipv.core.library.dto.CredentialIssuerConfig;
 import uk.gov.di.ipv.core.library.exceptions.ConfigException;
+import uk.gov.di.ipv.core.library.exceptions.NoConfigForConnectionException;
+import uk.gov.di.ipv.core.library.persistence.item.CriOAuthSessionItem;
 
 import java.net.URI;
 import java.nio.file.Path;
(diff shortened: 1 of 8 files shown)


kerrr (approved):

lib/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -240,13 +242,28 @@ public String getCriPrivateApiKey(String criId) {
...
+
+    public CredentialIssuerConfig getCriConfigForConnection(String connection, String criId) {
         final String pathTemplate =
                 ConfigurationVariable.CREDENTIAL_ISSUERS.getPath() + "/%s/connections/%s";
-        Map<String, String> result =
-                getSsmParameters(pathTemplate, false, credentialIssuerId, activeConnection);
+        Map<String, String> result = getSsmParameters(pathTemplate, false, criId, connection);
+
+        if (result == null || result.isEmpty()) {
+            throw new NoConfigForConnectionException(
```
TalatAydinGDS, line 260: Wouldn't it be better if we catch this exception in lambda and return it later with a specific error code and this message?

TalatAydinGDS (commented):
It looks for me and I have added jus one comment.

PR #1196, opened 2023-09-13
PYIC-3531: Use CRI api key for connection

Wynndow:
### What changed

Use CRI api key for connection

### Why did it change

The API key used to authenticate with the CRIs is stored in secrets
manager. This adds the ability to have an API key per CRI connection.
The new path mirrors how we structure the SSM config, but the value is
store in secrets manager.

The current API keys have been copied to the new paths in staging,
integration and prod. Effectively `connections/main/` has been added to
each API key, ready for when this code is used.

### Issue tracking

- [PYIC-3531](https://govukverify.atlassian.net/browse/PYIC-3531)

[PYIC-3531]: https://govukverify.atlassian.net/browse/PYIC-3531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS:
@Wynndow looks like this is slowly slipping out of date - do you think it's worth bringing up to date and merging, or shall we close this for now (can always keep the branch if it might be useful)?

As I understand it, this might have been needed for the passport CRI update, but actually wasn't necessary in the end?

Wynndow:
@Joe-Edwards-GDS I _think_ it's probably worth getting it merged. It would mean we're consistent in our approach to CRIs. The API keys are also already in the correct position in secrets manager, so it should just be a case of rebasing this and getting it in (with some double checking on those API keys).
I'll get back up to speed with core.

Joe-Edwards-GDS (approved):

PR #1209, opened 2023-09-21
BAU: Fix same session bug with journey map

Wynndow:
### What changed

Fix same session bug with journey map

### Why did it change

Enabling same session mitigations introduces a bug in separate and same session mitigation.

The way we activate separate session mitigation is by updating the CiMit config to send a user on a mitigation journey for same session, rather than failing them.

This means that if a user is at DCMAW on a mitigation journey (same or separate), and gets a failed response with no CIs, we attempt to send them back on the same session mitigation journey. This is because they still have the CI that needs mitigating and they still have a breaching score due to no mitigation check score being in place.

Attempting to send them on this journey fails as we don't define that event on the dcmaw mitigation states.

This scenario is the same as the thin file scenario. In DCMAW a user can get this by failing the liveness likeness check.

This fixes the immediate bug by updating the journey map to handle the mitigation event on the mitigation state. This is more of a band aid than a proper fix.

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -517,6 +517,8 @@ MITIGATION_01_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
     fail-with-no-ci:
       targetState: PYI_ANOTHER_WAY
+    mitigation02:
+      targetState: PYI_ANOTHER_WAY
 
 # Mitigation journey (02)
 MITIGATION_02_OPTIONS:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -555,6 +557,8 @@ MITIGATION_02_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
     fail-with-no-ci:
       targetState: PYI_ANOTHER_WAY
+    mitigation02:
+      targetState: PYI_ANOTHER_WAY
```
Joe-Edwards-GDS, line 561: Are there any other CRI states that could return `mitigation02` for the same reason?

I don't think so, because no other CRI should be used with a mitigation journey for now. When we add F2F I'm keen to normalise this and make sure we cover all the possible events in all cases, but it should be alright for now.

Joe-Edwards-GDS (approved):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: To return `mitigation02` the user would have to have the particular CI (unmitigated)we're interested in, and not be initiating a new journey, so no, shouldn't be.

PR #1210, opened 2023-09-21
PYIC-3505: On-the-fly VC Status (4): Get status of last visited CRI VC on-the-fly in EvaluateGpg45ScoresHandler

TalatAydinGDS:
### What changed

Add new method (and tests for) getVCSuccessStatus to UserIdentityService. The new method will

- fetch the VCs from credential store
- return an optional indicating whether the VC for the given user and issuer is successful or not, if it exists
- return an empty optional otherwise

### Why did it change

EvaluateGpg45ScoresHandler determines the status of the last visited CRI VC on-the-fly

### Issue tracking

- [PYIC-3505](https://govukverify.atlassian.net/browse/PYIC-3505)

### Other considerations

[PYIC-3505]: https://govukverify.atlassian.net/browse/PYIC-3505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
+++ b/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
@@ -22,6 +22,7 @@ import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.Name;
+import uk.gov.di.ipv.core.library.dto.CredentialIssuerConfig;
 import uk.gov.di.ipv.core.library.dto.VcStatusDto;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.NoVcStatusForIssuerException;
@@ -153,10 +154,12 @@ public class BuildProvenUserIdentityDetailsHandler
             throws ParseException, JsonProcessingException, ProvenUserIdentityDetailsException,
                     NoVcStatusForIssuerException {
         for (VcStoreItem item : credentialIssuerItems) {
+            CredentialIssuerConfig credentialIssuerConfig =
+                    configService.getCredentialIssuerActiveConnectionConfig(
+                            item.getCredentialIssuer());
             if (EVIDENCE_CRI_TYPES.contains(item.getCredentialIssuer())
                     && userIdentityService.isVcSuccessful(
-                            currentVcStatuses,
-                            configService.getComponentId(item.getCredentialIssuer()))) {
+                            currentVcStatuses, credentialIssuerConfig.getComponentId())) {
                 JsonNode vcSubjectNode =
                         mapper.readTree(
                                         SignedJWT.parse(item.getCredential())
@@ -198,10 +201,12 @@ public class BuildProvenUserIdentityDetailsHandler
             throws ParseException, JsonProcessingException, ProvenUserIdentityDetailsException,
                     NoVcStatusForIssuerException {
         for (VcStoreItem item : credentialIssuerItems) {
+            CredentialIssuerConfig credentialIssuerConfig =
+                    configService.getCredentialIssuerActiveConnectionConfig(
+                            item.getCredentialIssuer());
             if (item.getCredentialIssuer().equals(ADDRESS_CRI)
                     && userIdentityService.isVcSuccessful(
-                            currentVcStatuses,
-                            configService.getComponentId(item.getCredentialIssuer()))) {
+                            currentVcStatuses, credentialIssuerConfig.getComponentId())) {
                 JsonNode addressNode =
                         mapper.readTree(
                                         SignedJWT.parse(item.getCredential())
--- a/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
+++ b/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
@@ -22,7 +22,6 @@ import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.Name;
-import uk.gov.di.ipv.core.library.dto.CredentialIssuerConfig;
 import uk.gov.di.ipv.core.library.dto.VcStatusDto;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.NoVcStatusForIssuerException;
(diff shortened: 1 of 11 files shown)


lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -153,10 +154,12 @@ private NameAndDateOfBirth getProvenIdentityNameAndDateOfBirth(
...
                     NoVcStatusForIssuerException {
         for (VcStoreItem item : credentialIssuerItems) {
+            CredentialIssuerConfig credentialIssuerConfig =
+                    configService.getCredentialIssuerActiveConnectionConfig(
+                            item.getCredentialIssuer());
             if (EVIDENCE_CRI_TYPES.contains(item.getCredentialIssuer())
                     && userIdentityService.isVcSuccessful(
-                            currentVcStatuses,
-                            configService.getComponentId(item.getCredentialIssuer()))) {
+                            currentVcStatuses, credentialIssuerConfig.getComponentId())) {
```
Wynndow: I'd be inclined to stick with just calling `configService.getComponentId(item.getCredentialIssuer())`. It does the same thing under the hood (gets the active connection to get the config) but is a bit less verbose than fetching the config directly. Also, there might be a future where we don't want to use the active connection but would rather use whatever connection the user initiated a journey with a CRI with.

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -198,10 +201,12 @@ private List<Address> getProvenIdentityAddresses(
...
                     NoVcStatusForIssuerException {
         for (VcStoreItem item : credentialIssuerItems) {
+            CredentialIssuerConfig credentialIssuerConfig =
+                    configService.getCredentialIssuerActiveConnectionConfig(
+                            item.getCredentialIssuer());
             if (item.getCredentialIssuer().equals(ADDRESS_CRI)
                     && userIdentityService.isVcSuccessful(
-                            currentVcStatuses,
-                            configService.getComponentId(item.getCredentialIssuer()))) {
+                            currentVcStatuses, credentialIssuerConfig.getComponentId())) {
```
Wynndow: Same as above

libs/vc-helper/src/main/java/uk/gov/di/ipv/core/library/vchelper/VcHelper.java
```diff
@@ -37,9 +37,17 @@ public static void setConfigService(ConfigService configService) {
...
+    }
+
+    public static Set<String> getNonEvidenceCredentialIssuers(ConfigService configService) {
         return NON_EVIDENCE_CRI_TYPES.stream()
-                .map(credentialIssuer -> configService.getComponentId(credentialIssuer))
+                .map(
+                        credentialIssuer ->
+                                configService
+                                        .getCredentialIssuerActiveConnectionConfig(credentialIssuer)
+                                        .getComponentId())
```
Wynndow: I think this should stay as a call to `configService.getComponentId()`, rather than getting the active connection then getting the component ID. Functionally the same but less verbose and it might dry up a lot of the mocking you've had to do in the tests.

libs/vc-helper/src/main/java/uk/gov/di/ipv/core/library/vchelper/VcHelper.java
```diff
@@ -37,9 +37,17 @@ public static void setConfigService(ConfigService configService) {
         VcHelper.configService = configService;
     }
 
-    private static Set<String> getNonEvidenceCredentialIssuers() {
+    public static Set<String> getNonEvidenceCredentialIssuers() {
+        return VcHelper.getNonEvidenceCredentialIssuers(configService);
+    }
+
+    public static Set<String> getNonEvidenceCredentialIssuers(ConfigService configService) {
```
Wynndow: It looks like the only thing calling this now overloaded method is the version above that doesn't take a configService. Is this added for a reason? If not, we can just keep the original version and drop this new one.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -172,6 +172,19 @@ private List<VcStoreItem> getSuccessfulVCStoreItems(List<VcStoreItem> vcStoreIte
         return successfulVCStoreItems;
     }
 
+    public Optional<Boolean> getVCSuccessStatus(String userId, String criIssuer)
+            throws ParseException {
+        List<VcStoreItem> vcStoreItems = getVcStoreItems(userId);
```
Wynndow: We have a method in this class, `getVcStoreItem(String userId, String criId)`, which will get the VC for the CRI for the user. This would mean you wouldn't have to pull them all, deserialize them and filter them.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -172,6 +172,19 @@ private List<VcStoreItem> getSuccessfulVCStoreItems(List<VcStoreItem> vcStoreIte
...
+        for (VcStoreItem vcStoreItem : vcStoreItems) {
+            SignedJWT vc = SignedJWT.parse(vcStoreItem.getCredential());
+            if (vc.getJWTClaimsSet().getIssuer().equals(criIssuer)) {
+                return Optional.of(
```
Wynndow: Do we need to return an optional? The only place we use this we just call `get()` on the optional, so we should probably just return the boolean.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -172,6 +172,19 @@ private List<VcStoreItem> getSuccessfulVCStoreItems(List<VcStoreItem> vcStoreIte
...
+                        VcHelper.isSuccessfulVc(vc, VcHelper.getNonEvidenceCredentialIssuers()));
```
Wynndow: We have a method `VcHelper.isSuccessfulVc(SignedJWT vc)` which doesn't require you to pass in a list of non evidence CRIs. It actually uses the same method you're using here. Might be cleaner to just use that.

Wynndow (changes requested):
Just a few comments that might keep things a bit dryer.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: If we cannot find any VC Itesm, we return Optional.empty() with isPresent() in the EvaluateGpg45ScoresHandler. Returning a boolean here would break this structure and unit tests failed. I think, it's better for it to remain as 'Optional' .

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: I had to change the address after CRI, as it was breaking. I will test it again, as it was related to address and claim excluding CRIs.  Let me change it again and retest again.

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
TalatAydinGDS: Done.

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
TalatAydinGDS: Done.

libs/vc-helper/src/main/java/uk/gov/di/ipv/core/library/vchelper/VcHelper.java
TalatAydinGDS: Done.

libs/vc-helper/src/main/java/uk/gov/di/ipv/core/library/vchelper/VcHelper.java
TalatAydinGDS: Done.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: Done.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: Done.

Wynndow (approved):
Nice! Thanks for making the changes

PR #1216, opened 2023-09-21
PYIC-3301: Remove unused feature flags

Wynndow:
### What changed

Remove unused feature flags

### Why did it change

We have a number of feature flags that are no longer required as they are turned on all the way to production.

### Issue tracking

- [PYIC-3301](https://govukverify.atlassian.net/browse/PYIC-3301)

[PYIC-3301]: https://govukverify.atlassian.net/browse/PYIC-3301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -2231,22 +2231,6 @@ Resources:
           POWERTOOLS_SERVICE_NAME: !Sub ci-scoring-${Environment}
           IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           CLIENT_OAUTH_SESSIONS_TABLE_NAME: !Ref ClientOAuthSessionsTable
-          CI_STORAGE_GET_LAMBDA_ARN: !Sub
-            - "arn:aws:lambda:eu-west-2:${cimit_account_id}:function:getContraIndicators-${env}"
-            - cimit_account_id: !If
-                - UseIndividualCiMitStubs
-                - !Ref AWS::AccountId
-                - !FindInMap
-                  - EnvironmentConfiguration
-                  - !Ref AWS::AccountId
-                  - cimitAccountId
-              env: !If
-                - UseIndividualCiMitStubs
-                - !Sub ${Environment}
-                - !FindInMap
-                  - EnvironmentConfiguration
-                  - !Ref AWS::AccountId
-                  - cimitEnvironment
           CIMIT_GET_CONTRAINDICATORS_LAMBDA_ARN: !Sub
             - "arn:aws:lambda:eu-west-2:${cimit_account_id}:function:getContraIndicatorCredential-${env}"
             - cimit_account_id: !If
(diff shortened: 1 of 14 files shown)


deploy/template.yaml
```diff
@@ -2231,22 +2231,6 @@ Resources:
           POWERTOOLS_SERVICE_NAME: !Sub ci-scoring-${Environment}
           IPV_SESSIONS_TABLE_NAME: !Ref SessionsTable
           CLIENT_OAUTH_SESSIONS_TABLE_NAME: !Ref ClientOAuthSessionsTable
-          CI_STORAGE_GET_LAMBDA_ARN: !Sub
```
rpayal, line 2234: this was done in previous PR

Wynndow: Ah ok - was it merged because it's still in the template on main? - https://github.com/alphagov/di-ipv-core-back/blob/main/deploy/template.yaml#L2234

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CoreFeatureFlag.java
```diff
@@ -1,10 +1,7 @@
 package uk.gov.di.ipv.core.library.config;
 
 public enum CoreFeatureFlag implements FeatureFlag {
-    USE_CONTRA_INDICATOR_VC("useContraIndicatorVC"),
-    USE_POST_MITIGATIONS("usePostMitigations"),
-    MITIGATION_ENABLED("mitigationEnabled"),
-    BUNDLE_CIMIT_VC("bundleCimitVC");
+    UNUSED_PLACEHOLDER("unusedPlaceHolder");
```
rpayal, line 4: where is this enum used ?

Wynndow: It's only used in tests - it's just a placeholder so that the enum class itself can still exist. We also have a couple of tests that rely on a feature flag enum existing, so this feeding into them.

rpayal: this class can be removed if not needed.

rpayal (dismissed):

deploy/template.yaml
Wynndow: Oh, I see what you mean now. Yeah, I'll merge that other PR first I think. Not 100% why this clash has happened.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CoreFeatureFlag.java
Wynndow: Having the ability to add feature flags is a thing that we'll want in the future. So I want to keep that infrastructure (this enum class) about. And as I mentioned, some of our tests rely on having a feature flag available.

rpayal (approved):

PR #1221, opened 2023-09-26
Local running

Wynndow:
### What changed

This is an attempt at running core-front, core-back and the CRIs locally. They run in containers and are orchestrated by
docker compose. The benefits of this are quick(er) deployments and being able to get a debugger into the lambdas.

### What it's not

A replacement for deploying your stack to AWS and making sure things work properly in that environment.

### How it works

Core-front, orch-stub and the CRI stubs are pretty straight forward and just use the current Dockerfiles we have to build
images and run them.

Core-back has some extra code to replace the AWS step functions we use and the API gateways. A Spark application is spun
up with endpoints the same as the API gateways. The lamdbas are called from here, with the required inputs constructed
appropriately.

All other AWS services are still used - Dynamo, SSM, SQS etc. They're called by the lambdas as usual.


local-running/setSsmConfigForLocalOrCloudRunning.py
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env python
```
DanCorderIPV: I don't know if any of these config values are particularly sensitive, but would it be better to put this file in the core-common-infra repo?

Wynndow: No, they're not sensitive. And if this way of running core was actually accepted then it may make sense to put them in common-infra. But for a spikey thing that will probably die a death this seems like a reasonable way to optionally add them to your dev params.

Wynndow:
Closing in favour of https://github.com/govuk-one-login/ipv-core-back/pull/1240

PR #1222, opened 2023-09-27
BAU: Remove unused state from journey map

Wynndow:
### What changed

Remove unused state from journey map

### Why did it change

The `MITIGATION_02_POST_DCMAW_SUCCESS_PAGE` is not needed. The content in it is not appropriate as it mentions going on to provide your address and complete a fraud check. At this point in the journey the user already has those things. Instead we should just show them the success page.

Thankfully the page was plumed in wrong and was being skipped - the user was taken straight to the success page.

This removes it in from the journey map. It also removes the `next` event from the `MITIGATION_02_CRI_DCMAW` state. This is because the actual event provided on a successful mitigation in DCMAW is `end`. This is defined on the parent `CRI_STATE` state. This is the reason that we were skipping the unnecessary page in the first place.


Joe-Edwards-GDS (approved):
Sounds good - I'm always happy to see code being deleted :D

rpayal (approved):

PR #1224, opened 2023-09-27
PROBLEM-81: Reorder checks in eval-gpg45-scores lambda

Wynndow:
### What changed

Reorder checks in eval-gpg45-scores lambda

### Why did it change

[PROBLEM-81: Reorder checks in eval-gpg45-scores lambda](https://github.com/alphagov/di-ipv-core-back/pull/1224/commits/ca59c37ffdb8c13a4fad19c06e34b917f5a7b889) 

We've seen SPoT receive some iddentity bundles which it has then failed
due to name correlation issues. These identities shouldn't have made it
past core.

When we check for a matched profile in eval-gpg45-scores lamdba we set
the VOT value in the users session as P2 if we match a profile.

After we do this we were checking for name correlation issues. If the
names didn't correlate we'd return a no-match response to the user. But
the P2 was already set.

This meant that a user would be told they'd failed, then returned to
orchestrator. Their identity is then fetched and a bundle with P2 is
returned, despite failing name correlation. This was then sent to SPoT
and rejected.

This re-orders the checks in eval-gpg45-lambda to only set a P2 once the
other checks have passed. It also updates all the tests to ensure that
the correct value is set in the users session.

[PROBLEM-81: Check for data correlation in check-existing-identity](https://github.com/alphagov/di-ipv-core-back/pull/1224/commits/f41d4bd21e422388beec9cf866bc5078bf450c01) 

We were only checking for data correlation in the
check-existing-identity lambda when a user had just come back from F2F.
In that scenario we send the user to a special page before allowing them
to reset their identity.

If they had come back from somewhere else and had uncorrelated data we
weren't checking.

If a user had all the VCs needed for a profile but had a name
correlation error - picked up from their last VC, we were issuing an
identity. This was being pricked up by SPoT.

Rather than issuing a P2 we should just be resetting the users identity.

### Issue tracking

- [PROBLEM-81](https://govukverify.atlassian.net/browse/PROBLEM-81)

[PROBLEM-81]: https://govukverify.atlassian.net/browse/PROBLEM-81?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -167,13 +167,28 @@ public class CheckExistingIdentityHandler
                             userIdentityService.getUserIssuedCredentials(userId));
             updateSuccessfulVcStatuses(ipvSessionItem, credentials);
 
+            boolean dataCorrelates = vcDataCorrelates(userId);
+            if (!dataCorrelates && completedF2F(f2fRequest, f2fVc)) {
+                return JOURNEY_FAIL;
+            } else if (!dataCorrelates) {
+                LOGGER.info(
+                        new StringMapMessage()
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        "VC data does not correlate so resetting identity."));
+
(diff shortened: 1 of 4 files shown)


TalatAydinGDS (approved):
LGTM

kerrr (approved):
:rocket:

PR #1232, opened 2023-10-02
PYIC-3532: Shared Audit data - Milliseconds event timestamping

lwebb-iig:
Shared Audit data - Milliseconds event timestamping

### What changed

`event_timestamp_ms` field added to all audit events

### Why did it change

meet HMRCs shared signal millisecond requirement

### Issue tracking

- [PYIC-3532](https://govukverify.atlassian.net/browse/PYIC-3532)

### Other considerations

[PYIC-3532]: https://govukverify.atlassian.net/browse/PYIC-3532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionGpg45ProfileMatchedTest.java
+++ b/libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionGpg45ProfileMatchedTest.java
@@ -79,6 +79,8 @@ class AuditExtensionGpg45ProfileMatchedTest {
                         + "\"vcTxnIds\":[\"txn1\",\"txn2\",\"txn3\"]"
                         + "},"
                         + "\"timestamp\":1666170506"
+                        + ","
+                        + "\"event_timestamp_ms\":1666170506000"
                         + "}";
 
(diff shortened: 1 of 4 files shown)


libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionGpg45ProfileMatchedTest.java
```diff
@@ -79,6 +79,8 @@ void shouldSerializeToTheCorrectJson() throws Exception {
                         + "\"vcTxnIds\":[\"txn1\",\"txn2\",\"txn3\"]"
                         + "},"
                         + "\"timestamp\":1666170506"
+                        + ","
+                        + "\"event_timestamp_ms\":1666170506000"
```
Wynndow: I'm wondering if it's worth updating that spyClock mock to return an instant that isn't an exact second? That way you could have this test value end in something other than `000` to give confidence it's actually using milliseconds, and not just multiplying some value for seconds by 1000.

Wynndow (approved):

PR #1240, opened 2023-10-04
BAU: Local running

Wynndow:
### What changed

This adds a new gradle sub-project to allow starting the lambdas up
locally in a Spark web app.

It also a Dockerfile and docker-compose file to start all the services
required up in containers, and for them to communicate with each other.

### Why did it change

To allow an option to debug lambdas, and to iterate faster when developing.

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachineInitializer.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachineInitializer.java
@@ -1,5 +1,6 @@
 package uk.gov.di.ipv.core.processjourneyevent.statemachine;
 
+import com.amazonaws.util.IOUtils;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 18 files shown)


lwebb-iig (commented):
Looks good - did you want to capture any of the additional thoughts / future work around some of the docker quirks mentioned yesterday? the README.md strikes me as a good place.
From memory these were:
- Docker log format
- Gradle docker image

Wynndow:
> Looks good - did you want to capture any of the additional thoughts / future work around some of the docker quirks mentioned yesterday? the README.md strikes me as a good place. From memory these were:
> 
> * Docker log format
> * Gradle docker image

That's a good shout - I actually solved the gradle docker image thing though. I've reordered the steps in the core's dockerfile so that gradle gets installed _before_ we copy over the lamda/libs/local-running directories. That way the caching of that step doesn't get invalidated when the code changes.

Will add something about the log format.

Wynndow:
@lwebb-iig I've just been digging into the logs thing and have pushed a change to format how core-back's logs appear. Interested in your thoughts.

lwebb-iig (approved):

PR #1246, opened 2023-10-05
PYIC-3609: Allow all options to be used

Wynndow:
### What changed

Allow all options to be used

### Why did it change

Updates the html and scripts to allow multiple CRIs to be disabled and multiple flags to be activated.

Updates the way the CRI and flag options are gathered by recursively checking the events. Setting a disabled or flag check in the journey map allows for events to be defined underneath it. These may have options not used at the top level else where.

Update to consider these sub events when resolving the target states. Previously we were only resolving the top level.

### Issue tracking

- [PYIC-3609](https://govukverify.atlassian.net/browse/PYIC-3609)

[PYIC-3609]: https://govukverify.atlassian.net/browse/PYIC-3609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/public/index.html
+++ b/journey-map/public/index.html
@@ -40,7 +40,7 @@
             form select {
                 flex: 0 0 30%;
             }
-            span {
+            #checkbox-controls {
                 display: flex;
                 flex-direction: row;
                 justify-content: space-between;
@@ -51,7 +51,7 @@
         <div id="header">
             <h1>IPV Core Journey Map</h1>
             <form id="configuration-form">
-                <span>
+                <span id="checkbox-controls">
                     <fieldset id="disabledInput">
                         <legend>Disabled CRI:</legend>
                     </fieldset>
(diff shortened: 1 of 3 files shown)


journey-map/public/index.html
```diff
@@ -91,23 +96,52 @@ <h1>IPV Core Journey Map</h1>
...
+                    featureFlagInput.appendChild(span)
+                });
+            } else {
+                featureFlagInput.append("N/A");
+            }
+
 
             // Render the journey map SVG
             const renderSvg = async (formData) => {
-                const options = Object.fromEntries(formData || []);
```
Wynndow, line 109: Stopped converting to an object here, as when you have multiple checkboxes selected (with the same `name` attribute), you lose all values except for the last one. The key in the object can only hold one value. The `FormData` object maintains an array of values so we know what was selected.

Joe-Edwards-GDS:
Thanks Chris! I've made a few changes but will merge back into my branch unless you'd like to take a look in isolation.

I hadn't thought about recursive feature flag/disabled definitions, and I think this should also work with nesting between the two as well (i.e. you can nest a disabled check inside a feature flag or vice-versa)

Wynndow:
Ah yeah, yours is way nicer! Merge away...

PR #1247, opened 2023-10-05
PYIC-2928 Added a new audit event IPV_CORE_CRI_RESOURCE_RETRIEVED

rpayal:
## Proposed changes
Add a new audit event IPV_CORE_CRI_RESOURCE_RETRIEVED when resource retrieved from CRI.

### What changed
Add a new audit event IPV_CORE_CRI_RESOURCE_RETRIEVED
- Change made in lambda -  retrieve-cri-credential

### Issue tracking

- [PYIC-2928](https://govukverify.atlassian.net/browse/PYIC-2928)

### Other considerations

[PYIC-2928]: https://govukverify.atlassian.net/browse/PYIC-2928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TalatAydinGDS (dismissed):
LGTM

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -198,6 +222,17 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
+            try {
+                sendCriResRetrievedAuditEvent(
+                        auditEventUser,
+                        credentialIssuerId,
+                        CriResourceRetrievedType.ERROR.getType());
+            } catch (SqsException ex) {
+                LogHelper.logErrorMessage(
+                        "Failed to send cri_response_retrieve audit event to SQS queue.",
+                        ex.getMessage());
+            }
```
Joe-Edwards-GDS: Should we be catching errors in the success cases as well - should we be blocking the user if the Audit event fails to send (I'm not sure!)

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -170,24 +175,43 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
-                        verifiableCredentialResponse.getVerifiableCredentials(),
+                        verifiableCredentials,
                         clientOAuthSessionItem,
                         ipvSessionItem);
+                if (!verifiableCredentials.isEmpty()) {
+                    sendCriResRetrievedAuditEvent(
+                            auditEventUser,
+                            credentialIssuerId,
+                            CriResourceRetrievedType.VC.getType());
+                }
```
Joe-Edwards-GDS: What happens if the VCs are empty - is it correct to drop the audit event entirely? We've still collected the resources available, there just weren't any 🤔 

I wonder if we should return a different CRI_RESOURCE_RETRIEVED type ('Empty')? (Might need checking with TxMA though)

Joe-Edwards-GDS: This ordering means that we will log this event _after_ the VC received event - is that what we expect?

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -198,6 +222,17 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
                 | CiPostMitigationsException e) {
```
Joe-Edwards-GDS: Is it possible to get any of these exceptions before the `AuditEventUser` is defined - and will that be a problem?

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
rpayal: Currently when VCs are empty, we r not even sending event IPV_VC_RECEIVED. 

@kerrr In case of non pending vc-status  with empty VCs, do we need to send IPV_CORE_CRI_RESOURCE_RETRIEVED event ?

rpayal: yes currently this new event sent after IPV_VC_RECEIVED. Will check this.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
rpayal: what u mean by errors in the success cases? In case of failure of this audit event we are logging and returning JOURNEY_ERROR.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -222,17 +222,6 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
-            try {
-                sendCriResRetrievedAuditEvent(
-                        auditEventUser,
-                        credentialIssuerId,
-                        CriResourceRetrievedType.ERROR.getType());
-            } catch (SqsException ex) {
-                LogHelper.logErrorMessage(
-                        "Failed to send cri_response_retrieve audit event to SQS queue.",
-                        ex.getMessage());
-            }
```
Joe-Edwards-GDS: This means we're no longer sending a CRI resource retrieved error event - do you know when we're supposed to be sending one?

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
```diff
@@ -168,50 +167,51 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
                         criOAuthSessionItem.getCriOAuthSessionId(),
                         ipvSessionItem);
-                sendCriResRetrievedAuditEvent(
-                        auditEventUser,
-                        credentialIssuerId,
-                        CriResourceRetrievedType.PENDING.getType());
             } else {
                 List<SignedJWT> verifiableCredentials =
                         verifiableCredentialResponse.getVerifiableCredentials();
+                if (!verifiableCredentials.isEmpty()) {
```
Joe-Edwards-GDS: I'm still unsure about this - if the VC bundle is empty, I think we've still fetched a CRI resource, it's just empty.

It would be good to find out where this requirement came from, to determine what to do - I've asked in the thread to see if Simon knows.

rpayal: Yes agree we got the response with empty vc's. Here we can send type as 'empty' / 'vc-empty'.

As we are sending this event - I think we should decide what and when to pass. I'd checked with TxMA side they also told its team to decide what they want to pass.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
rpayal: I think it should be when the CRI response return as vc status as error, which currently it doesn't.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
Joe-Edwards-GDS: Sorry that was quite a confusing comment!

I meant, if sending the audit event fails in one of the other cases (for a successful VC retrieval), should we catch the exception as well?

I guess we should just follow the same pattern as all the other audit events.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
Joe-Edwards-GDS: Ah great - in that case I think we _should_ send an event when it's empty, but with an empty type as you suggest :)

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
Joe-Edwards-GDS: (as per the other thread, I think we should send the event, but with type 'empty')

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
Joe-Edwards-GDS: Ok - if you don't think we'll need to send error events for now, then we should probably just remove it as a possible status (YAGNI).

rpayal: Done and updated TxMA document.

lambdas/retrieve-cri-credential/src/main/java/uk/gov/di/ipv/core/retrievecricredential/RetrieveCriCredentialHandler.java
rpayal: done

Joe-Edwards-GDS (approved):

PR #1259, opened 2023-10-12
PYIC-3658: Update IPV_F2F_CRI_VC_CONSUMED audit event

Wynndow:
**Need to update the TXmA spreadsheet and inform TICF of the change before merging.**

### What changed

Update `IPV_F2F_CRI_VC_CONSUMED` audit event

### Why did it change

The way we were adding name parts to the restricted section of the audit event was out of line with the rest of the programme. This changes it from `restricted.nameParts` to `restricted.name.nameParts`.

Worth noting that this is the only place that we include restricted data in an audit event. We don't send this with the normal VC retrieval method.

Also, we include the `docExpiryDate` attribute in the restricted data. If the VC contains neither passport or driving license data, this attribute is included but set to null. We should potentially just exclude this attribute if its null, however I'm unsure how these events are used so am favouring leaving it in. In reality we shouldn't be getting VCs that don't include passport or DL data so this may be a non issue.

Update:
Looking at other audit events that include the name and name part
attributes in the restricted section of the audit event
(`DCMAW_CRI_VC_ISSUED` and `IPV_KBV_CRI_REQUEST_SENT` for example), the name
parts are objects within an outer array. If this work is to bring how we
are sending this data in line with other audit events, we should be
consistant.

#### Audit event issued in dev account:
```
{
    "event_name": "IPV_F2F_CRI_VC_CONSUMED",
    "component_id": "https://dev-chrisw.01.dev.identity.account.gov.uk",
    "user": {
        "user_id": "urn:uuid:f4628423-cdc6-42e0-b4af-8ba8105dc56c",
        "session_id": null,
        "govuk_signin_journey_id": "unknown",
        "ip_address": null
    },
    "restricted": {
        "name": [
            {
                "nameParts": [
                    {
                        "type": "GivenName",
                        "value": "Kenneth"
                    },
                    {
                        "type": "FamilyName",
                        "value": "Decerqueira"
                    }
                ]
            }
        ],
        "docExpiryDate": "2030-01-01"
    },
    "timestamp": 1697106527,
    "event_timestamp_ms": 1697106527818
}
```

### Issue tracking

- [PYIC-3658](https://govukverify.atlassian.net/browse/PYIC-3658)

[PYIC-3658]: https://govukverify.atlassian.net/browse/PYIC-3658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/auditing/AuditRestrictedVc.java
```diff
@@ -1,29 +1,28 @@
 package uk.gov.di.ipv.core.processasynccricredential.auditing;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.JsonNode;
+import com.nimbusds.jose.shaded.json.JSONArray;
```
Joe-Edwards-GDS, line 4: Why are we using a different JSON library here - feels a little odd to mix? Can we use the jackson version, or just a standard Java list?

Wynndow: It's because nimbus is the lib we use for handling JWTs, and it uses these JSON types when digging into the objects to pull out various elements.
Previously we were doing this, and then converting them to a string, then converting them back to a Jackson data type  to include in the AuditRestrictedVc object.
This change means we can skip the serialize/deserialize step and just use the nimbus object we already have. Jackson can serialize it without issue when the while audit event is serialized.
We could use a Java list, but then I think we'd end up with `List<JSONObject>` which is just using another nimbus object. Unless we want to reintroduce the serialize/deserialize.

Joe-Edwards-GDS: Ok fair enough - I don't love having both libraries mixed, but mapping to the other type is probably a waste of time.

It also means our audit event structure is tightly coupled to the VC structure, but that's probably not such a massive issue (if anything it makes sense for them to match).

Joe-Edwards-GDS (approved):

PR #1260, opened 2023-10-12
3611: Add test for journey map event consistency

MikeCollingwood:
### What changed

- Test all CRI states have the same events handled
- Test all states for the same page have the same events handled
- Test events consistent entering and exiting nested journeys
- Update CRI states with Error targets for events not handled

### Why did it change

- To make interactions with CRIs more consistent
- To know nested journeys have events covered
- To catch inconsistent journey map behaviour

### Issue tracking

- [PYIC-3611](https://govukverify.atlassian.net/browse/PYIC-3611)

### Other considerations

[PYIC-3611]: https://govukverify.atlassian.net/browse/PYIC-3611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -6,14 +6,22 @@ END_JOURNEY:
 
 CRI_STATE:
   events:
+    next:
+      targetState: ERROR
+    not-found:
+      targetState: ERROR
+    fail-with-no-ci:
+      targetState: ERROR
     error:
       targetState: ERROR
     access-denied:
-      targetState: PYI_NO_MATCH
+      targetState: ERROR
+    enhanced-verification:
+      targetState: ERROR
     temporarily-unavailable:
-      targetState: PYI_NO_MATCH
+      targetState: ERROR
     end:
-      targetState: IPV_SUCCESS_PAGE
+      targetState: ERROR
     pyi-no-match:
       targetState: PYI_NO_MATCH
     pyi-kbv-fail:
@@ -137,16 +145,6 @@ CRI_F2F:
       targetState: F2F_HANDOFF_PAGE
     access-denied:
       targetState: PYI_ANOTHER_WAY
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    fail-with-no-ci:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
 
 F2F_HANDOFF_PAGE:
   response:
@@ -187,12 +185,6 @@ CRI_DCMAW:
       checkIfDisabled:
         f2f:
           targetState: MULTIPLE_DOC_CHECK_PAGE
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 MULTIPLE_DOC_CHECK_PAGE:
   response:
@@ -333,16 +325,6 @@ CRI_UK_PASSPORT_J2:
       checkIfDisabled:
         f2f:
           targetState: MULTIPLE_DOC_CHECK_PAGE
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    fail-with-no-ci:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
 
 ADDRESS_AND_FRAUD_J2:
   nestedJourney: ADDRESS_AND_FRAUD
@@ -378,14 +360,6 @@ CRI_KBV_J2:
           targetState: PYI_CRI_ESCAPE_NO_F2F
     enhanced-verification:
       targetState: MITIGATION_02_OPTIONS
-    next:
-      targetState: ERROR
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 # Driving licence journey (J3)
 CRI_DRIVING_LICENCE_J3:
@@ -401,16 +375,6 @@ CRI_DRIVING_LICENCE_J3:
       checkIfDisabled:
         f2f:
           targetState: MULTIPLE_DOC_CHECK_PAGE
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    fail-with-no-ci:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
 
 ADDRESS_AND_FRAUD_J3:
   nestedJourney: ADDRESS_AND_FRAUD
@@ -418,7 +382,7 @@ ADDRESS_AND_FRAUD_J3:
     next:
       targetState: CHECK_FRAUD_SCORE_J3
     end:
-      targetState: END
+      targetState: ERROR
 
 CHECK_FRAUD_SCORE_J3:
   response:
@@ -458,12 +422,6 @@ CRI_KBV_J3:
       targetState: PYI_NO_MATCH
     enhanced-verification:
       targetState: MITIGATION_02_OPTIONS
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 # F2F journey (J4)
 CRI_CLAIMED_IDENTITY_J4:
@@ -474,18 +432,6 @@ CRI_CLAIMED_IDENTITY_J4:
   events:
     next:
       targetState: ADDRESS_AND_FRAUD_J4
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    fail-with-no-ci:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 ADDRESS_AND_FRAUD_J4:
   nestedJourney: ADDRESS_AND_FRAUD
@@ -508,14 +454,6 @@ CRI_DCMAW_J5:
       targetState: PYI_ANOTHER_WAY
     fail-with-no-ci:
       targetState: PYI_ANOTHER_WAY
-    temporarily-unavailable:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 POST_DCMAW_SUCCESS_PAGE_J5:
   response:
@@ -536,16 +474,6 @@ CRI_NINO_J6:
       targetState: CRI_HMRC_KBV_J6
     fail-with-no-ci:
       targetState: PRE_KBV_TRANSITION_PAGE_J2
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    end:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 CRI_HMRC_KBV_J6:
   response:
@@ -559,14 +487,6 @@ CRI_HMRC_KBV_J6:
       targetState: PRE_KBV_TRANSITION_PAGE_J3
     next:
       targetState: PYI_NO_MATCH
-    not-found:
-      targetState: ERROR
-    temporarily-unavailable:
-      targetState: ERROR
-    enhanced-verification:
-      targetState: ERROR
-    access-denied:
-      targetState: ERROR
 
 # Mitigation journey (01)
 MITIGATION_01:
@@ -605,8 +525,6 @@ MITIGATION_01_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
     enhanced-verification:
       targetState: PYI_ANOTHER_WAY
-    end:
-      targetState: ERROR
 
 # Mitigation journey (02)
 MITIGATION_02_OPTIONS:
@@ -645,7 +563,3 @@ MITIGATION_02_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
     enhanced-verification:
       targetState: PYI_ANOTHER_WAY
-    next:
-      targetState: ERROR
-    end:
-      targetState: ERROR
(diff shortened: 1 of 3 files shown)


lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+        List<StateAndEvents> criStates = new ArrayList<>();
+        Set<String> criStateEvents = new HashSet<>();
+        findCriStatesAndEvents(stateMachine, criStates, criStateEvents);
+
+        for (StateAndEvents stateAndEvents : criStates) {
+            assertEquals(
+                    criStateEvents,
+                    stateAndEvents.events,
+                    String.format("%s doesn't handle all CRI state events", stateAndEvents.state));
```
Joe-Edwards-GDS: Does the message describe which events are missing? Feels like that would be really useful if possible (but also not worth spending ages on)

(Same with the test below)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+    void shouldMatchNestedJourneyEntryEvents(IpvJourneyTypes journeyType) throws IOException {
+        StateMachineInitializer stateMachineInitialiser = new StateMachineInitializer(journeyType);
+
+        Map<String, State> stateMachine = stateMachineInitialiser.initialize();
+        Set<String> stateMachineKeys = stateMachine.keySet();
+
+        for (String key : stateMachineKeys) {
+            State state = stateMachine.get(key);
+
+            if (state instanceof NestedJourneyInvokeState) {
```
Joe-Edwards-GDS: I think we might be able to use some more fancy Java features here to avoid the explicit cast:

```
if (state instanceof NestedJourneyInvokeState nestedState) {
  Set<String> entryEvents = nestedState.getExitEvents().keySet();
  ...
}
```
I'm not super familiar with the syntax though!

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+                            if (event instanceof BasicEvent) {
+                                if (((BasicEvent) event).getTargetState() == key) {
+                                    assertTrue(entryEvents.contains(eventName));
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            ;
```
Joe-Edwards-GDS: Weird dangling semis?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+            State state = stateMachine.get(key);
+
+            if (state instanceof NestedJourneyInvokeState) {
+                Set<String> entryEvents =
+                        ((NestedJourneyInvokeState) state).getExitEvents().keySet();
+
+                for (String comparingStateKey : stateMachineKeys) {
+                    State comparingState = stateMachine.get(comparingStateKey);
+
+                    if (comparingState instanceof BasicState) {
```
Joe-Edwards-GDS: I think this does assume that we only ever link to nested states from basic states(?) - which is fine for now, but something to be aware of!

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
Joe-Edwards-GDS: (same applies elsewhere)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+    @ParameterizedTest
+    @EnumSource
+    void shouldMatchNestedJourneyEntryEvents(IpvJourneyTypes journeyType) throws IOException {
+        StateMachineInitializer stateMachineInitialiser = new StateMachineInitializer(journeyType);
+
+        Map<String, State> stateMachine = stateMachineInitialiser.initialize();
+        Set<String> stateMachineKeys = stateMachine.keySet();
+
+        for (String key : stateMachineKeys) {
+            State state = stateMachine.get(key);
```
Joe-Edwards-GDS: We could probably tighten up a few of these names to make it easier to understand what's going on here?

E.g. here we're looking for the `targetNestedState`, and further down we're looking for (any) `sourceBasicState` - maybe easier in combination with the pattern matching I describe below.

Joe-Edwards-GDS: (rationale being that we're verifying the events line up from 'source' to 'target')

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+                    if (comparingState instanceof BasicState) {
+                        Map<String, Event> eventMap = ((BasicState) comparingState).getEvents();
+
+                        for (Map.Entry<String, Event> entry : eventMap.entrySet()) {
+                            String eventName = entry.getKey();
+                            Event event = entry.getValue();
+
+                            if (event instanceof BasicEvent) {
+                                if (((BasicEvent) event).getTargetState() == key) {
+                                    assertTrue(entryEvents.contains(eventName));
```
Joe-Edwards-GDS: (Does this give a useful failure message?)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -0,0 +1,232 @@
...
+                        ((NestedJourneyInvokeState) state)
+                                .getNestedJourneyDefinition()
+                                .getNestedJourneyStates();
+                findPageSpecificStatesAndEvents(nestedStateMachine, pageStateMap);
+            }
+        }
+    }
+}
+
+class StateAndEvents {
```
Joe-Edwards-GDS: More modern Java, can we just make this a record class?

```
record StateAndEvents(String state, Set<String> events) {}
```

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -137,6 +137,16 @@ CRI_F2F:
...
+    not-found:
+      targetState: ERROR
+    temporarily-unavailable:
+      targetState: ERROR
+    fail-with-no-ci:
+      targetState: ERROR
+    end:
+      targetState: ERROR
+    enhanced-verification:
+      targetState: ERROR
```
Joe-Edwards-GDS: Interestingly, once we add F2F CIMIT this should become `F2F_HANDOFF_PAGE` (because the F2F is a valid enhanced verification journey).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -137,6 +137,16 @@ CRI_F2F:
       targetState: F2F_HANDOFF_PAGE
     access-denied:
       targetState: PYI_ANOTHER_WAY
+    not-found:
+      targetState: ERROR
+    temporarily-unavailable:
```
Joe-Edwards-GDS: I'm curious how you found this - I think there is default behaviour here from the parent state which should be enough to pass the test(?)

If we're actively changing behaviour, I think we should probably separate that into a different PR - to keep this one as 'test only'.

Joe-Edwards-GDS: The 'end' case is particularly interesting - and we probably want to talk to some other people about whether an unexpected 'end' should just complete the journey, continue the journey as-is, or be an error (acknowledging that the user can probably just restart the journey and get a valid profile)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -21,3 +33,13 @@ ADDRESS_AND_FRAUD:
...
+        not-found:
+          targetState: ERROR
+        temporarily-unavailable:
+          targetState: ERROR
+        fail-with-no-ci:
+          targetState: ERROR
+        enhanced-verification:
+          targetState: ERROR
+        access-denied:
+          targetState: ERROR
```
Wynndow: Part of me is wondering if rather than having all this repetition on every state, we could define one parent state that has all events we know about with a target state of error. And then all other states inherit it and override just the events we expect to receive. Might make the expected behaviour for each state more obvious while still handling events that *could* be thrown.

Joe-Edwards-GDS: Yeah - that could be a reasonable compromise. Having the repetition like this (and the test to validate) forces you to think about every state for every CRI, but perhaps it's unnecessarily tedious!

(It's still a lot better than having ad-hoc events all over the place)

Joe-Edwards-GDS: This is probably where I'd be keen for @blakeyp 's opinion, since HILUX might be at the receiving end of most changes like this :)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
MikeCollingwood: The assertion displays expected and actual, so the tester can understand what's needed but I display the difference easily so will do!

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
MikeCollingwood: I'm not familiar with the syntax either so I'll look into it because explicit casting didn't sit well - thank you!

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
MikeCollingwood: Hmm, maybe I'll check it's not an InvokeNestedJourneyState instead! At least it'll break if a new state type comes in (I think all states are Basic/ Invoke nested ...)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
MikeCollingwood: Good point I need to add it in

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
blakeyp: I like @Wynndow's suggestion of having the defaults on a parent state and each state just overrides any it needs to, will make it more readable and is consistent with how we've been using `CRI_STATE` as a parent

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: In the case where we're using the parent CRI_STATE to set the target state of the end event, I think it should be ERROR, since we don't have a consistent action on end

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
MikeCollingwood: I've moved additional events into the parent: CRI_STATE, to complete the set

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: Could it expected where it relies on the default of going to IPV_SUCCESS_PAGE?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -6,10 +6,18 @@ END_JOURNEY:
...
+    not-found:
+      targetState: ERROR
+    fail-with-no-ci:
+      targetState: ERROR
     error:
       targetState: ERROR
     access-denied:
       targetState: PYI_NO_MATCH
+    enhanced-verification:
+      targetState: ERROR
```
Joe-Edwards-GDS, line 20: I think we might want to change some of these defaults - e.g. so the failure cases route to `PYI_NO_MATCH` or `PYI_ANOTHER_WAY`, but in the spirit of separating the tests from functional changes, I'll raise a new ticket.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: Having talked to Patrick, since it shouldn't be reached we probably want to indicate this as an error so we can react to it. I'll update it to ERROR instead of the default IPV_SUCCESS_PAGE

Joe-Edwards-GDS (approved):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: Thank you - I think to avoid potentially breaking a flow, it'd be better done in a dedicated bit of work

PR #1262, opened 2023-10-13
PYIC-3607: Routing for same-session F2F CIMIT

lwebb-iig:
Implement the F2F routing for CIMIT behind a feature flag.

### What changed

Added two new states and respective pages:
- MITIGATION_02_OPTIONS_WITH_F2F (pyi-ipv-suggest-other-options)
- MITIGATION_02_SUGGEST_F2F (pyi-ipv-suggest-f2f)

Added new routing:
- If `f2fCimit` feature flag is enabled & F2F CRI is enabled route KBV mitigations to MITIGATION_02_OPTIONS_WITH_F2F
- Routes from new `pyi-ipv-suggest-other-options` page to dcmaw/f2f/another way
- If `f2fCimit` feature flag is enabled & F2F CRI is enabled route DCMAW mitigation to suggest f2f
- Failure/abandon/illegible mapping to suggest f2f page

Maintains:
- Default to current mitigation routing when `f2fCimit` or F2F CRI flags are disabled

Current mitigation journey
image

Mitigation journey with f2fCimit flag enabled
image

Mitigation journey with F2F cri disabled
image

### Why did it change

Enables routing for same-session F2F mitigations

### Issue tracking

- [PYIC-3607](https://govukverify.atlassian.net/browse/PYIC-3607)

### Other considerations

[PYIC-3607]: https://govukverify.atlassian.net/browse/PYIC-3607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -137,6 +137,8 @@ CRI_F2F:
       targetState: F2F_HANDOFF_PAGE
     access-denied:
       targetState: PYI_ANOTHER_WAY
+    enhanced-verification:
+      targetState: F2F_HANDOFF_PAGE
 
 F2F_HANDOFF_PAGE:
   response:
@@ -350,6 +352,9 @@ CRI_KBV_J2:
           targetState: PYI_CRI_ESCAPE_NO_F2F
     enhanced-verification:
       targetState: MITIGATION_02_OPTIONS
+      checkFeatureFlag:
+        f2fCimit:
+          targetState: MITIGATION_02_OPTIONS_WITH_F2F
 
 # Driving licence journey (J3)
 CRI_DRIVING_LICENCE_J3:
@@ -410,6 +415,10 @@ CRI_KBV_J3:
       targetState: PYI_NO_MATCH
     enhanced-verification:
       targetState: MITIGATION_02_OPTIONS
+      checkFeatureFlag:
+        f2fCimit:
+          targetState: MITIGATION_02_OPTIONS_WITH_F2F
+
 
 # F2F journey (J4)
 CRI_CLAIMED_IDENTITY_J4:
@@ -548,4 +557,24 @@ MITIGATION_02_CRI_DCMAW:
     fail-with-no-ci:
       targetState: PYI_ANOTHER_WAY
     enhanced-verification:
+      targetState: MITIGATION_02_SUGGEST_F2F
+
+MITIGATION_02_OPTIONS_WITH_F2F:
+  response:
+    type: page
+    pageId: pyi-suggest-other-options
+  events:
+    f2f:
+      targetState: CRI_F2F
+    dcmaw:
+      targetState: MITIGATION_02_CRI_ELIGIBILITY
+    end:
       targetState: PYI_ANOTHER_WAY
+
+MITIGATION_02_SUGGEST_F2F:
+  response:
+    type: page
+    pageId: pyi-ipv-suggest-f2f
+  events:
+    next:
+      targetState: CRI_F2F
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -585,7 +585,7 @@ MITIGATION_02_OPTIONS_WITH_F2F:
 MITIGATION_02_SUGGEST_F2F:
   response:
     type: page
-    pageId: pyi-ipv-suggest-f2f
+    pageId: pyi-suggest-f2f
   events:
     next:
       targetState: CRI_F2F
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -350,6 +352,12 @@ CRI_KBV_J2:
           targetState: PYI_CRI_ESCAPE_NO_F2F
     enhanced-verification:
       targetState: MITIGATION_02_OPTIONS
+      checkFeatureFlag:
+        f2fCimit:
```
Joe-Edwards-GDS, line 356: Do we need to add this feature flag to a list somewhere, or will it Just Work?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -549,3 +563,29 @@ MITIGATION_02_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
     enhanced-verification:
```
Joe-Edwards-GDS, line 588: Do you know which of these events corresponds to the 'ineligible/abandon' case in the UCD diagram?

In the main (non-mitigation) journey, we route all of these into the fallback passport/driving licence/f2f route - I wonder if we should be doing something similar here. I'm worried that it might need to be more than just the `enhanced-verification` case, but not 100% sure when each occurs.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
lwebb-iig: We add the feature flag via the common-infra repo https://github.com/govuk-one-login/ipv-core-common-infra/pull/733/files - the state machine will look for this as a parameter and will treat it as false if not present

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
lwebb-iig: I had assumed this would present itself under the `not-found` event but maybe that's worth double checking with UCD folks - this assumption is based on the fact that we already have these paths link to the same page in the figma diagram at the top of this https://www.figma.com/file/nglBHtbzJYEeST43Iu1jW3/PYI-Screen-Designs-(MVP-%26-Post-MVP)?type=design&node-id=3915-39707&mode=design (Same session app mitigation only)

Joe-Edwards-GDS: Yeah, it feels like we need to understand
- What causes DCMAW to return the different codes (e.g. not-found vs. access-denied)
- Which of these UCD want to route to F2F mitigation vs. 'no match' vs. 'another way'

I'm afraid I'm not 100% sure of the answer to either! - Are you happy to try and find this out, or would you like me to reach out to them. We could also try Simon first 🤷 

cc @rpayal

Joe-Edwards-GDS (dismissed):
As discussed - I think it's ok to merge now, but might be worth creating a subtask to track down the exact routing requirements for the DCMAW error events.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -549,3 +563,29 @@ MITIGATION_02_CRI_DCMAW:
       targetState: PYI_ANOTHER_WAY
```
rpayal, line 581: In case of error from DCMAW during mitigation the flow should take user to MITIGATION_02_SUGGEST_F2F page. So you need to handle this for all other error events from DCMAW.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: Agreed in discussion:
- pyi-no-match should route to 'no match' (handled in the CRI parent state)
- all the other error cases should route to 'suggest F2F' (unless F2F is disabled, in which case we go to 'another way')

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -543,9 +557,53 @@ MITIGATION_02_CRI_DCMAW:
       targetState: PYI_NO_MATCH
```
rpayal, line 563: Should DCMAW not found also lead to F2F ?

Joe-Edwards-GDS (dismissed):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
lwebb-iig: I don't think so no - I think we said that `not-found` represents this part of the figma diagram, therefore we should route to PYI_NO_MATCH
image

Joe-Edwards-GDS (approved):

rpayal (approved):

PR #1264, opened 2023-10-16
PYIC-3639: clear users identity if reset identity feature flag is set

shivanshuit914:
## Proposed changes

- Added a feature to reset users identity if it is already preset and reset identity featureflag is enabled.

### What changed

- Claimed identity lambda to reset users identity.

### Why did it change

- To allow QA to reset identity for the user so they can use same users multiple times.

Related PR: https://github.com/govuk-one-login/ipv-core-common-infra/pull/727

### Issue tracking

- [PYIC-3639](https://govukverify.atlassian.net/browse/PYIC-3639)

### Other considerations

[PYIC-3639]: https://govukverify.atlassian.net/browse/PYIC-3639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -50,6 +50,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.StringJoiner;
 
+import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.RESET_IDENTITY;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.F2F_CRI;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CLAIM;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_EVIDENCE;
@@ -67,7 +68,6 @@ import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_NEXT_PA
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_PENDING_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_RESET_IDENTITY_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_REUSE_PATH;
-import static uk.gov.di.ipv.core.library.service.ConfigService.featureFlags.RESET_IDENTITY;
 
 /** Check Existing Identity response Lambda */
 public class CheckExistingIdentityHandler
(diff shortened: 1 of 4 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -198,6 +198,11 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
                 return JOURNEY_FAIL;
             }
 
+            if (matchedProfile.isPresent() && configService.enabled("resetIdentity")) {
```
blakeyp: Can we make an enum for this? think can add it to `CoreFeatureFlag`

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -216,6 +217,35 @@ void shouldReturnJourneyReuseResponseIfScoresSatisfyM1AGpg45Profile() throws Exc
...
+        when(configService.enabled("resetIdentity")).thenReturn(true);
+
+        JourneyResponse journeyResponse =
+                toResponseClass(
+                        checkExistingIdentityHandler.handleRequest(event, context),
+                        JourneyResponse.class);
+
+        assertEquals(JOURNEY_RESET_IDENTITY, journeyResponse);
+
+        verify(userIdentityService, never()).deleteVcStoreItems(TEST_USER_ID);
```
blakeyp: ~~Hmm shouldn't this be verifying that we _do_ delete the VC store items?~~
Realised the reset-identity lambda now does this, in which case I think this assertion isn't needed and is probably misleading? (same goes for other uses of it in these tests)

shivanshuit914: Yes makes sense. thanks

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -61,6 +61,20 @@ public class ConfigService {
     private final SecretsProvider secretsProvider;
     private final ObjectMapper objectMapper = new ObjectMapper();
 
+    public enum featureFlags {
+        RESET_IDENTITY("resetIdentity");
```
blakeyp: We have an enum already for feature flags - in `library/config/CoreFeatureFlag.java` so I think we can just add this as a value there?

shivanshuit914: Thanks, I will add there.

blakeyp (approved):

PR #1268, opened 2023-10-18
PYIC-3610: publish journey map to github pages

Joe-Edwards-GDS:
Adds GitHub workflow to publish the journey map using GitHub pages.

The workflow is heavily based off the default GH pages publish workflow, with only a couple of small modifications:
- Limiting the paths to the journey map files
- Copying the yaml files into the publish directory

Visible at https://govuk-one-login.github.io/ipv-core-back/


lwebb-iig (dismissed):
LGTM

.github/workflows/journey-map.yml
```diff
@@ -0,0 +1,47 @@
...
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
Wynndow, line 32: I'm not super good with GHA. Did you have to create this environment, or does it already exist due to it being github-pages? 
The value for the URL comes from the deployment step, and seems to not be configurable. I guess I'm wondering why we need this environment at all? What would happen if we didn't have it? Would the URL output by deploy work? Would the deploy fail for some reason?

.github/workflows/journey-map.yml
```diff
@@ -0,0 +1,47 @@
...
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
```
Wynndow, line 26: By defining the concurrency at the workflow level, does every run belong to the "pages" group by default? I was sort of expecting to see this block defined, and then the job reference that group in some way. The docs don't suggest that's how it works though.

Joe-Edwards-GDS: It looks like you can specify concurrency at either the job or workflow level (and this is the latter) - so only one run of this workflow can happen at a time (and if we ever defined another 'pages' workflow, that wouldn't run concurrently either).

I will admit that I didn't actually write this - it's part of the default github pages workflow!

Wynndow: Ah gotcha. That makes sense now - I hadn't considered the concurrency of separate workflows.

.github/workflows/journey-map.yml
Joe-Edwards-GDS: I think this is referencing a github 'environment', as opposed to environment variables as such.

It links to the settings, which allows us to place protection on who can deploy to it etc., as well as secrets/variables (https://github.com/govuk-one-login/ipv-core-back/settings/environments/1454443961/edit) - I'm not sure what exactly would happen if we omitted it!

Joe-Edwards-GDS: I think `github-pages` is created automatically when you configure github pages - at the very least I didn't create it manually.

Wynndow: Ah ok, that makes sense. There's the `actions/configure-pages` step which I suppose did the initial configuration. I can see with the API that the env was created at `"created_at": "2023-10-12T15:42:40Z"`, which I'm guessing was the first run of the workflow maybe.

Wynndow (approved):

vam-gds (approved):
lgtm

PR #1270, opened 2023-10-18
PYIC-3608 Routing for separate-session F2F CIMIT

rpayal:
## Proposed changes
Update journey map for routing of separate session CI mitigation F2F.
### What changed
Updated journey map

### Issue tracking

- [PYIC-3608](https://govukverify.atlassian.net/browse/PYIC-3608)

### Other considerations

[PYIC-3608]: https://govukverify.atlassian.net/browse/PYIC-3608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -527,12 +527,6 @@ MITIGATION_01_CRI_DCMAW:
       targetState: POST_DCMAW_SUCCESS_PAGE
     not-found:
       targetState: PYI_NO_MATCH
-      checkFeatureFlag:
-        f2fCimit:
-          targetState: MITIGATION_01_SUGGEST_F2F
-          checkIfDisabled:
-            f2f:
-              targetState: PYI_NO_MATCH
     access-denied:
       targetState: PYI_ANOTHER_WAY
       checkFeatureFlag:
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -145,6 +145,8 @@ CRI_F2F:
       targetState: F2F_HANDOFF_PAGE
     access-denied:
       targetState: PYI_ANOTHER_WAY
+    enhanced-verification:
+      targetState: F2F_HANDOFF_PAGE
 
 F2F_HANDOFF_PAGE:
   response:
@@ -432,6 +434,8 @@ CRI_CLAIMED_IDENTITY_J4:
   events:
     next:
       targetState: ADDRESS_AND_FRAUD_J4
+    enhanced-verification:
+      targetState: ADDRESS_AND_FRAUD_J4
 
 ADDRESS_AND_FRAUD_J4:
   nestedJourney: ADDRESS_AND_FRAUD
@@ -506,6 +510,12 @@ MITIGATION_01_IDENTITY_START_PAGE:
       targetState: MITIGATION_01_CRI_DCMAW
     end:
       targetState: PYI_ANOTHER_WAY
+      checkFeatureFlag:
+        f2fCimit:
+          targetState: MITIGATION_01_F2F_START_PAGE
+          checkIfDisabled:
+            f2f:
+              targetState: PYI_ANOTHER_WAY
 
 MITIGATION_01_CRI_DCMAW:
   response:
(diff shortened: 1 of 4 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -517,14 +525,62 @@ MITIGATION_01_CRI_DCMAW:
...
+            f2f:
+              targetState: PYI_ANOTHER_WAY
     enhanced-verification:
       targetState: PYI_ANOTHER_WAY
+      checkFeatureFlag:
+        f2fCimit:
+          targetState: MITIGATION_01_SUGGEST_F2F
+          checkIfDisabled:
+            f2f:
+              targetState: PYI_ANOTHER_WAY
```
Joe-Edwards-GDS, line 561: We should make sure that these match the routing from the same-session mitigation journey, and that the right ones get routed to F2F - see https://github.com/govuk-one-login/ipv-core-back/pull/1262/files#r1363817667

cc @lwebb-iig

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -21,3 +25,5 @@ ADDRESS_AND_FRAUD:
           exitEventToEmit: end
         next:
           exitEventToEmit: next
+        enhanced-verification:
+          exitEventToEmit: next
```
Joe-Edwards-GDS, line 29: This is an interesting one - I think I'd feel slightly more comfortable if this emitted `enhanced-verification` as an exit event...

This will mean mapping `enhanced-verification` from each of the address + fraud steps, but avoids any risk that we ignore the V03 in cases where the address + fraud check happens last.

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -202,15 +203,17 @@ && doesEvidenceContainEvidenceType(evidenceItem, evidenceType)) {
...
-        for (RequiredGpg45ScoresDto gpg45Score : requiredGpg45Scores) {
-            Gpg45ScoresDto requiredScores = gpg45Score.getRequiredScores();
-            if (requiredScores.getFraud() > 0 || requiredScores.getActivity() > 0) {
-                continue;
-            }
-            List<EvidenceDto> evidences = requiredScores.getEvidences();
-            if (!evidences.isEmpty()
-                    && (strengthScore == 0 || evidences.get(0).getStrength() < strengthScore)) {
-                strengthScore = evidences.get(0).getStrength();
+        if (!CollectionUtils.isNullOrEmpty(requiredGpg45Scores)) {
```
Joe-Edwards-GDS, line 206: This seems a bit suspicious - when will this be null or empty?

rpayal: During test come across this NPE ...so applied this change.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: So here in case of event /enhanced-verification it get routed to F2F (MITIGATION_01_SUGGEST_F2F) but if F2F is disabled then will route to PYI_ANOTHER_WAY.

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
Joe-Edwards-GDS: Ok - do you know how that happened? If we're finding that the required scores are `null`, it suggests we're not calculating the diff as expected. We should generate and store this diff in the evaluate GPG45 lambda, but maybe we're skipping that step because of the CI (i.e. it's failing early from CI scoring)...

In any case, I don't think we should be sending a strength score of 0 to F2F - that doesn't sound right.

@MikeCollingwood  - this might be even stronger motivation to move the calculation here. It looks like we're skipping the evaluate GPG45 step when on a mitigation journey.

Joe-Edwards-GDS: It might be that we want to leave this in for now, but then I'd like a link to @MikeCollingwood 's ticket - PYIC-3753, so we don't forget.

e.g.
```
TODO: PYIC-3753 This check is temporarily required until we move the calculation here
```

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: I think we've agreed the way forward after this discussion this morning :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
Joe-Edwards-GDS: Now https://govukverify.atlassian.net/browse/PYIC-3780

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
Joe-Edwards-GDS: Now https://govukverify.atlassian.net/browse/PYIC-3779

Joe-Edwards-GDS (approved):

PR #1274, opened 2023-10-23
Upgrade Typescript, Prettier and ESLint

Joe-Edwards-GDS:
Upgrade Typescript, Prettier and ESLint dependencies - addressing Dependabot PRs, but fixing lint rules that have changed.

### DIFF
--- a/lambdas/delete-user-data/src/delete-data.ts
+++ b/lambdas/delete-user-data/src/delete-data.ts
@@ -7,7 +7,7 @@ import { VCItemKey } from "./types";
 const ddbDocClient = DynamoDBDocument.from(
   config.isLocalDev
     ? new DynamoDBClient({ endpoint: config.localDynamoDbEndpoint })
-    : new DynamoDBClient({ region: "eu-west-2" })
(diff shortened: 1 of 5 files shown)


lambdas/delete-user-data/src/delete-data.ts
```diff
@@ -7,7 +7,7 @@ import { VCItemKey } from "./types";
 const ddbDocClient = DynamoDBDocument.from(
   config.isLocalDev
     ? new DynamoDBClient({ endpoint: config.localDynamoDbEndpoint })
-    : new DynamoDBClient({ region: "eu-west-2" })
+    : new DynamoDBClient({ region: "eu-west-2" }),
```
Joe-Edwards-GDS, line 10: This default has changed in Prettier 3, and is definitely a matter of taste: https://prettier.io/docs/en/options.html#trailing-commas

We could also revert back to the original value with a `.prettierrc`

MikeCollingwood (approved):
LGTM! I can see you've upgraded the packages. I think enforcing trailing commas is fine.

PR #1280, opened 2023-10-24
PYIC-3753: Update requested evidence logic

MikeCollingwood:
### What changed

- Moved all logic for determining required evidence to build-cri-oauth-request
- Modify logic to enable multiple evidence to be considerred

### Why did it change

- To minimise & delay times Gpg45Scores are calculated
- To fix logic for multiple evidence
- To simplify logic around it

### Issue tracking

- [PYIC-3753](https://govukverify.atlassian.net/browse/PYIC-3753)

[PYIC-3753]: https://govukverify.atlassian.net/browse/PYIC-3753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -88,7 +86,12 @@ public int getVerification() {
     }
 
     public List<Evidence> getEvidences() {
-        return evidences;
+        return evidences.stream()
+                .sorted(
+                        Comparator.comparingInt(Gpg45Scores.Evidence::getStrength)
+                                .thenComparingInt(Gpg45Scores.Evidence::getValidity)
+                                .reversed())
+                .collect(Collectors.toList());
```
Joe-Edwards-GDS: It looks like one of the constructors already sorts the evidence - could we make all the other constructors use that one to avoid having to sort it every time you read this field?

Joe-Edwards-GDS: (it's pretty suspicious that this isn't already the case!)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -200,20 +197,34 @@ && doesEvidenceContainEvidenceType(evidenceItem, evidenceType)) {
...
-                continue;
-            }
-            List<EvidenceDto> evidences = requiredScores.getEvidences();
-            if (!evidences.isEmpty()
-                    && (strengthScore == 0 || evidences.get(0).getStrength() < strengthScore)) {
-                strengthScore = evidences.get(0).getStrength();
-            }
-        }
-        return strengthScore;
+    public List<List<Gpg45Scores.Evidence>> calculateEvidencesRequiredToMeetAProfile(
```
Joe-Edwards-GDS: I think it might worth extending this for completeness, since it shouldn't be too hard to do and makes it a lot more reusable - at the moment the signature of this method is a bit weird.

I also wonder if this would be more convenient as an instance method on `GPG45Scores`? i.e. so that it becomes:

```
public List<GPG45Scores> calculateRequiredScoresToMeetAProfile(List<GPG45Profile> profiles) { .. }

// In F2F call
var requiredScores = gpg45Scores.calculateRequiredScoresToMeetAProfile(ACCEPTED_PROFILES);
```

I don't think the implementation would need to change much - we'd just need to add the activity/fraud/verification scores to this block?

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -275,17 +287,17 @@ private JWEObject signEncryptJar(
...
             String criId)
-            throws HttpResponseExceptionWithErrorBody, ParseException, JOSEException {
+            throws HttpResponseExceptionWithErrorBody, ParseException, JOSEException,
+                    UnknownEvidenceTypeException {
+
+        List<SignedJWT> credentials = getSignedJWTs(userId);
+
         SharedClaimsResponse sharedClaimsResponse =
-                getSharedAttributesForUser(ipvSessionItem, userId, criId);
+                getSharedAttributesForUser(ipvSessionItem, credentials, criId);
```
Joe-Edwards-GDS, line 301: This will definitely conflict with Rohit's PR - you probably need to rebase after he merges

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -302,6 +314,44 @@ private JWEObject signEncryptJar(
...
+    private EvidenceRequest getEvidenceRequestForF2F(List<SignedJWT> credentials)
+            throws UnknownEvidenceTypeException, ParseException {
+        Gpg45Scores gpg45Scores = gpg45ProfileEvaluator.buildScore(credentials);
+        List<List<Gpg45Scores.Evidence>> requiredEvidences =
+                gpg45ProfileEvaluator.calculateEvidencesRequiredToMeetAProfile(
+                        gpg45Scores, ACCEPTED_PROFILES);
+
+        if (requiredEvidences == null) {
+            return null;
+        }
```
Joe-Edwards-GDS: When is this null?

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
Joe-Edwards-GDS: In any case I think you have to check activity/fraud/verification scores _somewhere_ - at the moment this isn't actually going to work.

In particular, the existing behaviour is deciding whether you need:
- strength 4 (to match M1A with a fraud score of 1)
- strength 3 (to match M1B with a fraud score of 2)

Since you're ignoring the fraud score, we'll always say that strength 3 is good enough - but it isn't if you only have a fraud score of 1!

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandlerTest.java
```diff
@@ -23,23 +23,16 @@
...
 import uk.gov.di.ipv.core.library.gpg45.exception.UnknownEvidenceTypeException;
 import uk.gov.di.ipv.core.library.helpers.SecureTokenHelper;
 import uk.gov.di.ipv.core.library.persistence.item.ClientOAuthSessionItem;
 import uk.gov.di.ipv.core.library.persistence.item.IpvSessionItem;
-import uk.gov.di.ipv.core.library.service.AuditService;
-import uk.gov.di.ipv.core.library.service.ClientOAuthSessionDetailsService;
-import uk.gov.di.ipv.core.library.service.ConfigService;
-import uk.gov.di.ipv.core.library.service.IpvSessionService;
-import uk.gov.di.ipv.core.library.service.UserIdentityService;
+import uk.gov.di.ipv.core.library.service.*;
```
Joe-Edwards-GDS: Doesn't matter so much in tests, but it looks like we generally avoid wildcard imports - and I think we should probably continue that practice. You might need to configure IntelliJ if it wasn't a deliberate change!

Means we can be precise about what's in scope, and avoid any risk of accidentally breaking something with a name clash.

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandlerTest.java
```diff
@@ -52,11 +55,7 @@
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.text.ParseException;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
Joe-Edwards-GDS: ditto

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -302,6 +314,44 @@ private JWEObject signEncryptJar(
...
+
+        OptionalInt minViableStrengthOpt =
+                requiredEvidences.stream()
+                        .filter(subList -> subList.size() == 1)
```
Joe-Edwards-GDS: (I think you'll also need to check we meet the fraud/activity/verification scores here, once we have them available)

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -302,6 +314,44 @@ private JWEObject signEncryptJar(
...
+                        .map(subList -> subList.get(0))
+                        .mapToInt(Gpg45Scores.Evidence::getStrength)
+                        .min();
+
+        return minViableStrengthOpt.isPresent()
+                ? new EvidenceRequest(minViableStrengthOpt.getAsInt())
+                : null;
```
Joe-Edwards-GDS: I think this is fine for now, since it replicates existing behaviour, but feels like this should actually be an error - it means we're about to send them to F2F but it's not possible for them to obtain a profile from it(!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
MikeCollingwood: That makes sense, and we assume this remains immutable

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
MikeCollingwood: Ah, this was a mistake. You're right it can't be null

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Joe-Edwards-GDS: Maybe we should add a warning log, which we can upgrade into an error if we confirm that it doesn't happen in production

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
MikeCollingwood: I think when I'm creating the list of required evidences I'll exclude cases where evidence cannot meet a profile because of the verif/fraud/activity scores - that should mean we won't need to check here. I'll call the method a less verbose version of: "calculateRequiredEvidencesToMeetAProfileGivenFraudActivityVerificationScores"

Joe-Edwards-GDS: I wonder if that's less flexible in the future though?

E.g. if we need to ask 'what verification score do we need to meet a profile' - we could use the same base method and filter for those that only have a nonzero verification score.

Maybe YAGNI

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -302,6 +321,46 @@ private JWEObject signEncryptJar(
...
+        OptionalInt minViableStrengthOpt =
+                requiredEvidences.stream()
+                        .filter(subList -> subList.size() == 1)
+                        .map(subList -> subList.get(0))
+                        .mapToInt(Gpg45Scores.Evidence::getStrength)
+                        .min();
+
+        System.out.println("Hey");
+        System.out.println(requiredEvidences);
+        System.out.println(minViableStrengthOpt);
```
Joe-Edwards-GDS: 👀

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -77,6 +81,8 @@ public class BuildCriOauthRequestHandler
         implements RequestHandler<JourneyRequest, Map<String, Object>> {
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String DCMAW_CRI_ID = "dcmaw";
+    private static final List<Gpg45Profile> ACCEPTED_PROFILES =
+            List.of(Gpg45Profile.M1A, Gpg45Profile.M1B, Gpg45Profile.M2B);
```
Joe-Edwards-GDS: I think Rohit moved this into a shared variable in `GPG45ProfileEvaluator`, so it's not repeated everywhere.

Also, I don't think we should be including M2B until we're ready to go live with that journey!

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -320,22 +322,40 @@ private JWEObject signEncryptJar(
...
+        List<SignedJWT> signedJWTs = new ArrayList<>();
+
+        for (String credential : credentials) {
+            try {
+                signedJWTs.add(SignedJWT.parse(credential));
+            } catch (ParseException e) {
+                throw new HttpResponseExceptionWithErrorBody(
+                        500, ErrorResponse.FAILED_TO_PARSE_ISSUED_CREDENTIALS);
+            }
+        }
```
Joe-Edwards-GDS, line 360: I'd forgotten how rubbish Java streams and checked exceptions are! 

This should really be a simple map operation, but it's really awkward to handle the exception - so just using a for loop is better 😢

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -320,22 +322,40 @@ private JWEObject signEncryptJar(
...
+                requiredEvidences.stream()
+                        .filter(subList -> subList.size() == 1)
+                        .map(subList -> subList.get(0))
+                        .mapToInt(Gpg45Scores.Evidence::getStrength)
+                        .min();
+
+        if (minViableStrengthOpt.isPresent()) {
+            return new EvidenceRequest(minViableStrengthOpt.getAsInt());
+        }
+        LOGGER.error("Minimum strength evidence required cannot be attained.");
```
Joe-Edwards-GDS: Minor, but I think it might be cleaner for this logic to be inverted - i.e. the 'normal' return at the end of the method is the happy path, and the early return is the exception?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -59,20 +62,17 @@ public Gpg45Scores(
...
         this.activity = activity;
         this.fraud = fraud;
         this.verification = verification;
-        this.evidences =
-                evidence.stream()
-                        .sorted(
-                                Comparator.comparing(Evidence::getStrength)
-                                        .thenComparing(Evidence::getValidity)
-                                        .reversed())
-                        .toList();
```
Joe-Edwards-GDS, line 75: Sorry, I think you misunderstood my comment - if you make all the other constructors delegate to this one, then you only need to do the sorting once :)

e.g.:
```
    public Gpg45Scores(Evidence evidence, int activity, int fraud, int verification) {
       this(List.of(evidence), activity, fraud, verification);
    }

    public Gpg45Scores(
            Evidence evidence1, Evidence evidence2, int activity, int fraud, int verification) {
        this(List.of(evidence1, evidence2), activity, fraud, verification);
    }

   // etc.

    public Gpg45Scores(List<Evidence> evidence, int activity, int fraud, int verification) {
        this.evidences = sortEvidence(evidence);
        this.activity = activity;
        this.fraud = fraud;
        this.verification = verification;
    }
```
I think this is a lot neater anyway, and good practice as it gives you a single place to add new constructor logic (like the sorting...).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -91,6 +91,41 @@ public List<Evidence> getEvidences() {
...
+                                        continue;
+                                    }
+                                }
+
+                                missingEvidence.add(requiredEvidences.remove(0));
+                            }
+
+                            return missingEvidence;
+                        })
+                .toList();
```
Joe-Edwards-GDS, line 125: Unless you want to argue for the contrary, I still think it's worth making this return the full profile, so this method is a bit more well-defined and reusable.

I think that means:
- Moving the filter from here into the F2F bit - making it clear that you need to have already met the activity/fraud/verification scores to get a profile
- Returning a GPG45Scores here, with activity/fraud/verification scores either zero (if met) or the required level (if unmet)

Hoping that won't take long!

Joe-Edwards-GDS (changes requested):
Looking good, just a few small changes...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
MikeCollingwood: Ah I see! I follow what you mean. I can make those changes

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
MikeCollingwood: Oh I understand. That makes sense, it would be more reusable

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -49,10 +49,7 @@ public Gpg45Scores(Evidence evidence, int activity, int fraud, int verification)
```
Joe-Edwards-GDS, line 43: I think we probably should update the other two constructors as well - even if the sorting isn't needed for them!

Joe-Edwards-GDS (approved):

PR #1292, opened 2023-10-26
PYIC-3773: Update core identity claim generation to evaluate all relevant VCs

lwebb-iig:
Refine the logic for selecting a VC to use for the identity claims.

### What changed

- Updated UserIdentityService to check all relevant VCs for a name and date of birth, find the first of each (if present) and generate identity containing a combination of the two. Maintains existing error states one or both are missing.
- Updated BuildProvenUserIdentityDetailsHandler to find first vc nodes containing name and date of birth and construct `NameAndDateOfBirth` if both are present. 

### Why did it change

Currently our core identity claim is generated by finding the first VC from the EVIDENCE_CRI_TYPES (which is currently passport, DCMAW, driving licence and F2F) and it reads both the name and DOB from this VC.

This isn’t really suitable for M2B:
- Adding BAV to the list won’t work, as it only contains the name (not DOB)
- Adding NINO to the list is inconsistent, as it is not always ‘evidence’ (and adding NINO without BAV is weird)

The same thing applies to getting the proven identity details during a reuse journey.

### Issue tracking

- [PYIC-3773](https://govukverify.atlassian.net/browse/PYIC-3773)

### Other considerations

[PYIC-3773]: https://govukverify.atlassian.net/browse/PYIC-3773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
+++ b/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
@@ -19,10 +19,12 @@ import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport
 import uk.gov.di.ipv.core.library.domain.Address;
 import uk.gov.di.ipv.core.library.domain.BirthDate;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.domain.IdentityClaim;
 import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.Name;
 import uk.gov.di.ipv.core.library.dto.VcStatusDto;
+import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.NoVcStatusForIssuerException;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
@@ -41,15 +43,13 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static uk.gov.di.ipv.core.library.domain.CriConstants.ADDRESS_CRI;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CLAIM;
 import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_CREDENTIAL_SUBJECT;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_ERROR_PATH;
-import static uk.gov.di.ipv.core.library.service.UserIdentityService.BIRTH_DATE_PROPERTY_NAME;
-import static uk.gov.di.ipv.core.library.service.UserIdentityService.EVIDENCE_CRI_TYPES;
-import static uk.gov.di.ipv.core.library.service.UserIdentityService.NAME_PROPERTY_NAME;
 
 public class BuildProvenUserIdentityDetailsHandler
         implements RequestHandler<JourneyRequest, Map<String, Object>> {
@@ -109,7 +109,7 @@ public class BuildProvenUserIdentityDetailsHandler
             List<VcStatusDto> currentVcStatuses = generateCurrentVcStatuses(credentials);
 
             NameAndDateOfBirth nameAndDateOfBirth =
-                    getProvenIdentityNameAndDateOfBirth(credentials, currentVcStatuses);
+                    getProvenIdentityNameAndDateOfBirth(credentials);
             provenUserIdentityDetailsBuilder.name(nameAndDateOfBirth.getName());
             provenUserIdentityDetailsBuilder.dateOfBirth(nameAndDateOfBirth.getDateOfBirth());
 
@@ -123,7 +123,7 @@ public class BuildProvenUserIdentityDetailsHandler
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
-        } catch (ParseException | JsonProcessingException e) {
+        } catch (ParseException | JsonProcessingException | CredentialParseException e) {
             LOGGER.error("Failed to parse credentials");
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH,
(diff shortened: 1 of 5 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -253,14 +253,36 @@ private IdentityClaim getIdentityClaim(
...
+        }
+
+        List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem item : successfulVCStoreItems) {
             if (EVIDENCE_CRI_TYPES.contains(item.getCredentialIssuer())) {
-                return Optional.of(
-                        getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), false));
+                identityClaims.add(
+                        getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true));
             }
```
Wynndow: I think it'd be really good if we could get rid of `EVIDENCE_CRI_TYPES` completely. At the moment it contains `PASSPORT_CRI, DCMAW_CRI, DRIVING_LICENCE_CRI, F2F_CRI`. We've agreed not to add BAV and NINO to it, so it's already a weirdly inconsistent thing.
One possibility it to check if the `VcStoreItem` contains an evidence and validity score, that way we know it's an evidence VC and can be used to build the identity claim.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -253,14 +253,36 @@ private IdentityClaim getIdentityClaim(
...
         }
-        LOGGER.warn("Failed to generate identity claim");
-        return Optional.empty();
+
+        Optional<IdentityClaim> claimWithName =
+                identityClaims.stream()
+                        .filter(identityClaim -> identityClaim.getName().size() >= 1)
```
Wynndow:
```suggestion
                        .filter(identityClaim -> !identityClaim.getName().isEmpty())
```

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -253,14 +253,36 @@ private IdentityClaim getIdentityClaim(
...
+            return Optional.empty();
+        }
+
+        List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem item : successfulVCStoreItems) {
             if (EVIDENCE_CRI_TYPES.contains(item.getCredentialIssuer())) {
-                return Optional.of(
-                        getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), false));
+                identityClaims.add(
+                        getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true));
```
Wynndow: We're now passing `true` here to avoid an error in the case of a missing property, and just get an empty collection back instead. This means that all uses of `getIdentityClaim` are now passing in true. We should probably refactor it out to make the method and signature simpler.

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -164,29 +165,32 @@ private NameAndDateOfBirth getProvenIdentityNameAndDateOfBirth(
...
-                return new NameAndDateOfBirth(nameBuilder.toString(), birthDate.getValue());
+                evidenceVcSubjectNodes.add(vcSubjectNode);
             }
         }
+
+        JsonNode nameValue = findFirstVcSubjectNode(evidenceVcSubjectNodes, NAME_PROPERTY_NAME);
+        JsonNode dateOfBirthValue =
+                findFirstVcSubjectNode(evidenceVcSubjectNodes, BIRTH_DATE_PROPERTY_NAME);
+
+        if (!nameValue.isNull() && !dateOfBirthValue.isNull()) {
```
Joe-Edwards-GDS: Very minor, but now we've updated the logic we could invert this condition - so the error case returns early, and the happy path continues

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
Joe-Edwards-GDS, line 1: Do you think it's worth trying to commonise the logic between these two processes?

The logic here _could_ call the user identity service method to get an `IdentityClaim`, and then just pick the first name and DOB from that? That way we have a consistent notion of how to read your core identity from your VCs.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -182,32 +185,24 @@ public Optional<Boolean> getVCSuccessStatus(String userId, String criId) throws
         return Optional.empty();
     }
 
-    private JsonNode getVCClaimNode(String credential) throws HttpResponseExceptionWithErrorBody {
+    private JsonNode getVCNode(String credential, String node)
```
Joe-Edwards-GDS: I think this is still getting a 'claim node' - the difference is that it's no longer just getting the 'subject' claim within it?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -260,9 +250,12 @@ private Optional<IdentityClaim> generateIdentityClaim(List<VcStoreItem> successf
 
         List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem item : successfulVCStoreItems) {
-            if (EVIDENCE_CRI_TYPES.contains(item.getCredentialIssuer())) {
-                identityClaims.add(
-                        getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true));
+            JsonNode vcEvidenceNode = getVCNode(item.getCredential(), VC_EVIDENCE);
+            for (JsonNode evidence : vcEvidenceNode) {
+                if (evidence.path(VC_EVIDENCE_VALIDITY).isInt()
+                        && evidence.path(VC_EVIDENCE_STRENGTH).isInt()) {
```
Joe-Edwards-GDS: Can we extract this out to a method so it's clear what we're trying to do (i.e. `isEvidenceVc`) or similar?

I'm slightly torn as to whether we should also be checking that these are both non-zero - we do check that already via the 'success' check, but I don't love relying on it.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
lwebb-iig: I've removed this in favor of a new `isEvidenceVc` method

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
lwebb-iig: Ultimately yes, I've had a go at this - it should also address your newer comments @Joe-Edwards-GDS 
Things to note, I had to shuffle the error handling around a bit as some userIdentityService methods throw http-related errors whereas the BuildProvenUserIdentityDetails lambda throws journey-related erorrs

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
lwebb-iig: At the moment I've opted to check whether they exist/are integers, it's easy enough to check if they are non-zero should we want that change

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -225,42 +215,67 @@ private <T> T getJsonProperty(
...
+
+        Optional<IdentityClaim> claimWithName =
+                identityClaims.stream()
+                        .filter(identityClaim -> !identityClaim.getName().isEmpty())
+                        .findFirst();
+        Optional<IdentityClaim> claimWithBirthDate =
+                identityClaims.stream()
+                        .filter(identityClaim -> !identityClaim.getBirthDate().isEmpty())
+                        .findFirst();
+        if (claimWithName.isEmpty() || claimWithBirthDate.isEmpty()) {
```
TalatAydinGDS, line 323: Is there a CRI type that we should exclude before throwing an errors here? For example BAV, Address CRI or, are we expecting at least one of the successful VCs to have a name and a birthdate?

lwebb-iig: My assumption for this ticket was that if we can't find a name & date of birth on the evidence VCs we should throw an error here. Do we need to handle the case where the `identityClaims` list is empty and return empty, similar to here [return Optional.of(identityClaim);](https://github.com/govuk-one-login/ipv-core-back/pull/1292/files/705778f645f1f3b4a101cfbbbebd58b3f8d055e6#diff-03e309f6bbe7b1e939e8d7a3f5ad31be61b375cf9c88a8a05048bed7970df470R246)

TalatAydinGDS: Yeah, that's ok. According to the task requirements, it appears that both name and dob are expected to be present in at least one successful VC, In Name and DOB correlations, I am excluding the following 

- DOB correlation excludes BAV and Address CRI
- Name correlation excludes only Address CRI

Because, Address CRI does not a name and DOB, BAV CRI does not a DOB.

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -150,46 +149,44 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
+                successfulVCStoreItems.add(credentialIssuerItem);
             }
         }
-        LOGGER.error("Failed to find name and date of birth of proven user identity");
-        throw new ProvenUserIdentityDetailsException(
-                "Failed to find name and date of birth of proven user identity");
+
+        try {
+            final IdentityClaim identityClaim =
+                    userIdentityService.findIdentityClaim(successfulVCStoreItems);
```
Wynndow: The `succesfulVCStoreItems` list here has been filtered by calling `isEvidenceVc()` on items before adding. That's happening again within `findIdentityClaim()`. You might be able to simplify the above by not doing the first filter?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -225,42 +215,67 @@ private <T> T getJsonProperty(
...
                                 .getTypeFactory()
-                                .constructCollectionType(List.class, BirthDate.class),
-                        validateCorrelation);
+                                .constructCollectionType(List.class, BirthDate.class));
 
         return new IdentityClaim(names, birthDates);
     }
 
-    private Optional<IdentityClaim> generateIdentityClaim(List<VcStoreItem> successfulVCStoreItems)
+    public Optional<IdentityClaim> generateIdentityClaim(List<VcStoreItem> successfulVCStoreItems)
```
Wynndow: Does this need to be made public? Looks like it's only called from within this class.

Wynndow: In fact, do we need it at all? Looks like a wrapper for `findIdentityClaim`, just with a guard clause

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
lwebb-iig: Good point - I've moved out the `isEvidenceVc()` such that `findIdentityClaim` is as method that just accepts a list of successful evidence vcs

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
lwebb-iig: Put this back to private - this method is doing slightly more now in filtering evidence vcs from its `successfulVcStoreItems` list so I'd say it probably warrants existence for now

Wynndow (approved):
Looks good to me. I think there is still some work to do, especially in the build proven user identity lambda, about standardising how we're passing "successful" VCs around. I think if we could remove the use of the "VC statues" that would be a good thing. But that's outside the scope of this ticket I think.

lwebb-iig:
> Looks good to me. I think there is still some work to do, especially in the build proven user identity lambda, about standardising how we're passing "successful" VCs around. I think if we could remove the use of the "VC statues" that would be a good thing. But that's outside the scope of this ticket I think.

That was my impression as well - I think we could remove vcStatuses from the rest of the lambda as well, but this felt like overreaching for the scope of this ticket as it would be getting into refactoring the address identity part of that lambda.

PR #1294, opened 2023-10-26
[PYIC-3700] VcHelper.isSuccessful always ignores CIs

Joe-Edwards-GDS:
### What changed

Combine the current `isSuccessfulVc` and `isSuccessfulVcIgnoringCis` into one method which always ignores CIs.

### Why did it change

We shouldn't be looking at CIs when considering VC 'success' - to do that we need to perform proper CI scoring.

In future we might want to rethink the whole approach, but keeping this minimal for now.

### Issue tracking

- [PYIC-3700](https://govukverify.atlassian.net/browse/PYIC-3700)

[PYIC-3700]: https://govukverify.atlassian.net/browse/PYIC-3700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
+++ b/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
@@ -37,27 +37,17 @@ public class VcHelper {
         VcHelper.configService = configService;
     }
 
-    public static boolean isSuccessfulVc(SignedJWT vc) throws ParseException {
-        return isSuccessfulVc(vc, getNonEvidenceCredentialIssuers(), true);
-    }
-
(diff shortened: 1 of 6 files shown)


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -37,27 +37,17 @@ public static void setConfigService(ConfigService configService) {
         VcHelper.configService = configService;
```
Joe-Edwards-GDS, line 37: This feels like a pretty dodgy pattern - having a static class that requires you to call a method on it before using. It should probably just be a non-static class (or this method could be absorbed into another service).

I'm not sure it's worth changing at this stage - we may be revising the concept of successful VCs entirely anyway.

lwebb-iig (approved):

PR #1296, opened 2023-10-26
PYIC-3661: Add ability to set "context" for a CRI in the journey map

MikeCollingwood:
### What changed

- Allowed states in the journey map to pass context to BuildCriOauthRequest to request from the CRI with the context included
- Enabled ProcessEventHandler tests to access test journey map

### Why did it change

- The CIC CRI is currently tailored for a photo ID journey, but it will also be used in No Photo ID journeys. The content on the screens needs to be adaptable based on the use case.

### Issue tracking

- [PYIC-3661](https://govukverify.atlassian.net/browse/PYIC-3661)

[PYIC-3661]: https://govukverify.atlassian.net/browse/PYIC-3661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
+++ b/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
@@ -116,6 +116,7 @@ class AuthorizationRequestHelperTest {
                         OAUTH_STATE,
                         TEST_USER_ID,
                         TEST_JOURNEY_ID,
+                        null,
                         null);
 
         assertEquals(IPV_ISSUER, result.getJWTClaimsSet().getIssuer());
@@ -147,6 +148,7 @@ class AuthorizationRequestHelperTest {
                         OAUTH_STATE,
                         TEST_USER_ID,
                         TEST_JOURNEY_ID,
+                        null,
                         null);
         assertNull(result.getJWTClaimsSet().getClaims().get(TEST_SHARED_CLAIMS));
     }
@@ -168,6 +170,7 @@ class AuthorizationRequestHelperTest {
                                         OAUTH_STATE,
                                         TEST_USER_ID,
                                         TEST_JOURNEY_ID,
+                                        null,
                                         null));
         assertEquals(500, exception.getResponseCode());
         assertEquals("Failed to sign Shared Attributes", exception.getErrorReason());
--- a/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
+++ b/lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
@@ -62,6 +62,7 @@ class AuthorizationRequestHelperTest {
     private static final String IPV_CLIENT_ID_VALUE = "testClientId";
     private static final String IPV_ISSUER = "http://example.com/issuer";
     private static final String AUDIENCE = "Audience";
+    private static final String BANK_ACCOUNT_CONTEXT = "bank_account";
     private static final String IPV_TOKEN_TTL = "900";
     private static final String MOCK_CORE_FRONT_CALLBACK_URL = "callbackUri";
     private static final String TEST_REDIRECT_URI = "http:example.com/callback/criId";
@@ -131,6 +132,30 @@ class AuthorizationRequestHelperTest {
                 IPV_CLIENT_ID_VALUE, result.getJWTClaimsSet().getClaims().get(CLIENT_ID_FIELD));
         assertEquals(TEST_REDIRECT_URI, result.getJWTClaimsSet().getClaims().get("redirect_uri"));
         assertTrue(result.verify(new ECDSAVerifier(ECKey.parse(EC_PUBLIC_JWK))));
+    }
+
+    @Test
+    void shouldCreateSignedJWTWithContextIfExists()
+            throws ParseException, HttpResponseExceptionWithErrorBody {
+        setupCredentialIssuerConfigMock();
+        setupConfigurationServiceMock();
+        when(credentialIssuerConfig.getComponentId()).thenReturn(AUDIENCE);
+        when(credentialIssuerConfig.getClientCallbackUrl())
+                .thenReturn(URI.create(TEST_REDIRECT_URI));
+
+        SignedJWT result =
+                AuthorizationRequestHelper.createSignedJWT(
+                        sharedClaims,
+                        signer,
+                        credentialIssuerConfig,
+                        configService,
+                        OAUTH_STATE,
+                        TEST_USER_ID,
+                        TEST_JOURNEY_ID,
+                        null,
+                        BANK_ACCOUNT_CONTEXT);
+
+        assertEquals(BANK_ACCOUNT_CONTEXT, result.getJWTClaimsSet().getStringClaim("context"));
     }
 
     @Test
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -23,6 +23,7 @@ import uk.gov.di.ipv.core.library.service.IpvSessionService;
 import uk.gov.di.ipv.core.processjourneyevent.exceptions.JourneyEngineException;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.StateMachine;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.StateMachineInitializer;
+import uk.gov.di.ipv.core.processjourneyevent.statemachine.StateMachineInitializerMode;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.StateMachineNotFoundException;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownEventException;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownStateException;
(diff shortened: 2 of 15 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -58,12 +59,27 @@ public ProcessJourneyEventHandler(
...
+            ConfigService configService,
+            ClientOAuthSessionDetailsService clientOAuthSessionService,
+            List<IpvJourneyTypes> journeyTypes)
+            throws IOException {
+        this(
+                ipvSessionService,
+                configService,
+                clientOAuthSessionService,
+                journeyTypes,
+                StateMachineInitializerMode.STANDARD);
```
Joe-Edwards-GDS: I don't think you need this additional constructor - it's only used in one test?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -227,14 +243,23 @@ private boolean sessionIsNewlyExpired(IpvSessionItem ipvSessionItem) {
...
+                    journeyType,
+                    new StateMachine(
+                            new StateMachineInitializer(journeyType, stateMachineInitializerMode)));
         }
         return stateMachinesMap;
     }
+
+    private Map<IpvJourneyTypes, StateMachine> loadStateMachines(List<IpvJourneyTypes> journeyTypes)
+            throws IOException {
+        return loadStateMachines(journeyTypes, StateMachineInitializerMode.STANDARD);
```
Joe-Edwards-GDS: Do we need this? Could we just pass the extra argument at it's usage?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
```diff
@@ -14,5 +14,6 @@ public class JourneyRequest {
     private String ipAddress;
     private String clientOAuthSessionId;
     private String journey;
+    private String context;
```
Joe-Edwards-GDS, line 17: I think this adds `context` to _all_ journey requests, rather than just ones to the `BuildCriOauthRequestHandler` - is that intentional?

Not necessarily a problem, and context could be useful for others in future, but a bit different to what we've done elsewhere. Compare the `ProcessRequest`, which is specific request class for the check GPG45 score lambda.

Joe-Edwards-GDS (commented):
A few simplifications, but generally looks sensible!

Joe-Edwards-GDS:
As mentioned on slack - I think you should check this in dev if possible, to work out whether you need to update the step function to handle passing the `context`.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
MikeCollingwood: I think there may be a desire to continue to use the test journey map because testing on a fixed journey map can only be so useful.
Is there a way I can mock/ build this without the constructor do you think?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
MikeCollingwood: Yep, I'll make the change

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: As in, you've got two constructors which only differ by the last parameter, and this version is only used in one test. Could you just specify the extra parameter in that test?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
MikeCollingwood: I'm on the fence about adding a new class for the request with context since it isn't particularly specialised, but I definitely can if you think it's worth while

MikeCollingwood:
I tested this by:
- adding context to each cri state
- deploying it to my dev environment
- watching the network tab in developer tools when trying to authorize with a cri
- using the request JWT into a temporary test method:
```@Test
void temp() throws Exception {
    JWEObject jweObject = JWEObject.parse("eyJjdQn1...Lyt9Vg");
    jweObject.decrypt(new RSADecrypter(KeyFactory.getInstance("RSA")
            .generatePrivate(
                    new PKCS8EncodedKeySpec(
                            Base64.getDecoder().decode(RSA_ENCRYPTION_PRIVATE_KEY)))));

    SignedJWT signedJWT = SignedJWT.parse(jweObject.getPayload().toString());
    System.out.println(signedJWT.getParsedString()); // put this into jwt decoder
}
```
- following this, I put the parsed string into https://jwt.io/ to visualise the decrypted jwt

MikeCollingwood:
Manual test shows that I could get context through to the request to the cri (dcmaw stub in my dev test):
```
{
  "sub": "urn:uuid:4c294365-df42-4367-81f0-2257008f140f",
  "shared_claims": {
    "address": [],
    "name": [],
    "birthDate": []
  },
  "iss": "https://dev-mikec.02.dev.identity.account.gov.uk",
  "response_type": "code",
  "client_id": "ipv-core-dev",
  "govuk_signin_journey_id": "cd63a67a-674f-4c56-b7d3-02de322cf983",
  "aud": "https://dcmaw-cri.stubs.account.gov.uk",
  "nbf": 1698665191,
  "context": "bank_account", <<<
  "redirect_uri": "https://dev-mikec.02.dev.identity.account.gov.uk/credential-issuer/callback/dcmaw",
  "state": "S2ClC1tPJM0nEZ3P06rpYq3iomVMeDn_m7zAhnaQj4Q",
  "exp": 1698666091,
  "iat": 1698665191
}
```

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
```diff
@@ -168,6 +170,7 @@ void shouldThrowExceptionWhenUnableToSignJwt() {
                                         OAUTH_STATE,
                                         TEST_USER_ID,
                                         TEST_JOURNEY_ID,
+                                        null,
```
Wynndow, line 198: Might be worth adding something into one of these tests, or creating a new one, that asserts that context gets added.

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/CriStepResponseTest.java
```diff
@@ -8,7 +8,7 @@
 
 public class CriStepResponseTest {
 
-    public static final CriStepResponse CRI_RESPONSE = new CriStepResponse("aCriId");
+    public static final CriStepResponse CRI_RESPONSE = new CriStepResponse("aCriId", null);
```
Wynndow, line 11: Probably worth adding a test case that does have a context.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
Wynndow: I'd be inclined to not add a new class. The `ProcessRequest` class isn't really implemented how I thought it should be - it's not generic. I'd intended that you'd be able to use that class for any "process" type state and freely define data to pass through, rather than having to update the class every time you wanted to use it to run a different lambda.

The new `context` attribute is only being used on the cri type states, which doesn't use the `ProcessRequest` state. I don't think I'd want to see a proliferation of classes for each bit of data we want to pass through from the journey map. And as you say, "context" is pretty generic and could be used for other things.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -111,6 +111,10 @@ public static String getJourney(JourneyRequest request) {
         return request.getJourney();
     }
 
+    public static String getContext(JourneyRequest request) {
+        return request.getContext();
+    }
```
Wynndow: I know this is following a pattern, but I'm not a big fan of it. This method is just a wrapper for the same method. I'd be tempted to just call `request.getContext()` when we need it, rather than using this helper.

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
```diff
@@ -108,6 +109,7 @@ public LambdaHandler() throws IOException {
                                     EMPTY_CONTEXT);
 
                     journey = (String) processJourneyEventOutput.get(JOURNEY);
+                    String context = (String) processJourneyEventOutput.get(CONTEXT);
```
Wynndow: I'd be tempted to just inline this where we need it.

Wynndow (dismissed):
Couple of small comments, but I think this is good.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
Joe-Edwards-GDS: Bear in mind that we're also about to add `scope` which is a bit more CRI-specific, so instead we're going to end up with a proliferation of fields that are only applicable to certain steps...

On balance, I think I'd slightly prefer a separate class, so you know exactly what each lambda expects, but I'm happy for you to leave if you'd like to move on.

cc @rpayal - this will affect you as well!

Joe-Edwards-GDS (approved):
Slight preference for splitting the classes, but happy for you to merge as-is.

PR #1300, opened 2023-10-27
PYIC-3774: Update name/dob correlation to properly handle missing claims

TalatAydinGDS:
### What changed

  We rewrote the correlation logic to ensure that each VC is associated via both name and DOB, except for BAV, which needs a special case to ignore the DOB.

### Why did it change

  Currently the name/dob correlation only checks that there is no more than one name/dob across all the VCs, it does not check that each VC contains a correlating name and DOB.

This is wrong - we need to positively correlate each VC we use to generate an identity. However, we also need to build in an exception for the BAV VC, which will not include a DOB.

### Issue tracking

- [PYIC-3774](https://govukverify.atlassian.net/browse/PYIC-3774)

### Other considerations

[PYIC-3774]: https://govukverify.atlassian.net/browse/PYIC-3774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -419,10 +419,10 @@ public class UserIdentityService {
     }
 
     // This method checks the birthdate requirement, which is in violation of GPG45 rules.
-    // However, considering that 'BAV' CRI does not conta'in a birthdate,
+    // However, considering that 'BAV' CRI does not contain a birthdate,
     // we add this special handling only for 'BAV' CRI. For other CRIs, we continue to validate this
     // requirement.
-    private List<IdentityClaim> getIdentityClaimsAndHandleDOBExceptionForBAVCRI(
+    private List<IdentityClaim> getIdentityClaimsForDOBCheckOnlyBAVCRI(
             List<VcStoreItem> vcStoreItems) throws HttpResponseExceptionWithErrorBody {
         List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem item : vcStoreItems) {
@@ -447,7 +447,7 @@ public class UserIdentityService {
         final List<VcStoreItem> successfulVCStoreItems =
                 getSuccessfulVCStoreItems(getVcStoreItems(userId));
         List<IdentityClaim> identityClaims =
-                getIdentityClaimsAndHandleDOBExceptionForBAVCRI(successfulVCStoreItems);
+                getIdentityClaimsForDOBCheckOnlyBAVCRI(successfulVCStoreItems);
         return identityClaims.stream()
                         .map(IdentityClaim::getBirthDate)
                         .flatMap(List::stream)
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -408,29 +408,36 @@ public class UserIdentityService {
                 .getIsSuccessfulVc();
     }
 
-    private List<IdentityClaim> getIdentityClaims(List<VcStoreItem> vcStoreItems)
-            throws HttpResponseExceptionWithErrorBody {
-        List<IdentityClaim> identityClaims = new ArrayList<>();
-        for (VcStoreItem item : vcStoreItems) {
-            identityClaims.add(
-                    getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true));
-        }
-        return identityClaims;
+    public boolean checkBirthDateCorrelationInCredentials(String userId)
+            throws HttpResponseExceptionWithErrorBody, CredentialParseException {
+        final List<VcStoreItem> successfulVCStoreItems =
+                getSuccessfulVCStoreItems(getVcStoreItems(userId));
+        List<IdentityClaim> identityClaims =
+                getIdentityClaimsForBirthDateCorrelation(successfulVCStoreItems);
+        return identityClaims.stream()
+                        .map(IdentityClaim::getBirthDate)
+                        .flatMap(List::stream)
+                        .map(BirthDate::getValue)
+                        .distinct()
+                        .count()
+                <= 1;
     }
 
     // This method checks the birthdate requirement, which is in violation of GPG45 rules.
     // However, considering that 'BAV' CRI does not contain a birthdate,
     // we add this special handling only for 'BAV' CRI. For other CRIs, we continue to validate this
     // requirement.
-    private List<IdentityClaim> getIdentityClaimsForDOBCheckOnlyBAVCRI(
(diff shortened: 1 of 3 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -415,11 +418,36 @@ private List<IdentityClaim> getIdentityClaims(List<VcStoreItem> vcStoreItems)
         return identityClaims;
     }
 
+    // This method checks the birthdate requirement, which is in violation of GPG45 rules.
+    // However, considering that 'BAV' CRI does not conta'in a birthdate,
+    // we add this special handling only for 'BAV' CRI. For other CRIs, we continue to validate this
+    // requirement.
+    private List<IdentityClaim> getIdentityClaimsAndHandleDOBExceptionForBAVCRI(
```
MikeCollingwood: - Is there an exception when there is no DOB? I can't see any try/ catch-ing. If not, then maybe the method name could be changed/ shortened to something like `getIdentityClaimsExceptFromBAV`
- Could the for loop be a stream which we filter with `BAV_CRI` before doing any `getIdentityClaim` logic? Would mean we wouldn't need the `List<IdentityClaim> identityClaims = new ArrayList<>();` variable.
- Could we also be asserting that the birthdate isn't empty for non-BAV CRIs?

MikeCollingwood (commented):
Looks good but just had some questions :))

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: 1-) Here, we are only returning the correlation result. The exception is thwron from where it is called, and we handle the correlation error caused by the BAV's DOB value being null. If DOB is included within the BAV in the future, the code will handle that as well. This is covered in the unit tests. For example : checkExisting IdentityHandler()
2 & 3)  Yes, you are right that can do it, but in the future, DOB might also be included within BAV. I didn't performa this check beforehand, keeping that possibility in mind. For the other cases, we will directly add the identityClaim for DOB correlation checks. Additionally,, as we'll see in the task's requirements, the exception of DOB null is only specific of BAV CRI according GPG45 rules.

TalatAydinGDS: The method name 'getIdentityClaimsForDOBCheckOnlyBAVCRI()' has been changed for clarity.

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -282,6 +320,83 @@ void checkNameCorrelationWithMissingBirthDateCredentialsForReturnTrue() throws E
...
+                                "ukPassport",
+                                SIGNED_PASSPORT_VC_MISSING_BIRTH_DATE,
+                                Instant.now()));
+
+        when(userIdentityService.getVcStoreItems(USER_ID_1)).thenReturn(vcStoreItems);
+        mockCredentialIssuerConfig();
+
+        boolean isValid = userIdentityService.checkBirthDateCorrelationInCredentials(USER_ID_1);
+
+        assertTrue(isValid);
```
Joe-Edwards-GDS, line 343: This should return false!

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -234,7 +251,28 @@ void checkNameCorrelationWithMissingNameCredentialsForReturnTrue() throws Except
...
+        mockCredentialIssuerConfig();
+
+        boolean isValid =
+                userIdentityService.checkNameAndFamilyNameCorrelationInCredentials(USER_ID_1);
+
+        assertFalse(isValid);
+    }
+
+    @Test
+    void checkBirthDateCorrelationWithSameBirthDatesAndMissingBirthDateCredentialsForReturnTrue()
```
Joe-Edwards-GDS, line 275: Similarly - I think this should return false?

I'm a bit scared by the fact there's an explicit test for this though, do you know if we're relying on this behaviour in a journey?

Joe-Edwards-GDS (changes requested):
Hi Talat, this doesn't actually address the issue that we also allow missing birth dates in any other VC.

I'm a bit worried that there are already explicit test cases for this, so I think we might need to work out where this is being relied upon, and what we can do about it.

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
TalatAydinGDS: Hi Joe, Thanks, you have brought up a very important point. There hasn't been a specific requirement from any of the other VC's to throw an error in the evidence when there is no name or birthdate. For name and birthdate, we gather and compare them for all VCs if they exist in the evidence, In summary, if name and birthdate not found in the evidence, the correlation check is not performed. I am sharing of the previous correlation epic and taks that we discussed. https://govukverify.atlassian.net/browse/PYIC-2600 https://docs.google.com/document/d/1LjET6Zw_JFwKNH2tODk8ZcZT2DLwMoWaEjeRIohXPWY/edit#heading=h.r15klbho5w4s  Additionally, some points were mentioned that we would address issues like pre-marital surname in the future. (https://gds.slack.com/archives/C0222KXB6RM/p1681832588075989) I have a concern. If we throw a correlation error or any other error when name and birthdate(not BAV) are not found in the evidence, it could impact a significant number of customer in production. Even when they enter different names in correlation at intervals, many people have asked us without understanding the error. I have concern about this. What do you think about getting approval from the Product team and Business teams and considering this in different task? Also, name correlation currently works on the same logic now.

Joe-Edwards-GDS: Yeah, we should make sure to confirm this change with product/business/architecture - do you want to do that, or shall I pursue?

The thing that reassures me is that SPOT have told me they _are_ currently checking that the name/birthdate exist on all VCs, and we should match their logic.

If we're really nervous, we could implement this as a two-step change:
- Add logic to check for missing name/birthdate, but log an _error_ in that case, rather than failing
- Deploy to production, and check how often we encounter that error
- If we don't see that error, then turn it into a proper failure - or handle the error properly

TalatAydinGDS: Joe, I would greatly appreciate it if you could discuss and obtain approval from the relevant teams.. I am going to add  some checks and logs for names/birthdates.

TalatAydinGDS:
I am closing this because we will raise these changes in the sub-task under PYIC-3858

PR #1304, opened 2023-10-31
PYIC-3844 Change to fix calculation for f2f evidenceRequest.

rpayal:
## Proposed changes
Failing while calculating F2F min strength score.

### What changed
Change to fix calculation for f2f evidenceRequest.

### Issue tracking

- [PYIC-3844](https://govukverify.atlassian.net/browse/PYIC-3844)

### Other considerations

[PYIC-3844]: https://govukverify.atlassian.net/browse/PYIC-3844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -334,10 +334,8 @@ private EvidenceRequest getEvidenceRequestForF2F(List<SignedJWT> credentials)
                 requiredEvidences.stream()
                         .filter(
                                 requiredScores ->
-                                        requiredScores.getEvidences().size() == 1
-                                                & requiredScores.getActivity() == 0
-                                                & requiredScores.getFraud() == 0
-                                                & requiredScores.getVerification() <= 3)
+                                        requiredScores.getActivity() == 0
+                                                & requiredScores.getFraud() == 0)
```
Joe-Edwards-GDS: The lines you've removed were intentional - if you require multiple pieces of evidence or verification score >3, then you can't achieve a profile with it.

I do think we should change the evidences size to `<= 1` rather than `== 1` though - it's still ok if you don't need any evidence (but still need the verification score).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -111,7 +111,9 @@ public List<Gpg45Scores> calculateGpg45ScoresRequiredToMeetAProfile(
                             }
 
                             return new Gpg45Scores(
-                                    missingEvidence,
+                                    missingEvidence.isEmpty()
+                                            ? Arrays.asList(EV_00)
```
Joe-Edwards-GDS: I think we should prefer `List.of` for immutable lists

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45Scores.java
```diff
@@ -111,7 +111,9 @@ public List<Gpg45Scores> calculateGpg45ScoresRequiredToMeetAProfile(
                             }
 
                             return new Gpg45Scores(
-                                    missingEvidence,
+                                    missingEvidence.isEmpty()
```
Joe-Edwards-GDS: If we change the `evidence.size() <= 1` above, I don't think we should need this - and I think it's more accurate to say no evidence is required(?)

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Joe-Edwards-GDS: Do you know which of these cases was causing the issue? I'm guessing it must have been the first one?

Joe-Edwards-GDS (changes requested):

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
rpayal: Ok will try out with evidences <= 1 and verification <= 3.   in dev env

Joe-Edwards-GDS (dismissed):

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ScoresTest.java
```diff
@@ -19,6 +27,50 @@ void shouldOrderEvidenceOnScoreAndValidity() {
         assertTrue(H2D.isSatisfiedBy(score2));
     }
 
+    @Test
+    void calculateGpg45ScoresRequiredToMeetAProfile() {
```
MikeCollingwood, line 31: Ideally these would be parameterised tests and have error messages

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -334,12 +334,18 @@ private EvidenceRequest getEvidenceRequestForF2F(List<SignedJWT> credentials)
                 requiredEvidences.stream()
                         .filter(
                                 requiredScores ->
-                                        requiredScores.getEvidences().size() == 1
-                                                & requiredScores.getActivity() == 0
-                                                & requiredScores.getFraud() == 0
-                                                & requiredScores.getVerification() <= 3)
-                        .map(requiredScores -> requiredScores.getEvidences().get(0))
-                        .mapToInt(Gpg45Scores.Evidence::getStrength)
+                                        requiredScores.getEvidences().size() <= 1
```
MikeCollingwood, line 333: This makes a lot of sense given our understanding that you can get verification scores from F2F , which means you may not need any evidence requirements to still make it possible F2F CRI completes the profile

MikeCollingwood (dismissed):
Tests could be improved but otherwise LGTM

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Joe-Edwards-GDS: Yep - I think this _might_ happen if you fail the 'liveness/likeness' check in DCMAW, so have a strength/validation score, but no verification (but I don't fully understand it tbh).

Joe-Edwards-GDS (approved):

PR #1314, opened 2023-11-01
Pyic 3858:  DOB correlation excludes BAV and Address CRI, while Name correlation excludes only Address CRI

TalatAydinGDS:
## Proposed changes
   We rewrote the correlation logic to ensure that each VC is associated via both name and DOB, except for BAV and Address CRI , which needs a special case to ignore the DOB and Name.
### Why did it change

   Currently the name/dob correlation only checks that there is no more than one name/dob across all the VCs, it does not check that each VC contains a correlating name and DOB.

This is wrong - we need to positively correlate each VC we use to generate an identity. However, we also need to build in an exception like  below

    - DOB correlation excludes BAV and Address CRI
    - Name correlation excludes only Address CRI

### Issue tracking

- [PYIC-3858](https://govukverify.atlassian.net/browse/PYIC-3858)

### Other considerations

[PYIC-3858]: https://govukverify.atlassian.net/browse/PYIC-3858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -61,7 +61,8 @@ public class UserIdentityService {
     public static final List<String> EVIDENCE_CRI_TYPES =
             List.of(PASSPORT_CRI, DCMAW_CRI, DRIVING_LICENCE_CRI, F2F_CRI);
 
-    public static final List<String> EXCLUDE_CRIS_TYPES_FOR_DOB_CORRELATION =
+    public static final List<String> CRI_TYPES_EXCLUDED_FOR_NAME_CORRELATION = List.of(ADDRESS_CRI);
+    public static final List<String> CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION =
             List.of(ADDRESS_CRI, BAV_CRI);
 
     private static final Logger LOGGER = LogManager.getLogger();
@@ -426,11 +427,6 @@ public class UserIdentityService {
                 <= 1;
     }
 
-    // This method checks the birthdate requirement, which is in violation of GPG45 rules.
-    // However, considering that BAV and Address CRI  does not contain a birthdate,
-    // we add this special handling only for BAV and Address CRI. For other CRIs, we continue to
-    // validate this
-    // requirement.
     private List<IdentityClaim> getIdentityClaimsForBirthDateCorrelation(
             List<VcStoreItem> vcStoreItems) throws HttpResponseExceptionWithErrorBody {
         List<IdentityClaim> identityClaims = new ArrayList<>();
@@ -438,8 +434,7 @@ public class UserIdentityService {
             IdentityClaim identityClaim =
                     getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true);
             if (isBirthDateEmpty(identityClaim.getBirthDate())) {
-                // excluded BAV and Address CRI
-                if (EXCLUDE_CRIS_TYPES_FOR_DOB_CORRELATION.contains(item.getCredentialIssuer())) {
+                if (CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION.contains(item.getCredentialIssuer())) {
                     continue;
                 }
                 addLogMessage(item, "Birthdate property is missing from VC");
@@ -449,9 +444,9 @@ public class UserIdentityService {
         return identityClaims;
     }
 
-    private boolean isBirthDateEmpty(List<BirthDate> birtDates) {
-        return CollectionUtils.isEmpty(birtDates)
-                || birtDates.stream().map(BirthDate::getValue).allMatch(StringUtils::isEmpty);
+    public boolean isBirthDateEmpty(List<BirthDate> birthDates) {
+        return CollectionUtils.isEmpty(birthDates)
+                || birthDates.stream().map(BirthDate::getValue).allMatch(StringUtils::isEmpty);
     }
 
     public boolean checkNameAndFamilyNameCorrelationInCredentials(String userId)
@@ -470,8 +465,7 @@ public class UserIdentityService {
             IdentityClaim identityClaim =
                     getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true);
             if (isNamesEmpty(identityClaim.getName())) {
-                // excluded Address CRI. Because Address CRI does not contain a name,
-                if (ADDRESS_CRI.equals(item.getCredentialIssuer())) {
+                if (CRI_TYPES_EXCLUDED_FOR_NAME_CORRELATION.contains(item.getCredentialIssuer())) {
                     continue;
                 }
                 addLogMessage(item, "Name property is missing from VC");
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -428,7 +428,8 @@ public class UserIdentityService {
 
     // This method checks the birthdate requirement, which is in violation of GPG45 rules.
     // However, considering that BAV and Address CRI  does not contain a birthdate,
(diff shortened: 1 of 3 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
...
+                    continue;
+                }
+                addLogMessage(item, "Birthdate property is missing from VC");
+            }
+            identityClaims.add(identityClaim);
+        }
+        return identityClaims;
+    }
+
+    private boolean isBirthDateEmpty(List<BirthDate> birtDates) {
```
rpayal: typo in parameter name

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
...
 
+    private List<IdentityClaim> getIdentityClaimsForNameCorrelation(List<VcStoreItem> vcStoreItems)
+            throws HttpResponseExceptionWithErrorBody {
+        List<IdentityClaim> identityClaims = new ArrayList<>();
+        for (VcStoreItem item : vcStoreItems) {
+            IdentityClaim identityClaim =
+                    getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true);
+            if (isNamesEmpty(identityClaim.getName())) {
+                // excluded Address CRI. Because Address CRI does not contain a name,
+                if (ADDRESS_CRI.equals(item.getCredentialIssuer())) {
```
Joe-Edwards-GDS: Can we still keep this in a variable alongside the DOB list - even though it's only one item, it makes it easier to see than being hardcoded inside this method :)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -58,6 +61,9 @@ public class UserIdentityService {
     public static final List<String> EVIDENCE_CRI_TYPES =
             List.of(PASSPORT_CRI, DCMAW_CRI, DRIVING_LICENCE_CRI, F2F_CRI);
 
+    public static final List<String> EXCLUDE_CRIS_TYPES_FOR_DOB_CORRELATION =
+            List.of(ADDRESS_CRI, BAV_CRI);
```
Joe-Edwards-GDS, line 66: Minor, but might be clearer as `CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION`?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
                 <= 1;
     }
 
+    // This method checks the birthdate requirement, which is in violation of GPG45 rules.
+    // However, considering that BAV and Address CRI  does not contain a birthdate,
+    // we add this special handling only for BAV and Address CRI. For other CRIs, we continue to
+    // validate this
+    // requirement.
```
Joe-Edwards-GDS: Maybe move this comment alongside the constant list of excluded CRIs?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
...
+    private List<IdentityClaim> getIdentityClaimsForBirthDateCorrelation(
+            List<VcStoreItem> vcStoreItems) throws HttpResponseExceptionWithErrorBody {
+        List<IdentityClaim> identityClaims = new ArrayList<>();
+        for (VcStoreItem item : vcStoreItems) {
+            IdentityClaim identityClaim =
+                    getIdentityClaim(item.getCredential(), item.getCredentialIssuer(), true);
+            if (isBirthDateEmpty(identityClaim.getBirthDate())) {
+                // excluded BAV and Address CRI
```
Joe-Edwards-GDS: (unnecessary comment?)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
...
+                addLogMessage(item, "Birthdate property is missing from VC");
+            }
+            identityClaims.add(identityClaim);
+        }
+        return identityClaims;
+    }
+
+    private boolean isBirthDateEmpty(List<BirthDate> birtDates) {
+        return CollectionUtils.isEmpty(birtDates)
+                || birtDates.stream().map(BirthDate::getValue).allMatch(StringUtils::isEmpty);
```
Joe-Edwards-GDS: I don't think you need the empty check? - `allMatch` should return true if the list is empty

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -429,14 +426,69 @@ public boolean checkBirthDateCorrelationInCredentials(String userId)
...
+        }
+        return identityClaims;
+    }
+
+    private boolean isNamesEmpty(List<Name> names) {
+        return CollectionUtils.isEmpty(names)
+                || names.stream()
+                        .flatMap(name -> name.getNameParts().stream())
+                        .map(NameParts::getValue)
+                        .allMatch(StringUtils::isEmpty);
```
Joe-Edwards-GDS, line 483: (ditto)

Joe-Edwards-GDS (commented):
A few minor changes to suggest, but looks broadly sensible.

I think there might be some more refactoring to do when we turn the missing name/dob into an error :)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: I think, we can remove this comment. Bacause, CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION list name is clear.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: If birthDates list comes as null., a NullPointerException will be thrown. I think, we should keep it.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: If names list comes as null., a NullPointerException will be thrown. I think, we should keep it.

Joe-Edwards-GDS: When is it `null`, rather than empty list?

TalatAydinGDS: I don't know in which scenario it might come as null, but wouldn't it be better to safely it, just in case?

Joe-Edwards-GDS: Not necessarily - if we don't expect this to return null, then it could be concealing an error.

However, if you think it might be null in certain cases that aren't a problem, then this is fine :)

(the 'billion-dollar mistake'...)

Joe-Edwards-GDS (approved):

PR #1316, opened 2023-11-01
PYIC-3829: Return exit code for contra indicators

Wynndow:
### What changed

Return exit code for contra indicators

### Why did it change

We want to return an exit code attribute in the user identity fetched by the orchestrator.

This change maps the CI codes a user has to the appropriate exit codes and adds them to the user identity. This is hidden behind a feature flag.

In the case that a user has achieved a P2, we only want to add exit codes for specific CIs. These are defined in config to avoid leaking sensitive information in this public repo.

In the case that a user has achieved a P0 due to breaching the CI threshold, we want to map all CI codes to their exit codes. This is looked up in a mapping table defined in config. This includes mitigated CIs. This is because a mitigated CI will still have a residual score, and so will be contributing to the threshold breach.

The case that a user has achieved a P0 but not due to a CI threshold breach will be handled in another ticket.

### Issue tracking

- [PYIC-3829](https://govukverify.atlassian.net/browse/PYIC-3829)

[PYIC-3829]: https://govukverify.atlassian.net/browse/PYIC-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
@@ -1,6 +1,8 @@
 package uk.gov.di.ipv.core.library.domain;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import lombok.Builder;
@@ -24,6 +26,7 @@ public class UserIdentity {
     private static final String PASSPORT_CLAIM_NAME = "https://vocab.account.gov.uk/v1/passport";
     private static final String DRIVING_PERMIT_CLAIM_NAME =
             "https://vocab.account.gov.uk/v1/drivingPermit";
+    public static final String EXIT_CODE_NAME = "exit_code";
 
     @JsonProperty(VCS_CLAIM_NAME)
     private List<String> vcs;
@@ -46,6 +49,10 @@ public class UserIdentity {
 
     @JsonProperty private String vtm;
 
+    @JsonInclude(Include.NON_NULL)
+    @JsonProperty(EXIT_CODE_NAME)
+    private List<String> exitCode;
+
     @JsonCreator
     public UserIdentity(
             @JsonProperty(value = VCS_CLAIM_NAME, required = true) List<String> vcs,
@@ -55,7 +62,8 @@ public class UserIdentity {
             @JsonProperty(value = DRIVING_PERMIT_CLAIM_NAME) JsonNode drivingPermitClaim,
             @JsonProperty(value = "sub", required = true) String sub,
             @JsonProperty(value = "vot", required = true) String vot,
-            @JsonProperty(value = "vtm", required = true) String vtm) {
+            @JsonProperty(value = "vtm", required = true) String vtm,
+            @JsonProperty(value = EXIT_CODE_NAME) List<String> exitCode) {
         this.vcs = new ArrayList<>(vcs);
         this.identityClaim = identityClaim;
         this.addressClaim = addressClaim;
@@ -64,5 +72,6 @@ public class UserIdentity {
         this.sub = sub;
         this.vot = vot;
         this.vtm = vtm;
+        this.exitCode = exitCode;
     }
 }
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -14,16 +14,20 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.StringMapMessage;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.domain.BirthDate;
+import uk.gov.di.ipv.core.library.domain.ContraIndicatorConfig;
+import uk.gov.di.ipv.core.library.domain.ContraIndicators;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.IdentityClaim;
 import uk.gov.di.ipv.core.library.domain.Name;
 import uk.gov.di.ipv.core.library.domain.NameParts;
 import uk.gov.di.ipv.core.library.domain.UserIdentity;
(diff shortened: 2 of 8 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
```diff
@@ -37,6 +38,36 @@ public boolean hasMitigations() {
                 .anyMatch(ci -> ci.getMitigation() != null && !ci.getMitigation().isEmpty());
     }
 
+    public boolean areBreaching(Map<String, ContraIndicatorConfig> ciConfig, int ciScoreThreshold)
+            throws UnrecognisedCiException {
+        return getContraIndicatorScore(ciConfig) >= ciScoreThreshold;
+    }
+
+    public List<String> ciCodesToExitCodes(Map<String, ContraIndicatorConfig> ciConfig)
```
Joe-Edwards-GDS: Could this just be `toExitCodes` (and below `toAlwaysRequiredExitCodes`)?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
```diff
@@ -37,6 +38,36 @@ public boolean hasMitigations() {
...
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    public List<String> ciCodesToAlwaysRequiredExitCodes(
+            Map<String, ContraIndicatorConfig> ciConfig, String alwaysRequiredExitCodes)
+            throws UnrecognisedCiException {
+        return ciCodesToExitCodes(ciConfig).stream()
+                .filter(alwaysRequiredExitCodes::contains)
+                .collect(Collectors.toList());
```
Joe-Edwards-GDS: I think we can just do `.toList()` these days? (and above)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -64,5 +72,6 @@ public UserIdentity(
         this.sub = sub;
         this.vot = vot;
         this.vtm = vtm;
+        this.exitCode = exitCode;
```
Joe-Edwards-GDS, line 75: (In case you were worried, I checked with orchestration and they've asserted that there won't be an issue if we start sending this before they're handling it)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -150,6 +157,22 @@ public UserIdentity generateUserIdentity(String userId, String sub, String vot)
             Optional<JsonNode> drivingPermitClaim =
                     generateDrivingPermitClaim(successfulVCStoreItems);
             drivingPermitClaim.ifPresent(userIdentityBuilder::drivingPermitClaim);
+            if (configService.enabled(EXIT_CODES)) {
+                userIdentityBuilder.exitCode(
+                        contraIndicators.ciCodesToAlwaysRequiredExitCodes(
+                                configService.getContraIndicatorConfigMap(),
+                                configService.getSsmParameter(ALWAYS_REQUIRED_EXIT_CODES)));
```
Joe-Edwards-GDS: I can't quite work out what the right answer is, but there seems to be something fishy about our abstractions - passing in the config to a method like this 🤔 

Should we actually have the `toExitCodes` methods in the `UserIdentityService` or `CimitService`, and just use `contraIndicators.getCiCodes()` here? Similarly with the calls below.

Maybe I'm used to a much more functional view of the world, and not as used to having model classes with business logic!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
Wynndow: They were going to be my next port of call! Good to know.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Yeah. I went back and forth on this a few times. Originally I was going to just pass in the config service, but that felt too tightly coupled. But I also don't love passing this stuff in.

The `ContraIndicators` class is more of a wrapper around a map, that extends its capabilities. The actual model is the `ContraIndicator` class that the map contains. I agree it's a bit weird.

Not sure I'd want to use the CimitService to do this, it's more about interfacing with the Cimit component.

I'll move the conversion to exit codes into the user identity service, see how it looks.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -150,11 +159,49 @@ public UserIdentity generateUserIdentity(String userId, String sub, String vot)
...
+            if (configService.enabled(EXIT_CODES)) {
+                userIdentityBuilder.exitCode(toAlwaysRequiredExitCode(contraIndicators));
+            }
+        } else {
+            if (configService.enabled(EXIT_CODES)
+                    && contraIndicators.getContraIndicatorScore(
+                                    configService.getContraIndicatorConfigMap())
+                            >= Integer.parseInt(
+                                    configService.getSsmParameter(CI_SCORING_THRESHOLD))) {
+                userIdentityBuilder.exitCode(toExitCode(contraIndicators));
```
Joe-Edwards-GDS, line 171: Interested to know what you think, but having these methods here feels cleaner to me :)

A minor point, but looking forward to the next ticket, maybe we should extract the breaching check as well, so we have two methods for `getSuccessExitCode` and `getFailedExitCode`, rather than being CI-specific? Happy for you to work it out there instead if you prefer.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Have pulled the `toExitCode` methods in now if you want to take another look.

Joe-Edwards-GDS (approved):

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Yeah, I agree this feels a lot cleaner. No need for the `ContraIndicators` class to know anything about exit codes. Less faff with calling config service and passing stuff through.

Not sure I follow on the breaching check - one for a face to face chat? Will get this in though and address in the next ticket I think.

PR #1320, opened 2023-11-02
PYIC-3869: Extend featureSets to backchannel lambdas

Wynndow:
### What changed

Extend featureSets to backchannel lambdas

### Why did it change

Feature sets are set in the config service by a value passed through in the request. This value comes from the users frontend session.

Lambdas that are called directly by clients don’t have access to the value from the frontend session, meaning that the feature sets can’t be used.

We can store the feature set in the users backend session at the last appropriate moment - in the build-client-oauth-response handler. This makes it available in those lambdas not called as part of a users request.

### Issue tracking

- [PYIC-3869](https://govukverify.atlassian.net/browse/PYIC-3869)

[PYIC-3869]: https://govukverify.atlassian.net/browse/PYIC-3869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
+++ b/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
@@ -124,6 +124,9 @@ public class BuildClientOauthResponseHandler
                         HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_SESSION_ID);
             }
 
+            ipvSessionItem.setFeatureSet(featureSet);
+            sessionService.updateIpvSession(ipvSessionItem);
+
             LogHelper.attachIpvSessionIdToLogs(ipvSessionId);
             LogHelper.attachClientSessionIdToLogs(clientOAuthSessionItem.getClientOAuthSessionId());
             LogHelper.attachClientIdToLogs(clientOAuthSessionItem.getClientId());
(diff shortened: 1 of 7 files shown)


lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
```diff
@@ -124,6 +124,9 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
                         HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_SESSION_ID);
             }
 
+            ipvSessionItem.setFeatureSet(featureSet);
+            sessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS, line 128: So the idea of this change is that we set the featureset on the session as we're building the OAuth response, then it'll be available to the two other endpoints that Orchestration will ultimately call (issue access token and build user identity), because they're called by orchestration (or its stub), rather than core-front?

Given that this introduces a slight inconsistency between stages in the journey, I'm thinking we should either:
- Make it work more uniformly - i.e. we update the featureset on the session _every_ time we kick off a lambda that is passed a feature set, and use that as the source of truth for the config service.
- Add a comment somewhere (perhaps on `IpvSessionItem`?) that says we _only_ set this at the very end of the journey - i.e. it's not suitable for generally working out what featureset is active.

I don't hugely mind which :)

Wynndow: The problem with the first option is that we have to retrieve the session in a lambda before we can make use of a feature set. If we carry on doing what we do we can set it right at the start of the lambda, before we do anything to get the session. That might be useful if we have features that require doing stuff before we get the session.

I'll stick a comment in.

Joe-Edwards-GDS (approved):

PR #1321, opened 2023-11-02
Pyic 3750 nino claim in identity

DanCorderIPV:
### What changed

If a successful NINO VC exists for a user when generating their UserIdentity a new claim containing the users NINO will be added to the UserIdentity 

### Why did it change

So the RPs can see the user's NINO

### Issue tracking

- [PYIC-3750](https://govukverify.atlassian.net/browse/PYIC-3750)

I have tested journeys both with and without the NINO CRI. When the NINO CRI isn't part of the journey the generated UserIdentity has no `https://vocab.account.gov.uk/v1/socialSecurityRecord` section at all. When the NINO CRI is in the journey the `https://vocab.account.gov.uk/v1/socialSecurityRecord` property exists and has the correct contents.

As the NINO CRI is behind a feature flag already this change shouldn't need it's own flag.

Also both the orchestrator and SPoT teams believe that even if the new claim was present it wouldn't cause them any issues.

[PYIC-3750]: https://govukverify.atlassian.net/browse/PYIC-3750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -48,11 +48,7 @@
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CORE_VTM_CLAIM;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EXIT_CODES;
 import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.USER_ISSUED_CREDENTIALS_TABLE_NAME;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.ADDRESS_CRI;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.BAV_CRI;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.DCMAW_CRI;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.DRIVING_LICENCE_CRI;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.PASSPORT_CRI;
+import static uk.gov.di.ipv.core.library.domain.CriConstants.*;
```
Joe-Edwards-GDS, line 51: It looks like elsewhere we're avoiding wildcard imports in favour of explicitly specifying them - might be worth keeping that consistent?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -333,131 +335,166 @@ public Optional<IdentityClaim> findIdentityClaim(List<VcStoreItem> vcStoreItems)
...
+                            SignedJWT.parse(ninoCredentialItem.getCredential())
+                                    .getPayload()
+                                    .toString())
+                    .path(VC_CLAIM)
+                    .path(VC_CREDENTIAL_SUBJECT)
+                    .path(criName);
+        } catch (JsonProcessingException | ParseException e) {
+            LOGGER.error("{}: '{}'", errorLog, e.getMessage());
+            throw new HttpResponseExceptionWithErrorBody(
+                    HttpStatus.SC_INTERNAL_SERVER_ERROR, errorResponse);
```
Joe-Edwards-GDS, line 492: I'm not sure if there's a sensible refactoring, but this is looking very similar to the `getVCClaimNode` method that @lwebb-iig modified recently.

Do you think there's scope to handle these more uniformly?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -333,131 +335,166 @@ public Optional<IdentityClaim> findIdentityClaim(List<VcStoreItem> vcStoreItems)
...
+    private Optional<VcStoreItem> findStoreItem(
+            List<String> criNames, List<VcStoreItem> vcStoreItems) {
+        return vcStoreItems.stream()
+                .filter(credential -> criNames.contains(credential.getCredentialIssuer()))
+                .findFirst();
+    }
+
+    private JsonNode extractCriNodeFromCredential(
+            String criName,
+            VcStoreItem ninoCredentialItem,
```
Joe-Edwards-GDS: should just be `credentialItem`

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: (and elsewhere)

DanCorderIPV: I didn't do this explicitly, not sure if it was spotless or IDEA but it seemed sensible. I can look into reversing it.

Joe-Edwards-GDS: Thanks - I expect it's IDEA (Mike had a similar thing)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
DanCorderIPV: I think it does make sense, the only difference is the error handling.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -333,131 +336,160 @@ public Optional<IdentityClaim> findIdentityClaim(List<VcStoreItem> vcStoreItems)
...
+
+    private Optional<VcStoreItem> findStoreItem(
+            List<String> criNames, List<VcStoreItem> vcStoreItems) {
+        return vcStoreItems.stream()
+                .filter(credential -> criNames.contains(credential.getCredentialIssuer()))
+                .findFirst();
+    }
+
+    private JsonNode extractCriNodeFromCredential(
+            String criName,
```
Joe-Edwards-GDS: Sorry, another small thing - this parameter isn't actually the CRI name, it's the property you're extracting. Perhaps `extractSubjectClaim()` or similar?

Joe-Edwards-GDS (dismissed):

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -274,9 +261,7 @@ void checkNameCorrelationWithMissingNameWithAddressCredentialsForReturnTrue() th
         List<VcStoreItem> vcStoreItems =
```
DanCorderIPV, line 261: I don't understand the name of this test so not sure what it is testing, but this is the change that was made on the main branch (removing the passport VC).

Joe-Edwards-GDS: I think the intent is that name correlation is happy to skip address credentials (which don't include a name)

Joe-Edwards-GDS (approved):

PR #1329, opened 2023-11-03
PYIC-3862: Upgrade the log into a correlation failure, unless we see any other legitimate instances where a VC is missing name/dob.

TalatAydinGDS:
### What changed

Upgrade the log into a correlation failure, unless we see any other legitimate instances where a VC is missing name/dob

### Why did it change

In accordance with GPG45 rules, except for exceptios, the 'name' and 'birthdate' should not empty or null. We throws exception to catch them.

### Issue tracking

- [PYIC-3862](https://govukverify.atlassian.net/browse/PYIC-3862)

### Other considerations

[PYIC-3862]: https://govukverify.atlassian.net/browse/PYIC-3862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -68,8 +68,9 @@ public class UserIdentityService {
     private static final List<String> DRIVING_PERMIT_CRI_TYPES =
             List.of(DCMAW_CRI, DRIVING_LICENCE_CRI);
 
-    public static final List<String> CRI_TYPES_EXCLUDED_FOR_NAME_CORRELATION = List.of(ADDRESS_CRI);
-    public static final List<String> CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION =
+    private static final List<String> CRI_TYPES_EXCLUDED_FOR_NAME_CORRELATION =
+            List.of(ADDRESS_CRI);
+    private static final List<String> CRI_TYPES_EXCLUDED_FOR_DOB_CORRELATION =
             List.of(ADDRESS_CRI, BAV_CRI);
 
     private static final Logger LOGGER = LogManager.getLogger();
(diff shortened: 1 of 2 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -510,6 +511,9 @@ private List<IdentityClaim> getIdentityClaimsForBirthDateCorrelation(
                     continue;
                 }
                 addLogMessage(item, "Birthdate property is missing from VC");
+                throw new HttpResponseExceptionWithErrorBody(
```
rpayal, line 514: Is method isBirthDateEmpty() be private ?

MikeCollingwood: I think CredentialParseException could make sense here

rpayal (approved):
LGTM

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -510,13 +511,16 @@ private List<IdentityClaim> getIdentityClaimsForBirthDateCorrelation(
...
+                        HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                        ErrorResponse.FAILED_BIRTHDATE_CORRELATION);
```
Joe-Edwards-GDS, line 516: I hadn't really considered, but worth noting that this means we treat missing fields differently to other correlation failures - in practice meaning that a missing birthdate will result in a technical error, but an incorrect birthdate will result in PYI_NO_MATCH.

That seems ok to me though - if a CRI starts omitting one of these fields, it's a bug that we want to notice!

PR #1330, opened 2023-11-03
PYIC-3661: Add context and scope to journey map

MikeCollingwood:
### What changed

Example:
Screenshot 2023-11-03 at 15 17 51

### Why did it change

- To visualise context and scope in the journey map, to be better informed when using the reference
- Was made to be simple, instead of using a tooltip, because this is sufficient and less code

### DIFF
--- a/journey-map/public/index.mjs
+++ b/journey-map/public/index.mjs
@@ -166,7 +166,9 @@ const renderStates = (journeyMap, states) => {
                     ? `    ${state}[${state}\\n${definition.response.pageId}]:::error_page`
                     : `    ${state}[${state}\\n${definition.response.pageId}]:::page`;
             case 'cri':
-                return `    ${state}([${state}\\n${definition.response.criId}]):::cri`;
+                let contextInfo = definition.response.context ? `\\n context: ${definition.response.context}` : "";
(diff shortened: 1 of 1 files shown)


journey-map/public/index.mjs
```diff
@@ -166,7 +166,9 @@ const renderStates = (journeyMap, states) => {
                     ? `    ${state}[${state}\\n${definition.response.pageId}]:::error_page`
                     : `    ${state}[${state}\\n${definition.response.pageId}]:::page`;
             case 'cri':
-                return `    ${state}([${state}\\n${definition.response.criId}]):::cri`;
+                let contextInfo = definition.response.context ? `\\n context: ${definition.response.context}` : "";
+                let scopeInfo = definition.response.scope ? `\\n scope: ${definition.response.scope}` : "";
```
Joe-Edwards-GDS: Prefer `const` unless you need a mutable variable :)

Joe-Edwards-GDS (commented):
It's not that beautiful, but probably fine for the time being :)

I think a redesign is a larger piece of work, that covers more than just this.

journey-map/public/index.mjs
MikeCollingwood: Ah yep!

Joe-Edwards-GDS (approved):

PR #1348, opened 2023-11-09
PYIC-3924: Return exit code for non breaching failures

Wynndow:
There is an outstanding question here about if the new scenario should *only* return the specified exit code, or if it should include any details of CIs the user may have that are still below the threshold.

### What changed

Return exit code for non breaching failures

### Why did it change

This extends the exit code functionality to return a specified exit code in the scenario that a user has failed to prove their identity (`P0`), but is not breaching the CI threshold limit. This would usually indicate that the user has chosen not to finish the journey. Possible reasons being they've go no documents, or don't want to/can't use the app or F2F journeys.

This is distinct from the scenario where a user abandons their journey entirely, which is not in scope.

### Issue tracking

- [PYIC-3924](https://govukverify.atlassian.net/browse/PYIC-3924)

[PYIC-3924]: https://govukverify.atlassian.net/browse/PYIC-3924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -43,9 +43,10 @@ import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.ALWAYS_REQUIRED_EXIT_CODES;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CI_SCORING_THRESHOLD;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CORE_VTM_CLAIM;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EXIT_CODES_ALWAYS_REQUIRED;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EXIT_CODES_NON_CI_BREACHING_P0;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EXIT_CODES;
 import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.USER_ISSUED_CREDENTIALS_TABLE_NAME;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.ADDRESS_CRI;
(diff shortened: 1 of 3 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -203,11 +214,9 @@ private List<String> toExitCode(ContraIndicators contraIndicators)
                 .toList();
     }
 
-    private List<String> toAlwaysRequiredExitCode(ContraIndicators contraIndicators)
-            throws UnrecognisedCiException {
-        return toExitCode(contraIndicators).stream()
-                .filter(configService.getSsmParameter(ALWAYS_REQUIRED_EXIT_CODES)::contains)
-                .toList();
+    private boolean breachingCiThreshold(ContraIndicators contraIndicators) {
```
Joe-Edwards-GDS, line 217: It feels a bit weird for this method to be here (surely we check breaching CI threshold elsewhere too), but perhaps it's not easy to share - and it's simple enough not to be a big deal.

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -1088,6 +1089,30 @@ void generateUserIdentityShouldNotSetExitCodeWhenBreachingCiThresholdAndExitCode
...
+                        .contraIndicatorsMap(
+                                Map.of("X01", ContraIndicator.builder().code("X01").build()))
+                        .build();
+
+        UserIdentity userIdentity =
+                userIdentityService.generateUserIdentity(
+                        USER_ID_1, "test-sub", "P0", contraIndicators);
+
+        assertEquals(List.of("🐧"), userIdentity.getExitCode());
+    }
```
Joe-Edwards-GDS, line 1116: Do you think it'd be useful to make explicit that it _doesn't_ include an 'always included' code either?

Joe-Edwards-GDS (dismissed):

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: It does feel like something we should "know" at this point from the place where we do this check - ci-scoring. But we don't store the result in the users session or anything. Perhaps we could - but then i have fears about that leading to hard to detect bugs later down the line when relying on it.

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
Wynndow: I supposed I could assert that there's no call to the config service to retrieve the "always included" code. I can't assert it's not directly in the returned list without defining what that code actually is (in this test case). The only way to do that is to mock the response for that call to the config service. But then if we don't actually call it (which we don't), then the tests get angry about unnecessary stubbings 🙂.

I'll add the negative assertion I think...

Joe-Edwards-GDS: Ah, yeah that's a bit annoying, but makes sense, thanks.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Yeah - I don't think we should try and cache it (since it's such a simple calculation), I was more thinking about sharing the code, so we have one method for checking CI score breaches.

It's one integer comparison though, so not that fussed about it :)

Joe-Edwards-GDS (approved):

PR #1351, opened 2023-11-09
PYIC-3947: Remove exit codes feature flag

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-core-common-infra/pull/766**

### What changed

Remove exit codes feature flag

### Why did it change

[PYIC-3947: Remove EXIT_CODES feature flag](https://github.com/govuk-one-login/ipv-core-back/commit/54954f9282ba94860344bdafce56ab13579cdfab) 

The flag is set to true all the way out to production. We can now remove
the flag itself.

This also removes the `NON_NULL` annotation from the exit code field in
the UserIdentity class. It was only there to prevent the field being
returned to orchestrator before we were ready.

[BAU: Remove use of wildcard imports](https://github.com/govuk-one-login/ipv-core-back/commit/2584977fd9df1d8fbea5813927e4833966f4412a) 

We don't use wildcard imports as a rule. So I've removed them. But I'm
open to being convinced otherwise. This makes a reasonably compelling
case to use them: https://www.baeldung.com/java-wildcard-imports

### Issue tracking

- [PYIC-3947](https://govukverify.atlassian.net/browse/PYIC-3947)

[PYIC-3947]: https://govukverify.atlassian.net/browse/PYIC-3947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
+++ b/lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
@@ -207,36 +207,6 @@ class BuildUserIdentityHandlerTest {
     }
 
     @Test
-    void shouldNotReturnExitCodeAttributeIfNullInUserIdentity() throws Exception {
-        APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
-        AccessToken accessToken = new BearerAccessToken(TEST_ACCESS_TOKEN);
-        Map<String, String> headers =
-                Map.of(
-                        "Authorization",
-                        accessToken.toAuthorizationHeader(),
-                        "ip-address",
-                        TEST_IP_ADDRESS);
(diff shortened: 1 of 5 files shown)


Joe-Edwards-GDS (approved):
Sounds like this might already be live anyway, but I assume we've tested that orchestration definitely doesn't fall over :)

Wynndow:
> Sounds like this might already be live anyway, but I assume we've tested that orchestration definitely doesn't fall over :)

This is just me getting my ducks in a row for after we've updated the config to turn the feature on all the way out to prod. Need to wait till the audit event work is in, and QA has given the nod to start using the thing.

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS:
This will now conflict with https://github.com/govuk-one-login/ipv-core-back/pull/1352

Joe-Edwards-GDS (approved):

PR #1355, opened 2023-11-13
BAU: Add orphan states option to journey map render

Wynndow:
### What changed

Add orphan states option to journey map render

### Why did it change

This adds a new option to the rendering of the journey map to show orphan states - states that have no other state targeting them.

We only have two states that we expect to be in this state - the `INITIAL_IPV_JOURNEY` state and the `CORE_SESSION_TIMEOUT` state. Both of these states are set directly in the core back code, rather than being transitioned to in the usual flow.

The `INITIAL_IPV_JOURNEY` state has been handled, but the `CORE_SESSION_TIMEOUT` state has been left in. Perhaps we should also exclude it...

This also removes the `PYI_KBV_THIN_FILE` state, as the new functionality showed it was an orphan.

### DIFF
--- a/journey-map/public/index.mjs
+++ b/journey-map/public/index.mjs
@@ -113,39 +113,20 @@ const resolveEventTarget = (definition, formData) => {
     return definition.targetState;
 };
 
-const resolveAllPossibleEventTargets = (eventDefinition) => {
-    let targets = [];
-    targets.push(eventDefinition.targetState);
-    if (eventDefinition.checkIfDisabled) {
-        Object.values(eventDefinition.checkIfDisabled).forEach((def) => {
-            targets.push(...resolveAllPossibleEventTargets(def));
-        })
-    }
-
-    if (eventDefinition.checkFeatureFlag) {
-        Object.values(eventDefinition.checkFeatureFlag).forEach((def) => {
-            targets.push(...resolveAllPossibleEventTargets(def));
-        })
-    }
-
-    return targets;
-};
+const resolveAllPossibleEventTargets = (eventDefinition) => [
+    eventDefinition.targetState,
+    ...Object.values(eventDefinition.checkIfDisabled || {}).flatMap(resolveAllPossibleEventTargets),
+    ...Object.values(eventDefinition.checkFeatureFlag || {}).flatMap(resolveAllPossibleEventTargets)
+];
 
 const calcOrphanStates = (journeyMap) => {
-    const targetedStates = [...initialStates]
-
-    Object.values(journeyMap).forEach((stateDefinition) => {
-        if (stateDefinition.events) {
-            Object.values(stateDefinition.events).forEach((eventDefinition) => {
-                targetedStates.push(...resolveAllPossibleEventTargets(eventDefinition));
-            })
-        }
-        if (stateDefinition.exitEvents) {
-            Object.values(stateDefinition.exitEvents).forEach((eventDefinition) => {
-                targetedStates.push(...resolveAllPossibleEventTargets(eventDefinition));
-            })
-        }
-    })
+    const targetedStates = [
+        ...initialStates,
+        ...Object.values(journeyMap).flatMap((stateDefinition) => [
+            ...Object.values(stateDefinition.events || {}).flatMap(resolveAllPossibleEventTargets),
+            ...Object.values(stateDefinition.exitEvents || {}).flatMap(resolveAllPossibleEventTargets)
+        ])
+    ]
 
     const uniqueTargetedStates = [...new Set(targetedStates)];
 
@@ -226,12 +207,9 @@ export const render = (journeyMap, nestedJourneys, formData = new FormData()) =>
     }
     expandParents(journeyMapCopy);
 
-    let { transitionsMermaid, states } = renderTransitions(journeyMapCopy, formData);
-
-    if (formData.has('onlyOrphanStates')) {
-        states = calcOrphanStates(journeyMapCopy)
-        transitionsMermaid = '';
-    }
+    const { transitionsMermaid, states } = formData.has('onlyOrphanStates')
+        ? { transitionsMermaid: '', states: calcOrphanStates(journeyMapCopy) }
+        : renderTransitions(journeyMapCopy, formData);
 
     const { statesMermaid } = renderStates(journeyMapCopy, states);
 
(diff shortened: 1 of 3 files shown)


journey-map/public/index.mjs
```diff
@@ -187,7 +226,13 @@ export const render = (journeyMap, nestedJourneys, formData = new FormData()) =>
     }
     expandParents(journeyMapCopy);
 
-    const { transitionsMermaid, states } = renderTransitions(journeyMapCopy, formData);
+    let { transitionsMermaid, states } = renderTransitions(journeyMapCopy, formData);
+
+    if (formData.has('onlyOrphanStates')) {
+        states = calcOrphanStates(journeyMapCopy)
+        transitionsMermaid = '';
+    }
```
Joe-Edwards-GDS: You may have an opinion on which is clearer, but you can also do this in one shot (to save you rendering the normal transitions if they aren't needed):

```

    const { transitionsMermaid, states } = formData.has('onlyOrphanStates')
        ? { transitionsMermaid: '', states: calcOrphanStates(journeyMapCopy) }
        : renderTransitions(journeyMapCopy, formData);
```

journey-map/public/index.mjs
```diff
@@ -113,6 +113,45 @@ const resolveEventTarget = (definition, formData) => {
...
 };
 
+const resolveAllPossibleEventTargets = (eventDefinition) => {
+    let targets = [];
+    targets.push(eventDefinition.targetState);
+    if (eventDefinition.checkIfDisabled) {
+        Object.values(eventDefinition.checkIfDisabled).forEach((def) => {
+            targets.push(...resolveAllPossibleEventTargets(def));
+        })
+    }
```
Joe-Edwards-GDS: You can simplify this a little if you like:
```
targets.push(...Object.values(eventDefinition.checkIfDisabled || {}).flatMap(resolveAllPossibleEventTargets));
```

Or combining it all together:
```
const resolveAllPossibleEventTargets = (eventDefinition) => [
    eventDefinition.targetState,
    ...Object.values(eventDefinition.checkIfDisabled || {}).flatMap(resolveAllPossibleEventTargets),
    ...Object.values(eventDefinition.checkFeatureFlag || {}).flatMap(resolveAllPossibleEventTargets),
];
```

journey-map/public/index.mjs
```diff
@@ -113,6 +113,45 @@ const resolveEventTarget = (definition, formData) => {
...
+            })
+        }
+        if (stateDefinition.exitEvents) {
+            Object.values(stateDefinition.exitEvents).forEach((eventDefinition) => {
+                targetedStates.push(...resolveAllPossibleEventTargets(eventDefinition));
+            })
+        }
+    })
+
+    const uniqueTargetedStates = [...new Set(targetedStates)];
```
Joe-Edwards-GDS, line 131: Extending the idea above, you could do this in one go as well.
```
const targetedStates = [
    ...initialStates,
    ...Object.values(journeyMap).flatMap((stateDefinition) => [
        ...Object.values(stateDefinition.events || {}).flatMap(resolveAllPossibleEventTargets),
        ...Object.values(stateDefinition.exitEvents || {}).flatMap(resolveAllPossibleEventTargets),
    ]),
];
```
(You could even inline the `Set`, but that feels like a distinct thing)

Joe-Edwards-GDS: Using functional methods effectively is one of the things I love doing :D - and one reason I like TS, as it works pretty well for it, much better than Java streams.

(My favourite language used to be Scala, which is even better :P - but haven't written any in years)

Joe-Edwards-GDS (commented):
You wanted me to comment on your JavaScript right :D

One question I have - is this the _right_ thing to be doing? We could also have a checkbox for 'include orphaned states' which would be much easier to implement. I suppose it makes it a bit harder to spot them, perhaps.

journey-map/public/index.mjs
Wynndow: I think the ternary is a little less clear in this scenario but it does solve something I was hoping you'd comment on.
The way I have it currently I'm mutating `transitionsMermaid` and `states` after the call to `renderTransitions()`. That force me to use `let` instead of `const`.
I was trying for a little while to do something like:
```
    let states;
    let transitionsMermaid;
    if (formData.has('onlyOrphanStates')) {
        states = calcOrphanStates(journeyMapCopy)
        transitionsMermaid = '';
    } else {
        { transitionsMermaid, states } = renderTransitions(journeyMapCopy, formData);
    }
```
But the else block doesn't work - variable might not have been initialized. I think the brace return thing is throwing me a little. How would you handle something like this? I guess use the ternary?

journey-map/public/index.mjs
Wynndow: ooOOooo....nice

journey-map/public/index.mjs
Wynndow: I only have one experience with Scala and that was years ago writing Gatling tests. And it burned me.

Wynndow:
> You wanted me to comment on your JavaScript right :D
> 
> One question I have - is this the _right_ thing to be doing? We could also have a checkbox for 'include orphaned states' which would be much easier to implement. I suppose it makes it a bit harder to spot them, perhaps.

I'm not sure that including orphan states would be _that_ useful. The motivation for doing this was to prevent myself having to manually check every state in the map to see if it was an orphan, and then remove any that were. I'm just interested in knowing what I can get rid of. If the all the other states were on the screen at the same time it's going to get confusing.

journey-map/public/index.mjs
Joe-Edwards-GDS: This is a case I had to look up, actually! Turns out you have to use an extra pair of brackets...
```
} else {
    ({ transitionsMermaid, states } = renderTransitions(journeyMapCopy, formData));
}
```
I prefer the ternary tbh - it makes it clear that both are being set in both branches

Joe-Edwards-GDS (dismissed):
I think this is ok, but will reapprove if you'd like to switch up the last bit again

journey-map/public/index.mjs
Wynndow: Ah yeah, of course, extra brackets! 🫤

But overall the ternary does seem like the best bet

Joe-Edwards-GDS: (I suppose making the `calcOrphanStates` method also return a blank `transitionsMermaid` would also be a neat, but not very satisfying, solution)

Joe-Edwards-GDS (approved):

PR #1357, opened 2023-11-13
PYIC-2892: Support loading CRI configuration from new connection parameter layout

lwebb-iig:
Support loading CRI configuration from new connection parameter layout while maintaining support for old parameter layout.

## Proposed changes

Support the adjustment of the layout of the connection parameters in the CRI configuration. New parameter layout for connection information is as follows:
```json
"{\"tokenUrl\":\"https://testTokenUrl\",\"credentialUrl\":\"https://testCredentialUrl\",\"authorizeUrl\":\"https://testAuthoriseUrl\",\"clientId\":\"ipv-core-test\",\"signingKey\":"\"{\\\"kty\\\":\\\"RSA\\\",\\\"e\\\":\\\"AQAB\\\",\\\"n\\\":\\\"vyapkvJXLwpYRJjbkQD99V2gcPEUKrO3dwjcAA9TPkLucQEZvYZvb7-wfSHxlvJlJcdS20r5PKKmqdPeW3Y4ir3WsVVeiht2iOZUreUO5O3V3o7ImvEjPS_2_ZKMHCwUf51a6WGOaDjO87OX_bluV2dp01n-E3kiIl6RmWCVywjn13fX3jsX0LMCM_bt3HofJqiYhhNymEwh39oR_D7EE5sLUii2XvpTYPa6L_uPwdKa4vRl4h4owrWEJaJifMorGcvqhCK1JOHqgknN_3cb_ns9Px6ynQCeFXvBDJy4q71clkBq_EZs5227Y1S222wXIwUYN8w5YORQe3M-pCIh1Q\\\"}\"",\"encryptionKey\":"\"{\\\"kty\\\":\\\"EC\\\",\\\"d\\\":\\\"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU\\\",\\\"crv\\\":\\\"P-256\\\",\\\"x\\\":\\\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\\\",\\\"y\\\":\\\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\\\"}\"",\"componentId\":\"https://testComponentId\",\"clientCallbackUrl\":\"https://testClientCallBackUrl\",\"requiresApiKey\":\"true\"}"
```

### What changed

Handles loading connection information from a single parameter.
Updates to the config service to first check if the new single parameter exists and if not, fall back to existing/old handling.

### Why did it change

By compressing connection information into a single parameter, we hope to reduce the number of CRI configuration parameter lookups on lambda cold starts

### Issue tracking

- [PYIC-2892](https://govukverify.atlassian.net/browse/PYIC-2892)

- [x] Documented in the [README](./blob/main/README.md)
- [x] Added to deployment repository

### Other considerations

[PYIC-2892]: https://govukverify.atlassian.net/browse/PYIC-2892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -251,29 +251,34 @@ public String getCriPrivateApiKey(CriOAuthSessionItem criOAuthSessionItem) {
...
-    public CredentialIssuerConfig getCriConfig(CriOAuthSessionItem criOAuthSessionItem) {
+    public CredentialIssuerConfig getCriConfig(CriOAuthSessionItem criOAuthSessionItem)
+            throws JsonProcessingException {
         return getCriConfigForConnection(
                 criOAuthSessionItem.getConnection(), criOAuthSessionItem.getCriId());
     }
 
-    public CredentialIssuerConfig getCriConfigForConnection(String connection, String criId) {
+    public CredentialIssuerConfig getCriConfigForConnection(String connection, String criId)
+            throws JsonProcessingException {
```
Joe-Edwards-GDS: If we never expect this to happen, I think it'd be fine to catch and rethrow some kind of `RuntimeException` (logging appropriately, of course). Saves an awful lot of checked exception plumbing.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CredentialIssuerConfig.java
```diff
@@ -17,8 +19,13 @@ public class CredentialIssuerConfig {
     private URI credentialUrl;
     private URI authorizeUrl;
     private String clientId;
+
+    @JsonDeserialize(using = KeepAsJsonDeserializer.class)
```
Joe-Edwards-GDS: I think this is an interesting question - should we embed the JWKs directly into the CRI config JSON, or should we double-encode it as a string?

I'd probably have a slight preference towards the double-encoding if we're treating as an opaque string in our code, as it's simpler to implement and more representative of how we're using it (we don't expect to edit the key content ourselves).

lwebb-iig: Makes sense - I'll go with the double encoding route and handle accordingly

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -264,16 +264,27 @@ public CredentialIssuerConfig getCriConfig(CriOAuthSessionItem criOAuthSessionIt
...
+        }
+        Map<String, String> parameters = getSsmParameters(pathTemplate, false, criId, connection);
+        if (parameters != null && !parameters.isEmpty()) {
+            return objectMapper.convertValue(parameters, CredentialIssuerConfig.class);
+        } else {
             throw new NoConfigForConnectionException(
                     String.format(
                             "No config found for connection: '%s' and criId: '%s'",
                             connection, criId));
         }
```
Joe-Edwards-GDS, line 287: I don't think you don't need the `else` here

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -264,16 +264,27 @@ public CredentialIssuerConfig getCriConfig(CriOAuthSessionItem criOAuthSessionIt
...
+                        "Failed to parse credential issuer configuration at parameter path {} because: {}",
+                        pathTemplate,
+                        e);
+                throw new RuntimeException(e);
+            }
+        }
+        Map<String, String> parameters = getSsmParameters(pathTemplate, false, criId, connection);
+        if (parameters != null && !parameters.isEmpty()) {
+            return objectMapper.convertValue(parameters, CredentialIssuerConfig.class);
+        } else {
```
Joe-Edwards-GDS: So the idea is that this will support either config mechanism? Does this mean the two-step deployment actually happens here - i.e. we can remove the 'old' method once the new config is in production?

Probably neater than switching via a temporary config set :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -264,16 +264,27 @@ public CredentialIssuerConfig getCriConfig(CriOAuthSessionItem criOAuthSessionIt
...
+        String parameter = getSsmParameter(resolvePath(pathTemplate, criId, connection));
+        if (parameter != null) {
+            try {
+                return objectMapper.readValue(parameter, CredentialIssuerConfig.class);
+            } catch (JsonProcessingException e) {
+                LOGGER.error(
+                        "Failed to parse credential issuer configuration at parameter path {} because: {}",
+                        pathTemplate,
+                        e);
+                throw new RuntimeException(e);
```
Joe-Edwards-GDS: I suspect Sonar will complain that this isn't a specific exception class - is there a convenient one we can use?

Joe-Edwards-GDS (dismissed):
Looks sensible I think.

Have you checked whether we're reading individual connection config values anywhere? It seems unlikely, but just want to make sure it's always being read as part of a single 'CRI connection config' bundle.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
lwebb-iig: Yes that's the idea:
- push these changes
- switch to new config / params (common-infra PR)
- remove old methods / old config

lwebb-iig:
> Looks sensible I think.
> 
> Have you checked whether we're reading individual connection config values anywhere? It seems unlikely, but just want to make sure it's always being read as part of a single 'CRI connection config' bundle.

That's a good point, as far as I'm aware we don't - I will check this when removing the old config/handling to make sure & update relevant code to pull the value out from the new configuration

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
lwebb-iig: Created a new `ConfigParseException` as there isn't one directly related to this that extends `RuntimeException`

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1358, opened 2023-11-13
Journey map selection highlights and detail

Joe-Edwards-GDS:
- Added click handlers to the journey map nodes:
  - Highlighting the nodes and connections in/out of the node
  - Displaying the response definition for the node
  - Providing a link to the page in build, if applicable
- Pulled out the CSS into a separate stylesheet
- Pulled out the main JS into a separate module

Example:

image

### DIFF
--- a/journey-map/public/index.mjs
+++ b/journey-map/public/index.mjs
@@ -180,19 +180,22 @@ const renderStates = (journeyMap, states) => {
 
         switch (definition.response?.type) {
             case 'process':
-                return `    ${state}(${state}\\n${definition.response.lambda}):::process`;
+                const processClickHandler = `click ${state} call onProcessClick(${state}, ${definition.response.lambda})`;
+                return `    ${state}(${state}\\n${definition.response.lambda}):::process\n${processClickHandler}`;
             case 'page':
+                const pageClickHandler = `click ${state} call onPageClick(${state}, ${definition.response.pageId})`;
                 return failureStates.includes(state)
-                    ? `    ${state}[${state}\\n${definition.response.pageId}]:::error_page`
-                    : `    ${state}[${state}\\n${definition.response.pageId}]:::page`;
+                    ? `    ${state}[${state}\\n${definition.response.pageId}]:::error_page\n${pageClickHandler}`
+                    : `    ${state}[${state}\\n${definition.response.pageId}]:::page\n${pageClickHandler}`;
             case 'cri':
+                const criClickHandler = `click ${state} call onCriClick(${state}, ${definition.response.criId})`;
                 const contextInfo = definition.response.context ? `\\n context: ${definition.response.context}` : "";
                 const scopeInfo = definition.response.scope ? `\\n scope: ${definition.response.scope}` : "";
-                return `    ${state}([${state}\\n${definition.response.criId}${contextInfo}${scopeInfo}]):::cri`;
+                return `    ${state}([${state}\\n${definition.response.criId}${contextInfo}${scopeInfo}]):::cri\n${criClickHandler}`;
             case 'error':
                 return `    ${state}:::error_page`
             default:
-                return `    ${state}`;
+                return `    ${state}\nclick ${state} call onOtherClick(${state})`;
         }
     });
 
--- a/journey-map/public/index.mjs
+++ b/journey-map/public/index.mjs
@@ -1,230 +1,120 @@
-const initialStates = ['INITIAL_IPV_JOURNEY'];
-const errorStates = ['ERROR'];
-const failureStates = ['PYI_KBV_FAIL', 'PYI_NO_MATCH', 'PYI_ANOTHER_WAY'];
+import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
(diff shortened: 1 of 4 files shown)


journey-map/public/index.mjs
Joe-Edwards-GDS, line 1: The diff won't be very useful unfortunately, as I renamed `index.mjs` -> `render.mjs` and moved the JS from the HTML into a new `index.mjs`...

journey-map/public/render.mjs
```diff
@@ -0,0 +1,219 @@
...
+            return `    ${state}:::other`;
+    }
+};
+
+const renderStates = (journeyMap, states) => {
+    const mermaids = states.flatMap((state) => {
+        const definition = journeyMap[state];
+        return [
+            renderState(state, definition),
+            renderClickHandler(state, definition),
```
Joe-Edwards-GDS, line 189: The changes in this file are really just adding the click handler here

journey-map/public/index.mjs
```diff
@@ -1,204 +1,120 @@
...
+    Array.from(document.getElementsByClassName(`LE-${state}`))
+        .forEach((edge) => edge.classList.add('highlight', 'incomingEdge'));
+    Array.from(document.getElementsByClassName('node'))
+        .filter((node) => node.id.startsWith(`flowchart-${state}-`))
+        .forEach((node) => node.classList.add('highlight'));
+};
 
-        switch (definition.response?.type) {
+// Set up the click handlers that mermaid binds to each node
+const setupMermaidClickHandlers = () => {
```
Joe-Edwards-GDS, line 75: This is the new bit

journey-map/public/style.css
```diff
@@ -0,0 +1,83 @@
...
+form input[type=checkbox] {
+    margin-left: 8px;
+}
+#checkbox-controls {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    gap: 16px;
+}
+#nodeInfo {
```
Joe-Edwards-GDS, line 48: New from here down

journey-map/public/style.css
```diff
@@ -0,0 +1,83 @@
...
+    flex-direction: column;
+    gap: 8px;
+}
+.header-split {
+    display: flex;
+    padding: 16px 0;
+}
+#diagramSvg .highlight.outgoingEdge {
+    stroke-width: 3px;
+    filter: drop-shadow(1px 0 0 #ffdd00) drop-shadow(-1px 0 0 #ffdd00) drop-shadow(0 1px 0 #ffdd00) drop-shadow(0 -1px 0 #ffdd00);
```
Joe-Edwards-GDS, line 74: This is a bit of a bodge really - it creates a 1px drop-shadow in each direction, which _roughly_ equates to a 1px border around the shape.

journey-map/public/index.mjs
```diff
@@ -1,204 +1,120 @@
...
+    const getDesc = (def) => {
```
Wynndow, line 76: I don't object to it, but why define the `getDesc` function within `setupMermaidClickHandlers`? Is it a way to effectively make it "private"?

journey-map/public/render.mjs
```diff
@@ -0,0 +1,219 @@
...
+
+    return { transitionsMermaid: stateTransitions.join('\n'), states };
+};
+
+const renderClickHandler = (state, definition) => {
+    if (!definition.response) {
+        definition.response = {};
+    }
+    // Click handler serializes the definition to Base64-encoded JSON to avoid escaping issues
+    return `    click ${state} call onStateClick(${JSON.stringify(state)}, ${btoa(JSON.stringify(definition.response))})`;
```
Wynndow, line 158: An unexpanded nested journey doesn't have a response defined on it, which means that when selected in the browser, you just get a `{}`. I'm not entirely sure what would be better, but it looks a little confusing. Maybe just what the nested journey is? Although that's pretty obvious from the state name currently. Maybe just the text "A nested journey"?

journey-map/public/style.css
Wynndow: You say it's a bodge, but it seems to work well!
If it saves this:

Wynndow (dismissed):
It looks good! Only real question is around the unexpanded nested journey definitions.

journey-map/public/index.mjs
Joe-Edwards-GDS: Yeah, there's no good reason really except that it felt specific to this method rather than a general thing. I could happily pull it out, but probably won't for now!

journey-map/public/render.mjs
Joe-Edwards-GDS: Yeah that's a good point - it'd be nice to describe a bit more what it's doing at the very least.

One thing we could do is add a 'description' property to nested journeys, which display here - I was imagining that in future each node could actually talk about what the thing is (e.g. what is the 'CIC CRI', and maybe a link to some other documentation).

I will probably leave that for another PR, unless you think it'll be really useful now.

Wynndow: No, leave it for another PR I think.

Wynndow (approved):

PR #1367, opened 2023-11-14
PYIC-3796: Add consolidated CRI return lambda

MikeCollingwood:
### What changed

Merge lambdas and create 3 services:
- CriApiService
- VcStoringService
- CriCheckingService

### Why did it change

- To simplify things in the long-run

### Issue tracking

- [PYIC-3798](https://govukverify.atlassian.net/browse/PYIC-3798)

### Other considerations

[PYIC-3798]: https://govukverify.atlassian.net/browse/PYIC-3798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS:
You'll also need to plumb this in to the CRI return API, so it actually gets called.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,344 @@
...
+    CriOAuthSessionItem criOAuthSessionItem = null;
+    ClientOAuthSessionItem clientOAuthSessionItem;
+    String criId;
+    String criOAuthSessionId;
+    String ipvSessionId;
+    String ipAddress;
+    String featureSet;
+    String error;
+    String errorDescription;
+    CredentialIssuerConfig criConfig;
```
Joe-Edwards-GDS: These feel weird to have as instance variables - could they just be local variables in the handler?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,344 @@
...
+    @Logging(clearState = true)
+    public Map<String, Object> handleRequest(CriCallbackRequest callbackRequest, Context context) {
+        try {
+            // Extract Callback Attributes
+            var authorisationCode = getValidAuthorisationCode(callbackRequest);
+            criId = getValidCriId(callbackRequest);
+            criOAuthSessionId = getValidCriOauthSessionId(callbackRequest);
+            ipvSessionId = getValidIpvSessionId(callbackRequest);
+            featureSet = callbackRequest.getFeatureSet();
+            error = callbackRequest.getError();
```
Joe-Edwards-GDS: If you've got an object containing the featureset etc., do we need to extract them into class-level variables at all?

I wonder if this should instead be one method to validate the callbackRequest (i.e. that all the required fields are not blank), and then you can access all the properties on that?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,344 @@
...
+
+            vcStoringService.storeVcResponse(
+                    vcResponse, callbackRequest, clientOAuthSessionItem, criConfig, ipvSessionItem);
+
+            var journeyResponse =
+                    criCheckingService.checkVcResponse(
+                            vcResponse, callbackRequest, clientOAuthSessionItem, ipvSessionItem);
+
+            // Update state (should I have shifted update to end?)
+            ipvSessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS: The main risk with shifting to the end is what happens if there is an error between the session object changing and this call? It looks like we update the session in some of the `catch` blocks as well, but we could consider moving to a `finally` if it should just always happen.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+                criConfig.getRequiresApiKey()
+                        ? configService.getCriPrivateApiKey(criOAuthSessionItem)
+                        : null;
+
+        var accessToken = exchangeCodeForToken(authorisationCode, criConfig, apiKey, criId);
+
+        return getVerifiableCredentialResponse(accessToken, criConfig, apiKey, criId);
+    }
+
+    public BearerAccessToken exchangeCodeForToken(
```
Joe-Edwards-GDS: Perhaps call this `fetchAccessToken`?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+    private final JWSSigner signer;
+
+    @ExcludeFromGeneratedCoverageReport
+    public CriApiService(ConfigService configService, JWSSigner signer) {
+        this.configService = configService;
+        this.signer = signer;
+    }
+
+    public VerifiableCredentialResponse getVcResponse(
+            String authorisationCode,
```
Joe-Edwards-GDS: I think we should be consistent here, and either:
- Pass `callbackRequest` and fish the auth code and CRI ID from it
- Pass both `authorisationCode` and `criId` explicitly

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+            CriCallbackRequest callbackRequest,
+            CriOAuthSessionItem criOAuthSessionItem,
+            CredentialIssuerConfig criConfig) {
```
Joe-Edwards-GDS: Can we get this from the `configService`, using the `criOauthSessionItem`?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
Joe-Edwards-GDS: Can this (and below) be `private`?

Might be convenient to have them public for testing idk

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+            return token;
+        } catch (IOException | ParseException | JOSEException e) {
+            LOGGER.error("Error exchanging token: {}", e.getMessage(), e);
+            throw new AuthCodeToAccessTokenException(
+                    HTTPResponse.SC_SERVER_ERROR,
+                    ErrorResponse.FAILED_TO_EXCHANGE_AUTHORIZATION_CODE);
+        }
+    }
+
+    public VerifiableCredentialResponse getVerifiableCredentialResponse(
```
Joe-Edwards-GDS: Should probably be named consistently with the method above - perhaps `fetchVerifiableCredential`?

Note that it's a POST request, so using get here is potentially misleading!

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+                        HTTPResponse.SC_BAD_REQUEST, ErrorResponse.INVALID_TOKEN_REQUEST);
+            }
+
+            BearerAccessToken token =
+                    tokenResponse.toSuccessResponse().getTokens().getBearerAccessToken();
+            LOGGER.info("Auth Code exchanged for Access Token.");
+            return token;
+        } catch (IOException | ParseException | JOSEException e) {
+            LOGGER.error("Error exchanging token: {}", e.getMessage(), e);
+            throw new AuthCodeToAccessTokenException(
```
Joe-Edwards-GDS: These two methods could probably just throw the same sort of exception - maybe `CriApiException` for issues with the API itself?

(The error message should make clear which endpoint failed, of course)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+        try {
+            HTTPResponse response = credentialRequest.send();
+            if (!response.indicatesSuccess()) {
+                LogHelper.logErrorMessage(
+                        "Error retrieving credential",
+                        response.getStatusCode(),
+                        response.getStatusMessage());
+                if (credentialIssuerId.equals(DCMAW_CRI)
+                        && response.getStatusCode() == HTTPResponse.SC_NOT_FOUND) {
+                    throw new VerifiableCredentialException(
```
Joe-Edwards-GDS: I think we've lost some exception handling for this case in the handler... If you look at `RetrieveCriCredentialHandler#L219`:
```
if (credentialIssuerId.equals(DCMAW_CRI)
                    && e instanceof VerifiableCredentialException vce
                    && vce.getHttpStatusCode() == HTTPResponse.SC_NOT_FOUND) {
                LogHelper.logErrorMessage(
                        "404 received from DCMAW CRI", vce.getErrorResponse().getMessage());
                return JOURNEY_NOT_FOUND;
            }
```

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,250 @@
...
+    private final ConfigService configService;
+    private final JWSSigner signer;
+
+    @ExcludeFromGeneratedCoverageReport
+    public CriApiService(ConfigService configService, JWSSigner signer) {
+        this.configService = configService;
+        this.signer = signer;
+    }
+
+    public VerifiableCredentialResponse getVcResponse(
```
Joe-Edwards-GDS: I'm slightly in two minds whether this is a useful method, or if we should just call the two API methods in the handler 🤔

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
```diff
@@ -0,0 +1,216 @@
...
+    IpvSessionItem ipvSessionItem;
+    String componentId;
+    String criId;
+    String criOAuthSessionId;
+    String ipvSessionId;
+    String ipAddress;
+    String govukSigninJourneyId;
+    String userId;
+    CredentialIssuerConfig criConfig;
+    AuditEventUser auditEventUser;
```
Joe-Edwards-GDS: I think I'd ask two questions about each of these:
- Does it need to be an instance variable (as opposed to a local variable)
- If so, can it be private?

Joe-Edwards-GDS: (Instance fields mean this class isn't thread-safe, which is obviously not an issue in lambdas, but might be surprising in general)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
```diff
@@ -0,0 +1,216 @@
...
+
+    private VisitedCredentialIssuerDetailsDto storeCreatedVcs(
+            VerifiableCredentialResponse vcResponse)
+            throws SqsException, ParseException, JsonProcessingException, CiPutException,
+                    CiPostMitigationsException, VerifiableCredentialResponseException {
+        var vcs = vcResponse.getVerifiableCredentials();
+
+        var issuers = new HashSet<String>();
+        for (SignedJWT vc : vcs) {
+            verifiableCredentialJwtValidator.validate(vc, criConfig, userId);
```
Joe-Edwards-GDS: Do you think it would be reasonable to pull this check out of here, so we have a separate method (perhaps in the checking service) that validates the VCs before storage?

I don't think it's terrible to do it here, if inconvenient for some reason.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
```diff
@@ -0,0 +1,216 @@
...
+
+        if (vcs.isEmpty()) {
+            sendAuditEventForProcessedVcResponse(
+                    CriResourceRetrievedType.EMPTY.getType(), criId, auditEventUser);
+            issuers.add(null);
+        } else {
+            if (issuers.size() != 1) {
+                throw new VerifiableCredentialResponseException(
+                        HttpStatus.SC_BAD_GATEWAY,
+                        ErrorResponse.INVALID_NUMBER_OF_ISSUERS_IN_VC_RESPONSE);
```
Joe-Edwards-GDS: Did you add this logic?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
```diff
@@ -0,0 +1,216 @@
...
+            }
+            sendAuditEventForProcessedVcResponse(
+                    CriResourceRetrievedType.VC.getType(), criId, auditEventUser);
+        }
+
+        return new VisitedCredentialIssuerDetailsDto(criId, issuers.iterator().next(), true, null);
```
Joe-Edwards-GDS: Do we actually _use_ the issuer returned here?

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -90,6 +90,35 @@ public Optional<JourneyResponse> getJourneyResponseForStoredContraIndicators(
...
     }
 
+    public boolean checkCiScoreBreaching(ContraIndicators contraIndicators)
+            throws UnrecognisedCiException {
+        LOGGER.info(
+                new StringMapMessage()
+                        .with(LOG_MESSAGE_DESCRIPTION.getFieldName(), "Retrieved user's CI items.")
+                        .with(
+                                LOG_NO_OF_CI_ITEMS.getFieldName(),
+                                contraIndicators.getContraIndicatorsMap().size()));
```
Joe-Edwards-GDS: This doesn't seem like the right place for this log message - this method doesn't actually retrieve the CIs!

I'd either delete it or move into the `CimitService` method (where we actually fetch the VCs).

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -90,6 +90,35 @@ public Optional<JourneyResponse> getJourneyResponseForStoredContraIndicators(
         return Optional.of(JOURNEY_RESPONSE_PYI_NO_MATCH);
     }
 
+    public boolean checkCiScoreBreaching(ContraIndicators contraIndicators)
```
Joe-Edwards-GDS: A few things:
- I would phrase boolean methods to more clearly indicate which way round they are - in this case `isCiScoreBreaching`
- Do you think the `Gpg45ProfileEvaluator` is the right place for this? Could it also be on `UserIdentityService`? Once you rebase, you'll find that there is a very similar `breachingCiThreshold` method which you probably want to combine with :)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -90,6 +90,35 @@ public Optional<JourneyResponse> getJourneyResponseForStoredContraIndicators(
...
+
+        if (ciScore <= Integer.parseInt(configService.getSsmParameter(CI_SCORING_THRESHOLD))) {
+            return false;
+        }
+
+        LOGGER.info(
+                new StringMapMessage()
+                        .with(
+                                LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                "Returning CI error response."));
```
Joe-Edwards-GDS: Not here it isn't! Do you think this log is useful?

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -90,6 +90,35 @@ public Optional<JourneyResponse> getJourneyResponseForStoredContraIndicators(
...
+                        .with(
+                                LOG_NO_OF_CI_ITEMS.getFieldName(),
+                                contraIndicators.getContraIndicatorsMap().size()));
+        final int ciScore =
+                contraIndicators.getContraIndicatorScore(
+                        configService.getContraIndicatorConfigMap());
+        LOGGER.info(
+                new StringMapMessage()
+                        .with(LOG_MESSAGE_DESCRIPTION.getFieldName(), "Calculated user's CI score.")
+                        .with(LOG_CI_SCORE.getFieldName(), ciScore));
```
Joe-Edwards-GDS: Can we downgrade this log to `debug` (or omit entirely) - I don't think we should be logging this regularly.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -607,6 +609,36 @@ private List<String> getFullNamesFromCredentials(List<IdentityClaim> identityCla
                 .toList();
     }
 
+    public boolean checkUncorrelated(String userId)
```
Joe-Edwards-GDS: Can we avoid the negative, and make it clear which way round it goes - perhaps `areVcsCorrelated` or similar?

At some point I'd like to avoid these methods re-fetching all the VCs every time, but that sounds like a separate ticket.

Joe-Edwards-GDS: This appears to be a copy of the method in `EvaluateGpg45ScoresHandler` - can we replace that with this one?

It's also _very similar_ to the `vcDataCorrelates` method in `CheckExistingIdentityHandler` - do you think we could use it there as well?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -65,6 +65,15 @@ public static boolean isSuccessfulVc(SignedJWT vc) throws ParseException {
         return isValidEvidence(credentialEvidenceList);
     }
 
+    public static boolean isSuccessfulVcs(List<SignedJWT> vcs) throws ParseException {
```
Joe-Edwards-GDS, line 68: `areSuccessfulVcs`?

I hate that you can't do this with a simple stream because Java :'(

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,344 @@
...
+    private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
+    private final IpvSessionService ipvSessionService;
+    private final AuditService auditService;
+    private final CriOAuthSessionService criOAuthSessionService;
+    private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
+    private final VerifiableCredentialService verifiableCredentialService;
+    private final VerifiableCredentialJwtValidator verifiableCredentialJwtValidator;
+    private final UserIdentityService userIdentityService;
+    private final CiMitService ciMitService;
+    private final CriResponseService criResponseService;
```
Joe-Edwards-GDS: I think some of these are not used within this class, and are only passed into dependencies - we can probably remove them as instance variables?

E.g. the `criResponseService` is probably only needed in the second constructor

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
MikeCollingwood: I think that could work, would create a neater distinction between the services

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
MikeCollingwood: I did. Previously it'd take the last issuer found from an unspecified-order list, which made it seem like we expect the same issuer

Joe-Edwards-GDS: I think what we _should_ be doing is confirming that the issuer matches what we expect under the CRI ID, but let's leave that for another ticket - maybe https://govukverify.atlassian.net/browse/PYIC-3907

I don't mind if you want to leave this in, but we could probably just remove the issuer logic entirely (if it's not actually used - as below)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
MikeCollingwood: After a check, the only thing that is used not just in tests is `criId`. After a look, turns out it's used in a method that'll be removed in the second stage of this work.

So unless you @Wynndow know any other reason we may want `VisitedCredentialIssuerDetailsDto` except for the soon-to-be removed `getJourneyResponseForFailWithNoCi`, I'll remove it in the next subticket

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
MikeCollingwood: I'll make it private. I thought it defaulted to that for some reason

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
MikeCollingwood: Had made this change in the meantime. Will be using just callbackRequest to provide all of the information. At the time I wanted the "important" input visible but in the end went for consistency like you suggest

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
Joe-Edwards-GDS: The default is 'package-private' - i.e. visible to other classes in the same package (which is rarely what you want).

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
MikeCollingwood: For reduced code, I'll remove this method and make the other two methods public

Joe-Edwards-GDS: :+1: I think that makes it a better abstraction for the API (i.e. one public method per API endpoint)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
MikeCollingwood: Now I'm removing the getVcResponse method, I don't have criOauthSessionItem, so it makes sense to continue to use the criConfig bc it's used by surrounding methods when called

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
MikeCollingwood: Oh I understand, yep that makes sense

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
MikeCollingwood: I previously didn't see VerifiableCredentialException used anywhere, but have since seen it. Thank you!

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
MikeCollingwood: I've rebased to use breachingCiThreshold since, without the logging (which I'm also keen to remove), it's the same as what I was using

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
MikeCollingwood: I was thinking something similar but looks like a not insignificant amount of work. I've created a subticket for the moment.

MikeCollingwood: I'll create a sub ticket unless told otherwise

MikeCollingwood: Just to clarify, do you mean `CheckExistingIdentityHandler` to use the same `areVcsCorrelated` method from `UserIdentityService`?

MikeCollingwood: That functionality in `EvaluateGpg45ScoresHandler` is removed in the second part of this ticket because it's incorporated into this `ProcessCriCallback` lambda. So will just look into removing `vcDataCorrelates` (?)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
MikeCollingwood: I'm double checking with Chris if VisitedCredentialIssuerDetailsDto is needed anywhere which won't be removed in the next part of this ticket. If so, I'll remove all logic related to it

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/VcStoringService.java
Wynndow: I think this was used quite heavily in the long departed select-cri lambda. If you've checked and there's no other uses of other than what you mentioned, then I don't know any reason to keep it hanging around.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Yeah exactly

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,314 @@
...
+    private final ConfigService configService;
+    private final CriApiService criApiService;
+    private final CriStoringService criStoringService;
+    private final CriCheckingService criCheckingService;
+    private final IpvSessionService ipvSessionService;
+    private final CriOAuthSessionService criOAuthSessionService;
+    private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
+
+    private IpvSessionItem ipvSessionItem = null;
+    private CriOAuthSessionItem criOAuthSessionItem = null;
```
Joe-Edwards-GDS: I think we could still make these local variables in the handler function (initialised to null)?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,314 @@
...
+                    .toObjectMap();
+        } catch (CredentialParseException e) {
+            LOGGER.error("Failed to parse successful VC Store items.", e);
+            return new JourneyErrorResponse(
+                            JOURNEY_ERROR_PATH,
+                            HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                            ErrorResponse.FAILED_TO_PARSE_SUCCESSFUL_VC_STORE_ITEMS)
+                    .toObjectMap();
+        } finally {
+            ipvSessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS: Do we need to check for `null` here? (If we haven't been able to initialise the `ipvSessionItem`)

We should also consider errors, but if it fails to update the session there's not much we can do except panic and return an error anyway 🤷

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,249 @@
...
+    public CriApiService(
+            ConfigService configService,
+            JWSSigner signer,
+            CriOAuthSessionService criOAuthSessionService) {
+        this.configService = configService;
+        this.signer = signer;
+        this.criOAuthSessionService = criOAuthSessionService;
+    }
+
+    public String getApiKey(CriCallbackRequest callbackRequest) {
```
Joe-Edwards-GDS: Should this be a `private` method that is called by each of the API methods?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,249 @@
...
+        var criOAuthSessionItem =
+                criOAuthSessionService.getCriOauthSessionItem(callbackRequest.getState());
```
Joe-Edwards-GDS: It feels a bit odd to fetch this from DynamoDB again here, when we already fetch it in the lambda handler. We might just have to add it as an extra parameter on the methods.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    public void validateVcResponse(
+            VerifiableCredentialResponse vcResponse,
+            CriCallbackRequest callbackRequest,
+            ClientOAuthSessionItem clientOAuthSessionItem)
+            throws VerifiableCredentialResponseException {
+        var userId = clientOAuthSessionItem.getUserId();
+        var criOAuthSessionItem =
+                criOAuthSessionService.getCriOauthSessionItem(callbackRequest.getState());
```
Joe-Edwards-GDS: Ditto - we're fetching this again here.

Fetching the CRI config repeatedly isn't too bad, because we cache the result, but I think this will end up making another call to DynamoDB.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -216,7 +218,7 @@ private List<String> mapCisToExitCodes(ContraIndicators contraIndicators)
                 .toList();
     }
 
-    private boolean breachingCiThreshold(ContraIndicators contraIndicators) {
+    public boolean breachingCiThreshold(ContraIndicators contraIndicators) {
```
Joe-Edwards-GDS: As Chris noted in his PR (https://github.com/govuk-one-login/ipv-core-back/pull/1374), this logic should be strictly greater than (not equal to).

Probably also worth adding some UTs just to make super clear :)

Joe-Edwards-GDS: (I also wouldn't mind renaming to `isBreachingCiThreshold`, but that's a matter of taste)

deploy/template.yaml
```diff
@@ -1162,6 +1162,114 @@ Resources:
...
 
+  ProcessCriCallbackFunction:
+    Type: AWS::Serverless::Function
+    DependsOn:
+      - "ProcessCriCallbackLogGroup"
+    Properties:
+      # checkov:skip=CKV_AWS_115: We do not have enough data to allocate the concurrent execution allowance per function.
+      # checkov:skip=CKV_AWS_116: Lambdas invoked via API Gateway do not support Dead Letter Queues.
+      # checkov:skip=CKV_AWS_117: Lambdas will migrate to our own VPC in future work.
+      FunctionName: !Sub "build-cri-oauth-request-${Environment}"
```
Wynndow:
```suggestion
      FunctionName: !Sub "process-cri-callback-${Environment}"
```

deploy/template.yaml
```diff
@@ -1162,6 +1162,114 @@ Resources:
...
+      Handler: uk.gov.di.ipv.core.processcricallback.ProcessCriCallback::handleRequest
+      Runtime: java17
+      PackageType: Zip
+      CodeUri: ../lambdas/process-cri-callback
+      MemorySize: 3008
```
Wynndow:
```suggestion
      MemorySize: 2048
```

deploy/template.yaml
```diff
@@ -1162,6 +1162,114 @@ Resources:
...
+      Tracing: Active
+      Environment:
+        # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
+        Variables:
+          ENVIRONMENT: !Sub "${Environment}"
+          POWERTOOLS_SERVICE_NAME: !Sub build-cri-oauth-request-${Environment}
```
Wynndow:
```suggestion
          POWERTOOLS_SERVICE_NAME: !Sub process-cri-callback-${Environment}
```

openAPI/core-back-internal.yaml
```diff
@@ -73,6 +73,59 @@ paths:
...
+        requestTemplates:
+          application/json:
+            Fn::Sub: |
+              #set($inputRoot = $input.path('$'))
+              {
+                "input": "{
+                  \"ipvSessionId\": \"$input.params('ipv-session-id')\", 
+                  \"featureSet\": \"$input.params('feature-set')\", 
+                  \"ipAddress\": \"$input.params('ip-address')\" #foreach($key in $inputRoot.keySet()),
+                  \"$key\": \"$inputRoot.get($key)\"#end
```
Joe-Edwards-GDS: Have you managed to test this yet? I wonder if instead you want to just pass the request directly to the lambda, and parse the request there (similar to e.g. the `InitializeIpvSessionHandler`)?

openAPI/core-back-internal.yaml
```diff
@@ -73,6 +73,59 @@ paths:
...
+                #elseif ($bodyObj.status == "FAILED")
+                  #set($context.responseOverride.status = 500)
+                  {
+                    "cause": "$bodyObj.cause",
+                    "error": "$bodyObj.error"
+                  }
+                #else
+                  #set($context.responseOverride.status = 504)
+                  $bodyObj
+                #end
```
Joe-Edwards-GDS: Similarly, I'd like to make sure this transformation is actually necessary.

(In general I'd like as much logic to be in Java and not AWS scripting...)

openAPI/core-back-internal.yaml
MikeCollingwood: I've been trying to test it this morning. Hmm, I hadn't considered that. I'll leave this unresolved because will look into it once it's working

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
MikeCollingwood: I'll add the check, but won't throw any new error because the error causing ipvSessionItem to be null will be caught

Joe-Edwards-GDS: What do you mean? Exceptions thrown in `finally` will supersede any that are thrown in the main body. (So in general you want to avoid throwing exceptions in `finally` unless you deliberately want to forget any being caught elsewhere.

MikeCollingwood: Sorry, yep I was saying I won't supersede any error by throwing an exception (about ipvSessionItem being null) in finally

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -0,0 +1,261 @@
...
+                    e,
+                    HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                    ErrorResponse.FAILED_TO_SEND_AUDIT_EVENT);
+        } catch (ParseException e) {
+            return buildErrorResponse(
+                    e,
+                    HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                    ErrorResponse.FAILED_TO_PARSE_ISSUED_CREDENTIALS);
+        } catch (VerifiableCredentialException e) {
+            // Removed check for DCMAW because output is now equal
```
Joe-Edwards-GDS: Are you sure about this? It looks like the DCMAW not found case is slightly different - it emits a `/journey/not-found` event, rather than an HTTP error.

(Annoyingly it looks like the RetrieveCriCredentialHandler error handling works differently from the others, but I don't mind turning the `/journey/error` responses into real errors)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
Joe-Edwards-GDS, line 1: I have a feeling that in future we may want to separate some of these methods, because at the moment this feels like a bit of a rag-tag bucket of validation/checking methods, rather than a coherent group.

_However_ I don't think we should spend too much effort doing that now - it'll be a bit easier to reason about once you've removed all the old code anyway.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -0,0 +1,232 @@
...
+import uk.gov.di.ipv.core.library.auditing.AuditEvent;
+import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
+import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionErrorParams;
+import uk.gov.di.ipv.core.library.cimit.exception.CiRetrievalException;
+import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
+import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.domain.JourneyResponse;
+import uk.gov.di.ipv.core.library.dto.VisitedCredentialIssuerDetailsDto;
+import uk.gov.di.ipv.core.library.exceptions.*;
```
Joe-Edwards-GDS: No wildcard imports please :)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -0,0 +1,259 @@
...
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_RESPONSE_CONTENT_TYPE;
+
+public class CriApiService {
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final String API_KEY_HEADER = "x-api-key";
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private final ConfigService configService;
+    private final JWSSigner signer;
+    private final CriOAuthSessionService criOAuthSessionService;
```
Joe-Edwards-GDS: Unused?

Joe-Edwards-GDS (commented):
Looking good - we obviously need tests, but keen to try and get it running now :)

deploy/template.yaml
```diff
@@ -1222,11 +1257,30 @@ Resources:
...
+#              Effect: Allow
+#              Action:
+#                - 'lambda:CreateFunction'
+#                - 'lambda:UpdateFunctionConfiguration'
+#              Resource:
+#                - "*"
+#              Condition:
+#                StringEquals:
+#                  "lambda:VpcIds":
+#                    - Fn::ImportValue: !Sub ${VpcStackName}-VpcId
```
Joe-Edwards-GDS: Do you know what this is for?

openAPI/core-back-internal.yaml
```diff
@@ -35,43 +35,50 @@ paths:
...
-
-                #if ($bodyObj.status == "SUCCEEDED")
-                  #set($lambdaOutputObj = $util.parseJson($bodyObj.output))
-                  #set($context.responseOverride.status = $lambdaOutputObj.statusCode)
-                  $bodyObj.output
-
-                #elseif ($bodyObj.status == "FAILED")
-                  #set($context.responseOverride.status = 500)
+        'Fn::If':
+          - IsSubscriptionEnviroment
```
Joe-Edwards-GDS: I think this is a bit of an abuse of this variable - the name implies it's controlling whether the logs should go to splunk.

We could add another condition into `template.yaml` with a more appropriate name?

deploy/template.yaml
MikeCollingwood: It enforces that lambda functions can only be created or updated within a specific VPC - which I felt could be outdated. I saw it in other lambdas (e.g. ValidateOAuthCallbackFunction) so didn't want to forget about it, hence the comment. Will ask Chris when he's back

Joe-Edwards-GDS: Cool - could also be one for @vam-gds

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
MikeCollingwood: That makes sense!

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -0,0 +1,249 @@
...
+            throw new InvalidCriCallbackRequestException(ErrorResponse.MISSING_IPV_SESSION_ID);
+        }
+        if (StringUtils.isBlank(state)) {
+            throw new InvalidCriCallbackRequestException(ErrorResponse.MISSING_OAUTH_STATE);
+        }
+        var persistedOauthState = getPersistedOauthState(criOAuthSessionItem);
+        if (!callbackRequest
+                .getState()
+                .equals(persistedOauthState)) { // might be ipvSessionItem.getCriOAuthSessionId() !=
+            // null ? ... : null
```
Joe-Edwards-GDS: Will it ever be valid for this to be be null? I wonder if we just make that explicit:
```
if (criOauthSessionItem == null || !state.equals(criOauthSessionItem.getCriOAuthSessionId()) {
  throw new InvalidCriCallbackRequestException(ErrorResponse.INVALID_OAUTH_STATE);
}
```

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandlerTest.java
Joe-Edwards-GDS, line 1: Assuming this isn't final..!

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -136,6 +138,11 @@ public ContraIndicators getContraIndicators(SignedJWT contraIndicatorsVc)
         return mapToContraIndicators(parseContraIndicatorEvidence(contraIndicatorsVc));
     }
 
+    public boolean isCiMitigatable(ContraIndicator ci) throws ConfigException {
+        var cimitConfig = configService.getCimitConfig();
+        return cimitConfig.containsKey(ci.getCode()) && CollectionUtils.isEmpty(ci.getMitigation());
```
Joe-Edwards-GDS: I think the second half can be `!ci.isMitigated()`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -72,8 +72,14 @@ public enum ErrorResponse {
...
     FAILED_TO_GENERATE_NINO_CLAIM(1062, "Failed to generate the NINO claim"),
-    MISSING_IS_USER_INITIATED_PARAMETER(1059, "Missing isUserInitiated in request"),
-    MISSING_VTR(1063, "The 'vtr' claim is required and was not provided in the request.");
+    MISSING_IS_USER_INITIATED_PARAMETER(1063, "Missing isUserInitiated in request"),
+    MISSING_VTR(1064, "The 'vtr' claim is required and was not provided in the request."),
+    NO_IPV_FOR_CRI_OAUTH_SESSION(1065, "No ipvSession for existing CriOAuthSession."),
+    FAILED_TO_GET_VALID_CREDENTIAL_STATUS(
+            1066, "Failed to get credential from issuer with valid status."),
+    INVALID_NUMBER_OF_ISSUERS_IN_VC_RESPONSE(
+            1067, "Incorrect number of CRI issuers in VC response from a CRI."),
```
Joe-Edwards-GDS: I don't think these two are used any more?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -219,11 +221,22 @@ private List<String> mapCisToExitCodes(ContraIndicators contraIndicators)
...
 
+    public boolean isBreachingCiThresholdIfMitigated(ContraIndicator ci, ContraIndicators cis) {
+        var scoreOnceMitigated =
+                cis.getContraIndicatorScore(configService.getContraIndicatorConfigMap())
+                        + configService
+                                .getContraIndicatorConfigMap()
+                                .get(ci.getCode())
+                                .getCheckedScore();
+        var threshold = Integer.parseInt(configService.getSsmParameter(CI_SCORING_THRESHOLD));
+        return scoreOnceMitigated > threshold; // This will be commonised
```
Joe-Edwards-GDS: Will it 👀

Joe-Edwards-GDS (commented):
Looks like a few of the test classes are still empty...

At this stage I'm quite keen to get this merged, so I don't mind too much if that's done as a quick follow-up, but I also still haven't spoken to Vam. I'll see if I can schedule that in this afternoon.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -222,6 +225,33 @@ public VerifiableCredentialResponse fetchVerifiableCredential(
...
 
+    public HTTPRequest buildFetchVerifiableCredentialRequest(
+            BearerAccessToken accessToken,
+            CriCallbackRequest callbackRequest,
+            CriOAuthSessionItem criOAuthSessionItem) {
+        var criId = callbackRequest.getCredentialIssuerId();
+        var criConfig = configService.getCriConfig(criOAuthSessionItem);
+        var apiKey = getApiKey(criConfig, criOAuthSessionItem);
+
+        LOGGER.error("HALLO, {} {} {}", criId, criConfig, apiKey);
```
Joe-Edwards-GDS: !

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -162,10 +162,7 @@ public void validateCallbackRequest(
             throw new InvalidCriCallbackRequestException(ErrorResponse.MISSING_OAUTH_STATE);
         }
         var persistedOauthState = getPersistedOauthState(criOAuthSessionItem);
-        if (!callbackRequest
-                .getState()
-                .equals(persistedOauthState)) { // might be ipvSessionItem.getCriOAuthSessionId() !=
-            // null ? ... : null
+        if (!callbackRequest.getState().equals(persistedOauthState)) {
```
Joe-Edwards-GDS: (not important, but you have pulled out a variable for state already, so perhaps this is just `state.equals(persistedOauthState)`)

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingServiceTest.java
```diff
@@ -1,11 +1,586 @@
...
 
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.*;
```
Joe-Edwards-GDS: We've got a few wildcards in this review - it should be possible to configure IntelliJ to avoid this. If you look in settings Code Style -> Java -> Imports, and set the wildcard threshold to something high.

Joe-Edwards-GDS (commented):
Almost there I think!

Joe-Edwards-GDS (dismissed):

Wynndow (approved):

PR #1372, opened 2023-11-15
PYIC-3860: Added journey map config for reset user identity flow

RogVanGDS:
### What changed

Added journey map config for delete-details flow 

### Why did it change

Integrating the new delete user details pages into a journey flow.

Added a test route that can be enabled by feature set, to start the flow from the existing reuse details page

### Issue tracking

- [PYIC-3860](https://govukverify.atlassian.net/browse/PYIC-3860)

[PYIC-3860]: https://govukverify.atlassian.net/browse/PYIC-3860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -370,12 +370,12 @@ PYI_CONFIRM_DELETE_DETAILS_PAGE:
   events:
     next:
       targetState:
-        RESET_IDENTITY_V2
+        RESET_IDENTITY_USER_INITIATED
     end:
       targetState:
         IPV_IDENTITY_REUSE_PAGE
 
-RESET_IDENTITY_V2:
+RESET_IDENTITY_USER_INITIATED:
   response:
     type: process
     lambda: reset-identity
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -83,6 +83,10 @@ CHECK_EXISTING_IDENTITY:
       targetState: RESET_IDENTITY
     reuse:
       targetState: IPV_IDENTITY_REUSE_PAGE
+      checkFeatureFlag:
+        deleteDetailsEnabled:
+          targetState: IPV_IDENTITY_REUSE_PAGE_TEST
+
     pending:
       targetState: IPV_IDENTITY_PENDING_PAGE
     pyi-no-match:
@@ -100,13 +104,26 @@ RESET_IDENTITY:
     lambda: reset-identity
   events:
     next:
-      targetState: IPV_IDENTITY_START_PAGE
+      targetState: PYI_DETAILS_DELETED_PAGE
 
 IPV_IDENTITY_REUSE_PAGE:
   response:
     type: page
     pageId: page-ipv-reuse
   parent: END_JOURNEY
+
+IPV_IDENTITY_REUSE_PAGE_TEST:
+  response:
+    type: page
+    pageId: page-ipv-reuse
+  parent: END_JOURNEY
+  events:
+    next:
+      targetState:
+        END_JOURNEY
+    reset:
+      targetState:
+        PYI_NEW_DETAILS_PAGE
 
 IPV_IDENTITY_PENDING_PAGE:
   response:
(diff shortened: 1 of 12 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -100,13 +104,26 @@ RESET_IDENTITY:
     lambda: reset-identity
   events:
     next:
-      targetState: IPV_IDENTITY_START_PAGE
+      targetState: PYI_DETAILS_DELETED_PAGE
```
blakeyp: This will affect existing journeys that go through `RESET_IDENTITY` so I think we'll need a new state specifically for the user initiated reset flow

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -294,6 +311,39 @@ IPV_SUCCESS_PAGE:
...
+        IPV_IDENTITY_REUSE_PAGE
+
+PYI_CONFIRM_DELETE_DETAILS_PAGE:
+  response:
+    type: page
+    pageId: pyi-confirm-delete-details
+  events:
+    reset:
+      targetState:
+        RESET_IDENTITY
```
blakeyp: Would like to have an info log in this lambda so we can track user-initiated identity resets (probably also an audit event but TBC). To do so I think we'll need to use `lambdaInput` to distinguish this from the other routes that go to reset-identity. (We'll also need this for the email sending as we only want to send an email for this route)

Could you have a look at how we can do this?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
RogVanGDS: Do we not want the delete confirmation screen to show everywhere we delete a user's details?

RogVanGDS: Oh I see, you are talking about times when we delete the users details without the user initiating it themselves?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
RogVanGDS: Yep, I'll take a look

RogVanGDS: We probably just want to have 2 separate states for the reset-identity lambda. One that we use in user initiated journeys, and one that we use in non-user initiated.

Then we can pass in a boolean flag as a lambda Input similar to the check-gpg45-score lambda so we can then log inside the lambda what the trigger was.

Do you want me to do this as part of this ticket? Or shall we raise a separate one?

blakeyp: Sounds good, I think we should just do it as part of this ticket

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -332,6 +340,61 @@ CRI_EXPERIAN_KBV:
...
+    pageId: pyi-confirm-delete-details
+  events:
+    next:
+      targetState:
+        RESET_IDENTITY_V2
+    end:
+      targetState:
+        IPV_IDENTITY_REUSE_PAGE
+
+RESET_IDENTITY_V2:
```
blakeyp: Could we name this `RESET_IDENTITY_USER_INITIATED` ?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -98,6 +101,8 @@ RESET_IDENTITY:
   response:
     type: process
     lambda: reset-identity
+    lambdaInput:
+      isUserInitiated: false
```
blakeyp, line 105: Just checking does this have to be provided? (or can it just default to false?)

blakeyp:
Could you just pop a title on the PR please

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
RogVanGDS: I think it does have to be provided

blakeyp (approved):

PR #1374, opened 2023-11-15
PYIC-3991: Fix exit code breaching threshold logic

Wynndow:
### What changed

Fix exit code breaching threshold logic

### Why did it change

When we check a users CI scores against the threshold we only fail them if the score is *greater* than the threshold. Equal to is ok.

When calculating the exit code we were deciding the scores were breaching if they were greater than *or* equal to. Which is wrong.

This pulls the logic for determining if CI scores are breaching to a shared place. This will ensure that we're always determining it the same way.

The exit codes are not being consumed yet, so this isn't a wider problem.

### Issue tracking

- [PYIC-3991](https://govukverify.atlassian.net/browse/PYIC-3991)

[PYIC-3991]: https://govukverify.atlassian.net/browse/PYIC-3991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -48,6 +48,7 @@ import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.CORE_VTM_C
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EXIT_CODES_ALWAYS_REQUIRED;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EXIT_CODES_NON_CI_BREACHING_P0;
 import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.USER_ISSUED_CREDENTIALS_TABLE_NAME;
+import static uk.gov.di.ipv.core.library.domain.CiScoreThresholdChecker.ciScoreBreachesThreshold;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.ADDRESS_CRI;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.BAV_CRI;
 import static uk.gov.di.ipv.core.library.domain.CriConstants.DCMAW_CRI;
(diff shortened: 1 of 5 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -217,8 +218,10 @@ private List<String> mapCisToExitCodes(ContraIndicators contraIndicators)
     }
 
     private boolean breachingCiThreshold(ContraIndicators contraIndicators) {
-        return contraIndicators.getContraIndicatorScore(configService.getContraIndicatorConfigMap())
-                >= Integer.parseInt(configService.getSsmParameter(CI_SCORING_THRESHOLD));
+        return ciScoreBreachesThreshold(
+                contraIndicators.getContraIndicatorScore(
+                        configService.getContraIndicatorConfigMap()),
+                configService.getSsmParameter(CI_SCORING_THRESHOLD));
```
Joe-Edwards-GDS, line 224: It feels a bit weird to pull out just the comparison here - you've got two suspiciously similar methods: `breachingCiThreshold` and `ciScoreBreachesThreshold`, but the second one is actually just comparing the two arguments.

Perhaps we could make this method public and test it directly?

Also worth noting that @MikeCollingwood is already going to do that (I hope) as part of https://github.com/govuk-one-login/ipv-core-back/pull/1367/files#r1394057532 - reusing this in the main CI scoring step.

Joe-Edwards-GDS: If you feel that the integer conversion is a distraction, you could add an extra method to the config service - `getIntegerSsmParameter` or similar, but not too fussed.

Joe-Edwards-GDS (commented):
I think we want to commonise the CI scoring logic overall, and that might need coordination with Mike.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: I intentionally left the wrapper method in place to keep the `getFailExitCode()` more readable. Including the two calls to the config service in the predicate for the ternary got really busy. I was on the fence for a while though. Happy to ditch it if you think it's better.

Not too worried about the integer conversion as it's tucked away in the new class.

Will catch up with Mike about what he's up to and see if we can combine.

Wynndow: @Joe-Edwards-GDS I've had a chat with Mike and he's happy to incorporate this into his work, as it's a larger piece.

PR #1385, opened 2023-11-17
PYIC-3979: Save 'vtr' from Orchestration into the ClientAuthSessionItem

TalatAydinGDS:
### What changed

The 'vtr' is checked in the coming request. If the vtr is null or empty, a bad request is returned from the Initialise Ipv Session Handler. If the vtr is provided, it it stored in the ClientOAuthSessionItem document. 

### Why did it change

Save 'vtr' from Orchestration into the ClientAuthSessionItem

### Issue tracking

- [PYIC-3979](https://govukverify.atlassian.net/browse/PYIC-3979)

### Other considerations

[PYIC-3979]: https://govukverify.atlassian.net/browse/PYIC-3979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -41,11 +41,13 @@ import uk.gov.di.ipv.core.library.service.ConfigService;
 import uk.gov.di.ipv.core.library.service.IpvSessionService;
 
 import java.text.ParseException;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.JAR_KMS_ENCRYPTION_KEY_ID;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_LAMBDA_RESULT;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
 
 public class InitialiseIpvSessionHandler
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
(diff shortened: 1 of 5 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -71,7 +71,8 @@ public enum ErrorResponse {
     MISSING_SCORE_THRESHOLD(1059, "Missing score type in request"),
     UNKNOWN_SCORE_TYPE(1060, "Unknown score type in request"),
     FAILED_TO_PARSE_SUCCESSFUL_VC_STORE_ITEMS(1061, "Failed to parse successful VC Store items."),
-    FAILED_TO_GENERATE_NINO_CLAIM(1062, "Failed to generate the NINO claim");
+    FAILED_TO_GENERATE_NINO_CLAIM(1062, "Failed to generate the NINO claim"),
+    MISSING_VTR(1063, "The 'vtr' field is required and was not provided in the request.");
```
Joe-Edwards-GDS: Technically this is `The 'vtr' claim is required ...`, but not super important

Joe-Edwards-GDS (commented):
This looks good, but I just want to check with orchestration that this is definitely not going to break anything by making it mandatory.

If we're not sure, we might need to do something similar to the correlation change, where we only log to start with, and upgrade to an error once we're sure it's safe.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -118,10 +123,23 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            if (vtr == null || vtr.isEmpty() || vtr.stream().allMatch(String::isEmpty)) {
+                var message =
+                        new StringMapMessage()
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        ErrorResponse.MISSING_VTR.getMessage());
+                LOGGER.error(message);
+
+                return ApiGatewayResponseGenerator.proxyJsonResponse(
+                        HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_VTR);
```
Joe-Edwards-GDS: Ok, I've spoken to orchestration and I think I'd like to do this in two stages, as we did before:
- For now, only log the error message
- Once we're confident it's not breaking anything in production, turn it into a real error

Sorry we didn't capture this as a requirement at the start!

Joe-Edwards-GDS (changes requested):

TalatAydinGDS:
@Joe-Edwards-GDS  if 'vtr' is empty, should we handle it by throwing and exception within in same task and creating another sub-task, or would you prefer creating a separate task in the backlog for this?

Joe-Edwards-GDS (approved):

PR #1411, opened 2023-11-23
BAU: Combine catch blocks in check-existing-identity

Wynndow:
### What changed

Combine catch blocks in check-existing-identity

### Why did it change

The check-existing-identity handler is currently getting flagged for cognitive complexity by sonar. Each catch block counts as a cognitive "point". By combining them like this it might be easy to avoid getting flagged... It also might be really horrible.

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -107,6 +107,17 @@ public class CheckExistingIdentityHandler
     private final AuditService auditService;
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
     private final CiMitService ciMitService;
+    private static final Map<Class<?>, ErrorResponse> ERROR_RESPONSE_MAP =
+            Map.of(
+                    CiRetrievalException.class, ErrorResponse.FAILED_TO_GET_STORED_CIS,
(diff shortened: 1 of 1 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -400,8 +405,10 @@ private Map<String, Object> buildProfileMatchedResponse(
         return JOURNEY_REUSE;
     }
 
-    private Map<String, Object> buildErrorResponse(ErrorResponse errorResponse, Exception e) {
+    private Map<String, Object> buildErrorResponse(Exception e) {
+        ErrorResponse errorResponse = ERROR_RESPONSE_MAP.get(e.getClass());
```
Joe-Edwards-GDS, line 409: I think we need to be slightly more careful here
- This won't match on subclasses (but they will match the catch block)
- I'd feel more comfortable if we had some kind of default case, if nothing matches*

Perhaps a `switch` statement would be a more natural way of achieving both?

*I think what we really want is a proper error handler that can handle a wide range of errors in a consistent way, including unexpected errors. A significant portion of each handler class is very repetitive error handling :(

Wynndow: Oop yeah that's a good point. This is mostly the completionist in me wanting to remove that last reported code smell. I'm also inclinded to just add an ignore to this method so that it won't get flagged.

PR #1412, opened 2023-11-23
PYIC-4043: force identity reset if reprove identity flag is true

shivanshuit914:
## Proposed changes

- Reset users identity when `reprove_identity` flag is `true`.

### What changed

- Resetting users identity when `reprove_identity` flag is `true`. Allowing user to prove identity again.

### Why did it change

- When orch API passes `reprove_identity: "true"` claim we let user to prove identity.
- In case if its false or not present user can continue journey as it is.

### Issue tracking

- [PYIC-4043](https://govukverify.atlassian.net/browse/PYIC-4043)

## Checklists

[PYIC-4043]: https://govukverify.atlassian.net/browse/PYIC-4043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -202,8 +202,10 @@ public class CheckExistingIdentityHandler
             }
 
             // Reset identity if reprove is true.
-            Boolean reproveIdentity = clientOAuthSessionItem.getReproveIdentity();
-            if (configService.enabled(ACCOUNT_INTERVENTIONS.getName()) && reproveIdentity) {
+            String reproveIdentity = clientOAuthSessionItem.getReproveIdentity();
+            if (configService.enabled(ACCOUNT_INTERVENTIONS.getName())
+                    && !Objects.isNull(reproveIdentity)
+                    && reproveIdentity.equals("true")) {
                 return buildForceResetResponse();
             }
 
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -200,6 +200,12 @@ public class CheckExistingIdentityHandler
                         : buildNotCorrelatedResponse(auditEventUser);
             }
 
+            // Reset identity if reprove is true.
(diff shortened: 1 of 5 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
```diff
@@ -24,6 +24,7 @@ public class ClientOAuthSessionItem implements DynamodbItem {
     private String state;
     private String userId;
     private String govukSigninJourneyId;
+    private String reproveIdentity;
```
DanCorderIPV: Can this be a Boolean instead of a string here?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -200,6 +200,12 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
                         : buildNotCorrelatedResponse(auditEventUser);
             }
 
+            // Reset identity if reprove is true.
```
DanCorderIPV: Should this happen earlier in this function? e.g. What should happen if a user is part way through a F2F journey when the flag is set?

shivanshuit914: hm not sure need to check with Patrick, at the moment if user hasn't finished f2f journey it redirect to f2f page and takes the priority over `reprove_identity`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
shivanshuit914: yes Thanks

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
blakeyp: Had a chat with @shivanshuit914, yeah the reprove_identity check should come first before the F2F checks

shivanshuit914: Thank you, updated it :)

DanCorderIPV (approved):

PR #1420, opened 2023-11-27
BAU: Run gradle tasks in parallel

Wynndow:
### What changed

Run gradle tasks in parallel

### Why did it change

Our gradle project consists of lots of separate sub-projects, without too many dependencies on each other. This makes it a good candidate to run separate sub-project tasks in parallel.

Currently, on my machine, running `gradlew clean build` takes 2m12s. Applying the parallel config reduces that to 59s.

We'd need to keep an eye on how this affects build in GHA for the pipeline, but hopefully it should be fine.

See the gradle docs for this config here: https://docs.gradle.org/current/userguide/performance.html#parallel_execution

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+org.gradle.parallel=true


TalatAydinGDS (approved):

Joe-Edwards-GDS (approved):
Happy to give this a try :)

As you say, I think we need to keep an eye on this to make sure there are no unintended side-effects (flaky tests, or degraded performance on build agents perhaps) - but hopefully gradle is clever enough to do it sensibly.

PR #1423, opened 2023-11-28
PYIC-3853

TalatAydinGDS:
### What changed

Implement rules within IPV Core to avoid specific CRIs being used as a single piece of evidence.

### Why did it change

  While certain Credential Issuers like "Bank Account CRI" can technically contribute to profiles like M1B, the programme has determined that these should not be used for certain identity validations. Therefore, rules are needed to restrict or specify which CRIs can contribute to which GPG45 profiles. 

### Issue tracking

- [PYIC-3853](https://govukverify.atlassian.net/browse/PYIC-3853)

### Other considerations

[PYIC-3853]: https://govukverify.atlassian.net/browse/PYIC-3853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -259,6 +261,29 @@ private JourneyResponse checkForMatchingGpg45Profile(
...
+            if (credentialIssuer != null) {
+                boolean requiresAdditionalEvidence =
+                        configService
+                                .getCredentialIssuerActiveConnectionConfig(credentialIssuer)
+                                .isRequiresAdditionalEvidence();
+                LOGGER.info(
+                        new StringMapMessage()
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        "This credential, which contains the only valid evidence, should not any GPG45 profile.")
```
DanCorderIPV: This message seems to be missing a word.
Also do we want to log this even if requiresAdditionalEvidence is false?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -527,6 +528,34 @@ private boolean isEvidenceVc(VcStoreItem item) throws CredentialParseException {
         return false;
     }
 
+    private boolean isNonZeroInt(JsonNode node) {
+        return node.isInt() && node.asInt() != 0;
+    }
+
+    private boolean isValidEvidence(VcStoreItem item) {
```
DanCorderIPV: There is already a function `isEvidenceVc()` could these be made more consistent or even combined?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -527,6 +528,34 @@ private boolean isEvidenceVc(VcStoreItem item) throws CredentialParseException {
...
+        try {
+            JsonNode vcEvidenceNode = getVCClaimNode(item.getCredential(), VC_EVIDENCE);
+            return ((vcEvidenceNode.size() > 0)
+                    && isNonZeroInt(vcEvidenceNode.get(0).path(VC_EVIDENCE_VALIDITY))
```
DanCorderIPV: Is this a sufficient check, or should we also check for CIs?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
TalatAydinGDS: The message has been changed to simpler and more understandable one. In case of a 'false' condition, we don't want to add log.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -527,6 +528,34 @@ private boolean isEvidenceVc(VcStoreItem item) throws CredentialParseException {
...
+    }
+
+    public String getCredentialIssuerIfSingleValidEvidence(List<VcStoreItem> vcStoreItems) {
+        Map<String, Long> issuerCountMap =
+                vcStoreItems.stream()
+                        .filter(item -> isValidEvidence(item))
+                        .map(item -> item.getCredentialIssuer())
+                        .collect(Collectors.groupingBy(issuer -> issuer, Collectors.counting()));
+        if (issuerCountMap.size() == 1
+                && issuerCountMap.values().stream().anyMatch(count -> count >= 1)) {
```
DanCorderIPV: This doesn't match the name of the function - you could have multiple bits of evidence from the same issuer and that issuer would be returned.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -527,6 +528,34 @@ private boolean isEvidenceVc(VcStoreItem item) throws CredentialParseException {
...
+            JsonNode vcEvidenceNode = getVCClaimNode(item.getCredential(), VC_EVIDENCE);
+            return ((vcEvidenceNode.size() > 0)
+                    && isNonZeroInt(vcEvidenceNode.get(0).path(VC_EVIDENCE_VALIDITY))
+                    && isNonZeroInt(vcEvidenceNode.get(0).path(VC_EVIDENCE_STRENGTH)));
+        } catch (CredentialParseException e) {
+            return false;
+        }
+    }
+
+    public String getCredentialIssuerIfSingleValidEvidence(List<VcStoreItem> vcStoreItems) {
```
DanCorderIPV: This seems like it might be a bit over-specialised as a function. It's only returning a credential issuer because that's what the calling code in this case needs. It might be more re-usable if it just returned a single evidence item and the caller then called `.getCredentialIssuer()` on the return item.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -232,7 +234,7 @@ private JourneyResponse checkForMatchingGpg45Profile(
                 gpg45ProfileEvaluator.getFirstMatchingProfile(
                         gpg45Scores, CURRENT_ACCEPTED_GPG45_PROFILES);
 
-        if (matchedProfile.isPresent()) {
+        if (matchedProfile.isPresent() && !checkRequiresAdditionalEvidence(userId)) {
```
Joe-Edwards-GDS: Do you think we should move this check up - is there any need to check GPG45 profiles if we know they require additional evidence?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -259,6 +261,29 @@ private JourneyResponse checkForMatchingGpg45Profile(
         }
     }
 
+    private boolean checkRequiresAdditionalEvidence(String userId) {
+        List<VcStoreItem> vcStoreItems = userIdentityService.getVcStoreItems(userId);
```
Joe-Edwards-GDS: This will fetch all the credentials from dynamodb again, but the method this is called from is already being passed a list of credentials...

Can we just use the `List<SignedJWT>` here instead? That means we won't need to pass the `userId` in either.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Yes, I think they should probably be combined

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: We don't need to check CIs - those are checked elsewhere. (Nonzero should be enough)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -259,6 +261,29 @@ private JourneyResponse checkForMatchingGpg45Profile(
...
+        if (!vcStoreItems.isEmpty()) {
+            String credentialIssuer =
+                    userIdentityService.getCredentialIssuerIfSingleValidEvidence(vcStoreItems);
```
Joe-Edwards-GDS: Do you think it would be simpler to approach this from a slightly different direction - first identifying if we only have a single evidence credential?

i.e.:
```
var evidenceCredentials = ...; // filter credential list by 'isEvidenceVc'
if (evidenceCredentials.size() == 1) {
  var credentialIssuer = evidenceCredentials.get(0).getCredentialIssuer();
  var criConfig = configService.getCredentialIssuerActiveConnectionConfig(credentialIssuer);
  if (criConfig.isRequiresAdditionalEvidence()) {
    return true;
  }
}
return false;
```

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: Yes, I've considered that. However, in the isEvidenceVC(), it only checks whether there is an int value. More importantly, the CredentialParseException error is propagated to the caller method without being caught. In the isValidEvidence(), it checks for credentials with valid evidence only. If a CredentialParseException error is encountered, it ignores that invalid credential. Therefore, I added a new method for the sake of clarify in these details.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: I don't know; this wasn't included in the task requirements.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: Yes, you're right. I'm updating the method name accordingly.  Thank you.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: We added a new property name 'requiredAdditionalEvidence' under CRI in our config param files. Since we only need credential to access this property, that's what is being retuned. Yes, you are right, this method can be used when needed. I am not sure whether we should do tis now or when needed.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Hmm, I think actually the other method _should_ include the non-zero check. I think the exception difference isn't hugely important - in practice if we get an invalid credential we'll fail on CRI return, and won't reach any of these code paths anyway.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
TalatAydinGDS: I thought moving this check after the matched profile would be more efficient in terms of performance. Because, currently, it should only be true for the BAV CRI? But, if I'm mistaken, I can pull check forward before matching the profile.

Joe-Edwards-GDS: I guess it doesn't matter too much.

With Mike's change, I think we expect 'most' journeys to successfully match a profile, which means it'll have to do this check either way (so we may as well do it first), but at the moment that's not the case.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
TalatAydinGDS: Yes, you're right. We are making DynamoDB calls twice. I also considered this, but I thought it might be challenging to access the config CRI parameter the JWT's "iss" value if we go through credentials. This is because "iss" field contains full path. For example, "iss": "https://address-cri.stubs.account.gov.uk". I did it this way to ensure certainly. Let me look any solution with credential.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -259,6 +260,30 @@ private JourneyResponse checkForMatchingGpg45Profile(
...
+                var credentialIssuer = evidenceCredentials.get(0).getJWTClaimsSet().getIssuer();
+                boolean requiresAdditionalEvidence =
+                        configService
+                                .getCredentialIssuerActiveConnectionConfig(credentialIssuer)
+                                .isRequiresAdditionalEvidence();
+                LOGGER.info(
+                        new StringMapMessage()
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        "GPG45 profile should not not be set for this credential.")
```
Joe-Edwards-GDS: I don't think this message is right - this is only true if `requiresAdditionalEvidence = true`. If that's not the case, I don't think we need a log (since most profiles only use a single piece of evidence).

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -516,17 +516,36 @@ public boolean isVcSuccessful(List<VcStatusDto> currentVcStatuses, String criIss
                 .getIsSuccessfulVc();
     }
 
-    private boolean isEvidenceVc(VcStoreItem item) throws CredentialParseException {
-        JsonNode vcEvidenceNode = getVCClaimNode(item.getCredential(), VC_EVIDENCE);
+    private boolean isEvidenceVc(String credential, boolean isNonZero)
```
Joe-Edwards-GDS: Sorry I didn't make this clear - I think we _should_ check for nonzero in all cases (including the existing usage).

If a VC has zero scores, we shouldn't be using it to generate the user identity!

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -516,17 +516,36 @@ public boolean isVcSuccessful(List<VcStatusDto> currentVcStatuses, String criIss
...
         }
         return false;
     }
 
+    public List<SignedJWT> filterValidCredentials(List<SignedJWT> credentials) {
+        return credentials.stream()
+                .filter(
+                        item -> {
+                            try {
+                                return isEvidenceVc(item.serialize(), true);
```
Joe-Edwards-GDS: Hmm - it's a bit annoying that we have to keep serializing and deserializing the VC, but hopefully that'll be solved by https://govukverify.atlassian.net/browse/PYIC-3908

Feels like it'll be painful to address now

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
TalatAydinGDS: I added to understand why the profile couldn't be matched. For tracking through the log. Ok. I am deleting the log now.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -266,24 +270,38 @@ private boolean checkRequiresAdditionalEvidence(List<SignedJWT> credentials)
...
+                                            "requiresAdditionalEvidence",
+                                            requiresAdditionalEvidence));
+                    return requiresAdditionalEvidence;
+                }
             }
         }
         return false;
     }
 
+    public String getCRIWithCredentialIssuer(String iss) {
```
Joe-Edwards-GDS: I don't think we can do this - the issuer id doesn't necessarily match the ID in the configuration, especially in production. E.g. the `ukPassport` CRI has issuer ID `https://review-p.account.gov.uk`

If you'd like to fix it now, then you'll need to pass around `VcStoreItem`s and refactor the other places that use the credentials accordingly. If you think that's too painful, then you can add your new call to the list on https://govukverify.atlassian.net/browse/PYIC-3997

Joe-Edwards-GDS (changes requested):
One thing that occurs to me - we should update the logic in `CheckExistingIdentityHandler` as well, to make sure it's consistent.

Joe-Edwards-GDS (dismissed):
Thanks for making the changes, sorry it ended up going in circles a bit!

Joe-Edwards-GDS (approved):

PR #1424, opened 2023-11-28
PYIC-4056: Create lambda to call the TICF CRI

Wynndow:
### What changed

Create lambda to call the TICF CRI.

It's probably best to review this commit by commit, as there were a series of changes before the main logic was implemented.

### Why did it change

[PYIC-4056: Move CriStoringService to a lib](https://github.com/govuk-one-login/ipv-core-back/commit/71534fba8556ac1e7d36fddeca6cdd0bf447271c)

The logic in the CriStoringService can be re-used for the ticf CRI
lambda, so this moves it out of the process-cri-callback lambda and to a
shared lib.

[PYIC-4056: Make VerifiableCredentialException checked](https://github.com/govuk-one-login/ipv-core-back/commit/0351b5140d9cb7b2d25fc32fee51f025a0751b4c)

This exception was extending a RuntimeException, rather than just
Exception. This meant that it wasn't a checked exception. There was no
real reason for this - one piece of logic needed to be rewritten as a
simple for loop.
Making it a checked exception means that we aren't going to miss
handling it.

[PYIC-4056: Fix CheckExistingIdentityHandlerTest](https://github.com/govuk-one-login/ipv-core-back/commit/aa1b1f81719197a69a9d72ac9b66ce6e58254d54)

A recent change was made to the structure of the UserIdentityService.
Specifically we only call the `areVcsCorrelated` method to check
correlation. This means the mocking we had set up was a bit wrong - it
was mocking the internal methods called by that method. This was messing
up return value from the mock and breaking the test.

Not entirely sure why these tests passed in the first place.

[PYIC-4056: Add method to ConfigService for active connection API key](https://github.com/govuk-one-login/ipv-core-back/commit/ef9a7deb5bdf1e723e4699ce97000c15233b0e96)

A backend only CRI will only ever need to use the active connection
configuration for that CRI. A typical OAuth based CRI needs to use the
same connection for all interactions across the users session. This is
not the case for back end only, which just calls the CRI once.

This adds a new method to get the API key from secrets manager for the
active connection for a CRI.

It also tidies some private methods to beneath the public.

[PYIC-4056: Add ConfigService method for back end cri config](https://github.com/govuk-one-login/ipv-core-back/commit/6ef1c37685fb89ce2e49bae3a36bef74592da860)

The config required for a back end only CRI is a lot simpler than the
usual config for a CRI. This is because it doesn't use OAuth.

This adds a class and method to represent the required config. This
should make working with the config object simpler and less prone to
errors.

[PYIC-4056: Pass lambdaInput in process request directly](https://github.com/govuk-one-login/ipv-core-back/commit/e62e1d507d03592dd3d816bc73ace968d5de15ce)

This changes the way the lambdaInput, defined in the journey map for a
process step response type, is passed to the lambda via the process
request.

It passes the object through as a whole, rather than pulling out
individual elements. This makes it a lot more flexible. Otherwise we'd
need to add new fields to the ProcessRequest class every time we wanted
to pass some data through from the map.

[PYIC-4056: Add logic for calling TICF CRI](https://github.com/govuk-one-login/ipv-core-back/commit/af16984bfd1c89270836d1999cfc25ab7994b513)

This fleshes out the handler, which orchestrates various services, and
adds a new service for talking to the TICF CRI.

This still needs plumbing in to the journey map, so this commit on its
own won't make any functional changes.

When adding to the journey map it will need to have a `lambdaInput`
block defined, to pass through the `journeyType` value. This is due to
a requirement that reuse journeys don't send any VCs to the TICF CRI.

### Issue tracking

- [PYIC-4056](https://govukverify.atlassian.net/browse/PYIC-4056)

[PYIC-4056]: https://govukverify.atlassian.net/browse/PYIC-4056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/journeyEngineStepFunction.asl.json
```diff
@@ -74,7 +74,7 @@
         "ipAddress.$": "$$.Execution.Input.ipAddress",
         "clientOAuthSessionId.$": "$$.Execution.Input.clientOAuthSessionId",
         "featureSet.$": "$$.Execution.Input.featureSet",
-        "isUserInitiated.$": "$.lambdaInput.isUserInitiated"
+        "lambdaInput.$": "$.lambdaInput"
```
Joe-Edwards-GDS, line 82: I think this is a sensible change :)

We could also consider matching the way we pass scope/context to the CRI states (i.e. using query params in the journey string), but there's not particular need.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -271,18 +272,19 @@ public JourneyResponse getJourneyResponse(CriCallbackRequest callbackRequest)
...
-                                    assert criOAuthSessionItem != null;
-                                    verifiableCredentialJwtValidator.validate(
-                                            vc,
-                                            configService.getCriConfig(criOAuthSessionItem),
-                                            clientOAuthSessionItem.getUserId());
-                                });
-                criStoringService.storeCreatedVcs(
-                        vcResponse, callbackRequest, clientOAuthSessionItem);
+                for (SignedJWT vc : vcResponse.getVerifiableCredentials()) {
+                    assert criOAuthSessionItem != null;
```
Joe-Edwards-GDS: Sorry I think I missed this in the @MikeCollingwood 's review - I don't think we should be using `assert`, as it behaves rather surprisingly (i.e. it does nothing by default).

Either we should check it explicitly and throw an error, or not.

lambdas/check-gpg45-score/src/test/java/uk/gov/di/ipv/core/checkgpg45score/CheckGpg45ScoreHandlerTest.java
```diff
@@ -80,7 +81,7 @@ static void setUp() throws Exception {
                         .ipAddress(TEST_CLIENT_SOURCE_IP)
                         .ipvSessionId(TEST_SESSION_ID)
                         .clientOAuthSessionId(TEST_CLIENT_OAUTH_SESSION_ID)
-                        .scoreThreshold(2)
+                        .lambdaInput(new HashMap<>(Map.of("scoreThreshold", 2)))
```
Joe-Edwards-GDS, line 83: sneaky :eyes:

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/BackEndCriConfig.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.di.ipv.core.library.dto;
+
+import com.nimbusds.jose.jwk.ECKey;
+import lombok.Getter;
+
+import java.net.URI;
+import java.text.ParseException;
+
+@Getter
+public class BackEndCriConfig implements CriConfig {
```
Joe-Edwards-GDS: I think my personal taste is `Backend` rather than `BackEnd`, but I'll get over it

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CredentialIssuerConfig.java
Joe-Edwards-GDS, line 1: Worth noting that @TalatAydinGDS changed the CRI config class a little bit in his last change - hopefully not too much of a conflict.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -2,17 +2,173 @@
...
+                    clientOAuthSessionItem.getGovukSigninJourneyId());
+
+            List<SignedJWT> ticfVcs =
+                    ticfCriService.getTicfVc(
+                            clientOAuthSessionItem,
+                            ipvSessionItem,
+                            journeyType.equals(REUSE_JOURNEY_TYPE)
+                                    ? List.of()
+                                    : userIdentityService.getUserIssuedCredentials(
+                                            clientOAuthSessionItem.getUserId()));
```
Joe-Edwards-GDS, line 125: Interesting - so the plan is that we'll attach a 'journey type' to the state in the journey map :+1:

It's worth noting that this logic may have to change in future - it _should_ be 'send all VCs collected in this session', but that does boil down to 'is it reuse or not'.

I think that's fine for now, and we can update later - we might have to add journey ID to the VC store item to allow us to do so (which may be useful for other reasons as well).

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -2,17 +2,173 @@
...
+
+            if (ticfVcs.isEmpty()) {
+                LOGGER.info("No VC to process - returning next");
+                return JOURNEY_NEXT;
```
Joe-Edwards-GDS, line 129: I think this is probably an error - SPOT is likely to fail us if we don't send a TICF VC

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
```diff
@@ -0,0 +1,133 @@
...
+            HttpResponse<String> ticfCriHttpResponse;
+            try {
+                ticfCriHttpResponse = sendHttpRequest(httpRequestBuilder.build());
+                checkStatusCode(ticfCriHttpResponse);
+            } catch (IOException
+                    | InterruptedException
+                    | IllegalArgumentException
+                    | SecurityException
+                    | TicfCriServiceException e) {
+                // In the case of unavailability, the TICF CRI is deemed optional.
```
Joe-Edwards-GDS, line 91: Have you checked this with David Page?

I think the idea is that the TICF CRI should _always_ return a response to us, it may just not contain any evidence if the service wasn't able to respond. This allows us to prove that we tried to run the check to SPOT etc.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
```diff
@@ -0,0 +1,133 @@
...
+        this.jwtValidator = jwtValidator;
+    }
+
+    public List<SignedJWT> getTicfVc(
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            IpvSessionItem ipvSessionItem,
+            List<String> credentials)
+            throws TicfCriServiceException {
+        try {
+            BackEndCriConfig ticfCriConfig = configService.getBackEndCriConfig(TICF_CRI);
```
Joe-Edwards-GDS: Question about our code style - I notice you don't use `var` very much?

I'd probably lean towards using it for pretty much everything, unless there's a need to cast to a wider type, but keen to hear arguments. (We should probably also _try_ and be consistent - although I know at the moment we aren't at all!)

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
```diff
@@ -0,0 +1,133 @@
...
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static uk.gov.di.ipv.core.library.domain.CriConstants.TICF_CRI;
+
+public class TicfCriService {
```
Joe-Edwards-GDS, line 28: (Some day maybe this will become `BackendCriService`, but I think it's fine to keep it specific for now - we don't know how similar other such CRIs might be, or if we'll ever need them!)

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
Wynndow: I got this from [ADR 114](https://github.com/govuk-one-login/architecture/blob/main/adr/0114-ipvcore-to-ticf-cri.md#adr-0114-the-ipv-journey-and-ticf-risk-assessment). "TICF risk assessments as part of an IPV journey are deemed optional behaviour in the event of unavailability of the TICF risk components".
Perhaps my interpretation is a little off though. Will check in with David.

[RFC 64](https://github.com/govuk-one-login/architecture/blob/main/rfc/0064-ipv-ticf-risk-assessment.md#request-error-responses) also mentions that requests that indicate that the TICF CRI failed to recieve the request should raise a warning - not fail the journey. I read that as any failure in the request. I'm sure I read something about TICF replaying requests after a period of unavailablity somewhere. Would have to read all the RFC/ADR's againt to put my finger on it.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
Wynndow: Yep, I'll yank it. I think it's there to silence a compile time warning.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/BackEndCriConfig.java
Wynndow: Yeah, that looks a bit nicer, happy to change it out.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CredentialIssuerConfig.java
Wynndow: I'll take a look

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: Interesting. I went this way to avoid having to look at the users current state to determine if they're on a reuse journey or not. But the journey ID thing is interesting. Not for now though, as you say.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: The ADRs seem to suggest this should be fine - see comment below.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
Wynndow: I'm not a fan of using `var`. I much prefer being able to immediately see what type something is. When var is used I'll often have to go find a method definition to determine what type something is. I'll _occasionally_ use it if a type is some obnoxiously long generic type of something. But I usually don't find an issue with being explicit and more informative.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
Wynndow: I did think about this, but the logic inside seems *very* TICF specific...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CredentialIssuerConfig.java
Wynndow: Quick rebase and all is well

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
lwebb-iig, line 1: Are these one-liner methods all that useful? `getCriPrivateApiKeyForActiveConnection`, `getCriPrivateApiKey` & `getBackendCriConfig` seem a bit redundant now that the error handling has been moved into `getApiKeyFromSecretManager` and `getCriConfigForType` respectively. Does it make sense to just directly call these methods in the respective lambdas (especially as `getActiveConnection` is public?)

Wynndow: Possibly not - I think I like them though as my feeling is that when you want some config for a thing, you just want to call the config service. If it can abstract some of the fiddly bits away from you (like getting the right connection, or giving it the expected class object) and they don't have to clutter the calling code, then that feels cleaner. Happy to be wrong though.

lwebb-iig: If it makes sense to handle the connection information in the ConfigService then happy to keep as is

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
Joe-Edwards-GDS: Ok, definitely check this with David. My understanding is that the TICF CRI should always give us a response, and it will handle issues with the underlying risk engine is unresponsive, but he should be able to confirm.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
Joe-Edwards-GDS: Ok interesting - I mostly rely on my IDE to tell me the type of variables (which it can still do with `var`) and in my opinion makes it cleaner to write, but it's definitely a preference. It's already a mix so won't try and do anything about this now, but at some point I might try to write a style-guide for us.

(I guess also as an observation, a lot of modern languages have shifted towards var-esque syntax!)

Wynndow: Maybe my IDE isn’t set up right 😅. But if I’m reviewing a PR in GitHub, I don’t really want to have to check out the PR into my IDE. Although GitHub sides have some fancy new features - perhaps it can tell me.

Joe-Edwards-GDS: Yeah, I mostly meant in-IDE rather than GitHub... If it's a complex PR I generally check out the branch locally anyway, but that is annoying.

You get this sort of thing on hover, and it shows the type in autocomplete dialogs etc.
image

Hopefully in most cases it should be obvious - e.g. in this case it calls `getBackendConfig` and returns a `BackendConfig`, but that's not always true.

Joe-Edwards-GDS (approved):
As discussed, happy for you to merge to avoid merge conflicts - we can update after the call this afternoon :)

PR #1427, opened 2023-11-29
PYIC-3875: wire in HMRC KBV to driving licence journey

AmritSidhu:
### What changed

When the HMRC KBV CRI is enabled, you can only get to it by going down the passport route. If you use a driving licence then you are taken to the Experian KBVs.
In both cases you should be taken the NINO CRI and then (if appropriate) the HMRC KBVs.

### Issue tracking

- [PYIC-3875](https://govukverify.atlassian.net/browse/PYIC-3875)

[PYIC-3875]: https://govukverify.atlassian.net/browse/PYIC-3875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -487,7 +487,10 @@ CHECK_FRAUD_SCORE_J3:
       scoreThreshold: 2
   events:
     met:
-      targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE
+      targetState: CRI_NINO_J6
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE
     unmet:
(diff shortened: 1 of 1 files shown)


DanCorderIPV (dismissed):

DanCorderIPV:
Not strictly part of this review, but is it OK that we are dropping the fraud score check before directing users to Experian?

blakeyp:
We still need the fraud score check for the DL route, context for why:
> When users are on a driving license journey (after the multidoc page, not the app journey), they are targeting an M1B profile. That profile requires a fraud score of 2. The fraud CRI can issue a user a fraud score of 1.
> If that happens to a user on the driving license journey they will continue on to the KBV CRI. If they get a successful VC from KBV they will still fail to match the profile. 
> We should fail them earlier, after the fraud CRI, rather than making then unnecessarily go through the KBV CRI.

(from https://govukverify.atlassian.net/browse/PYIC-3283)

shivanshuit914 (dismissed):

blakeyp (approved):

PR #1428, opened 2023-11-29
Pyic 3884: Added dynamic content backend framework 

RogVanGDS:
### What changed

Added dynamic content backend framework and added context to pyi-suggest-other-options page

### Why did it change

We are setting up a framework to allow for front-end pages to display more dynamic content. This requires us passing through a context value from the backend journey-map.

### Issue tracking

- [PYIC-3884](https://govukverify.atlassian.net/browse/PYIC-3884)

[PYIC-3884]: https://govukverify.atlassian.net/browse/PYIC-3884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private final HashMap<String, List<String>> expectedStateContexts =
+            new HashMap<>() {
+                {
+                    put("no-photo-id", List.of("MITIGATION_02_OPTIONS_WITH_F2F_J7"));
+                }
+            };
+
+    @ParameterizedTest
+    @EnumSource
+    void shouldHandleSameContextForSamePage(IpvJourneyTypes journeyType) throws IOException {
```
DanCorderIPV: I think this needs a big comment.
As far as I can see this is testing that anywhere that `MITIGATION_02_OPTIONS_WITH_F2F_J7` is used in the map it should have a context of `no-photo-id` but I'm still not 100% sure even after staring at this code for a while.
Relatedly I'm unsure what this test is supposed to catch exactly. AFAIK we can't add a second `MITIGATION_02_OPTIONS_WITH_F2F_J7` to the journey map so this won't catch people adding that and forgetting the context. It will only catch new contexts if the person adding the new context updates this test but then forgets to include the context in the journey map.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -123,7 +123,7 @@ private Map<String, Object> executeJourneyEvent(
         String currentUserState = ipvSessionItem.getUserState();
         if (sessionIsNewlyExpired(ipvSessionItem)) {
             updateUserSessionForTimeout(currentUserState, ipvSessionItem);
-            return new PageStepResponse(PYIC_TIMEOUT_UNRECOVERABLE_ID).value();
+            return new PageStepResponse(PYIC_TIMEOUT_UNRECOVERABLE_ID, "error").value();
```
DanCorderIPV: This doesn't seem to be mentioned on the ticket?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -525,6 +525,8 @@ CRI_HMRC_KBV_J6:
     criId: hmrcKbv
   parent: CRI_STATE
   events:
+    enhanced-verification:
+      targetState: MITIGATION_03_OPTIONS
```
DanCorderIPV: Should this be part of this ticket?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
RogVanGDS: Ah yep good point, the test should be changed to not check Events against Contexts, but about PageId against Contexts.

My only thought around adding this test was to avoid somebody removing or changing a context value in the JourneyMapPage, as that would result in the front-end displaying the wrong content, or no content at all.

Perhaps this would be better covered as an integration test anyway. @blakeyp, thoughts?

blakeyp: By integration test do you mean the end to end UI tests? It'd be good to prove that the dynamic pages mechanism is working by checking one instance of it, but I don't think we'd want to test every page as that would mean a lot of hardcoding of content in the tests

RogVanGDS: Ok that makes sense, it probably is worth at least having this unit test in there to provide a slight safety net

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
RogVanGDS: Ah I shouldn't have committed that

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -574,14 +574,59 @@ ADDRESS_AND_FRAUD_J7:
...
     end:
       targetState: ERROR
     enhanced-verification:
       targetState: CRI_HMRC_KBV_J6
       checkIfDisabled:
         hmrcKbv:
-          targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE_J7
+
+PRE_EXPERIAN_KBV_TRANSITION_PAGE_J7:
```
Joe-Edwards-GDS: I'm not a massive fan of propagating these incrementing journey numbers - are they actually documented anywhere?

Can we come up with a more descriptive name?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        assertTrue(
+                invalidContexts.isEmpty(),
+                String.format(
+                        "Some journey map contexts are not currently supported in ipv-core-front: %s",
+                        invalidContexts));
+    }
+
+    private void findPagesWithContexts(
+            Map<String, State> stateMachine, List<PageStepResponse> pageContextMap) {
```
DanCorderIPV: `pageContextMap` isn't a map? Maybe `pageStepResponsesWithContexts` or similar?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
```diff
@@ -0,0 +1,87 @@
...
+    private final HashMap<String, List<String>> acceptedStateContexts =
+            new HashMap<>() {
+                {
+                    put("pyi-suggest-other-options", List.of("no-photo-id"));
+                }
+            };
+
+    @ParameterizedTest
+    @EnumSource
+    void shouldHandleSameContextForSamePage(IpvJourneyTypes journeyType) throws IOException {
```
DanCorderIPV: I'm still not clear what this is testing, I don't think the name of test makes it clear.

DanCorderIPV: I think that this will catch cases where someone adds an unexpected context to a page.

Is there anyway we can catch people forgetting to add a context to a page that needs one (that seems more likely to be a problem)? I suspect not, unless we force those pages always take a context e.g. add a `photo-id` possible context for `pyi-suggest-other-options` and use that in all the other places.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -574,14 +574,59 @@ ADDRESS_AND_FRAUD_J7:
       targetState: CRI_HMRC_KBV_J6
```
Joe-Edwards-GDS, line 574: Do we also need to update the HMRC KBV route as well - what if you get a V03 there? (Or is that a separate ticket?)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
RogVanGDS: > I think that this will catch cases where someone adds an unexpected context to a page.

This was what I concluded was the only useful test to have here.

The ticket originally asked if we could use ENUM values for contexts we use in the journey-map config, and this test is trying to enforce a similar behaviour (because we can't use ENUMS)

Enforcing a mandatory context isn't a bad idea, but it would involve a larger amount of refactoring everytime somebody wanted to make a page dynamic (when the purpose of the ticket is to make the process a lot smoother)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -658,6 +703,16 @@ MITIGATION_01_PYI_POST_OFFICE:
     end:
       targetState: PYI_ANOTHER_WAY
 
+MITIGATION_03_OPTIONS:
```
Joe-Edwards-GDS: Unused?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapPageContextTest.java
Joe-Edwards-GDS: The existing journey map tests follow a similar principle - checking that states are consistent with each other, rather than trying to define specific rules.

The rationale is that when adding new context, you aren't going to forget to add it the first one, and this test will make sure you add it to others of the same type (now and in future).

Maintaining a separate list of which pages need what context feels painful to maintain!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
RogVanGDS: That's a separate ticket

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandlerTest.java
Joe-Edwards-GDS, line 1: Did you mean to delete this one?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -535,53 +535,98 @@ CRI_HMRC_KBV_J6:
     next:
       targetState: PYI_NO_MATCH
 
-# No photo id M2B journey (J7)
-CRI_CLAIMED_IDENTITY_J7:
+# No photo id journey (M2B)
+CRI_CLAIMED_IDENTITY_M2B:
```
Joe-Edwards-GDS, line 539: (You probably already know, but in general renaming states needs a two-step deployment - no need in this case because it's not live :) )

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandlerTest.java
RogVanGDS: 🤦‍♂️

Joe-Edwards-GDS (approved):

PR #1432, opened 2023-11-30
PYIC-3799: Remove old CRI return lambdas and tidy up

MikeCollingwood:
### What changed

- Removed cri return step function
- Remove associated lambdas, exc CheckCiScore and EvaluateGpg45Scores which are still used
- Remove redundant code based
- Recoup tests from the old lambdas into new lambda test suites

### Why did it change

- To tidy up after PYIC-3798, making the code easier to develop afterwards

### Issue tracking

- [PYIC-3799](https://govukverify.atlassian.net/browse/PYIC-3799)
- [PYIC-3796](https://govukverify.atlassian.net/browse/PYIC-3796)

### Other considerations

[PYIC-3799]: https://govukverify.atlassian.net/browse/PYIC-3799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-3796]: https://govukverify.atlassian.net/browse/PYIC-3796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -888,336 +879,6 @@ Resources:
       FilterPattern: ""
       LogGroupName: !Ref InitialiseIpvSessionFunctionLogGroup
 
-  RetrieveCriOauthAccessTokenFunction:
```
Joe-Edwards-GDS, line 891: Hard question alert - do you think there's any risk that we could tear down the old lambdas while users are still using the old CRI return?

I don't immediately know the answer... If we're worried we can deploy the API GW part of the change first, but it's a bit faffy. If it does go wrong it should only be a small number of people, but still :/

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -134,16 +120,6 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
                     gpg45ProfileEvaluator.parseCredentials(
                             userIdentityService.getUserIssuedCredentials(userId));
 
-            if (!checkCorrelation(userId)) {
-                return JOURNEY_PYI_NO_MATCH.toObjectMap();
-            }
```
Joe-Edwards-GDS, line 139: I'm not sure we should remove this check - it feels quite good to double-check correlation at the end as well (as it's part of the GPG45 algorithm), what do you think?

(We should be able to use the commonised checking logic in `UserIdentityService` though)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -21,10 +21,7 @@ CRI_STATE:
     temporarily-unavailable:
       targetState: PYI_NO_MATCH
     end:
```
Joe-Edwards-GDS: Can CRI states actually result in an `end` event any more?

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
```diff
@@ -176,46 +169,11 @@ public LambdaHandler() throws IOException {
 
     private final Route criCallBack =
```
Joe-Edwards-GDS, line 169: Do you also need to update the path in `CoreBack.java`?

settings.gradle
Joe-Edwards-GDS, line 1: Do we use `check-ci-score` any more either?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsService.java
```diff
@@ -80,8 +80,4 @@ public ClientOAuthSessionItem generateErrorClientSessionDetails(
 
         return clientOAuthSessionErrorItem;
     }
-
-    public void updateClientOAuthSession(ClientOAuthSessionItem updatedClientOAuthSessionItem) {
-        dataStore.update(updatedClientOAuthSessionItem);
-    }
```
Joe-Edwards-GDS, line 86: (I think this is a reasonable method for the service to have, even if it's currently unused -  but I don't mind either way)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -301,12 +301,6 @@ public String getComponentId(String credentialIssuerId) {
         return getSsmParameterWithOverride(pathTemplate, credentialIssuerId, activeConnection);
     }
 
-    public boolean isUnavailable(String credentialIssuerId) {
-        final String pathTemplate =
-                ConfigurationVariable.CREDENTIAL_ISSUERS.getPath() + "/%s/unavailable";
-        return Boolean.parseBoolean(getSsmParameterWithOverride(pathTemplate, credentialIssuerId));
```
Joe-Edwards-GDS, line 307: @Wynndow do you know what this was used for? It looks like we have some config for 'unavailable' CRIs, but we don't actually use it for anything?

(We already have enabled/disabled for turning off CRIs)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
```diff
@@ -63,16 +63,4 @@ public CriOAuthSessionItem persistCriOAuthSession(
                                 criOAuthSessionItem.getCriOAuthSessionId()));
         return criOAuthSessionItem;
     }
-
-    public void updateCriOAuthSessionItem(CriOAuthSessionItem updatedCriOAuthSessionItem) {
```
Joe-Edwards-GDS, line 67: (ditto)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: They can't, good point! Will also remove pyi-kbv-fail

deploy/template.yaml
MikeCollingwood: I'll make a PR to merge the open api spec in first, that should avoid the issue and won't be too much faff

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
MikeCollingwood: Ok, that makes sense! I don't feel strongly about it, so will add in the check again

settings.gradle
MikeCollingwood: Ah good point, we don't

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsService.java
MikeCollingwood: That's fair, I deleted it to highlight the option mostly

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -134,23 +124,17 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
-
             return checkForMatchingGpg45Profile(
                             userId, ipvSessionItem, clientOAuthSessionItem, credentials, ipAddress)
                     .toObjectMap();
         } catch (HttpResponseExceptionWithErrorBody e) {
             LOGGER.error("Received HTTP response exception", e);
             return new JourneyErrorResponse(
-                            JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
+                            JOURNEY_ERROR_PATH, HttpStatus.SC_BAD_REQUEST, e.getErrorResponse())
```
Joe-Edwards-GDS: Why has this changed?

MikeCollingwood: It works in the context, since any errors come from the request. That said, I meant to revert this change, thanks for the spot

Joe-Edwards-GDS (approved):
Will probably need rebasing again once the other PR is merged

PR #1437, opened 2023-12-01
PYIC-3799: Add clientOAuthSessionId attribute to pageOutput response …

MikeCollingwood:
…on NO_IPV_FOR_CRI_OAUTH_SESSION

### Issue tracking

### Other considerations

### DIFF
--- a/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
+++ b/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
@@ -145,12 +145,16 @@ public class ProcessCriCallbackHandler
         } catch (InvalidCriCallbackRequestException e) {
             if (e.getErrorResponse() == ErrorResponse.NO_IPV_FOR_CRI_OAUTH_SESSION) {
                 LOGGER.error(e.getErrorResponse(), e);
+                var pageOutput = StepFunctionHelpers.generatePageOutputMap(
+                        "error",
+                        HttpStatus.SC_UNAUTHORIZED,
+                        PYI_TIMEOUT_RECOVERABLE_PAGE_ID);
+                assert callbackRequest != null;
+                var criOAuthSessionItem = criOAuthSessionService.getCriOauthSessionItem(callbackRequest.getState());
+                pageOutput.put("clientOAuthSessionId", criOAuthSessionItem.getClientOAuthSessionId());
                 return ApiGatewayResponseGenerator.proxyJsonResponse(
                         HttpStatus.SC_UNAUTHORIZED,
-                        StepFunctionHelpers.generatePageOutputMap(
-                                "error",
-                                HttpStatus.SC_UNAUTHORIZED,
-                                PYI_TIMEOUT_RECOVERABLE_PAGE_ID));
+                        pageOutput);
             }
             if (e.getErrorResponse() == ErrorResponse.INVALID_OAUTH_STATE) {
                 LOGGER.error(e.getErrorResponse(), e);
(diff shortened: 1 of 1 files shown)


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -145,12 +145,16 @@ public APIGatewayProxyResponseEvent handleRequest(
         } catch (InvalidCriCallbackRequestException e) {
             if (e.getErrorResponse() == ErrorResponse.NO_IPV_FOR_CRI_OAUTH_SESSION) {
                 LOGGER.error(e.getErrorResponse(), e);
+                var pageOutput = StepFunctionHelpers.generatePageOutputMap(
+                        "error",
+                        HttpStatus.SC_UNAUTHORIZED,
+                        PYI_TIMEOUT_RECOVERABLE_PAGE_ID);
+                assert callbackRequest != null;
```
MikeCollingwood, line 152: We know this isn't null, because we can get the criOauthSessionId from it

Joe-Edwards-GDS: I don't think we should be using `assert`, it's weird and confusing!

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -145,12 +145,16 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                var criOAuthSessionItem = criOAuthSessionService.getCriOauthSessionItem(callbackRequest.getState());
```
MikeCollingwood: The only thing that could go wrong is if criOAuthSessionService.getCriOauthSessionItem can't get the session item from the id

Joe-Edwards-GDS: If you're worried, you could wrap that in a try catch as well, and just swallow the error for now (so at worst it does what it already does)

MikeCollingwood: I don't think I'm worried because it's just this:
```
    public CriOAuthSessionItem getCriOauthSessionItem(String criOAuthSessionId) {
        return dataStore.getItem(criOAuthSessionId);
    }
```
so would throw a runtime error which isn't caught anyway (my only concern was now we're running this inside a catch statement, what if we rethrow a caught error)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -145,12 +145,17 @@ public APIGatewayProxyResponseEvent handleRequest(
...
                         HttpStatus.SC_UNAUTHORIZED,
-                        StepFunctionHelpers.generatePageOutputMap(
-                                "error",
-                                HttpStatus.SC_UNAUTHORIZED,
-                                PYI_TIMEOUT_RECOVERABLE_PAGE_ID));
+                        PYI_TIMEOUT_RECOVERABLE_PAGE_ID);
+                if (callbackRequest != null) {
+                    var criOAuthSessionItem = criOAuthSessionService.getCriOauthSessionItem(callbackRequest.getState());
+                    pageOutput.put("clientOAuthSessionId", criOAuthSessionItem.getClientOAuthSessionId());
+                    return ApiGatewayResponseGenerator.proxyJsonResponse(
```
Joe-Edwards-GDS: Can we move this return out of the if, so it always returns an error, just only includes the clientOAuthSessionId if available?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
Joe-Edwards-GDS: ok sure, if we get a dynamodb error that's not an issue, I guess it could also return null though?

MikeCollingwood: I don't think it was caught before

Joe-Edwards-GDS (approved):

PR #1438, opened 2023-12-02
BAU: Fix NPE and IAE in refactored CRI callback

Joe-Edwards-GDS:
The new CRI callback has a couple of exceptions which are causing errors (in other error cases), this should resolve them.

### DIFF
--- a/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
+++ b/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
@@ -154,8 +154,11 @@ public class ProcessCriCallbackHandler
                     var criOAuthSessionItem =
                             criOAuthSessionService.getCriOauthSessionItem(
                                     callbackRequest.getState());
-                    pageOutput.put(
-                            "clientOAuthSessionId", criOAuthSessionItem.getClientOAuthSessionId());
+                    if (criOAuthSessionItem != null) {
+                        pageOutput.put(
+                                "clientOAuthSessionId",
+                                criOAuthSessionItem.getClientOAuthSessionId());
+                    }
                 }
                 return ApiGatewayResponseGenerator.proxyJsonResponse(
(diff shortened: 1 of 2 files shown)


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -154,8 +154,11 @@ public APIGatewayProxyResponseEvent handleRequest(
                     var criOAuthSessionItem =
                             criOAuthSessionService.getCriOauthSessionItem(
                                     callbackRequest.getState());
-                    pageOutput.put(
-                            "clientOAuthSessionId", criOAuthSessionItem.getClientOAuthSessionId());
+                    if (criOAuthSessionItem != null) {
+                        pageOutput.put(
+                                "clientOAuthSessionId",
+                                criOAuthSessionItem.getClientOAuthSessionId());
+                    }
```
Joe-Edwards-GDS, line 161: Although we know the `clientOAuthSessionId` exists to get to this point, we can't guarantee that the corresponding session still does (it may have expired).

Previously this caused a `NullPointerException`

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -90,7 +91,9 @@ public JourneyResponse handleCallbackError(
         var criId = callbackRequest.getCredentialIssuerId();
         var ipAddress = callbackRequest.getIpAddress();
         var error = callbackRequest.getError();
-        var errorDescription = callbackRequest.getErrorDescription();
+        var errorDescription =
+                Objects.toString(
+                        callbackRequest.getErrorDescription(), "No error description provided");
```
Joe-Edwards-GDS, line 96: Sometimes no error description is provided, and this breaks the logger method with an `IllegalArgumentException`.

AmritSidhu (approved):

PR #1445, opened 2023-12-04
PYIC-3748: change default error state for journey map events

shivanshuit914:
## Proposed changes

- Change defaults to CRI journey map events, change error maps.

### What changed

- Changed defaults journey map for the cri state errors.

### Why did it change

- To trace error correctly in logs.

### Issue tracking

- [PYIC-3748](https://govukverify.atlassian.net/browse/PYIC-3748)

[PYIC-3748]: https://govukverify.atlassian.net/browse/PYIC-3748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -7,19 +7,19 @@ END_JOURNEY:
 CRI_STATE:
   events:
     next:
-      targetState: ERROR
+      targetState: UNEXPECTED_EVENT
     not-found:
-      targetState: ERROR
+      targetState: PYI_NO_MATCH
     fail-with-no-ci:
-      targetState: ERROR
+      targetState: PYI_NO_MATCH
     error:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -7,19 +7,19 @@ END_JOURNEY:
 CRI_STATE:
   events:
     next:
-      targetState: ERROR
+      targetState: UNEXPECTED_EVENT
```
Joe-Edwards-GDS, line 10: Is it worth adding a comment here to describe what you should do if this occurs? i.e. 'you must define this route explicitly for each CRI'

I also think it'd be fine to just omit this (especially now we've removed `end`), but being explicit is nice too :)

Joe-Edwards-GDS (approved):
This looks fine, but a couple of considerations

PR #1451, opened 2023-12-06
PYIC-4013: Added reprove_identity analytics parameter to IPV_JOURNEY_…

RogVanGDS:
…START event

### What changed

Added reprove_identity analytics parameter to IPV_JOURNEY_START event

### Why did it change

To facilitate the generation of reports that track how many identity proofing journeys are initiated as a result of account interventions.

### Issue tracking

- [PYIC-4013](https://govukverify.atlassian.net/browse/PYIC-4013)

[PYIC-4013]: https://govukverify.atlassian.net/browse/PYIC-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -162,19 +162,21 @@ public class InitialiseIpvSessionHandler
 
             String reproveIdentity = claimsSet.getStringClaim(REPROVE_IDENTITY_KEY);
 
-            AuditExtensionsReproveIdentity auditExtension = reproveIdentity == null ? null :
-                    new AuditExtensionsReproveIdentity(reproveIdentity);
+            AuditExtensionsReproveIdentity reproveAuditExtension =
+                    reproveIdentity == null
+                            ? null
+                            : new AuditExtensionsReproveIdentity(reproveIdentity);
 
-            AuditEvent auditEvent = new AuditEvent(
-                    AuditEventTypes.IPV_JOURNEY_START,
-                    configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
-                    auditEventUser,
-                    auditExtension);
+            AuditEvent auditEvent =
+                    new AuditEvent(
+                            AuditEventTypes.IPV_JOURNEY_START,
+                            configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser,
+                            reproveAuditExtension);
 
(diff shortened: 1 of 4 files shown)


DanCorderIPV:
This looks good to me, are there any issues/considerations with deploying it in sync with code that will understand/report on the new data?

DanCorderIPV (dismissed):

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEvent.java
```diff
@@ -33,13 +33,19 @@ public class AuditEvent {
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private final AuditRestricted restricted;
 
+    @JsonProperty("reprove_identity")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private final Boolean reproveIdentity;
```
Joe-Edwards-GDS: Are you sure this should go on the top-level audit event, as opposed to the `extensions`? Usually our custom fields go into the extensions object, which can vary across events.

Joe-Edwards-GDS (commented):
You will also need to go through the TxMA process for updating an audit event, if you haven't already (https://govukverify.atlassian.net/wiki/spaces/TMA/pages/3597762595/How+To+Guides+for+Other+Teams)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -126,6 +126,8 @@ public APIGatewayProxyResponseEvent handleRequest(
             String govukSigninJourneyId =
                     claimsSet.getStringClaim(REQUEST_GOV_UK_SIGN_IN_JOURNEY_ID_KEY);
             String emailAddress = claimsSet.getStringClaim(REQUEST_EMAIL_ADDRESS_KEY);
+            String isReproveIdentity = claimsSet.getStringClaim("reprove_identity");
```
Joe-Edwards-GDS: Do we have this key as a constant anywhere?

Joe-Edwards-GDS (dismissed):
Looks cleaner now, thanks. As above, you'll still need to go through the TxMA process, but hopefully someone in Hilux can help you out with that if needed (otherwise we've also done it a few times in Jet)

Joe-Edwards-GDS (approved):
If the ticket explicitly says we shouldn't provide the extension, then I think this makes sense.

(I'm a bit surprised, because I would have thought it'd be more useful to always return the extension but with `true` vs. `false` - but not my ticket!)

PR #1456, opened 2023-12-07
PYIC-4152: exitCode -> returnCode

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-core-common-infra/pull/804**

### What changed

exitCode -> returnCode

### Why did it change

The requirement for the format of the codes returned in the identity bundle has changed.

There will be a follow up PR to this once we've switched over to the new config, and the config repo has been updated.

### Issue tracking

- [PYIC-4152](https://govukverify.atlassian.net/browse/PYIC-4152)

[PYIC-4152]: https://govukverify.atlassian.net/browse/PYIC-4152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorConfig.java
@@ -1,5 +1,6 @@
 package uk.gov.di.ipv.core.library.domain;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -7,10 +8,10 @@ import lombok.NoArgsConstructor;
 @Getter
 @NoArgsConstructor
 @AllArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class ContraIndicatorConfig {
     private String ci;
     private Integer detectedScore;
     private Integer checkedScore;
-    private String exitCode;
     private String returnCode;
 }
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -22,6 +22,7 @@ import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.IdentityClaim;
 import uk.gov.di.ipv.core.library.domain.Name;
 import uk.gov.di.ipv.core.library.domain.NameParts;
+import uk.gov.di.ipv.core.library.domain.ReturnCode;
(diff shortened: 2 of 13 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -181,29 +182,38 @@ public UserIdentity generateUserIdentity(
...
             throws UnrecognisedCiException {
-        return mapCisToExitCodes(contraIndicators).stream()
-                .filter(configService.getSsmParameter(EXIT_CODES_ALWAYS_REQUIRED)::contains)
+        List<ReturnCode> returnCodes = mapCisToReturnCodes(contraIndicators);
+        return returnCodes.stream()
+                .filter(
+                        returnCode ->
+                                returnCode
+                                        .code()
+                                        .equals(
```
Joe-Edwards-GDS: Should this be equals, or contains?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorConfig.java
```diff
@@ -12,4 +12,5 @@ public class ContraIndicatorConfig {
     private Integer detectedScore;
     private Integer checkedScore;
     private String exitCode;
```
Joe-Edwards-GDS: Can we just delete the old `exitCode`, if we have both in the config now? (and then remove it from the config later)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Ah yep, good catch. It would be fine for now, but if we ever add more it'll fail. I'll update.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorConfig.java
Wynndow: There's an new PR which I'm about to merge that adds returnCode independantly of this PR, as it's needed to consume the new config. So this is going to get a rebase, and will remove exitCode then.

Joe-Edwards-GDS: Ok cool

I wasn't sure how it would deal with extra properties (I think it depends on the JSON deserialisation settings), so hoped it might _just work_ - not too hard to do a two-step deploy though.

Wynndow: No, it'll bork on unexpected properties. 
I've just pushed a change to this PR that drops the exitCode fields and also ignores unknown properties.

Joe-Edwards-GDS (approved):

PR #1460, opened 2023-12-08
PYIC-4076: Send new audit event on mitigationStart

MikeCollingwood:
### What changed

- Added mitigationStart attributes to states at the start of mitigation journeys
- Added conditional sending of audit event IPV_MITIGATION_START when mitigationStart is true

### Why did it change

- This will enable us to calculate conversion rates for the app and f2f routes a user completes.

### Issue tracking

- [PYIC-4076](https://govukverify.atlassian.net/browse/PYIC-4076)

### Other considerations

- [ ] [Follow process to add a new audit event](https://govukverify.atlassian.net/wiki/spaces/TMA/pages/3626696705/1.+Follow+the+process+outlined+here#Adding-new-events)

[PYIC-4076]: https://govukverify.atlassian.net/browse/PYIC-4076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/PageStepResponse.java
```diff
@@ -13,10 +13,9 @@ public class PageStepResponse implements StepResponse {
 
     private String pageId;
     private String context = "";
+    private String mitigationStart = "";
```
MikeCollingwood: These 3 could be made into a single URI. We'd have to extract `context` in `processJourneyEvent` and add it to the response, if it's there. I'd also have to take the base URI for `page` before sending off

lambdas/process-journey-event/build.gradle
```diff
@@ -10,11 +10,14 @@ repositories {
...
-	implementation "com.amazonaws:aws-lambda-java-core:$rootProject.ext.dependencyVersions.awsLambdaJavaCore",
+	implementation "com.amazonaws:aws-java-sdk-sqs:$rootProject.ext.dependencyVersions.awsJavaSdkSqs",
+			"com.amazonaws:aws-lambda-java-core:$rootProject.ext.dependencyVersions.awsLambdaJavaCore",
 			"com.amazonaws:aws-lambda-java-events:$rootProject.ext.dependencyVersions.awsLambdaJavaEvents",
 			"com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:$rootProject.ext.dependencyVersions.jacksonDataformatYaml",
 			"com.nimbusds:oauth2-oidc-sdk:$rootProject.ext.dependencyVersions.nimbusdsOauth2OidcSdk",
-			project(":libs:common-services")
+			project(":libs:common-services"),
+			project(":libs:audit-service"),
+			project(":libs:audit-service")
```
Joe-Edwards-GDS, line 19: Duplicated line?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -85,35 +104,43 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
                 throw new HttpResponseExceptionWithErrorBody(
                         HttpStatus.SC_BAD_REQUEST, ErrorResponse.INVALID_SESSION_ID);
             }
+            ClientOAuthSessionItem clientOAuthSessionItem =
+                    clientOAuthSessionService.getClientOAuthSession(
+                            ipvSessionItem.getClientOAuthSessionId());
+            configService.setFeatureSet(featureSet);
 
-            ClientOAuthSessionItem clientOAuthSessionItem;
-            if (ipvSessionItem.getClientOAuthSessionId() != null) {
```
Joe-Edwards-GDS, line 101: Are you confident we don't need this null check? (Do we also need to check if `clientOAuthSessionItem` is `null`?)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -245,4 +272,52 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
         }
         return stateMachinesMap;
     }
+
+    @Tracing
+    private Map<String, Object> checkForQueryParameters(
```
Joe-Edwards-GDS: I don't think this is a great name - what is it checking, and what does it return?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -245,4 +272,52 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
...
+                            ipAddress);
+
+            auditService.sendAuditEvent(
+                    new AuditEvent(
+                            AuditEventTypes.IPV_MITIGATION_START,
+                            configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser));
+        }
+
+        // Strip queries if page
```
Joe-Edwards-GDS: This feels pretty awkward - we're doing a bunch of work to serialize `mitigationStart` into the step response and then intercepting it here to remove it again.

Instead, can we return a `StepResponse` from `executeJourneyStep` (which feels more natural anyway!), and then you can check that for the `mitigationStart` property without impacting the `value()` - either by adding a new method to the interface or explicitly matching to the relevant subclass.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/PageStepResponse.java
```diff
@@ -3,20 +3,48 @@
...
 @Data
 @NoArgsConstructor
 @AllArgsConstructor
 public class PageStepResponse implements StepResponse {
+    private static final Logger LOGGER = LogManager.getLogger();
 
     private String pageId;
-    private String context = "";
+    private String context;
+    private String mitigationStart;
```
Joe-Edwards-GDS: Should this be a `Boolean`?

Joe-Edwards-GDS (commented):
I think you might be able to simplify this a bit if you don't try and include this in the response value

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
MikeCollingwood: Would the variables not have to be initialised? The context had been initialised as "" before. So I wonder if, like before, we can't optional attributes

Joe-Edwards-GDS: Not sure what you mean - which variables?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -245,4 +265,46 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
...
+
+    @Tracing
+    private void checkIfStartingMitigationJourney(
+            Map<String, String> input,
+            StepResponse stepResponse,
+            ClientOAuthSessionItem clientOAuthSessionItem)
+            throws HttpResponseExceptionWithErrorBody, SqsException {
+
+        String ipvSessionId = StepFunctionHelpers.getIpvSessionId(input);
+        String ipAddress = StepFunctionHelpers.getIpAddress(input);
```
Joe-Edwards-GDS: Can we just pass these as parameters?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
MikeCollingwood: I think practically we don't need to check for it, the id is always generated when an ipv session is made & to check - there aren't any null values in build dynamodb

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -245,4 +265,46 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
...
+    }
+
+    private Boolean getMitigationStart(StepResponse stepResponse) {
+        if (stepResponse instanceof PageStepResponse) {
+            return ((PageStepResponse) stepResponse).getMitigationStart();
+        } else if (stepResponse instanceof ProcessStepResponse) {
+            return ((ProcessStepResponse) stepResponse).getMitigationStart();
+        }
+        return false;
+    }
```
Joe-Edwards-GDS: You can cast and assign in a single expression to simplify this. You might also be able to use one of a few tricks to simplify the boolean check (here are two options, you can decide which you think is cleaner...)
```
if (stepResponse instanceof PageStepResponse pageStepResponse) {
    return Objects.requireNonNullElse(pageStepResponse.getMitigationStart(), false);
} else if (stepResponse instanceof ProcessStepResponse processStepResponse) {
    return Boolean.TRUE.equals(processStepResponse.getMitigationStart());
}

return false;
```

Incidentally, if we were using Java 21, this could be a neat use for a pattern matching expression. Something like:
```
return Boolean.TRUE.equals(switch (stepResponse) {
    case PageStepResponse page -> page.getMitigationStart();
    case ProcessStepResponse process -> process.getMitigationStart();
    default -> null;
});
```

Joe-Edwards-GDS: You could also consider whether you want to limit `mitigationStart` to just `page` and `process` states - if it applies to everything then you can add it to the interface and no casting is needed :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/PageStepResponse.java
```diff
@@ -12,11 +14,14 @@
...
+    private String context;
+    @Getter private Boolean mitigationStart;
 
     public Map<String, Object> value() {
-        return Map.of(
-                "page", pageId,
-                "context", context);
+        Map<String, Object> response = new HashMap<>();
+        response.put("page", pageId);
+        response.put("context", context);
```
Joe-Edwards-GDS, line 23: Why has this changed - do you need the response to be mutable?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachineInitializerTest.java
```diff
@@ -79,7 +77,8 @@ void stateMachineInitializerShouldCorrectlyDeserializeJourneyMaps()
                 (NestedJourneyInvokeState) journeyMap.get("NESTED_JOURNEY_INVOKE_STATE");
 
         // page state assertions
-        assertEquals("page-id-for-some-page", pageState.getResponse().value().get("page"));
+        assertEquals(
+                "page-id-for-some-page?context=test", pageState.getResponse().value().get("page"));
```
Joe-Edwards-GDS: (probably needs reverting?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -288,7 +288,7 @@ public Map<String, ContraIndicatorConfig> getContraIndicatorConfigMap() {
             }
             return configMap;
         } catch (JsonProcessingException e) {
-            LOGGER.error("Failed to parse contra-indicator config");
+            LOGGER.error("Failed to parse contra-indicator config: {}", e.getMessage());
```
Joe-Edwards-GDS: Do you think this error message will ever include details of our CIMIT configuration?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -245,4 +265,46 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
         }
         return stateMachinesMap;
     }
+
+    @Tracing
+    private void checkIfStartingMitigationJourney(
```
Joe-Edwards-GDS: I still don't love this name - can we call it something more like `sendMitigationStartAuditEvent` (or if you insist, `sendMitigationStartAuditEventIfRequired`), so it's clear what it does?

(The fact it sends the audit event is the important bit - the fact it checks the journey step property is an implementation detail!)

Joe-Edwards-GDS (commented):
Thanks, I think this is looking a lot better

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Ok makes sense. I'm also wondering if there's any way e.g. the client Oauth session could expire before the IPV session, but we didn't actually catch that before either.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/PageStepResponse.java
MikeCollingwood: I changes it so the context wouldn't have to be initialised. Map.of requires non-null values

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
MikeCollingwood: It does, I didn't realise we're keeping them secret, but that would make sense! I'll revert this edit

Joe-Edwards-GDS: Yeah, it's one of those slightly fuzzy lines - we could definitely log it if it's useful (since access to the logs is controlled), but in general it's good to avoid logging more sensitive info than we need.

Joe-Edwards-GDS (approved):
Assuming you've followed the proper TxMA process, this should be good

PR #1468, opened 2023-12-11
BAU: Add not correlated event

Wynndow:
### What changed

Add not correlated event

### Why did it change

We check a users VCs to ensure the data withing is correlated. We do this in the evaluate-gpg45-scores lambda and the process-cri-callback lambdas. If the data wasn't correlated we would thrown a `/journey/pyi-no-match` event. This is the only place this event is used. We weren't handling it in the journey map on either of the required states.

This removes the old journey event and replaces it with something more descriptive.


Joe-Edwards-GDS (dismissed):
Nice :)

Joe-Edwards-GDS (approved):

PR #1472, opened 2023-12-12
PYIC-4139: Create lambda and scripts to re-send VCs to SPOT

lwebb-iig:
Creates a lambda and a collection of scripts to pull relevant VCs from DynamoDB, query them and send to CIMIT/SPOT.

### What changed

- Adds new `replay-cimit-vcs` lambda that accepts a list of VC UUIDs, looks them up and sends to CIMIT/SPOT via SQS.
- Adds bash scripts to scan `user-issued-credentials-v2` dynamo table for VC details in a specific time range and groups them by CRI.

### Why did it change

- [Limitations on the ability to query by date in dynamoDB
](https://stackoverflow.com/questions/14836600/querying-dynamodb-by-date)
- Due to the size of the VC table in production, performing one-off scans to facilitate this task and query data locally was deemed an appropriate response to this specific issue

### Issue tracking

- [PYIC-4139](https://govukverify.atlassian.net/browse/PYIC-4139)

### Other considerations

[PYIC-4139]: https://govukverify.atlassian.net/browse/PYIC-4139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -2410,11 +2492,22 @@ Resources:
...
         - AttributeName: "credentialIssuer"
           AttributeType: "S"
+        - AttributeName: "dateCreated"
+          AttributeType: "S"
       KeySchema:
         - AttributeName: "userId"
           KeyType: "HASH"
         - AttributeName: "credentialIssuer"
           KeyType: "RANGE"
+      GlobalSecondaryIndexes:
```
Joe-Edwards-GDS: Are we still keeping this?

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
```diff
@@ -0,0 +1,90 @@
...
+         Env=$OPTARG;;
+      p) # Enter an AWS profile
+         Profile=$OPTARG;;
+     \?) # Invalid option
+         echo "Error: Invalid option"
+         exit;;
+   esac
+done
+
+echo Performing scan for $Profile in $Env
```
Joe-Edwards-GDS, line 42: Call me paranoid, do you think it's worth showing a warning before using this script? Something like 'this may take several minutes and use up read capacity etc., do you wish to continue? (Y/n)'

Alternatively, putting a similar warning in the README?

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Scan script
+aws dynamodb scan \
+--table-name user-issued-credentials-v2-$Env \
+--filter-expression "attribute_exists(dateCreated) AND credentialIssuer = :vc" \
```
Joe-Edwards-GDS: Can we filter on date here as well? Presumably as written, this will dump the details of every single VC in the store, which will be massive!

Even if it doesn't change the cost, filtering dynamo-side will save a load of unnecessary data transfer.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -128,6 +128,29 @@ public List<T> getItemsWithAttributeLessThanOrEqualValue(
                 .toList();
     }
 
+    public List<T> getItemsByRange(
```
Joe-Edwards-GDS: Is this right? I thought the idea was that we are using the bash script to run the scan to get the list of VCs, and then we don't need to rescan in the lambda? (i.e. we just pass the lambda a list of keys to resend)

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
```diff
@@ -0,0 +1,90 @@
...
+--projection-expression "#ci, #dc, #uid" \
+--expression-attribute-names file://expression-attribute-names.json \
+--expression-attribute-values file://expression-attribute-values-address.json > ./$Env/address-results.json \
+--profile $Profile &&
```
Joe-Edwards-GDS: Do we need to chain these together with `&&` if it's in a script? Can we just run them as sequential steps?

You might need to set the 'e' flag at the top (i.e. `set -e`) so it exits if any command fails

deploy/template.yaml
lwebb-iig: Nope, removed

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
lwebb-iig: Yes that's right, this code will be removed once I've updated the lambda

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
lwebb-iig: I'll have a go at adjusting as you've suggested here and run against build again.

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
lwebb-iig: Good idea - it's pretty quick on build, but that obviously won't be the case in prod

deploy/template.yaml
Joe-Edwards-GDS: I think you removed it before I'd even submitted the comment :D

lambdas/replay-cimit-vcs/dynamo-scripts/scan.sh
Joe-Edwards-GDS: If you're feeling fancy you could create a shared function so it's clearer that it's doing the same thing for each CRI. Coding in bash is a pain though :(

lambdas/replay-cimit-vcs/dynamo-scripts/expression-attribute-values-address.json
Joe-Edwards-GDS, line 1: There's obviously a lot we could do to avoid all this repetition, but I think it's not worth spending any time on it until we actually need to do it again!

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
```diff
@@ -0,0 +1,83 @@
...
+    public ReplayCimitVcsHandler() {
+        this.configService = new ConfigService();
+        this.ciMitService = new CiMitService(configService);
+        this.verifiableCredentialService = new VerifiableCredentialService(configService);
+    }
+
+    @Override
+    @Tracing
+    @Logging(clearState = true)
+    public void handleRequest(InputStream inputStream, OutputStream outputStream, Context context) {
```
Joe-Edwards-GDS, line 58: Is there a reason we're using the `InputStream` signature here, rather than allowing AWS to deserialize into the `ReplayRequest` type for us?

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
```diff
@@ -0,0 +1,83 @@
...
+        for (ReplayItem item : request.getItems()) {
+            VcStoreItem vcStoreItem =
+                    this.verifiableCredentialService.getVcStoreItem(
+                            item.getUserId().get("S"), item.getCredentialIssuer().get("S"));
+            vcStoreItems.add(vcStoreItem);
+        }
+        List<String> vcs = vcStoreItems.stream().map(VcStoreItem::getCredential).toList();
+        LOGGER.info("Submitting {} VCs to CIMIT", vcs.size());
+        try {
+            ciMitService.submitMitigatingVcList(vcs, null, null);
```
Joe-Edwards-GDS: Do you know what the max size of this request is? IIRC lambda has a limit of 6MB - is there any risk we'll hit this?

If there is, we may want to send to CIMIT in batches (say 100 VCs at a time). This has the advantage that we might be able to resume if the replay fails part-way through.

Joe-Edwards-GDS: I think we really want to call `ciMitService.submitVC` (i.e. to detect CIs)

We probably want to submit as a possible mitigation as well, but that's the less important case tbh

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
```diff
@@ -0,0 +1,83 @@
...
+            LOGGER.error("Failed to map request to valid replay event", e);
+            throw new FailedVcReplayException(
+                    String.format("Failed to replay VCs to CIMIT because: '%s'", e));
+        }
+        LOGGER.info("Retrieving {} VCs", request.getItems().size());
+        List<VcStoreItem> vcStoreItems = new ArrayList<>();
+        for (ReplayItem item : request.getItems()) {
+            VcStoreItem vcStoreItem =
+                    this.verifiableCredentialService.getVcStoreItem(
+                            item.getUserId().get("S"), item.getCredentialIssuer().get("S"));
```
Joe-Edwards-GDS, line 94: Is this reasonably fast? It's going to make a separate call to DynamoDB for each VC, which could add up quickly...

If it's a problem, you might want to batch them up (i.e. https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html) - these could be the batches that you send to CIMIT as well.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ReplayRequest.java
```diff
@@ -0,0 +1,26 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ReplayRequest {
+    @JsonProperty("Items")
+    private List<ReplayItem> items;
+
+    @JsonProperty("Count")
+    private Integer count;
+
+    @JsonProperty("ScannedCount")
+    private Integer scannedCount;
```
Joe-Edwards-GDS, line 25: What are the `Count` and `ScannedCount` fields for? (Are they just leftovers from the scan output?)

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
lwebb-iig: I had a go at doing at using the in built AWS deserializer but it seemed to be falling over when trying to deserialize into the ReplayRequest object - in the interests of time I resorted to using an InputStream

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
lwebb-iig: Good idea to send in batches, I'll update - I think we should be well within the 6mb limit based on the size of the files grabbed from prod (max is 1700 VCs in one file)

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
lwebb-iig: I considered this, the current implementation seems quick, although I've only tried in a relatively small table on dev - something to look at when updating to send in batches

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ReplayRequest.java
lwebb-iig: Leftovers from the scan output - we don't strictly need these, but have left them in as might be useful for debugging

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
Joe-Edwards-GDS: Ok - if we pick this up again I'd like to figure that out, but it's probably not worth it now :)

Joe-Edwards-GDS (dismissed):
As discussed, there are various optimisations we _might_ be able to make here, but I think this should be good enough to handle this case.

(The batching means even if it goes wrong a couple of times, we can resume fairly easily)

Joe-Edwards-GDS (approved):

PR #1473, opened 2023-12-12
PYIC-4189: handled to the new property 'issuers' and changed response IPV_IDENTITY_ISSUED event for fail return code

TalatAydinGDS:
### What changed

D&A need to have access to metrics around Failure Codes, however, it is important that they are not only able to see how many failure codes have been raised, but they also need to be able to see what CRI’s resulted in those Failure codes. 

This story builds on the existing feature to add information about which CRI’s produced failure codes to the event sent to TxMA

### Why did it change

To facilitate future analysis around the performance of the CRI’s and more specifically which ones are resulting in certain failure codes

### Issue tracking

- [PYIC-4189](https://govukverify.atlassian.net/browse/PYIC-4189)

### Other considerations

[PYIC-4189]: https://govukverify.atlassian.net/browse/PYIC-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
+++ b/lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
@@ -23,8 +23,11 @@ import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsUserIdentity;
 import uk.gov.di.ipv.core.library.cimit.exception.CiRetrievalException;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
+import uk.gov.di.ipv.core.library.domain.AuditEventReturnCode;
+import uk.gov.di.ipv.core.library.domain.ContraIndicatorConfig;
 import uk.gov.di.ipv.core.library.domain.ContraIndicators;
 import uk.gov.di.ipv.core.library.domain.UserIdentity;
+import uk.gov.di.ipv.core.library.domain.VectorOfTrust;
 import uk.gov.di.ipv.core.library.dto.AccessTokenMetadata;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
@@ -44,8 +47,15 @@ import uk.gov.di.ipv.core.library.service.IpvSessionService;
 import uk.gov.di.ipv.core.library.service.UserIdentityService;
 
 import java.time.Instant;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.RETURN_CODES_ALWAYS_REQUIRED;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.RETURN_CODES_NON_CI_BREACHING_P0;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_LAMBDA_RESULT;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_VOT;
@@ -156,7 +166,7 @@ public class BuildUserIdentityHandler
             userIdentity.getVcs().add(signedCiMitJwt.serialize());
 
             sendIdentityIssuedAuditEvent(
-                    ipvSessionItem, auditEventUser, contraIndicators, userIdentity);
+                    ipvSessionItem, ipvSessionItem.getVot(), auditEventUser, contraIndicators);
 
             ipvSessionService.revokeAccessToken(ipvSessionItem);
 
@@ -191,16 +201,16 @@ public class BuildUserIdentityHandler
 
     private void sendIdentityIssuedAuditEvent(
             IpvSessionItem ipvSessionItem,
+            String vot,
             AuditEventUser auditEventUser,
-            ContraIndicators contraIndicators,
-            UserIdentity userIdentity)
+            ContraIndicators contraIndicators)
             throws SqsException {
         AuditExtensionsUserIdentity extensions =
                 new AuditExtensionsUserIdentity(
                         ipvSessionItem.getVot(),
                         ciMitUtilityService.isBreachingCiThreshold(contraIndicators),
                         contraIndicators.hasMitigations(),
-                        userIdentity.getReturnCode());
+                        getReturnCodes(vot, contraIndicators));
 
         LOGGER.info(
                 new StringMapMessage()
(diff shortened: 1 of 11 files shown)


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
```diff
@@ -2,15 +2,25 @@
...
-        @JsonProperty boolean ciFail,
-        @JsonProperty boolean hasMitigations,
-        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<ReturnCode> returnCode)
-        implements AuditExtensions {}
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class AuditExtensionsUserIdentity<T> implements AuditExtensions {
+    @JsonProperty private String levelOfConfidence;
```
Joe-Edwards-GDS, line 20: Why is this `private`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -15,7 +15,7 @@
 @Setter
 @ExcludeFromGeneratedCoverageReport
 @Builder
-public class UserIdentity {
+public class UserIdentity<T> {
```
Joe-Edwards-GDS: Sorry I think I might have been unclear - I don't think this should be generic:
- The return code in the user identity bundle (this one) should stay the same
- The return code info in the audit event should use the new format for _all_ return codes (failure or success). Each of these will have `issuers` except `X`, which can leave it as null.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -152,12 +157,74 @@ public UserIdentity generateUserIdentity(
 
             userIdentityBuilder.returnCode(getSuccessReturnCode(contraIndicators));
         } else {
-            userIdentityBuilder.returnCode(getFailReturnCode(contraIndicators));
```
Joe-Edwards-GDS: I think it was quite neat having a single method call here to get the return code, because it keeps this main method simpler - is there a reason you brought the logic back in here?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS, line 1: Because we now need different structures for the audit event compared to the user identity, you're going to have to separate:
- Logic to determine the return codes (for the user identity bundle)
- Constructing the identity bundle info (adding in the `issuers`)

This probably means a helper method in `BuildUserIdentityHandler` (or perhaps here) to do the conversion, given the return codes and contraindicators.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
```diff
@@ -2,15 +2,25 @@
...
+    @JsonProperty boolean ciFail;
+    @JsonProperty boolean hasMitigations;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonProperty("https://vocab.account.gov.uk/v1/returnCode")
```
Joe-Edwards-GDS: Sorry I think this is wrong on the ticket, I don't think it's helpful to use this name in the audit event - we can just use `returnCodes`.

Joe-Edwards-GDS: I've updated the ACs

Joe-Edwards-GDS (changes requested):

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
TalatAydinGDS: I thought, here, based on the CI codes, returncodes are determined in the config according to spesific logic, and ReturnCode is set accordingly. That's why, I added all logic here. I thought, before task requirement success and fail audit event return code is different. That's why, I used 2 different objects. ReturnCode, FailureEventReturnCode.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -215,6 +225,80 @@ private void sendIdentityIssuedAuditEvent(
                         extensions));
     }
 
+    private List<AuditEventReturnCode> getReturnCodes(
```
Joe-Edwards-GDS: Do you think we can simplify this if we pass the return codes from the user identity here? That way we don't need to duplicate that logic that determines the return codes themselves, and we can be sure that the return codes in the audit event match the user identity.

I'm imagining something like this in the audit event method
```
var auditEventReturnCodes = userIdentity.getReturnCode().stream()
        .map((returnCode -> getAuditEventReturnCode(returnCode, contraIndicators))).toList();
```

With this as the interesting part:
```
private AuditEventReturnCode getAuditEventReturnCode(
        ReturnCode returnCode,
        ContraIndicators contraIndicators
) {
    var ciConfig = configService.getContraIndicatorConfigMap();
    var issuers = contraIndicators.getContraIndicatorsMap().values().stream()
            .filter(ci -> ciConfig.get(ci.getCode()).getReturnCode().equals(returnCode.code()))
            .flatMap(ci -> ci.getIssuers().stream())
            .toList();
    return new AuditEventReturnCode(returnCode.code(), issuers);
}
```
(There might need to be some additional null-checking in there somewhere, but you get the idea...)

Joe-Edwards-GDS: This means we'll iterate through the config map once per return code, but in 99% of cases we only expect to have 0 or 1 return codes, so that seems fine!

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -215,6 +225,80 @@ private void sendIdentityIssuedAuditEvent(
...
+                                                        && contraIndicator.getCode() != null
+                                                        && contraIndicatorConfig != null
+                                                ? contraIndicatorConfig.getReturnCode()
+                                                : configService.getSsmParameter(
+                                                        RETURN_CODES_NON_CI_BREACHING_P0);
+                            }
+
+                            List<String> issuers =
+                                    contraIndicator.getIssuers() != null
+                                                    && contraIndicators.hasMitigations()
```
MikeCollingwood: If one contraindicator has a mitigation, but another doesn't, will this ignore that unmitigated contraindicator's issuer?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -215,6 +225,80 @@ private void sendIdentityIssuedAuditEvent(
...
+                                        .contains(returnCode.code()))
+                .toList();
+    }
+
+    private List<AuditEventReturnCode> getAuditEventReturnCodes(
+            Map<String, ContraIndicatorConfig> configMap,
+            String vot,
+            ContraIndicators contraIndicators,
+            boolean isBreachingCiThreshold) {
+        return contraIndicators.getContraIndicatorsMap().values().stream()
```
MikeCollingwood: could we define a variable for `contraIndicators.getContraIndicatorsMap().values()` before, to make it slightly less nested?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -215,6 +225,80 @@ private void sendIdentityIssuedAuditEvent(
...
+                            }
+
+                            List<String> issuers =
+                                    contraIndicator.getIssuers() != null
+                                                    && contraIndicators.hasMitigations()
+                                            ? contraIndicator.getIssuers()
+                                            : Collections.emptyList();
+                            return Map.entry(
+                                    Objects.requireNonNull(returnCode),
+                                    issuers.stream().distinct().collect(Collectors.toList()));
```
MikeCollingwood: Could the getIssuers() method be overridden in the ContraIndicator class to return a distinct list? Rather than having the logic here

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
TalatAydinGDS: I think so, when trying to reach ContraIndicator through the method of reversing the return code, we cannot directly go from the CI.code. Because, If we want to access ContraIndicator from the return code, unfortunately, we should be working in the same business here like UserIndentityService.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -215,6 +225,80 @@ private void sendIdentityIssuedAuditEvent(
...
+                        })
+                .collect(
+                        Collectors.toMap(
+                                Map.Entry::getKey,
+                                Map.Entry::getValue,
+                                (existing, replacement) -> {
+                                    existing.addAll(replacement);
+                                    return existing;
+                                }))
```
MikeCollingwood: This this would be more readable if we defined a variable before it and then called this variable "merged...".

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/AuditEventReturnCode.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.di.ipv.core.library.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+import java.util.List;
+
+@ExcludeFromGeneratedCoverageReport
+public record AuditEventReturnCode(
+        @JsonProperty("code") String code, @JsonProperty("issuers") List<String> issuers) {}
```
MikeCollingwood, line 10: Do you need to specify the @JsonProperty tag if they're the same as the property name?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Joe-Edwards-GDS: I'm not sure I follow - the `filter` line in the code above should match against CIs based on their return code (not CI code)?

PR #1474, opened 2023-12-12
PYIC-4200: Serve well known file from S3

Wynndow:
### What changed

Serve well known file from S3

### Why did it change

This updates core's template to serve the well-known jwks endpoint from an S3 bucket instead of from the stubbed mock data.

~This is dependant on that file existing. The mechanism for that hasn't been decided yet and this should not be merged until it is, or bad things™️  will happen.~
The file now exists in all the buckets up to and including prod.

### Issue tracking

- [PYIC-4200](https://govukverify.atlassian.net/browse/PYIC-4200)

[PYIC-4200]: https://govukverify.atlassian.net/browse/PYIC-4200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openAPI/core-back-external.yaml
+++ b/openAPI/core-back-external.yaml
@@ -50,19 +50,30 @@ paths:
             application/json:
               schema:
                 type: "object"
+        500:
+          description: 500 response
+          content: {}
       x-amazon-apigateway-integration:
-        type: "MOCK"
-        requestTemplates:
-          application/json: "{\"statusCode\":200}"
+        type: aws
+        credentials:
+          Fn::GetAtt: ExternalApiGatewayJwksS3Role.Arn
+        httpMethod: GET
+        uri:
+          Fn::Sub:
+            - arn:aws:apigateway:eu-west-2:s3:path/ipv-core-well-known-jwks-${env}/well-known.json
+            - env:
+                Fn::If:
+                  - IsDevelopment
+                  - Fn::If:
+                      - IsDev01
+                      - dev01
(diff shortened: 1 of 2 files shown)


lwebb-iig (dismissed):

openAPI/core-back-external.yaml
```diff
@@ -50,19 +50,35 @@ paths:
...
+          default:
             statusCode: 200
-            responseTemplates:
-              application/json:
-                Fn::FindInMap:
-                  - EnvironmentConfiguration
-                  - Ref: AWS::AccountId
-                  - jwksJson
+          4\d{2}:
+            statusCode: 400
```
Joe-Edwards-GDS: Are you sure this is right? If we get (e.g.) a 4XX from S3 that's very likely a problem on our side (either bad permissions or a missing file), so should probably be a 500 to the client.

Wynndow: Good shout, I'll fix it up

Joe-Edwards-GDS (approved):
I think this seems reasonable, assuming you'll wait to merge until we've got the integration test as well.

Wynndow:
> I think this seems reasonable, assuming you'll wait to merge until we've got the integration test as well.

Yep - the new tests have just been approved so waiting for the new image to hit, then I'll merge this.

PR #1477, opened 2023-12-14
Pyic 3865 email sending

DanCorderIPV:
### What changed

For user requested VC resets we try to send a confirmation email

### Why did it change

User requested VC resets are a new feature and we want to send a confirmation email when they happen

### Issue tracking

- [PYIC-3865](https://govukverify.atlassian.net/browse/PYIC-3865)

New secret value for GovUK Notify API key has been added to all environments - will need updating in production when we have the live key.

New parameter added for GovUK Notify email template ID - will need updating in production if we go with a separate production template.

[PYIC-3865]: https://govukverify.atlassian.net/browse/PYIC-3865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
```diff
@@ -24,4 +25,22 @@ public List<Name> getName() {
     public List<BirthDate> getBirthDate() {
         return birthDate;
     }
+
+    // Concatenate the first name we have into a single string
+    @JsonIgnore
+    public String getFullName() {
+        StringBuilder nameBuilder = new StringBuilder();
+        name.get(0)
```
Joe-Edwards-GDS: Do you think this should be a method on the `Name` class instead?

It might also be worth comparing to the logic in `getFullNamesFromCredentials`, which is slightly different (filtering for `GivenName`s and `FamilyName`s first). See also https://govukverify.atlassian.net/browse/PYIC-4217 which might touch this...

lambdas/reset-identity/src/test/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandlerTest.java
```diff
@@ -83,18 +80,61 @@ void setUpEach() {
     }
 
     @Test
-    void shouldDeleteUsersVcsAndReturnNext() throws SqsException {
+    void handleRequest_whenFraudReset_shouldDeleteUsersVcsAndReturnNext() throws SqsException {
```
blakeyp: Not a big deal but it's not just fraud reset that calls this lambda so maybe just `whenNotUserInitiated` ?

lambdas/reset-identity/src/test/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandlerTest.java
```diff
@@ -83,18 +80,61 @@ void setUpEach() {
...
+        // Act
         JourneyResponse journeyResponse =
                 objectMapper.convertValue(
                         resetIdentityHandler.handleRequest(event, context), JourneyResponse.class);
 
+        // Assert
         verify(verifiableCredentialService).deleteVcStoreItems(TEST_USER_ID, true);
         verify(criResponseService).deleteCriResponseItem(TEST_USER_ID, F2F_CRI);
         verify(mockAuditService, times(1)).sendAuditEvent((AuditEvent) any());
+        verify(emailService, times(1)).sendUserTriggeredIdentityResetConfirmation(any(), any());
```
blakeyp: Would it be worth asserting the actual email and name here?

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -0,0 +1,97 @@
...
+    private void SendEmail(
+            String templateId, String toAddress, Map<String, Object> personalisation) {
+
+        var retries = 0;
+
+        while (true) {
+            try {
+                LOGGER.info("About to send email");
+                notificationClient.sendEmail(templateId, toAddress, personalisation, null, null);
+                LOGGER.info("Email sent");
```
blakeyp: I'd probably make these debug level (or remove)?

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -0,0 +1,97 @@
...
+import uk.gov.service.notify.NotificationClientException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EmailService {
+
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final int NUMBER_OF_RETRIES = 3;
+    private static final int RETRY_WAIT_MILLISECONDS = 3000;
```
blakeyp, line 17: Could we make these config params (in SSM) ?

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -0,0 +1,97 @@
...
+        // hardcoded
+        final String templateId =
+                configService.getSsmParameter(
+                        ConfigurationVariable
+                                .GOV_UK_NOTIFY_TEMPLATE_ID_USER_TRIGGERED_IDENTITY_RESET_CONFIRMATION);
+        LOGGER.info("Got template ID {}", templateId);
+        SendEmail(templateId, userEmailAddress, templateParameters);
+    }
+
+    private void SendEmail(
```
blakeyp: Is there a reason for using pascal case for this method name?

DanCorderIPV: Nope, just C# habits creeping in!

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
DanCorderIPV: Good point, these were inserted when trying to find out exactly what was going wrong.

lambdas/reset-identity/src/test/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandlerTest.java
DanCorderIPV: The email address is simple, but the name would require quite a lot of mocking out, so maybe not worth it.

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
DanCorderIPV: Happy to do so, but are they ever likely to change? We already have a lot of parameters in SSM and I was thinking that we could maybe avoid adding more if they weren't going to change.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/IdentityClaim.java
DanCorderIPV: Oh, that looks useful, if we can filter out middle names that would make the email nicer

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
blakeyp: Thinking we might want to be able to easily tweak them and test with different values in lower envs?

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -165,22 +164,10 @@ private NameAndDateOfBirth getProvenIdentityNameAndDateOfBirth(
                         500, ErrorResponse.FAILED_TO_GENERATE_IDENTIY_CLAIM);
             }
 
-            Name name = mapper.convertValue(identityClaim.get().getName().get(0), Name.class);
             BirthDate birthDate =
                     mapper.convertValue(identityClaim.get().getBirthDate().get(0), BirthDate.class);
```
MikeCollingwood, line 168: Would this suit being a BirthDate class method, if getFullName is going to be?

DanCorderIPV: I'm not sure that this call to `mapper` is actually doing anything at all here, so the only thing I think you could refactor would be to put a method on the IdentityClaim to do the `get(0)`.
I believe that this is going to be properly refactored in a later ticket so I'm planning to leave it alone.

blakeyp (approved):

PR #1479, opened 2023-12-15
PYIC-4184: Added passportCri pact test token and credential issue

RogVanGDS:
### What changed

Added pact test framework, as well as initial contract tests covering PassportCri - both token and credential issuer endpoint calls.

### Why did it change

So we can start building contract test coverage for points of integration between IPV Core and other services.

### Issue tracking

- [PYIC-4184](https://govukverify.atlassian.net/browse/PYIC-4184)

[PYIC-4184]: https://govukverify.atlassian.net/browse/PYIC-4184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -142,7 +142,7 @@ public class InitialiseIpvSessionHandler
                         HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_VTR);
             }
 
-            String clientOAuthSessionId = SecureTokenHelper.generate();
+            String clientOAuthSessionId = SecureTokenHelper.getInstance().generate();
 
             IpvSessionItem ipvSessionItem =
                     ipvSessionService.generateIpvSession(clientOAuthSessionId, null, emailAddress);
@@ -194,7 +194,7 @@ public class InitialiseIpvSessionHandler
             LogHelper.logErrorMessage(
                     "Recoverable Jar validation failed.", e.getErrorObject().getDescription());
 
-            String clientOAuthSessionId = SecureTokenHelper.generate();
+            String clientOAuthSessionId = SecureTokenHelper.getInstance().generate();
 
             IpvSessionItem ipvSessionItem =
                     ipvSessionService.generateIpvSession(
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -174,8 +174,6 @@ public class InitialiseIpvSessionHandler
                             auditEventUser,
                             reproveAuditExtension);
 
-            LOGGER.warn("Audit Event: " + auditEvent.getExtensions());
-
             auditService.sendAuditEvent(auditEvent);
 
             Map<String, String> response =
--- a/lambdas/process-cri-callback/build.gradle
+++ b/lambdas/process-cri-callback/build.gradle
@@ -26,8 +26,6 @@ dependencies {
 			project(":libs:user-identity-service"),
 			project(":libs:verifiable-credentials")
 
-	testImplementation 'junit:junit:4.13.1'
-
 	aspect "software.amazon.lambda:powertools-logging:$rootProject.ext.dependencyVersions.powertoolsLogging",
 			"software.amazon.lambda:powertools-tracing:$rootProject.ext.dependencyVersions.powertoolsTracing"
 	aspect('org.aspectj:aspectjrt') {
--- a/lambdas/process-cri-callback/build.gradle
+++ b/lambdas/process-cri-callback/build.gradle
@@ -42,6 +42,7 @@ dependencies {
 
 	testImplementation "org.junit.jupiter:junit-jupiter:5.10.0",
 			"org.mockito:mockito-junit-jupiter:5.8.0",
+			"au.com.dius.pact.consumer:junit5:4.6.3",
 			project(path: ':libs:common-services', configuration: 'tests')
 	testRuntimeOnly("org.junit.platform:junit-platform-launcher")
 }
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
@@ -102,12 +102,9 @@ public class ContractTest {
     @Pact(provider = "PassportCriProvider", consumer = "IpvCoreBack")
     public RequestResponsePact validRequestReturnsIssuedCredential(PactDslWithProvider builder)
             throws Exception {
-        return builder.given("dummyAuthCode is a valid authorization code")
-                .given("dummyApiKey is a valid api key")
-                .given("dummyPassportComponentId is the passport CRI component ID")
-                .given(
-                        "Passport CRI uses CORE_BACK_SIGNING_PRIVATE_KEY_JWK to validate core signatures")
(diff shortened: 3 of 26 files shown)


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -174,8 +174,6 @@ public APIGatewayProxyResponseEvent handleRequest(
                             auditEventUser,
                             reproveAuditExtension);
 
-            LOGGER.warn("Audit Event: " + auditEvent.getExtensions());
-
```
DanCorderIPV, line 178: Is this deliberate?

lambdas/process-cri-callback/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
 
 	testImplementation "org.junit.jupiter:junit-jupiter:5.10.0",
 			"org.mockito:mockito-junit-jupiter:5.8.0",
+			"au.com.dius.pact.consumer:junit5:4.6.3",
```
DanCorderIPV, line 45: Is this going to clash with the junit definition on line 29?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
RogVanGDS: I accidentally left this in on the previous PR that got merged in, was just using it to debug and feels unnecessary given we can access the audit events from the auditing lambda anyway

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -0,0 +1,260 @@
...
+                        "Passport CRI uses CORE_BACK_SIGNING_PRIVATE_KEY_JWK to validate core signatures")
+                .uponReceiving("Valid auth code")
+                .path("/credential")
+                .method("POST")
+                .headers("x-api-key", PRIVATE_API_KEY, "Authorization", "Bearer dummyAccessToken")
+                .willRespondWith()
+                .status(200)
+                .body(
+                        generateVerifiableCredential(
+                                vcClaim(CREDENTIAL_ATTRIBUTES_2), TEST_USER, TEST_ISSUER),
```
DanCorderIPV: I'm a little concerned about using something called `CREDENTIAL_ATTRIBUTES_2` there's no guarantee that this will remain an expected set of attributes for the passport CRI.
For these tests we're probably going to need a number of different attribute sets to match the various valid VCs that the passport CRI might return.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -0,0 +1,260 @@
...
+                                        })
+                                .build())
+                .toPact();
+    }
+
+    @Pact(provider = "PassportCriProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact validRequestReturnsIssuedCredential(PactDslWithProvider builder)
+            throws Exception {
+        return builder.given("dummyAuthCode is a valid authorization code")
+                .given("dummyApiKey is a valid api key")
```
DanCorderIPV: You'll need to change these `given()` methods to match this test.
e.g. I believe you should be sending an access token not an API key with this request.
These `given()` methods are used by the provider to set up data in their test so you need to specify anything they will need to know to be able to generate the response you're expecting.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
RogVanGDS: I'm not entirely sure how this is going to work in the consumer / provider relationship we have with contract testing.

I thought that the only thing that was coordinated between the consumer and the provider was the overall structure of the response/request objects? 

But in this instance the stipulation would be that the response object contains a JWT, not that the JWT contains specific claims?

We will only know what attributes are present in the claims by trying to validate it, which we can only test against the mock response that we provide, not the actual response that we are sent from the CRI?

RogVanGDS: We already have a unit test suite covering the VC validator logic, so I guess I'm not sure how this is any different?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -0,0 +1,260 @@
...
+                .getVerifiableCredentials()
+                .forEach(
+                        credential -> {
+                            try {
+                                verifiableCredentialJwtValidator.validate(
+                                        credential, credentialIssuerConfig, TEST_USER);
+                            } catch (VerifiableCredentialException e) {
+                                throw new RuntimeException(e);
+                            }
+                        });
```
DanCorderIPV, line 275: This is a bit of a fuzzy line, but I think I'd expect these assertions to at least cover things like checking that this has the claims we'd expect from the passport CRI.
e.g. assuming that `CREDENTIAL_ATTRIBUTES_2` is realistic then we should be checking that there are name, DoB, and address claims present with some values in them.

DanCorderIPV:
I've left comments but I can't really do the full review as some of this code is mine

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
blakeyp: We don't ever want to log audit events/extensions in prod given they may contain PII - I don't think it's an issue in this case as it will only have been logging null in prod here but one to be careful of

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
RogVanGDS: @blakeyp Do you think it is worth having a higher level of validation in the test than in the actual codebase? 

If we want to validate to a higher level in the test, then it feels like we should spin out another ticket to increase the validation in the code itself. (I think Joe was also looking at making a ticket for this)

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -0,0 +1,294 @@
...
+                                        })
+                                .build())
+                .toPact();
+    }
+
+    @Pact(provider = "PassportCriProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact validRequestReturnsIssuedCredential(PactDslWithProvider builder)
+            throws Exception {
+        return builder.given("dummyApiKey is a valid api key")
+                .given("dummyAccessToken is a valid access token")
```
RogVanGDS, line 112: Do we need any additional pact conditions to these? Like fixing the clock at the tight time? Or the dummy signing and secure tokens we are using?

DanCorderIPV: Only if they are used in the request or the response. e.g. if the response is signed we'll need to specify a signing key.

RogVanGDS: Well it's only used in the test implementation below, so I guess it's ok then?

blakeyp (approved):

PR #1481, opened 2023-12-18
PYIC-4189: handled to the new property 'issuers' and changed response IPV_IDENTITY_ISSUED event for fail return code

TalatAydinGDS:
### What changed

D&A need to have access to metrics around Failure Codes, however, it is important that they are not only able to see how many failure codes have been raised, but they also need to be able to see what CRI’s resulted in those Failure codes.

This story builds on the existing feature to add information about which CRI’s produced failure codes to the event sent to TxMA

### Why did it change

To facilitate future analysis around the performance of the CRI’s and more specifically which ones are resulting in certain failure codes

### Issue tracking

- [PYIC-4189](https://govukverify.atlassian.net/browse/PYIC-4189)

### Other considerations

[PYIC-4189]: https://govukverify.atlassian.net/browse/PYIC-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -200,10 +202,432 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+        assertEquals(extensions.getReturnCodes().size(), 3);
+        assertTrue(extensions.getReturnCodes().stream().anyMatch(rt -> rt.code().equals("2")));
+        assertTrue(
+                extensions.getReturnCodes().stream()
+                        .anyMatch(
+                                rt ->
+                                        rt.issuers()
+                                                .toString()
+                                                .equals(
+                                                        "[https://review-z.account.gov.uk, https://review-f.account.gov.uk]")));
```
Wynndow: I think this could be tightend up a bit - could you assert that the issuers for the three different return codes are correct, rather than just asserting that one of the return codes has those issuers?

I think there's also a mismatch in the return codes returned in the mocked userIdentity and those defined in the mocked CO config map which makes the test a little confusing, especially if you look at all the audit event return codes.

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -200,10 +202,432 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+        AuditEvent capturedAuditEvent = auditEventCaptor.getValue();
+        assertEquals(AuditEventTypes.IPV_IDENTITY_ISSUED, capturedAuditEvent.getEventName());
+        AuditExtensionsUserIdentity extensions =
+                (AuditExtensionsUserIdentity) capturedAuditEvent.getExtensions();
+        assertEquals(VectorOfTrust.P0.toString(), extensions.getLevelOfConfidence());
+        assertFalse(extensions.isCiFail());
+        assertFalse(extensions.isHasMitigations());
+        assertEquals(extensions.getReturnCodes().size(), 3);
+        assertTrue(extensions.getReturnCodes().stream().anyMatch(rt -> rt.code().equals("1")));
+        assertTrue(extensions.getReturnCodes().stream().anyMatch(rt -> rt.issuers().isEmpty()));
```
Wynndow: Same here - it would be good to be explicit about all the return codes and which have which issuers.

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -200,10 +202,432 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+                                                        .build(),
+                                        "Z03",
+                                                ContraIndicator.builder()
+                                                        .code("Z03")
+                                                        .issuers(
+                                                                List.of(
+                                                                        "https://review-z.account.gov.uk",
+                                                                        "https://review-f.account.gov.uk"))
+                                                        .build()))
+                        .build();
```
Wynndow: This, or somethinve very similar is getting repeated a lot. Probably worth defining it once and then referencing it in the tests.

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -200,10 +202,432 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+        AuditEvent capturedAuditEvent = auditEventCaptor.getValue();
+        assertEquals(AuditEventTypes.IPV_IDENTITY_ISSUED, capturedAuditEvent.getEventName());
+        AuditExtensionsUserIdentity extensions =
+                (AuditExtensionsUserIdentity) capturedAuditEvent.getExtensions();
+        assertEquals(VectorOfTrust.P0.toString(), extensions.getLevelOfConfidence());
+        assertTrue(extensions.isCiFail());
+        assertFalse(extensions.isHasMitigations());
+        assertEquals(extensions.getReturnCodes().size(), 3);
+        assertTrue(extensions.getReturnCodes().stream().anyMatch(rt -> rt.code().equals("3")));
+        assertFalse(extensions.getReturnCodes().stream().allMatch(rt -> rt.issuers().isEmpty()));
```
Wynndow: Same again here.

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -200,10 +202,432 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+        AuditEvent capturedAuditEvent = auditEventCaptor.getValue();
+        assertEquals(AuditEventTypes.IPV_IDENTITY_ISSUED, capturedAuditEvent.getEventName());
+        AuditExtensionsUserIdentity extensions =
+                (AuditExtensionsUserIdentity) capturedAuditEvent.getExtensions();
+        assertEquals(VectorOfTrust.P2.toString(), extensions.getLevelOfConfidence());
+        assertTrue(extensions.isCiFail());
+        assertTrue(extensions.isHasMitigations());
+        assertEquals(extensions.getReturnCodes().size(), 3);
+        assertTrue(extensions.getReturnCodes().stream().anyMatch(rt -> rt.code().equals("1")));
+        assertTrue(extensions.getReturnCodes().stream().allMatch(rt -> rt.issuers().isEmpty()));
```
Wynndow: And here

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
```diff
@@ -2,15 +2,26 @@
...
-        @JsonProperty boolean ciFail,
-        @JsonProperty boolean hasMitigations,
-        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<ReturnCode> returnCode)
-        implements AuditExtensions {}
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class AuditExtensionsUserIdentity implements AuditExtensions {
+    @JsonProperty private String levelOfConfidence;
```
Wynndow: Does this need to be private here?

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
```diff
@@ -2,15 +2,26 @@
...
+    @JsonProperty boolean ciFail;
+    @JsonProperty boolean hasMitigations;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonProperty("returnCodes")
```
Wynndow: I think you can drop this line as it matches the field name - also consistent with the other fields.

libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/service/AuditServiceTest.java
```diff
@@ -161,15 +169,52 @@ void shouldSendMessageToSqsQueueWithAuditExtensionsUserIdentityWithoutExitCode()
                 AuditEventTypes.IPV_JOURNEY_START.toString(),
                 messageBody.get("event_name").asText());
         JsonNode auditExtensionsUserIdentity = messageBody.get("extensions");
-        assertNull(auditExtensionsUserIdentity.get("returnCode"));
+        assertNull(auditExtensionsUserIdentity.get(RETURN_CODE_KEY));
     }
 
     @Test
     void shouldSendMessageToSqsQueueWithAuditExtensionsUserIdentityWithExitCode()
```
Wynndow: I think this test is the same as the one below, except the data isn't quite right. The expeced value is `SUCCESS_RETURN_CODES_TEST`, but it doesn't really represent that. If a user had a succesful journey they would either have no return codes at all, or one of the always required return codes which would have an issuer.

I'd be inclided to drop this test and stick with just the next one.

libs/cimit-service/src/test/java/uk/gov/di/ipv/core/library/service/CiMitServiceTest.java
```diff
@@ -219,7 +219,7 @@ void getContraIndicatorsVC() throws Exception {
...
+                "ContraIndicators(contraIndicatorsMap={D01=ContraIndicator(code=D01, issuers=null, issuanceDate=2022-09-20T15:54:50.000Z, document=passport/GBR/824159121, txn=[abcdef], mitigation=[Mitigation(code=M01, mitigatingCredential=[MitigatingCredential(issuer=https://credential-issuer.example/, validFrom=2022-09-21T15:54:50.000Z, txn=ghij, id=urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6)])], incompleteMitigation=[Mitigation(code=M02, mitigatingCredential=[MitigatingCredential(issuer=https://another-credential-issuer.example/, validFrom=2022-09-22T15:54:50.000Z, txn=cdeef, id=urn:uuid:f5c9ff40-1dcd-4a8b-bf92-9456047c132f)])])})",
```
Wynndow: We should probably update the `SIGNED_CONTRA_INDICATOR_VC` with issuers, so this doesn't come back as null

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -517,11 +302,19 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
...
+        assertEquals(extensions.getReturnCodes().get(1).code(), "2");
+        assertEquals(
+                extensions.getReturnCodes().get(1).issuers(),
+                List.of("https://review-q.account.gov.uk", "https://review-f.account.gov.uk"));
+        assertEquals(extensions.getReturnCodes().get(2).code(), "3");
+        assertEquals(
+                extensions.getReturnCodes().get(2).issuers(),
+                List.of("https://review-z.account.gov.uk", "https://review-f.account.gov.uk"));
         verify(mockClientOAuthSessionDetailsService, times(1)).getClientOAuthSession(any());
         verify(mockCiMitService, times(1)).getContraIndicatorsVCJwt(any(), any(), any());
```
Wynndow, line 320: I'm not entirely sure the test below (line 325) is needed. The test above checks the code path where the return code isn't found in the CI config map already. The test below is just an extreme example of that, one that wouldn't actually happen?

Wynndow (commented):
Just a comment on maybe not needing one of the test, otherwise I think this is good. I'm going to deploy it to my dev env to check it out.

We'll also need to think about the acceptance tests - this change will break them. Should be a pretty straightforward fix though. It'll need to be ready before this is merged though or we'll block the pipeline.

Wynndow:
> Just a comment on maybe not needing one of the test, otherwise I think this is good. I'm going to deploy it to my dev env to check it out.
> 
> We'll also need to think about the acceptance tests - this change will break them. Should be a pretty straightforward fix though. It'll need to be ready before this is merged though or we'll block the pipeline.

Have deployed to my dev env, but we need the stub to be able to test this before merging.

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
TalatAydinGDS: Ok, I deleted this test.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -195,12 +199,22 @@ private void sendIdentityIssuedAuditEvent(
             ContraIndicators contraIndicators,
             UserIdentity userIdentity)
             throws SqsException {
+
+        Map<String, ContraIndicatorConfig> configMap = configService.getContraIndicatorConfigMap();
+        var auditEventReturnCodes =
+                userIdentity.getReturnCode().stream()
```
MikeCollingwood, line 205: Is the property called returnCode on UserIdentity, instead of returnCodes, because it'd break some agreed schema?

Wynndow: In the user identity bundle that's returned to core, it looks (something, working from memory) like this:
```
{
...
"https://vocab.account.gov.uk/v1/returnCode": [{"code": "X"}, {"code": "Y"}]
...
}
```

So the return code, can contain mutiple, a single, or no actual codes.

So to answer your question, sort of yes. The return code is all the codes, but we override the property name to something more complicated anyway.

MikeCollingwood: Thank you for the run-through!

Wynndow (approved):
Have deployed this to my dev account and run some journeys. The audit events look good to me.

PR #1482, opened 2023-12-18
INCIDENT-599: Stop parsing reprove identity claim

Wynndow:
### What changed

Stop parsing reprove identity claim

### Why did it change

A change has been made in Orch that has meant that the value coming through is no longer a string - possibly a boolean.

This is causing the `getStringClaim()` method to throw. This has caused a P1 outage.

The change on orch side is being reverted - this change is a backup just in case to return functionality.

### Issue tracking

- [INCIDENT-599](https://govukverify.atlassian.net/browse/INCIDENT-599)


Wynndow:
Closing due to the Orch reversion working

PR #1483, opened 2023-12-18
PYIC-3897 Updated journey map to call TICF cri for reuse and other journeys, all behind feature flag.

rpayal:
## Proposed changes
Add TICF CRI to all existing Journeys

### What changed
Updated journey map to call TICF cri for reuse and other journeys, all behind feature flag.

### Issue tracking

- [PYIC-3897](https://govukverify.atlassian.net/browse/PYIC-3897)

### Other considerations

[PYIC-3897]: https://govukverify.atlassian.net/browse/PYIC-3897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -81,6 +81,8 @@ CHECK_EXISTING_IDENTITY:
       checkFeatureFlag:
         ticfCriBeta:
           targetState: CALL_TICF_CRI_ON_REUSE
+        deleteDetailsEnabled:
+          targetState: IPV_IDENTITY_REUSE_PAGE_TEST
     pending:
       targetState: IPV_IDENTITY_PENDING_PAGE
     fail-with-ci:
@@ -391,9 +393,6 @@ PYI_NEW_DETAILS_PAGE:
         PYI_CONFIRM_DELETE_DETAILS_PAGE
     end:
       targetState: IPV_IDENTITY_REUSE_PAGE
-      checkFeatureFlag:
-        ticfCriBeta:
-          targetState: CALL_TICF_CRI_ON_REUSE
 
 PYI_CONFIRM_DELETE_DETAILS_PAGE:
   response:
@@ -405,9 +404,6 @@ PYI_CONFIRM_DELETE_DETAILS_PAGE:
         RESET_IDENTITY_USER_INITIATED
     end:
       targetState: IPV_IDENTITY_REUSE_PAGE
-      checkFeatureFlag:
-        ticfCriBeta:
-          targetState: CALL_TICF_CRI_ON_REUSE
 
 RESET_IDENTITY_USER_INITIATED:
   response:
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -79,8 +79,8 @@ CHECK_EXISTING_IDENTITY:
     reuse:
       targetState: IPV_IDENTITY_REUSE_PAGE
       checkFeatureFlag:
-        deleteDetailsEnabled:
-          targetState: IPV_IDENTITY_REUSE_PAGE_TEST
+        ticfCriBeta:
+          targetState: CRI_TICF_ON_REUSE
     pending:
       targetState: IPV_IDENTITY_PENDING_PAGE
     fail-with-ci:
@@ -286,10 +286,19 @@ EVALUATE_GPG45_SCORES:
   events:
     met:
       targetState: IPV_SUCCESS_PAGE
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF
     unmet:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF
     vcs-not-correlated:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF
 
 IPV_SUCCESS_PAGE:
   response:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -302,6 +311,35 @@ END:
     type: process
     lambda: build-client-oauth-response
 
+CRI_TICF:
+  response:
+    type: process
+    lambda: call-ticf-cri
+    lambdaInput:
+      journeyType: main
+  parent: CRI_STATE
```
rpayal: Not sure of this parent ?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -302,6 +311,35 @@ END:
...
+    fail-with-ci:
+      targetState: PYI_NO_MATCH
+
+CRI_TICF_ON_REUSE:
+  response:
+    type: process
+    lambda: call-ticf-cri
+    lambdaInput:
+      journeyType: reuse
+  parent: CRI_STATE
```
rpayal: Not sure of this parent ?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: Does the lambda only respond with `next` or `fail-with-ci`? 

If it's not getting events back from the CRI itself, it probably wont have the same event mappings as other cri states, which have defaults defined in the CRI_STATE parent:
```
CRI_STATE:
  events:
    next:
      targetState: UNEXPECTED_EVENT
    not-found:
      targetState: PYI_NO_MATCH
    ...
    vcs-not-correlated:
      targetState: PYI_NO_MATCH
```

I also think, in either case the name of the state might be misleading as it's of type process, rather than CRI, so maybe a better state name would be like "CALL_TICF_CRI" - a similar style to CHECK_EXISTING_IDENTITY which is also a process

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -79,8 +79,8 @@ CHECK_EXISTING_IDENTITY:
     reuse:
       targetState: IPV_IDENTITY_REUSE_PAGE
       checkFeatureFlag:
-        deleteDetailsEnabled:
-          targetState: IPV_IDENTITY_REUSE_PAGE_TEST
+        ticfCriBeta:
```
MikeCollingwood, line 82: I'm unsure if it's the convention, but I think a more informative flag names could explicitly specify that it enables the ticf cri to be called, which might be a bit obscured behind the term "beta" otherwise

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -302,6 +311,35 @@ END:
     type: process
     lambda: build-client-oauth-response
 
+CRI_TICF:
+  response:
+    type: process
+    lambda: call-ticf-cri
+    lambdaInput:
+      journeyType: main
```
Wynndow, line 321: I'd be tempted to call this something like `identity-proving` to be explicit, although it technically doesn't matter

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: It can also respond with `/journey/error`, which we should handle. But I don't think it needs a parent.
And I agree with Mike about the naming thing.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: Doesn't need one I don't think, but it should also handle an `error` event

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -350,8 +388,10 @@ PYI_NEW_DETAILS_PAGE:
       targetState:
         PYI_CONFIRM_DELETE_DETAILS_PAGE
     end:
-      targetState:
-        IPV_IDENTITY_REUSE_PAGE
+      targetState: IPV_IDENTITY_REUSE_PAGE
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF_ON_REUSE
```
Wynndow: I thnk if we get to this state, we will have already called the TICF CRI on a reuse journey. Adding this here would result in calling it twice. Maybe that's ok though, and would prevent us accidentally missing it iof the config changes in future.

rpayal: Calling ticf twice is ok alrady been confirm by David (and informed TICF).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Calling ticf twice is ok alrady been confirm by David (and been informed to ticf team).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: done

Wynndow (approved):

PR #1484, opened 2023-12-18
PYIC-3897 Updated journey map to call TICF cri for all error journeys, all behind feature flag.

rpayal:
## Proposed changes
Call TICF before all error (PYI_NO_MATCH and PYI_ANOTHER_WAY)

### What changed
Updated journey map to call TICF cri for all error journeys, all behind feature flag.

### Issue tracking

- [PYIC-3897](https://govukverify.atlassian.net/browse/PYIC-3897)

### Other considerations

[PYIC-3897]: https://govukverify.atlassian.net/browse/PYIC-3897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
@@ -10,20 +10,38 @@ CRI_STATE:
       targetState: UNEXPECTED_EVENT
     not-found:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF_ON_NO_MATCH_ERR
     fail-with-no-ci:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -302,6 +338,59 @@ END:
     type: process
     lambda: build-client-oauth-response
 
+CRI_TICF:
+  response:
+    type: process
+    lambda: call-ticf-cri
+    lambdaInput:
+      journeyType: main
```
MikeCollingwood: (Just a comment) Outside the scope of the ticket, but it could be nice to use this input to reduce the number of separately-defined states. We could have more events, and the journeyType being main/ reuse/ no-match/ another-way can define the event outputted.

That being said, it would hide some information from the journey map which we can now visualise well

PR #1489, opened 2023-12-19
PYIC-4252: use boolean instead of string to fetch reprove identity flag

shivanshuit914:
## Proposed changes

- Update reprove identity flag to boolean from String

### What changed

- Updated reprove identity flag to boolean from String

### Why did it change

- To match with auth, fraud service claim.

### Issue tracking

- [PYIC-4252](https://govukverify.atlassian.net/browse/PYIC-4252)

[PYIC-4252]: https://govukverify.atlassian.net/browse/PYIC-4252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsService.java
@@ -51,8 +51,7 @@ public class ClientOAuthSessionDetailsService {
         clientOAuthSessionItem.setGovukSigninJourneyId(
                 claimsSet.getStringClaim("govuk_signin_journey_id"));
         clientOAuthSessionItem.setVtr(claimsSet.getStringListClaim("vtr"));
-        Boolean reproveIdentity =
-                Boolean.parseBoolean(claimsSet.getStringClaim("reprove_identity"));
+        Boolean reproveIdentity = claimsSet.getBooleanClaim("reprove_identity");
         clientOAuthSessionItem.setReproveIdentity(reproveIdentity);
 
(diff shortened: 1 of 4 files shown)


RogVanGDS (dismissed):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsService.java
```diff
@@ -51,8 +51,7 @@ public ClientOAuthSessionItem generateClientSessionDetails(
         clientOAuthSessionItem.setGovukSigninJourneyId(
                 claimsSet.getStringClaim("govuk_signin_journey_id"));
         clientOAuthSessionItem.setVtr(claimsSet.getStringListClaim("vtr"));
-        Boolean reproveIdentity =
-                Boolean.parseBoolean(claimsSet.getStringClaim("reprove_identity"));
+        Boolean reproveIdentity = claimsSet.getBooleanClaim("reprove_identity");
```
blakeyp, line 54: What is the behaviour of this if the value is null/missing? do we have some unit test cases

shivanshuit914: It returns null in case of null or value not present. No we don't have unit tests. Let. me add it thanks

shivanshuit914: Added unit tests. Thanks

Wynndow (approved):

PR #1494, opened 2023-12-20
PYIC-4273 Updated journey map to call TICF cri for all error journeys, all behind feature flag

rpayal:
## Proposed changes
Call TICF before all error (PYI_NO_MATCH, PYI_ANOTHER_WAY, PYI_TECHNICAL)

### What changed
Updated journey map to call TICF cri for all error journeys, all behind feature flag.

### Issue tracking

- [PYIC-4273](https://govukverify.atlassian.net/browse/PYIC-4273)

### Other considerations

[PYIC-4273]: https://govukverify.atlassian.net/browse/PYIC-4273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -8,22 +8,49 @@ CRI_STATE:
   events:
     next:
       targetState: UNEXPECTED_EVENT
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_TECHNICAL_ERR
```
Wynndow: This now ommits the UNEXPECTED_EVENT state completely (when the ticfCriBeta flag is on). We'd go CALL_TICF_CRI_ON_TECHNICAL_ERR -> ERROR.

The UNEXPECTED_EVENT state was added recently as a more explicit version of ERROR, although [the PR comments](https://github.com/govuk-one-login/ipv-core-back/pull/1445/files) from when it was added suggest it may not be neccessary at all.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -87,10 +114,16 @@ CHECK_EXISTING_IDENTITY:
       targetState: IPV_IDENTITY_PENDING_PAGE
     fail-with-ci:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_NO_MATCH_ERR
```
Wynndow: In this scenario, I think that this would be treated as a reuse journey. The user hasn't picked up any new VCs, so we don't want to send anything new to TICF CRI. The CRI_TICF_ON_NO_MATCH_ERR state doesn't set the journey type to `reuse`, so the lambda will send all the users current VC to TICF which is wrong.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -87,10 +114,16 @@ CHECK_EXISTING_IDENTITY:
...
     f2f-fail:
       targetState: PYI_F2F_TECHNICAL
```
Wynndow, line 119: In this scenario, the user might have a new VC from F2F. Which won't get sent to the TICF CRI. I've checked with Simon Parsons and the current thinking seems to be that F2F VC don't go to TICF. So this is maybe ok. I think it's worth double checking that though.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -87,10 +114,16 @@ CHECK_EXISTING_IDENTITY:
...
     error:
       targetState: ERROR
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_TECHNICAL_ERR
```
Wynndow: This needs to call TICF as a `reuse` journey so that no VCs are sent in the request. If the user experiences an error at this point they won't have collected any new VCs. `CALL_TICF_CRI_ON_TECHNICAL_ERR` will call TICF with all the users current VCs

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -135,6 +168,9 @@ IPV_IDENTITY_START_PAGE:
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: This call to TICF needs to be `reuse` as the user hasn't collected any new VCs at this point.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -146,6 +182,9 @@ CRI_F2F:
       targetState: F2F_HANDOFF_PAGE
```
Wynndow, line 193: I have a question about what we do in this scenario. In this journey the user will have collected a bunch of VCs, but won't yet have a f2f VC. Should we be calling the TICF CRI with them here?
We've said that maybe F2F VCs shouldn't go to TICF, so when the user comes back to us after the F2F VC has been issued, we'd need to send all these VCs at that point (but not the F2F?).
Feels like we should send everything gathered so far at this point, and then treat the return as a reuse journey which would mean not sending the new F2F VC.
An alternative is to not send them now, wait for the return journey (after F2F has been issued), and then send all VCs, including the F2F to TICF. But that's a new behaviour in how we send stuff to TICF (sending VCs not gathered in the current session).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -146,6 +182,9 @@ CRI_F2F:
...
     access-denied:
       targetState: PYI_ANOTHER_WAY
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: This isn't really an error I don't think, it just means the user can't use F2F, maybe because of a lack of appropriate docs. The state name might need a rethink

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -344,6 +383,48 @@ CALL_TICF_CRI_ON_REUSE:
     error:
       targetState: ERROR
 
+CALL_TICF_CRI_ON_NO_MATCH_ERR:
```
Wynndow: This isn't really an error scenario - I'd maybe rename the state

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -344,6 +383,48 @@ CALL_TICF_CRI_ON_REUSE:
...
+  response:
+    type: process
+    lambda: call-ticf-cri
+    lambdaInput:
+      journeyType: error
```
Wynndow: I'd maybe change this journeyType. While it makes no difference technically what it is, other than reuse, having a bunch of values would be confusing.

Thinking about it, I think this is showing that the mechanism for passing the journey type into the lambda isn't fit for purpose, and that the lambda should be able to determine which VCs to send without any help.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -344,6 +383,48 @@ CALL_TICF_CRI_ON_REUSE:
...
+  events:
+    next:
+      targetState: PYI_NO_MATCH
+    fail-with-ci:
+      targetState: PYI_NO_MATCH
+    error:
+      targetState: ERROR
+
+CALL_TICF_CRI_ON_ANOTHER_WAY_ERR:
```
Wynndow: Again, not really an error. Maybe needs a new name.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -650,6 +740,9 @@ MITIGATION_01_IDENTITY_START_PAGE:
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: The MITIGATION_01 journey is the separate session mitigation route, meaning the user will be on a new session. That means that at this point the user hasn't collected and new VCs in this session. So we'll need to call TICF as a reuse journey.

Having said that, we will have just reset their identity so they won't actually have any to submit anyway...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -661,26 +754,41 @@ MITIGATION_01_CRI_DCMAW:
       targetState: POST_DCMAW_SUCCESS_PAGE
     not-found:
       targetState: PYI_NO_MATCH
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_NO_MATCH_ERR
```
Wynndow: As above, if we get here then they won't have any new VCs this session, so we should probably be calling TICF as a reuse journey. But they won't have any VCs anyway....

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -661,26 +754,41 @@ MITIGATION_01_CRI_DCMAW:
...
     access-denied:
       targetState: MITIGATION_01_PYI_POST_OFFICE
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: As above

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -661,26 +754,41 @@ MITIGATION_01_CRI_DCMAW:
...
     temporarily-unavailable:
       targetState: MITIGATION_01_PYI_POST_OFFICE
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: As above

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -661,26 +754,41 @@ MITIGATION_01_CRI_DCMAW:
...
     fail-with-no-ci:
       targetState: MITIGATION_01_PYI_POST_OFFICE
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: Here however they will have collected a VC so it does need to be submitted, which will happen from the TICF state defined.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -661,26 +754,41 @@ MITIGATION_01_CRI_DCMAW:
...
     enhanced-verification:
       targetState: MITIGATION_01_PYI_POST_OFFICE
       checkIfDisabled:
         f2f:
           targetState: PYI_ANOTHER_WAY
+          checkFeatureFlag:
+            ticfCriBeta:
+              targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: As above, they'll have a VC that needs submitting.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -691,6 +799,9 @@ MITIGATION_01_F2F_START_PAGE:
       targetState: CRI_CLAIMED_IDENTITY_J4
     end:
       targetState: PYI_ANOTHER_WAY
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_ON_ANOTHER_WAY_ERR
```
Wynndow: Another that should be a reuse type.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -701,6 +812,9 @@ MITIGATION_01_PYI_POST_OFFICE:
...
```
Wynndow: Another reuse case

Wynndow (commented):
There's a lot of comments here and a lot of those comment are actually unanswered questions, so sorry about that. But it's a place to have a conversation about it. Below are some more things I noticed but GitHub wouldn't let me comment on them, so adding here:

- The `PYI_ESCAPE` state allows a user to end their journey. It'll need a call to TICF CRI including.
- The `PYI_CRI_ESCAPE` and `PYI_CRI_ESCAPE_NO_F2F` states both allow a user to end. These states come after a thin file response from the KBV CRIs. Will need to call TICF before ending.
- The `CORE_SESSION_TIMEOUT` state is a bit of a unique one. We don't transition to it from any other state, instead we set it as the users state in the backend directly if their session timesout. We'll need to call TICF though. And it'll be hard to know if it's a reuse journey or not.
- There might be an issue with the flow around the `EVALUATE_GPG45_SCORES` logic. If a user hasn't matched a profile but has no CIs, and no CIs are issued by the TICF CRI we'll go `EVALUATE_GPG45_SCORES` -> `unmet` -> `CALL_TICF_CRI` -> `next` -> `IPV_SUCCESS`. Which isn't right. We won't actually issue an identity, but the user shouldn't see that success page.

Sorry about all the comments, but I think it's important to thrash out.
I'm definitely thinking that the call-ticf-cri lambda needs a chagne so that we don't have to pass in the journey type in the map...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Yeah noticed that done this as both these states lead to pyi-technical page. Will check again.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: Good spot - I hadn't considered this scenario. Your draft PR should simplify this, as promised :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: As discussed, this is fine - we don't need to worry about sending F2F VCs to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: (ditto)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: (ditto)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: The description in confluence is to send everything to TICF _before_ routing the user to F2F: https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3763505535/IPV+TICF+CRI+Interaction#Exceptions-for-Face-to-Face-(F2F)

We shouldn't try to fail them _after_ going to the F2F CRI, as they will already have set up an appointment at the post office.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: As above, these cases probably don't need to go to TICF (but we should have already been to TICF before reaching F2F).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: ditto

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: This'll be taken care by 4331 when we'll only send VC's collected for the session to TICF.

rpayal:
Need to discuss following states;
- for state PYI_ESCAPE don't think we need to call TICF 
- from state PYI_CRI_ESCAPE when f2f event is raised then TICF is called before calling F2F_CRI
- state CORE_SESSION_TIMEOUT

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -264,6 +298,9 @@ PYI_CRI_ESCAPE:
       targetState: CRI_DCMAW_PYI_ESCAPE
     f2f:
       targetState: CRI_F2F
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CALL_TICF_CRI_BEFORE_F2F
     end:
       targetState: END
```
Joe-Edwards-GDS: Not really part of your ticket, but can `PYI_CRI_ESCAPE` actually result in an 'end' event? Looks like the page only has options for `dcmaw` and `f2f`.

(Can we just delete it :) )

Joe-Edwards-GDS: (I think it can still occur in the `PYI_CRI_ESCAPE_NO_F2F` route though, which we can just leave as-is)

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/utils/ProcessJourneyStepPages.java
```diff
@@ -1,10 +1,5 @@
 package uk.gov.di.ipv.core.processjourneyevent.utils;
 
 public class ProcessJourneyStepPages {
-    public static final String IPV_IDENTITY_REUSE_PAGE = "page-ipv-reuse";
-    public static final String IPV_IDENTITY_START_PAGE = "page-ipv-identity-document-start";
-    public static final String PRE_KBV_TRANSITION_PAGE = "page-pre-kbv-transition";
-    public static final String PYI_NO_MATCH_PAGE = "pyi-no-match";
-    public static final String PYI_TECHNICAL_ERROR_PAGE = "pyi-technical";
     public static final String PYI_UNRECOVERABLE_TIMEOUT_ERROR_PAGE = "pyi-timeout-unrecoverable";
```
Joe-Edwards-GDS: This last one is only used in a test - I wonder if it should just be a test variable there instead?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS, line 1: I think this looks good - I'm just wondering if we want to make all the TICF states follow a consistent pattern (including the success/reuse ones).

What do you think of something like:
- `CRI_TICF_BEFORE_SUCCESS`
- `CRI_TICF_BEFORE_REUSE`
- `CRI_TICF_BEFORE_F2F`
- `CRI_TICF_BEFORE_NO_MATCH`
- `CRI_TICF_BEFORE_ANOTHER_WAY`
- `CRI_TICF_BEFORE_ERROR`

with `CRI_TICF` matching the other CRIs?

(cc @Wynndow you called out naming below as well)

Joe-Edwards-GDS (commented):
I think it would be good to update the list of 'failure' and 'error' states in the journey map so we don't see the TICF versions.

This should be pretty easy - basically updating `errorStates` and `failureStates` at the top of `render.mjs` to include your new TICF variants.

Joe-Edwards-GDS (dismissed):
Thanks for this - I know it's a bit of a fiddly one!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Done - removed event 'end' for state PYI_CRI_ESCAPE

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Done

rpayal:
> I think it would be good to update the list of 'failure' and 'error' states in the journey map so we don't see the TICF versions.
> 
> This should be pretty easy - basically updating `errorStates` and `failureStates` at the top of `render.mjs` to include your new TICF variants.

Done.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -396,6 +396,18 @@ CRI_TICF_BEFORE_F2F:
...
 
+CRI_TICF_BEFORE_F2F_MITIGATION:
+  response:
+    type: process
+    lambda: call-ticf-cri
+  events:
+    next:
+      targetState: CRI_F2F
+    fail-with-ci:
+      targetState: CRI_F2F
```
Joe-Edwards-GDS: Ah, I hadn't thought about this case! I think this isn't quite right though - it's probably _ok_, but we should really handle `enhanced-verification` properly.

This probably means:
- Update the `CallTicfCriHandler` code to return `ciMitUtilityService.getCiMitigationJourneyStep` (as in all the other places we call `isBreachingCiThreshold`)
- Update the TICF journey steps to route `enhanced-verification`... Usually to the same as `fail-with-ci`, but to F2F in the F2F case

(Then there is no need for a separate state here)

libs/gpg45-evaluator/src/test/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45TicfValidatorTest.java
```diff
@@ -0,0 +1,26 @@
...
+    @Test
+    void isSuccessfulShouldReturnTrueOnValidCredential() {
+        CredentialEvidenceItem credentialEvidenceItem =
+                CredentialEvidenceItem.builder().type(TICF_EVIDENCE_TYPE).build();
+
+        assertTrue(Gpg45TicfValidator.isSuccessful(credentialEvidenceItem));
+    }
+
+    @Test
+    void isSuccessfulShouldReturnFalseOnValidCredential() {
```
Joe-Edwards-GDS: ...OnInvalidCredential?

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -374,7 +376,7 @@ private EvidenceRequest getEvidenceRequestForF2F(List<SignedJWT> credentials)
     private List<SignedJWT> getSignedJWTs(String userId) throws HttpResponseExceptionWithErrorBody {
         List<String> credentials =
                 userIdentityService.getUserIssuedCredentials(
-                        verifiableCredentialService.getVcStoreItems(userId));
+                        excludeTicfVC(verifiableCredentialService.getVcStoreItems(userId)));
```
Joe-Edwards-GDS: I wonder if we should actually move this check inside `userIdentityService.getUserIssuedCredentials`, and perhaps rename to `getIdentityCredentials`? - I don't think we'll ever want to use the TICF credential in any of our checks/identity operations.

That might actually mean we can remove the TICF evidence type from the GPG45 evaluator, because it'll never get that far :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Done. and accordingly updated jouney map.

libs/gpg45-evaluator/src/test/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45TicfValidatorTest.java
rpayal: Done

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
rpayal: Refactor done.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -351,11 +381,69 @@ CALL_TICF_CRI_ON_REUSE:
...
+  response:
+    type: process
+    lambda: call-ticf-cri
+  events:
+    next:
+      targetState: PYI_ANOTHER_WAY
+    enhanced-verification:
+      targetState: PYI_ANOTHER_WAY
+    fail-with-ci:
+      targetState: PYI_ANOTHER_WAY
```
Joe-Edwards-GDS: Is there a reason you changed this back? I think we agreed that `fail-with-ci` (and `enhanced-verification` except in the F2F case) should always go to `PYI_NO_MATCH`.

As per
> I've just spoken to David and he's confirmed that if we get a CI from TICF in an error scenario, we should route to NO_MATCH

(I hope I haven't confused things!)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -351,11 +381,69 @@ CALL_TICF_CRI_ON_REUSE:
...
+  response:
+    type: process
+    lambda: call-ticf-cri
+  events:
+    next:
+      targetState: ERROR
+    enhanced-verification:
+      targetState: ERROR
+    fail-with-ci:
+      targetState: ERROR
```
Joe-Edwards-GDS: Ditto here

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: Done change event 'enhanced-verification' will go to the original error page raised before calling TICF and event 'fail-with-ci' will always go to PYI_NO_MATCH.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
rpayal: done - as above

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1495, opened 2023-12-20
BAU: Add TICF CRI stub to local running setup

Wynndow:
### What changed

Add TICF CRI stub to local running setup

### Why did it change

As the TICF CRI is not OAuth based (and so no callbacks to worry about), we can just use the deployed stub when running locally, rather than running it as another container like the other stubs.


TalatAydinGDS (approved):
LGTM

PR #1497, opened 2023-12-20
BAU: Track VC received for TICF CRI

Wynndow:
### What changed

Track VC received for TICF CRI

### Why did it change

When calling the TICF CRI, what we're actually interested in sending is all the VCs received so far this session. We've modelled this as a reuse/other journey so far, with a reuse journey not sending any VCs, and anything else sending all of them.

Due to how the journey map works and how many places we need to update to integrate the TICF CRI into all journeys it's very fiddly to get this right and not make a mistake.

This is an attempt to alleviate having to think about that. This is a speculative, untested, not even deployed stab at tracking the VCs received in the users session, and then using that to inform which VCs are sent to TICF.


Joe-Edwards-GDS:
@Wynndow as discussed, this seems like a reasonable approach to me - I'll raise a new task to track this properly.

PR #1509, opened 2024-01-03
BAU: Remove DDB TTL from UserIssuedCredentialsV2Table

Wynndow:
### What changed

Remove DDB TTL from UserIssuedCredentialsV2Table

### Why did it change

This removes the config from the SAM template that enables the DynamoDB TTL feature for the UserIssuedCredentialsV2Table table. This will have the effect of no more VCs being removed automatically from the table once the timestamp in the ttl attribute of the record has been reached.

It also removes the writing of the ttl timestamp for any future VCs stored in the table as this is no longer required.

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -2315,12 +2315,6 @@ Resources:
           KeyType: "HASH"
         - AttributeName: "credentialIssuer"
           KeyType: "RANGE"
-      TimeToLiveSpecification:
-        AttributeName: "ttl"
-        Enabled: !If
-          - IsProduction
-          - true #false set to true until user privacy notice is updated
-          - true
       SSESpecification:
         SSEEnabled: true
         SSEType: KMS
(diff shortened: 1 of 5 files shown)


Joe-Edwards-GDS (dismissed):
Looks good, a couple of thoughts:

- If you wanted to be particularly complete, you could remove the `ttl` property on `VcStoreItem` and add an explicit `setTtl` method that logs a warning (or even throws)
- Are you planning to do any checks in lower envs first (beyond integration tests)? I can't really think what they would be tbh

Wynndow:
> Looks good, a couple of thoughts:
> 
>     * If you wanted to be particularly complete, you could remove the `ttl` property on `VcStoreItem` and add an explicit `setTtl` method that logs a warning (or even throws)
> 
>     * Are you planning to do any checks in lower envs first (beyond integration tests)? I can't really think what they would be tbh

Ah, good point. I'll remove the attribute. Might be confusing to leave it in. I was debating about adding a comment somewhere not to use ttl, and the setter would be a good place to communcate this.

I've already deployed this to my dev account to make sure it the change doesn't blow up - it seems fine. Other than that I'm not sure what else could be done. I would say I could check the oldest VCs at the point the change goes out and then make sure they're still there in a couple of days, but we already know how hard that is to do...

But thinking about it, I could look at the metric for TTL deletions and check it doesn't go up. Won't block the change going to prod but will give is confidence we've stop deleting stuff.

Joe-Edwards-GDS (approved):

PR #1512, opened 2024-01-04
PYIC-4331: Track VCs to submit to TICF

Wynndow:
### What changed

Track VCs to submit to TICF

### Why did it change
We want to simplify the journey map when invoking the call-ticf-cri lambda by not having to supply a journey type. The journey type was being used to determine if we should send all the users VC (an IPV journey), or none (a reuse journey).

What we're actually interested in sending to the TICF VC is the VC gathered during the users current session. The previous behaviour was just a proxy for that.

This change keeps track of the VC the user has received in the current session, and then only sends those to the TICF CRI when appropriate.

I had originally planned to just store the JTI from the VC in the users session, and use that to filter the users VCs from the store. However not all CRIs currently provide a JTI claim (even thought they should), so have fallen back to storing the entire serialized VC in the session.

I've maintained the filtering behaviour - the VC in the users session are compared against the VC from the VC store before sending. We could in theory just use the VC in the store directly and skip the check of the VC in the store - this would save a round trip to the DB. It feels safer however to do the check, although I don't have a good example of why this is beneficial.

One consequence of this change is that on a reuse journey, where the user decides to reset their identity (an upcoming feature) we could send TICF their own VC. Hopefully this is ok.

### Issue tracking

- [PYIC-4331](https://govukverify.atlassian.net/browse/PYIC-4331)

[PYIC-4331]: https://govukverify.atlassian.net/browse/PYIC-4331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
+++ b/lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
@@ -36,6 +36,7 @@ import uk.gov.di.ipv.core.library.service.UserIdentityService;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
 
 import java.text.ParseException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -51,7 +52,6 @@ public class CallTicfCriHandler implements RequestHandler<ProcessRequest, Map<St
             new JourneyResponse(JOURNEY_FAIL_WITH_CI_PATH).toObjectMap();
     private static final Map<String, Object> JOURNEY_NEXT =
             new JourneyResponse(JOURNEY_NEXT_PATH).toObjectMap();
-    public static final String REUSE_JOURNEY_TYPE = "reuse";
 
     private final ConfigService configService;
     private final IpvSessionService ipvSessionService;
@@ -106,57 +106,16 @@ public class CallTicfCriHandler implements RequestHandler<ProcessRequest, Map<St
     @Override
     @Tracing
     @Logging(clearState = true)
-    public Map<String, Object> handleRequest(ProcessRequest input, Context context) {
+    public Map<String, Object> handleRequest(ProcessRequest request, Context context) {
         LogHelper.attachComponentIdToLogs(configService);
         LogHelper.attachCriIdToLogs(TICF_CRI);
 
+        IpvSessionItem ipvSessionItem = null;
         try {
-            String ipvSessionId = RequestHelper.getIpvSessionId(input);
-            String featureSet = RequestHelper.getFeatureSet(input);
-            String journeyType = RequestHelper.getJourneyType(input);
-            configService.setFeatureSet(featureSet);
-            IpvSessionItem ipvSessionItem = ipvSessionService.getIpvSession(ipvSessionId);
-            ClientOAuthSessionItem clientOAuthSessionItem =
-                    clientOAuthSessionDetailsService.getClientOAuthSession(
-                            ipvSessionItem.getClientOAuthSessionId());
-            LogHelper.attachGovukSigninJourneyIdToLogs(
-                    clientOAuthSessionItem.getGovukSigninJourneyId());
+            ipvSessionItem =
+                    ipvSessionService.getIpvSession(RequestHelper.getIpvSessionId(request));
 
-            List<SignedJWT> ticfVcs =
-                    ticfCriService.getTicfVc(
-                            clientOAuthSessionItem,
-                            ipvSessionItem,
-                            journeyType.equals(REUSE_JOURNEY_TYPE)
-                                    ? List.of()
-                                    : userIdentityService.getUserIssuedCredentials(
-                                            verifiableCredentialService.getVcStoreItems(
-                                                    clientOAuthSessionItem.getUserId())));
-
-            if (ticfVcs.isEmpty()) {
-                LogHelper.logMessage(Level.INFO, "No VC to process - returning next");
-                return JOURNEY_NEXT;
-            }
-
-            criStoringService.storeVcs(
-                    TICF_CRI, input.getIpAddress(), ipvSessionId, ticfVcs, clientOAuthSessionItem);
-
-            ContraIndicators cis =
-                    ciMitService.getContraIndicatorsVC(
-                            clientOAuthSessionItem.getUserId(),
-                            clientOAuthSessionItem.getGovukSigninJourneyId(),
-                            input.getIpAddress());
-
-            if (ciMitUtilityService.isBreachingCiThreshold(cis)) {
-                LogHelper.logMessage(
-                        Level.INFO, "CI score is breaching threshold - setting VOT to P0");
-                ipvSessionItem.setVot(VOT_P0);
-                ipvSessionService.updateIpvSession(ipvSessionItem);
-
-                return JOURNEY_FAIL_WITH_CI;
-            }
-
-            LogHelper.logMessage(Level.INFO, "CI score not breaching threshold");
-            return JOURNEY_NEXT;
+            return callTicfCri(ipvSessionItem, request);
 
         } catch (HttpResponseExceptionWithErrorBody e) {
             LogHelper.logErrorMessage("Error calling TICF CRI", e);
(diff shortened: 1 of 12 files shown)


lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -177,6 +136,74 @@ public Map<String, Object> handleRequest(ProcessRequest input, Context context)
...
+        List<String> vcToSendToTicf = new ArrayList<>();
+        List<String> vcReceivedThisSession = ipvSessionItem.getVcReceivedThisSession();
+
+        if (vcReceivedThisSession != null && !vcReceivedThisSession.isEmpty()) {
+            for (String storeVc : vcInStore) {
+                if (vcReceivedThisSession.contains(storeVc)) {
+                    vcToSendToTicf.add(storeVc);
+                }
+            }
+        }
```
Joe-Edwards-GDS, line 200: Is there a reason not to use streams here? I think it makes the intent a bit clearer:
```
return vcInStore.stream().filter(vcReceivedThisSession::contains).toList(); // or the expanded lambda syntax
```
(You'll still need the null check, but don't really need the empty check)

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
```diff
@@ -120,7 +127,9 @@ public void storeVcs(
                     List.of(vc.serialize()), govukSigninJourneyId, ipAddress);
 
             verifiableCredentialService.persistUserCredentials(vc, criId, userId);
+            vcReceivedThisSession.add(vc.serialize());
```
Joe-Edwards-GDS, line 130: Originally we were only storing the VC ID, which I kinda liked for a couple of reasons:
- We're storing less data (a full VC is much bigger)
- Once we move to encrypted VC storage, we won't have any VCs in IPV Core controlled stores

If it's much easier to do this way I think it's not the end of the world.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -177,6 +136,74 @@ public Map<String, Object> handleRequest(ProcessRequest input, Context context)
...
+        LogHelper.attachGovukSigninJourneyIdToLogs(
+                clientOAuthSessionItem.getGovukSigninJourneyId());
+
+        List<String> vcToSendToTicf = getVcToSendToTicf(clientOAuthSessionItem, ipvSessionItem);
+
+        List<SignedJWT> ticfVcs =
+                ticfCriService.getTicfVc(clientOAuthSessionItem, ipvSessionItem, vcToSendToTicf);
+
+        if (ticfVcs.isEmpty()) {
+            LogHelper.logMessage(Level.INFO, "No VC to process - returning next");
```
rpayal: Msg can be - "No TICF VC to process....

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -177,6 +136,74 @@ public Map<String, Object> handleRequest(ProcessRequest input, Context context)
...
+            return JOURNEY_FAIL_WITH_CI;
+        }
+
+        LogHelper.logMessage(Level.INFO, "CI score not breaching threshold");
+        return JOURNEY_NEXT;
+    }
+
+    @Tracing
+    private List<String> getVcToSendToTicf(
+            ClientOAuthSessionItem clientOAuthSessionItem, IpvSessionItem ipvSessionItem) {
```
rpayal: We can just pass userId instead of ClientOAuthSessionItem

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -333,8 +333,6 @@ CALL_TICF_CRI:
   response:
     type: process
     lambda: call-ticf-cri
-    lambdaInput:
-      journeyType: main
```
rpayal, line 337: Do we need removal of lambdaInput from journeyEngineStepFunction.asl.json?

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: No, no reason. Will switch it up.

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
rpayal, line 130: QQ - above line 125-126 - Why we need 2 calls to cimitService   (ciMitService.submitVC & ciMitService.submitMitigatingVcList) ?

Wynndow: Ah yeah, this is the thing I mentioned in the PR comment. Turns out that not all CRIs are returning a JTI claim in their VCs, even though they're supposed to. Another option could be to just store the signature from the VC and use that as a unique identfier, but it's a bit meh.

Storing the whole thing gives us the option to just use the values in the session rather than going back to the table. I've left the implementation here to go and fetch from the store, bit only because I have a uneasy feeling it's safer. What do you reckon?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: Ooo, good shout

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
Wynndow: So the first call is to submit the new VC so that any new CIs it contains can be stored.
The second call is to submit the same VC to CiMits mitigations endpoint - that's where we send VC that _could_ help mitigate a users CIs. 
They could really have been the same endpoint, but it was designed separately for reasons I don't know.

We could in theory only call that second endpoint when we know the user has a mitigating VC;  when they're on a mitigations journey and come back from DCMAW for example. But for ease sake, we decided to just send everything we get to the second endpoint too. CiMit will just ignore anything that isn't actually a mitigating VC.

rpayal: Got it....in those cimit service lambdas are called putContraIndicators & postMitigations.
Yeah sure postMitigations could have handled both as we are calling them together and passing the same payload.

Joe-Edwards-GDS: Sorry, I should have read the description! Do we know which CRIs don't issue a JTI, and why not (can we report it as a bug?)

In any case, they probably won't fix it soon enough for us here :( - maybe we can add it as a tech debt change later on. Given that, I agree that re-fetching from the store feels better, although it shouldn't actually make any difference.

Wynndow: I think it's a known about thing - it's mentioned in this RFC: https://github.com/govuk-one-login/architecture/blob/a6314577bcf8e49e6ff07b147b7ae51f749738e5/rfc/0041-CI-mitigation-VC.md?plain=1#L39
 
That RFC is 8 months old. Realisticly it's never getting fixed...

Joe-Edwards-GDS (approved):
Happy to approve, but let's try and get to the bottom of the missing JTIs anyway

rpayal (approved):

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
Wynndow: Have created a tech-debt ticket to fix it up when we can: https://govukverify.atlassian.net/browse/PYIC-4375

PR #1513, opened 2024-01-05
PYIC-4258: Validate and store inherited identity with new CriConfig

MikeCollingwood:
### What changed

- Added new class, `CriConfig`, to be able to validate the signed inherited identity JWT
- Made configs extend this new class instead of implement interface.
- Renamed these configs to OauthCriConfig and RestConfig.
- Added in functionality to validate and store inheritedIdentityJWT, behind a feature flag

### Why did it change

- We need to be able to process inherited identities received from the RP as part of the request claims set

### Issue tracking

- [PYIC-4258](https://govukverify.atlassian.net/browse/PYIC-4258)

See this PR for changes to parameters and the addition of the feature flag:
https://github.com/govuk-one-login/ipv-core-common-infra/pull/819

In theory, this sh/could work straight out of the box, though it needs manual testing. This could be done by deploying `orch-stub` hardcoded to respond with the inherited identity. Seeing the VC in DynamoDB would prove success.

Talat is working on the stub, which would be a great way to test this code. In addition, George Nemeshanyi is working on giving us the inherited identity JWT from orchestration, probably good to talk to them at some point!

Another thing, Sonar is breaking checks. Something to check with an SRE because looks like it could be an env problem.

[PYIC-4258]: https://govukverify.atlassian.net/browse/PYIC-4258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CriConfig.java
```diff
@@ -1,14 +1,23 @@
...
 
 import java.net.URI;
 import java.text.ParseException;
 
-public interface CriConfig {
-    URI getCredentialUrl();
+@AllArgsConstructor
+@NoArgsConstructor
+@SuperBuilder
+public class CriConfig {
```
MikeCollingwood, line 17: Classes which previously implemented the CriConfig interface now just extend this class. This is useful because this class is what is needed for validating the inheritedIdentityJWT

The classes which now extend this class were renamed for clarity:
- "CredentialIssuerConfig" => "OauthCriConfig"
- "BackendCriConfig" => "RestCriConfig"

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -155,6 +171,15 @@ public APIGatewayProxyResponseEvent handleRequest(
                             govukSigninJourneyId,
                             ipAddress);
 
+            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)) {
```
MikeCollingwood, line 177: Here's the important stuff!

Behind a feature flag it tries to validate and story any inherited identities.

The userId passed in comes from the claimSet sub of the parent VC. When validating the inherited identity jwt, we check the subject is equal to this id. This should reassure you that we're storing it under the correct userId!

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -345,6 +356,59 @@ void shouldReturnIpvSessionIdWhenRecoverableErrorFound()
...
+                Map.of("clientId", "test-client", "request", signedEncryptedJwt.serialize());
+        event.setBody(objectMapper.writeValueAsString(sessionParams));
+        event.setHeaders(Map.of("ip-address", TEST_IP_ADDRESS));
+
+        // Act
+        initialiseIpvSessionHandler.handleRequest(event, mockContext);
+
+        // Assert
+        verify(mockVerifiableCredentialJwtValidator, times(1))
+                .validate(any(), eq(testCriConfig), eq("test-user-id"));
```
MikeCollingwood, line 407: Ah if this method were actually called it would fail because the inherited identity sub !== "test-user-id"

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -224,9 +249,26 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+    private void validateAndStoreHMRCInheritedIdentity(String inheritedIdentityJWT, String userId)
+            throws ParseException, VerifiableCredentialException {
+        SignedJWT signedInheritedIdentityJWT = SignedJWT.parse(inheritedIdentityJWT);
+        CriConfig inheritedIdentityCriConfig = configService.getCriConfig(HMRC_MIGRATION_CRI);
+
+        verifiableCredentialJwtValidator.validate(
+                signedInheritedIdentityJWT, inheritedIdentityCriConfig, userId);
+
+        verifiableCredentialService.persistUserCredentials(
+                signedInheritedIdentityJWT, HMRC_MIGRATION_CRI, userId);
```
Wynndow: I'm going to hazard a guess we're going to need a new audit event in here.

Joe-Edwards-GDS: Audit event is covered by https://govukverify.atlassian.net/browse/PYIC-4248 (or perhaps one of the associated tickets)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -224,9 +249,26 @@ public APIGatewayProxyResponseEvent handleRequest(
...
                     HttpStatus.SC_BAD_REQUEST, ErrorResponse.MISSING_IP_ADDRESS);
+        } catch (VerifiableCredentialException e) {
+            LogHelper.logErrorMessage(
+                    "Inherited identity validation failed.", e.getErrorResponse().getMessage());
+            return ApiGatewayResponseGenerator.proxyJsonResponse(
+                    HttpStatus.SC_BAD_REQUEST, ErrorResponse.INVALID_SESSION_REQUEST);
         }
     }
 
+    private void validateAndStoreHMRCInheritedIdentity(String inheritedIdentityJWT, String userId)
```
Joe-Edwards-GDS: We'll also need to include in the `vcReceivedThisSession` array (for TICF)

Wynndow: Yep, good spot.

This raises an interesting question though. A user comes in on a migration journey with an inherited identity VC with a `vot` of `PCL250` and a `vtr` including `PCL250`. In theory we happily match the operational profile and send the user back to the service as a successful journey.

Before we issue the identity we'll send the inherited identity VC to TICF. What happens if TICF responds with a VC containing a CI? Do we fail the journey? Because presumably the identity is already issued and has no concept of CIs. It may be that TICF would never issue a CI in this scenario? Or do we ignore the response from TICF in that case?

Joe-Edwards-GDS: In general we respect CIs as normal, so in this case we'd fail the journey - TICF may be using other information to make that judgement (not sure what exactly...), but you're probably right that it's unlikely :)

Wynndow: I guess I'm wondering if that's acceptable to HMRC? A user may be blocked using our system that previously wasn't using their system.

Joe-Edwards-GDS: Yep - the latest direction is that if we've identified breaching CIs, then HMRC would like to know about it.

(They went back and forth on this a bit - as you say, it will block access to some existing users)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -249,24 +246,53 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-        SignedJWT signedInheritedIdentityJWT = SignedJWT.parse(inheritedIdentityJWT);
-        CriConfig inheritedIdentityCriConfig = configService.getCriConfig(HMRC_MIGRATION_CRI);
+    @Tracing
+    private void validateAndStoreHMRCInheritedIdentity(String userId, JWTClaimsSet claimsSet)
+            throws RecoverableJarValidationException, ParseException {
+        SignedJWT signedInheritedIdentityJWT;
+        try {
+            signedInheritedIdentityJWT =
+                    SignedJWT.parse(
+                            ((JarClaims) claimsSet.getClaim("claims"))
```
Joe-Edwards-GDS: I'm not convinced this will work - I think you need to explicitly map this as a JSON value (you can't just cast it).

Wynndow: Yep, you're quite right. The tests are all constructing a claimsSet with the right objects in which is why the tests pass. Need to make sure it's been deserialized/serialized to be a valid test.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +251,76 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+
+        var userInfo = claims.userInfo();
+        if (userInfo == null) {
+            return;
+        }
+
+        var inheritedIdentityJwtClaim = userInfo.inheritedIdentityClaim();
+        if (inheritedIdentityJwtClaim == null) {
+            return;
+        }
```
Joe-Edwards-GDS: Can we separate this logic into a different method (which will make it much more reusable if we introduce other inherited ids) - i.e. separate `getInheritedIdentityClaim` and `validateAndStoreHMRCInheritedIdentity`?

Joe-Edwards-GDS: These null checks are pretty ugly :( - shame Java doesn't let you do conditional access.

One slightly sneaky way of simplifying is leveraging optionals, for example:
```
var inheritedIdClaim = Optional.ofNullable(claims)
        .map(JarClaims::userInfo)
        .map(JarUserInfo::inheritedIdentityClaim)
        .map(InheritedIdentityJwtClaim::value)
        .flatMap((values) -> values.stream().findFirst());
```

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +251,76 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+        } catch (VerifiableCredentialException e) {
+            throw buildRecoverableException(
+                    OAuth2Error.SERVER_ERROR
+                            .setDescription("Failed to persist inherited identity JWT - ")
+                            .appendDescription(e.getMessage()),
+                    claimsSet);
+        }
+    }
+
+    private RecoverableJarValidationException buildRecoverableException(
```
Joe-Edwards-GDS: So this exception will (after some back and forth) result in the user going straight back to the RP with an error?

Joe-Edwards-GDS: Do you think this should be a constructor for `RecoverableJarValidationException` (or factory method if the exception is a problem)?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +251,76 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+        if (userInfo == null) {
+            return;
+        }
+
+        var inheritedIdentityJwtClaim = userInfo.inheritedIdentityClaim();
+        if (inheritedIdentityJwtClaim == null) {
+            return;
+        }
+
+        SignedJWT signedInheritedIdentityJWT;
```
Joe-Edwards-GDS: Whenever I see this pattern it smells of needing to extract a method - if you have a method, you can return from within the block.

(In truth, I just hate any non-final variables :) )

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +251,76 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                            .setDescription("Inherited identity JWT failed to validate - ")
+                            .appendDescription(e.getMessage()),
+                    claimsSet);
+        }
+
+        try {
+            verifiableCredentialService.persistUserCredentials(
+                    signedInheritedIdentityJWT, HMRC_MIGRATION_CRI, userId);
+            ipvSessionItem.setVcReceivedThisSession(
+                    List.of(signedInheritedIdentityJWT.serialize()));
```
Joe-Edwards-GDS: I wonder if we should add an `addVcReceivedThisSession` method to `IpvSessionItem` - I think it might simplify the code in the other use-case as well...

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -169,11 +173,13 @@ private JWTClaimsSet getValidatedClaimSet(SignedJWT signedJWT, String clientId)
                                 JWTClaimNames.SUBJECT));
 
         try {
-            verifier.verify(signedJWT.getJWTClaimsSet(), null);
+            JWTClaimsSet claimsSet = signedJWT.getJWTClaimsSet();
+            verifier.verify(claimsSet, null);
 
-            validateMaxAllowedJarTtl(signedJWT.getJWTClaimsSet());
+            validateMaxAllowedJarTtl(claimsSet);
+            validateInheritedIdentityJwtClaim(claimsSet);
```
Joe-Edwards-GDS: Can we put this behind the feature flag, to avoid the risk of breaking any prod journeys prematurely?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -223,4 +229,44 @@ private URI validateRedirectUri(JWTClaimsSet claimsSet, String clientId)
                             "Failed to parse JWT claim set in order to access redirect_uri claim"));
         }
     }
+
+    private void validateInheritedIdentityJwtClaim(JWTClaimsSet claimsSet)
```
Joe-Edwards-GDS: It feels a bit unsatisfactory to convert the claims object and perform almost all the same checks twice, but it looks ugly to avoid :(

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -80,33 +116,50 @@ class InitialiseIpvSessionHandlerTest {
...
+        jarClaims.put(AUDIENCE, "test-audience");
+        jarClaims.put(ISSUER, "test-issuer");
+        jarClaims.put(RESPONSE_TYPE, "code");
+        jarClaims.put(REDIRECT_URI, "https://example.com");
+        jarClaims.put(STATE, "test-state");
+        jarClaims.put(CLIENT_ID, "test-client");
+        jarClaims.put(VTR, List.of("Cl.Cm.P2", "Cl.Cm.PCL200"));
+        jarClaims.put(
+                CLAIMS,
+                new JarClaims(
```
Joe-Edwards-GDS: Is this allowed - can we put a custom class here? I assume it works, but strictly speaking I think these should be maps.

MikeCollingwood: Is there a reason we wouldn't want to build it like this:
```
         JWTClaimsSet claimsSet =
                new JWTClaimsSet.Builder()
                        .expirationTime(new Date(Instant.now().plusSeconds(1000).getEpochSecond()))
                        .issueTime(new Date())
                        .notBeforeTime(new Date())
                        .subject("test-user-id")
                        .audience("test-audience")
                        .issuer("test-issuer")
                        .claim("response_type", "code")
                        .claim("redirect_uri", "http://example.com")
                        .claim("state", "test-state")
                        .claim("client_id", "test-client")
                        .claim("vtr", List.of("Cl.Cm.P2", "Cl.Cm.PCL200"))
                        .build();
```

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood: I've made it final by nesting a try/ catch, though could do this another way

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
MikeCollingwood: Have moved validation to only be in `validateAndStoreHMRCInheritedIdentity`

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +258,88 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            }
+        } catch (VerifiableCredentialException | ParseException e) {
+            throw buildRecoverableException(
+                    OAuth2Error.INVALID_REQUEST_OBJECT
+                            .setDescription("Inherited identity JWT failed to validate - ")
+                            .appendDescription(e.getMessage()),
+                    claimsSet);
+        } catch (JarValidationException e) {
+            throw buildRecoverableException(e.getErrorObject(), claimsSet);
+        }
```
Joe-Edwards-GDS, line 301: If you separate this into two methods, I think you can avoid the nested try/catch? i.e. `validateHmrcInheritedIdentity` and `storeHmrcInheritedIdentity` (the 'and' in this method name hints at it too!)

Joe-Edwards-GDS: (With some subtlety to distinguish validating that the JAR claims structure is valid, and that the VC contained in the claim is valid)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +258,88 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                                        "Inherited identity jwt claim received but value is null")));
+        if (inheritedIdentityJwtList.size() != 1) {
+            throw new JarValidationException(
+                    OAuth2Error.INVALID_REQUEST_OBJECT.setDescription(
+                            String.format(
+                                    "%d inherited identity jwts received - one expected",
+                                    inheritedIdentityJwtList.size())));
+        }
+        return inheritedIdentityJwtList;
+    }
```
Joe-Edwards-GDS, line 332: Given that this method validates that the list contains exactly one element, should it just return that directly?

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
Joe-Edwards-GDS: I think it's because there are some tests that might modify these. It might be neater to have a 'default' builder which can be appended to somehow. I haven't thought through all the details though...

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -227,6 +258,88 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            var signedInheritedIdentityJWT = SignedJWT.parse(inheritedIdentityJwtList.get(0));
+            var inheritedIdentityCriConfig = configService.getCriConfig(HMRC_MIGRATION_CRI);
+
+            verifiableCredentialJwtValidator.validate(
+                    signedInheritedIdentityJWT, inheritedIdentityCriConfig, userId);
+            LogHelper.logMessage(Level.INFO, "Migration VC successfully validated");
+
+            // Store
+            try {
+                verifiableCredentialService.persistUserCredentials(
```
rpayal: When we r persisting the migrated hmrc VC........let say the req was for vtr PCL250. 
Do we here also need to check for any previous hmrc vc for vtr PCL200 (i.e all lower profile vc's) and delete it ?

MikeCollingwood: Think we have a [ticket](https://govukverify.atlassian.net/browse/PYIC-4384https://govukverify.atlassian.net/browse/PYIC-4384) for looking into this

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -231,6 +261,97 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            throw new RecoverableJarValidationException(e.getErrorObject(), claimsSet, e);
+        }
+    }
+
+    private SignedJWT validateHmrcInheritedIdentity(
+            String userId, InheritedIdentityJwtClaim inheritedIdentityJwtClaim)
+            throws JarValidationException, ParseException, VerifiableCredentialException {
+        // Validate JAR claims structure is valid
+        var inheritedIdentityJwtList =
+                Optional.of(inheritedIdentityJwtClaim.value())
```
Joe-Edwards-GDS: You need to use `ofNullable` if you think it may be nullable (`of` will throw an NPE)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -288,30 +295,30 @@ private void validateAndStoreHMRCInheritedIdentity(
 
             verifiableCredentialJwtValidator.validate(
                     signedInheritedIdentityJWT, inheritedIdentityCriConfig, userId);
-            LogHelper.logMessage(Level.INFO, "Migration VC successfully validated");
+            LOGGER.info("Migration VC successfully validated");
```
Joe-Edwards-GDS: Can we use `LogHelper.buildLogMessage` here?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/domain/InheritedIdentityJwtClaim.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.ipv.core.initialiseipvsession.domain;
+
+import java.util.List;
+
+public record InheritedIdentityJwtClaim(List<String> value) {}
```
Joe-Edwards-GDS: Sorry Mike - this has just been changed to `values` in the RFC 🤦 

(kinda makes sense, given it's a list...)

https://github.com/govuk-one-login/architecture/pull/554/files

Joe-Edwards-GDS (commented):
I think this is almost there!

Joe-Edwards-GDS (approved):
I think this is good, thanks, and we can come back to it as part of the other tickets if anything comes up. :)

PR #1516, opened 2024-01-08
PYIC-4121: Fix passport pact test

DanCorderIPV:
Update passport pact test to not change the pact file each time the test runs, and to ignore the JWT signature in the provider test (because the signature changes each time it is generated)

The changes happened for two reasons
1. The generated VC had a timestamp inside it that changed
2. The JWT signature is generated with a random seed and so changes even if the JWT content is identical

### What changed

- Change the way that the test VC is created - we now hardcode the JSON string in the test which should make updating it in the future simpler. This also fixes the timestamp.
- Create a fixed valid signature so that the PACT file doesn't change each time the test is run
- Tell the PACT framework to validate the provider response with a regex that will ignore the signature block

### Why did it change

Because our response is a JWT (which the PACT framework doesn't understand) the expected response needs to match the provider response exactly. That means that things like timestamps need to be fixed so the the consumer and provider side tests can both use the same timestamp.
The PACT file should also not change every time we run the test. If it did it would add a lot of noise to the PACT server and the server would probably run lots of unnecessary on tests on pacts that would be functionally identical.

### Issue tracking

- [PYIC-4124](https://govukverify.atlassian.net/browse/PYIC-4121)

[PYIC-4124]: https://govukverify.atlassian.net/browse/PYIC-4124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/PactJwtIgnoreSignatureBodyBuilder.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/PactJwtIgnoreSignatureBodyBuilder.java
@@ -1,0 +1,79 @@
+package uk.gov.di.ipv.core.processcricallback.pact;
+
+import au.com.dius.pact.consumer.dsl.BodyBuilder;
+import au.com.dius.pact.core.model.ContentType;
+import au.com.dius.pact.core.model.generators.Generators;
+import au.com.dius.pact.core.model.matchingrules.MatchingRuleCategory;
+import au.com.dius.pact.core.model.matchingrules.MatchingRuleGroup;
+import au.com.dius.pact.core.model.matchingrules.RegexMatcher;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.nimbusds.jose.util.Base64URL;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.List;
+
+// This class can generate an example signed JWT for use in a test, but will also tell the PACT
+// framework that the provider signature does not need to match the example signature (as JWT
+// signatures are randomised)
+public class PactJwtIgnoreSignatureBodyBuilder implements BodyBuilder {
+    private final String minifiedHeaderJson;
+    private final String minifiedBodyJson;
+    private final String signature;
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    public PactJwtIgnoreSignatureBodyBuilder(String headerJson, String bodyJson, String signature) {
+        this.minifiedHeaderJson = minifyJson(headerJson);
+        this.minifiedBodyJson = minifyJson(bodyJson);
+        this.signature = signature;
+    }
+
+    @Override
+    public MatchingRuleCategory getMatchers() {
+        var noSignatureRegex = new RegexMatcher(createJwtRegex());
+        return new MatchingRuleCategory(
+                "full body matcher",
+                Collections.singletonMap(
+                        "ignore JWT signature", new MatchingRuleGroup(List.of(noSignatureRegex))));
+    }
+
+    @Override
+    public Generators getGenerators() {
+        return new Generators();
+    }
+
+    @Override
+    public ContentType getContentType() {
+        return new ContentType("application/jwt; charset=UTF-8");
+    }
+
+    @Override
+    public byte[] buildBody() {
+        return (Base64URL.encode(minifiedHeaderJson)
+                        + "."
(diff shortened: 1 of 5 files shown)


lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/PactJwtIgnoreSignatureBodyBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+    public MatchingRuleCategory getMatchers() {
+        var noSignatureRegex = new RegexMatcher(createJwtRegex());
+        return new MatchingRuleCategory(
+                "full body matcher",
+                Collections.singletonMap(
+                        "ignore JWT signature", new MatchingRuleGroup(List.of(noSignatureRegex))));
+    }
+
+    @Override
+    public Generators getGenerators() {
```
DanCorderIPV, line 43: I don't know what this is for. Leaving it empty seems to be OK

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -64,13 +64,94 @@
...
-    public static final String CRI_RSA_ENCRYPTION_PUBLIC_JWK =
-            "{\"kty\":\"RSA\",\"e\":\"AQAB\",\"n\":\"vyapkvJXLwpYRJjbkQD99V2gcPEUKrO3dwjcAA9TPkLucQEZvYZvb7-wfSHxlvJlJcdS20r5PKKmqdPeW3Y4ir3WsVVeiht2iOZUreUO5O3V3o7ImvEjPS_2_ZKMHCwUf51a6WGOaDjO87OX_bluV2dp01n-E3kiIl6RmWCVywjn13fX3jsX0LMCM_bt3HofJqiYhhNymEwh39oR_D7EE5sLUii2XvpTYPa6L_uPwdKa4vRl4h4owrWEJaJifMorGcvqhCK1JOHqgknN_3cb_ns9Px6ynQCeFXvBDJy4q71clkBq_EZs5227Y1S222wXIwUYN8w5YORQe3M-pCIh1Q\"}";
+    private static final String IPV_CORE_CLIENT_ID = "ipv-core";
+    private static final String PRIVATE_API_KEY = "dummyApiKey";
+    private static final Clock CURRENT_TIME =
+            Clock.fixed(Instant.parse("2099-01-01T00:00:00.00Z"), ZoneOffset.UTC);
+    private static final String CRI_SIGNING_PRIVATE_KEY_JWK =
+            """
+            {"kty":"EC","d":"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU","crv":"P-256","x":"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM","y":"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04"}
```
DanCorderIPV, line 73: Having the string in this form in the code means that we can just copy and paste it into JWT.io

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java
```diff
@@ -64,13 +64,94 @@
...
+    // sent by the CRI team
+    private static final String VALID_VC_HEADER =
+            """
+            {
+              "typ": "JWT",
+              "alg": "ES256"
+            }
+            """;
+    // 2099-01-01 00:00:00 is 4070908800 in epoch seconds
+    private static final String VALID_VC_BODY =
```
DanCorderIPV, line 91: It's quite unlikely that I've got all the JSON below in the correct order, so this will probably need to be updated once Jamie writes the provider side test

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/FixedTimeJWTClaimsVerifier.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Date currentTime;
+
+    public FixedTimeJWTClaimsVerifier(
+            JWTClaimsSet exactMatchClaims, HashSet<String> requiredClaims, Date currentTime) {
+        super(exactMatchClaims, requiredClaims);
+        this.currentTime = currentTime;
+    }
+
+    @Override
+    protected Date currentTime() {
```
DanCorderIPV, line 26: According to a comment in the JWT code this method is overrideable for precisely this reason

shivanshuit914 (approved):

PR #1518, opened 2024-01-09
PYIC-4217: Fix NPE when name parts are null

Wynndow:
### What changed

Fix NPE when name parts are null

### Why did it change
The code we had to create a full name from a VC's name parts was throwing NPEs. This would happen if one of the name parts was null. We were calling equals on it when trying to filter out values that were equal to an empty string.

We've determined that it is not currently acceptable for a VC to have a missing name part, so we should be catching and logging this scenario in advance of getting to the stage of throwing this NPE.

And we were infact already attemting to do this when retrieving the name parts from the VC. We were only throwing though if ALL name parts were missing. This updates that check to catch if ANY of the name parts are missing. It also refines that check to check for null, empty string OR pure whitespace.

This change also removes the code that was originally throwing the NPE as it was unnecessary. We should have already caught any of these values. And if we hadn't, we'd still want to fail correlation for this.

This also updates the tests to be more flexible. Rather than relying on hardcoded VCs, which can be difficult to understand what they're doing, it allows VCs to be constructed with specific names and DOBs.

### Issue tracking

- [PYIC-4217](https://govukverify.atlassian.net/browse/PYIC-4217)

[PYIC-4217]: https://govukverify.atlassian.net/browse/PYIC-4217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -329,31 +329,25 @@ public class UserIdentityService {
 
     private List<String> getFullNamesFromCredentials(List<IdentityClaim> identityClaims) {
         return identityClaims.stream()
-                .flatMap(id -> id.getName().stream())
+                .flatMap(claim -> claim.getName().stream())
                 .map(Name::getNameParts)
                 .map(
                         nameParts -> {
                             String givenNames =
                                     nameParts.stream()
                                             .filter(
-                                                    nameParts1 ->
+                                                    namePart ->
                                                             GIVEN_NAME_PROPERTY_NAME.equals(
-                                                                            nameParts1.getType())
-                                                                    && !nameParts1
-                                                                            .getValue()
-                                                                            .equals(""))
+                                                                    namePart.getType()))
                                             .map(NameParts::getValue)
                                             .collect(Collectors.joining(" "));
 
                             String familyNames =
                                     nameParts.stream()
                                             .filter(
-                                                    nameParts1 ->
+                                                    namePart ->
                                                             FAMILY_NAME_PROPERTY_NAME.equals(
-                                                                            nameParts1.getType())
-                                                                    && !nameParts1
-                                                                            .getValue()
-                                                                            .equals(""))
+                                                                    namePart.getType()))
                                             .map(NameParts::getValue)
                                             .collect(Collectors.joining(" "));
 
@@ -672,7 +666,7 @@ public class UserIdentityService {
                 || names.stream()
                         .flatMap(name -> name.getNameParts().stream())
                         .map(NameParts::getValue)
-                        .allMatch(StringUtils::isEmpty);
+                        .anyMatch(StringUtils::isBlank);
     }
 
     private void addLogMessage(VcStoreItem item, String error) {
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
@@ -294,11 +294,12 @@ public class UserIdentityService {
         List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem item : vcStoreItems) {
             IdentityClaim identityClaim = getIdentityClaim(item.getCredential());
-            if (isNamesEmpty(identityClaim.getName())) {
+            String missingNames = getMissingNames(identityClaim.getName());
+            if (!missingNames.isBlank()) {
                 if (CRI_TYPES_EXCLUDED_FOR_NAME_CORRELATION.contains(item.getCredentialIssuer())) {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -672,7 +666,7 @@ private boolean isNamesEmpty(List<Name> names) {
                 || names.stream()
                         .flatMap(name -> name.getNameParts().stream())
                         .map(NameParts::getValue)
-                        .allMatch(StringUtils::isEmpty);
+                        .anyMatch(StringUtils::isBlank);
```
Joe-Edwards-GDS: I think this works, but does mean that we don't get any additional information in the error message - i.e. to tell us which VC or name part is empty.

I suppose it should be relatively clear if you dive into the logs, but do you think we can add it easily?

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -1352,4 +1586,74 @@ private void mockParamStoreCalls(Map<ConfigurationVariable, String> params) {
...
+        return createCredentialWithNameAndBirthDate(
+                givenName, null, familyName, birthDateList, isSuccessful);
+    }
+
+    private String createCredentialWithNameAndBirthDate(
+            String givenName, String familyName, List<String> birthDates) throws Exception {
+        return createCredentialWithNameAndBirthDate(givenName, null, familyName, birthDates, true);
+    }
+
+    private String createCredentialWithNameAndBirthDate(
```
Joe-Edwards-GDS, line 1621: I think these are much better ways of constructing test credentials, since it allows us to easily see what's in them. They are definitely not specific to this test suite though (apart from the fact we only care about name and DOB).

Perhaps link this to https://govukverify.atlassian.net/browse/PYIC-4294, so they can be combined?

Joe-Edwards-GDS: (Another one that might end up being easier after https://govukverify.atlassian.net/browse/PYIC-3908 as well - since we'll get classes in the right shape)

Joe-Edwards-GDS (dismissed):
Looks good, a couple of comments

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: We're logging the issuer of the VC, but it would be good to show which name part... I'll wrangle it in somehow.

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
Wynndow: Yep, I'll link something on that ticket.

It would be nice to have a lib for the generation of testing VCs, which can be used across all the lambdas tests.

I think there is also value in having the hardcoded test fixtures, but they need to be up to date and easily viewable...

Joe-Edwards-GDS: Thanks - I think having a preset list of 'useful test VCs' is good, but we could still generate them with the shared lib rather than using a hardcoded string?

```
public static String DCMAW_BIOMETRIC_PASSPORT_VC = createTestVc(...);
```
etc.

Wynndow: Yeah that's possible. I think the thing in my head is that the hardcoded stuff should be examples given to us (or gathered) of *actual* VCs issued by the CRIs. So that we can be sure that core will work with them, or catch issues in them.
The fixtures we currently have aren't really that - lots of incomplete of just wrong examples in there which makes me worry a little that we could easily miss something when using them.
That's my concern about generating them with the lib - it'd be our interpretation of what they should be, and will be more fallible.

Some of this stuff will just naturally get tested by the acceptance tests, but then we don't run as many scenarios (rightly) in staging. I'm not sure what the right balance is...

Joe-Edwards-GDS: Hmm, I think the difficulty will be keeping those updated.

I suppose in principle this is where the contract tests could help, but we're struggling to really leverage them against the VC structure, which is the important part.

If we generate classes from the JSON schema then I'd feel a lot more confident we're handling the right _structure_, and we're just left with issues around specific values. This shouldn't be such a big deal - we have a lot of specs for this stuff, and if we tear out all the existing evidence parsing (where it deduces the type of VC from the scores present..!) it should become much more robust against different values.

Joe-Edwards-GDS (approved):

PR #1519, opened 2024-01-10
BAU: Fix Sonar code smells and bugs

Wynndow:
### What changed

Since Sonar has not been reporting issues on PRs we've built up a few code smells and bugs. This address them.

### Why did it change

To keep the codebase at the standard we've agreed on.

Probably best to review commit by commit. They're all pretty small.


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -125,27 +125,26 @@ public static String getJourneyParameter(URI journeyUri, String key) {
...
 
-    public static Boolean getIsUserInitiated(ProcessRequest request)
+    public static boolean getIsUserInitiated(ProcessRequest request)
             throws HttpResponseExceptionWithErrorBody {
-        return extractValueFromLambdaInput(
-                request,
-                "isUserInitiated",
-                ErrorResponse.MISSING_IS_USER_INITIATED_PARAMETER,
-                Boolean.class);
+        return Boolean.TRUE.equals(
```
Joe-Edwards-GDS, line 139: This doesn't hurt, but it looks like `extractValueFromLambdaInput` can never actually return `null` anyway?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -125,27 +125,26 @@ public static String getJourneyParameter(URI journeyUri, String key) {
...
+                extractValueFromLambdaInput(
+                        request,
+                        "isUserInitiated",
+                        ErrorResponse.MISSING_IS_USER_INITIATED_PARAMETER));
     }
 
     private static <T> T extractValueFromLambdaInput(
-            ProcessRequest request, String key, ErrorResponse errorResponse, Class<T> returnType)
```
Joe-Edwards-GDS, line 148: Presumably this was intended to check that the value actually had the right type, but now we just don't bother? - Relying on a runtime exception in the (unlikely) case that it has the wrong type

Joe-Edwards-GDS (approved):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
Wynndow: No it can't, it does a null check internally and throws. But Sonar doesn't detect that null check. Returning a primitive instead keeps it happy and means we don't have to do the null check everywhere we used the Boolean in the handler.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
Wynndow: It was included to set the value for <T>. I think it was me failing to realised that <T> would be set by the return type defined when calling the method. This change is how I should have done it the first time.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
Joe-Edwards-GDS: Yeah ok - I guess the classic solution to this is to create specialisations for primitives, but this is fine. Those damn nulls...

PR #1524, opened 2024-01-12
PYIC-4425 Revise log helper so it is more extensible and does not lose context

Joe-Edwards-GDS:
### What changed

LogHelper methods now only build the log messages, but do not perform the logging themselves.

### Why did it change

Putting all the log messages in a helper means that all the messages appear to come from `LogHelper.java`, which makes it much harder to identify which piece of code the log originates from.

### Issue tracking

- [PYIC-4425](https://govukverify.atlassian.net/browse/PYIC-4425)

[PYIC-4425]: https://govukverify.atlassian.net/browse/PYIC-4425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
+++ b/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
@@ -188,16 +188,18 @@ public class BuildClientOauthResponseHandler
 
             return clientResponse.toObjectMap();
         } catch (ParseException e) {
-            LogHelper.logErrorMessage("Authentication request could not be parsed", e);
+            LOGGER.error(
+                    LogHelper.buildErrorMessage("Authentication request could not be parsed", e));
             return buildJourneyErrorResponse(
                     HttpStatus.SC_BAD_REQUEST,
                     ErrorResponse.FAILED_TO_PARSE_OAUTH_QUERY_STRING_PARAMETERS);
         } catch (SqsException e) {
-            LogHelper.logErrorMessage("Failed to send audit event to SQS queue.", e.getMessage());
+            LOGGER.error(
+                    LogHelper.buildErrorMessage("Failed to send audit event to SQS queue.", e));
             return buildJourneyErrorResponse(
                     HttpStatus.SC_INTERNAL_SERVER_ERROR, ErrorResponse.FAILED_TO_SEND_AUDIT_EVENT);
         } catch (URISyntaxException e) {
-            LogHelper.logErrorMessage("Failed to construct redirect uri.", e.getMessage());
+            LOGGER.error(LogHelper.buildErrorMessage("Failed to construct redirect uri.", e));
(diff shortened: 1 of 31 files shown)


lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
```diff
@@ -188,16 +188,18 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
                     ErrorResponse.FAILED_TO_PARSE_OAUTH_QUERY_STRING_PARAMETERS);
         } catch (SqsException e) {
-            LogHelper.logErrorMessage("Failed to send audit event to SQS queue.", e.getMessage());
+            LOGGER.error(
+                    LogHelper.buildErrorMessage("Failed to send audit event to SQS queue.", e));
             return buildJourneyErrorResponse(
                     HttpStatus.SC_INTERNAL_SERVER_ERROR, ErrorResponse.FAILED_TO_SEND_AUDIT_EVENT);
         } catch (URISyntaxException e) {
-            LogHelper.logErrorMessage("Failed to construct redirect uri.", e.getMessage());
+            LOGGER.error(LogHelper.buildErrorMessage("Failed to construct redirect uri.", e));
```
Joe-Edwards-GDS, line 202: There are a whole bunch of other instances of this, where we were only including the exception message, but not the stack trace.

I'm not sure there's any reason to exclude the stack trace, unless these are very frequent and well-understood cases (and I don't think they generally are...)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -322,29 +321,31 @@ private String getSecretValue(String secretId) {
...
             LOGGER.error(
-                    new StringMapMessage()
-                            .with(
-                                    LOG_MESSAGE_DESCRIPTION.getFieldName(),
-                                    "Failed to find the resource within Secrets manager.")
-                            .with(LOG_SECRET_ID.getFieldName(), secretId)
-                            .with(LOG_ERROR_DESCRIPTION.getFieldName(), e.getMessage()));
+                    LogHelper.buildErrorMessage(
+                                    "Failed to find the resource within Secrets manager.", e)
+                            .with(LOG_SECRET_ID.getFieldName(), secretId));
```
Joe-Edwards-GDS, line 348: This is the other advantage of this pattern - we can easily extend the message with additional properties.

(There are almost certainly other examples where we can do this, but I haven't been through them...)

rpayal: Yeah this way we are adding the various attributes to the message field in the log message.
"message": {
    "description": "ContraIndicators Verifiable Credential validated.",
    "criId": "3123213"
  },

rpayal (approved):

PR #1545, opened 2024-01-18
PYIC-4515: Evaluate vtr and include operational profiles

Wynndow:
### What changed

Evaluate vtr and include operational profiles

### Why did it change

This updates the check-existing-identity lambda to take account of the vtr value received in the JAR request from orch. It will now check if the user has met any of the vtr received, in strength order, and return a reuse journey if the have.

It creates some enums for the vot we will support, and maps them to the profiles that are currently supported for a user to meet that vot.

For a GPG45 vot (currently only P2), this is done in the usual way. For an HMRC vot this is done by checking if the user has an inherited identity VC that has the correct level of vot set on it.

### Issue tracking

- [PYIC-4515](https://govukverify.atlassian.net/browse/PYIC-4515)

[PYIC-4515]: https://govukverify.atlassian.net/browse/PYIC-4515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -350,22 +361,17 @@ private Map<String, Object> buildNoMatchResponse(
...
-                        auditEventUser));
+    private Map<String, Object> buildReuseResponse(Vot vot, AuditEventUser auditEventUser)
+            throws SqsException {
+        if (vot.requiresProfileType(GPG_45)) {
+            auditService.sendAuditEvent(
+                    new AuditEvent(
+                            AuditEventTypes.IPV_IDENTITY_REUSE_COMPLETE,
+                            configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser));
+        }
```
Wynndow: I've kept this audit event only for P2 vots in the assumption we're going to need a different audit event for inherited identity reuse journeys.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -185,7 +185,7 @@ private Map<String, Object> callTicfCri(IpvSessionItem ipvSessionItem, ProcessRe
             LOGGER.info(
                     LogHelper.buildLogMessage(
                             "CI score is breaching threshold - setting VOT to P0"));
-            ipvSessionItem.setVot(VOT_P0);
+            ipvSessionItem.setVot(Vot.P0.name());
```
Joe-Edwards-GDS, line 188: Do you think it'd be convenient to make `IpvSessionItem#vot` enum-valued? (e.g. with https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBTypeConvertedEnum.html)

I think there are some merits to using a plain string (it better represents what we're _actually_ storing and forces us to handle unsupported values more explicitly) but might make a lot of code cleaner, and reduce the number of places we call `.name()`.

I'm not sure!

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -346,26 +358,21 @@ private Map<String, Object> buildNoMatchResponse(
...
-        LOGGER.info(message);
-
-        auditService.sendAuditEvent(
-                new AuditEvent(
-                        AuditEventTypes.IPV_IDENTITY_REUSE_COMPLETE,
-                        configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
-                        auditEventUser));
+    private Map<String, Object> buildReuseResponse(Vot vot, AuditEventUser auditEventUser)
+            throws SqsException {
+        if (vot.hasGpg45Profiles()) {
```
Joe-Edwards-GDS: This is an interesting question - we'll have to determine whether we still want this audit event for reuse of an operational profile. Perhaps worth asking Simon Parsons?

It might become less relevant once we've got our journey split, and we have a new explicit event that captures different scenarios.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -410,4 +417,102 @@ private List<String> extractTxnIdsFromCredentials(List<SignedJWT> credentials)
...
+            // Successful match
+            if (matchedOperationalProfile.isPresent()) {
+                return Optional.of(
+                        new VotProfilePair(requestedVot, matchedOperationalProfile.get()));
+            }
+        }
+        return Optional.empty();
+    }
+
+    private List<SignedJWT> removeOperationalProfileVcs(List<SignedJWT> credentials)
```
Joe-Edwards-GDS, line 502: Looks like @rpayal has implemented a similar thing as part of his ticket as well, so we should make sure to commonise when merging the two.

I probably didn't do a good enough job of separating the implementation details for the different tickets 😬

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -410,4 +417,102 @@ private List<String> extractTxnIdsFromCredentials(List<SignedJWT> credentials)
...
+
+                return attainedVotAndProfile;
+            }
+        }
+        return Optional.empty();
+    }
+
+    private Optional<VotProfilePair> matchGpg45Profile(
+            List<SignedJWT> credentials,
+            List<VcStoreItem> vcStoreItems,
```
Joe-Edwards-GDS, line 461: It's pretty suspicious that this requires both `credentials` and `vcStoreItems`, but not sure if you can easily untangle this.

Feels like it _should_ just be `credentials`, but `checkRequiresAdditionalEvidence` uses the CRI Id instead :/

Something I'd like to fix as part of the proposed VC model refactoring, so don't worry _too_ much about it now.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -0,0 +1,34 @@
...
+    PCL250(List.of(), List.of(OperationalProfile.PCL250)),
+    PCL200(List.of(), List.of(OperationalProfile.PCL250, OperationalProfile.PCL200));
+
+    private final List<Gpg45Profile> validGpg45Profiles;
+    private final List<OperationalProfile> validOperationalProfiles;
+
+    Vot(List<Gpg45Profile> validGpg45Profiles, List<OperationalProfile> validOperationalProfiles) {
+        this.validGpg45Profiles = validGpg45Profiles;
+        this.validOperationalProfiles = validOperationalProfiles;
+    }
```
Joe-Edwards-GDS, line 23: Might be cleaner to have two constructors to enforce that you only ever have one of the two lists?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public List<Gpg45Profile> getValidGpg45Profiles() {
```
Joe-Edwards-GDS: Is this actually `getSupportedGpg45Profiles`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -0,0 +1,34 @@
...
+        return validGpg45Profiles;
+    }
+
+    public List<OperationalProfile> getValidOperationalProfiles() {
+        return validOperationalProfiles;
+    }
+
+    public boolean hasGpg45Profiles() {
+        return validGpg45Profiles != null && !validGpg45Profiles.isEmpty();
```
Joe-Edwards-GDS: If we do the constructor change above, I wonder if this should be `isGpg45()` and check for the presence of a list (but not check its emptiness)?

This could introduce the possibility to (for example) add P3 as a recognised VOT with an empty list of supported profiles.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -223,37 +234,37 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
-            return buildProfileMatchedResponse(matchedProfile.get(), auditEventUser);
         } catch (HttpResponseExceptionWithErrorBody e) {
             LOGGER.error(LogHelper.buildErrorMessage(e.getErrorResponse().getMessage(), e));
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
         } catch (CiRetrievalException e) {
             return buildErrorResponse(ErrorResponse.FAILED_TO_GET_STORED_CIS, e);
-        } catch (ParseException e) {
+        } catch (ParseException | JwtClaimsSetParseException e) {
```
Joe-Edwards-GDS: If you're catching this here, should it really be a runtime exception?

Joe-Edwards-GDS (commented):
I think this is looking good, but a few comments

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VectorOfTrust.java
Joe-Edwards-GDS, line 1: You'll probably want to rebase once Mike merged his changes, as he's updated this file.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: I think it would be conveniant - and presumably values already stored will be correctly unmarshalled back to an enum at the point that change was introduced.
I think the majority of the code affected will be in tests - I think we only set the Vot in a small number of places, but that still might be valuable.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Dropping this here for context: https://github.com/govuk-one-login/ipv-core-back/pull/1548

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Yeah, this annoyed me. I didn't want to have to parse all the credentials again just to reduce the number of params.
I think it would be interesting to understand exactly how heavy a process that is to do...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Wynndow: Ha. That's exacly what I started with and then switched to `valid` at the last minute. I think mostly because it was less verbose. Maybe that was a bad decision. I'll switch it back.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Wynndow: Nice, I'll do that.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: It's a runtime exception as it's getting thrown in a stream, and you can't throw a checked exception from a stream. Unless you know a sneaky/better way?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Wynndow: Interestingly this isn't super straightforward, as the two constructors have the same signature after type erasure.

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: It might be worth making this change as a separate PR - i think it might end up being a bit chunky and better reviewed separately.

Joe-Edwards-GDS: Yep, that might make it easier to tell if it's a good idea as well

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: I don't think it's very heavy - string splitting and presumably base64 decoding (it's not doing signature validation or anything). Ideally I think we'd have one model class, which handles the 'parse once' logic, but not worth doing now.

In the meantime, I don't mind too much - either re-parse or have two args 🤷

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Joe-Edwards-GDS: Ah of course...

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Bah 😞 

Yeah, the main alternative is just to avoid using a stream, which is almost as disappointing.

Wynndow: Want me to remove the stream?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Wynndow: We can switch the empty lists in the calls to the constructor to nulls though, and then still make the change you mention below.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Yeah, I think that's probably slightly more maintainable (but I'm not happy about it!)

Joe-Edwards-GDS (dismissed):

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -410,4 +416,98 @@ private List<String> extractTxnIdsFromCredentials(List<SignedJWT> credentials)
...
+
+            return Optional.of(new VotProfilePair(requestedVot, matchedGpg45Profile.get().label));
+        }
+        return Optional.empty();
+    }
+
+    private Optional<VotProfilePair> matchOperationalProfile(
+            List<SignedJWT> credentials, Vot requestedVot) throws ParseException {
+        for (SignedJWT cred : credentials) {
+            String credentialVot = cred.getJWTClaimsSet().getStringClaim(VOT_CLAIM);
```
Joe-Edwards-GDS, line 486: Do you think this should use the `getVot` method in `UserIdentityService`?

The interesting difference is if the VOT is invalid - this will ignore it, whereas that method will throw...

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -410,4 +416,98 @@ private List<String> extractTxnIdsFromCredentials(List<SignedJWT> credentials)
...
+            }
+        }
+        return Optional.empty();
+    }
+
+    private List<SignedJWT> removeOperationalProfileVcs(List<SignedJWT> credentials)
+            throws ParseException {
+        var gpg45Vcs = new ArrayList<SignedJWT>();
+        for (var credential : credentials) {
+            var credVot = credential.getJWTClaimsSet().getStringClaim(VOT_CLAIM);
```
Joe-Edwards-GDS, line 506: (ditto regarding `getVot`)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: I don't think so - only the inherited identity VCs have a `vot` defined. The method in UserIdentityService will throw a NPE if it's used with a cred that doesn't contain a VOT.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: As above

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Yeah, I guess maybe that method should handle the null case as well - but also happy to leave.

Joe-Edwards-GDS (approved):

PR #1547, opened 2024-01-18
PYIC-4384: Add check for existing inherited identity with stronger vot

MikeCollingwood:
### What changed

- Added `getVot` method to user identity service to get a VC's vot value and parse as a VectorOfTrust
- Added `isHmrcInheritedIdentityWithStrongerVotPresent` to check if existing vc exists for `userId` and `HMRC_MIGRATION_CRI` which has a stronger vot than the incoming vc
- Added PCL200 and PCL250 enum values to VectorOfTrust enum - which is now ordered for logic on strength

### Why did it change

- We want to prevent writing new VCs to the vc store for the cri hmrcMigration and userId when am inherited identity with a stronger vot already exists

### Issue tracking

- [PYIC-4384](https://govukverify.atlassian.net/browse/PYIC-4384)

[PYIC-4384]: https://govukverify.atlassian.net/browse/PYIC-4384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -394,12 +394,11 @@ public class InitialiseIpvSessionHandler
                                 existingInheritedIdentityVot, incomingInheritedIdentityVot));
             }
 
-            return indexOfExistingVot - indexOfIncomingVot < 0;
+            return indexOfIncomingVot > indexOfExistingVot;
         } catch (ParseException | IllegalArgumentException e) {
             throw new CredentialParseException(
-                    String.format(
-                            "Encountered a parsing error while attempting to parse or compare credentials: %s",
-                            e.getMessage()));
+                    "Encountered a parsing error while attempting to parse or compare credentials: %s",
+                    e);
         }
     }
 
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -33,6 +33,7 @@ import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsReproveIdent
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
 import uk.gov.di.ipv.core.library.config.CoreFeatureFlag;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.domain.VectorOfTrust;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.SqsException;
@@ -72,6 +73,8 @@ public class InitialiseIpvSessionHandler
     private static final String REQUEST_GOV_UK_SIGN_IN_JOURNEY_ID_KEY = "govuk_signin_journey_id";
     private static final String REQUEST_EMAIL_ADDRESS_KEY = "email_address";
     private static final String REQUEST_VTR_KEY = "vtr";
+    private static final List<VectorOfTrust> HMRC_PROFILES_BY_STRENGTH =
+            List.of(VectorOfTrust.PCL250, VectorOfTrust.PCL200);
 
     private final ConfigService configService;
     private final IpvSessionService ipvSessionService;
@@ -379,11 +382,24 @@ public class InitialiseIpvSessionHandler
             var incomingInheritedIdentityVot =
                     userIdentityService.getVot(incomingInheritedIdentity);
 
-            return existingInheritedIdentityVot.compareTo(incomingInheritedIdentityVot) > 0;
-        } catch (JsonProcessingException | ParseException | IllegalArgumentException e) {
+            var indexOfExistingVot =
+                    HMRC_PROFILES_BY_STRENGTH.indexOf(existingInheritedIdentityVot);
+            var indexOfIncomingVot =
+                    HMRC_PROFILES_BY_STRENGTH.indexOf(incomingInheritedIdentityVot);
+
+            if (indexOfExistingVot == -1 || indexOfIncomingVot == -1) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "At least one of the existing (%s) or incoming (%s) VoTs from hmrc aren't expected",
+                                existingInheritedIdentityVot, incomingInheritedIdentityVot));
+            }
+
+            return indexOfExistingVot - indexOfIncomingVot < 0;
+        } catch (ParseException | IllegalArgumentException e) {
             throw new CredentialParseException(
-                    "Encountered a parsing error while attempting to parse credentials: "
-                            + e.getMessage());
+                    String.format(
+                            "Encountered a parsing error while attempting to parse or compare credentials: %s",
+                            e.getMessage()));
         }
     }
 
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -397,7 +397,7 @@ public class InitialiseIpvSessionHandler
             return indexOfIncomingVot > indexOfExistingVot;
         } catch (ParseException | IllegalArgumentException e) {
             throw new CredentialParseException(
-                    "Encountered a parsing error while attempting to parse or compare credentials: %s",
+                    "Encountered a parsing error while attempting to parse or compare credentials",
                     e);
         }
     }
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -33,6 +33,7 @@ import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsReproveIdent
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
 import uk.gov.di.ipv.core.library.config.CoreFeatureFlag;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
+import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.SqsException;
 import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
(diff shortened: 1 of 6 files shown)


libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -733,15 +779,50 @@ void shouldNotSetIdentityClaimWhenVotIsP0() throws Exception {
...
+        // Act
         UserIdentity credentials =
                 userIdentityService.generateUserIdentity(
                         USER_ID_1, "test-sub", "P0", emptyContraIndicators);
 
+        // Assert
         assertNull(credentials.getIdentityClaim());
     }
 
+    @Test
```
MikeCollingwood, line 791: Here are the tests I added for the `getVot` method I introduced for the `UserIdentityService`

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -352,6 +367,26 @@ private void storeHmrcInheritedIdentity(
...
+            var incomingInheritedIdentityVot =
+                    userIdentityService.getVot(incomingInheritedIdentity);
+
+            return existingInheritedIdentityVot.compareTo(incomingInheritedIdentityVot) > 0;
+        } catch (JsonProcessingException | ParseException | IllegalArgumentException e) {
+            throw new CredentialParseException(
+                    "Encountered a parsing error while attempting to parse credentials: "
+                            + e.getMessage());
+        }
+    }
```
MikeCollingwood, line 388: Here's the important stuff ^. We check if there's a stored inherited identity with a stronger vot already

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VectorOfTrust.java
```diff
@@ -1,6 +1,9 @@
 package uk.gov.di.ipv.core.library.domain;
 
+// In order of strength
```
Joe-Edwards-GDS: We won't necessarily be able to strictly order all our Vectors of Trust, so I think we should _not_ attempt to order them (instead, we can probably just list `P0`, `P1`, `P2`, `PCL200`, `PCL250`.

In particular, once we introduce P1, it will be incomparable with PCL200/PCL250

cc @Wynndow , I think you might be enhancing this enum anyway(?)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -236,6 +242,7 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             return ApiGatewayResponseGenerator.proxyJsonResponse(HttpStatus.SC_OK, response);
         } catch (ParseException e) {
+            // Doesn't match all the potential causes
```
Joe-Edwards-GDS, line 248: Is this WIP?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -481,6 +482,14 @@ private IdentityClaim getIdentityClaim(String credential)
         return new IdentityClaim(names, birthDates);
     }
 
+    public VectorOfTrust getVot(SignedJWT credential)
+            throws JsonProcessingException, IllegalArgumentException {
+        var votJson =
+                objectMapper.readTree(credential.getPayload().toString()).path(VOT_CLAIM_NAME);
```
Joe-Edwards-GDS: Do you think `VectorOfTrust.valueOf(credential.getJWTClaimsSet().getStringClaim(VOT_CLAIM_NAME));` is cleaner here - doesn't require us to do any explicit JSON mapping.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -352,6 +367,26 @@ private void storeHmrcInheritedIdentity(
...
+            var vcStoreItem =
+                    verifiableCredentialService.getVcStoreItem(HMRC_MIGRATION_CRI, userId);
+            SignedJWT existingInheritedIdentity = SignedJWT.parse(vcStoreItem.getCredential());
+
+            var existingInheritedIdentityVot =
+                    userIdentityService.getVot(existingInheritedIdentity);
+            var incomingInheritedIdentityVot =
+                    userIdentityService.getVot(incomingInheritedIdentity);
+
+            return existingInheritedIdentityVot.compareTo(incomingInheritedIdentityVot) > 0;
```
Joe-Edwards-GDS: As described below, I think it might be good to make this a bit more explicit - e.g. introduce a:
```
private static final var HMRC_PROFILES_BY_STRENGTH = List.of(PCL250, PCL200);
```
or something

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood: It's not for this ticket, but I might raise a tech debt ticket to review error handling, and maybe to try to make it more standardised at the same time (since there seem to be quite a few ways one can raise an error)?

Joe-Edwards-GDS: Can you elaborate?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
MikeCollingwood: Ah definitely, I forgot about `getStringClaim`!

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood: Here are a few thoughts explaining why I think a review/ some carved-out time to try to understand exceptions use could help:

1. We don't want to miss potential causes like ParseException, IllegalArgumentException and exceptions extending RuntimeException when handling exceptions
2. Should we try to ensure we don't have exceptions extending RuntimeException?
3. (Something we've discussed before) Should we review errors are raised with a message relevant to the cause just once, to avoid duplicated logs
6. In InitialiseIpvSessionHandler JarValidationException is only used for it’s description despite needing to define an ErrorObject for it, so maybe we should be looking at it. I see some uses of the errorObject, but do wonder whether there's a better way of defining it. Maybe to be closer to HttpResponseExceptionWithErrorBody?
7. HttpResponseExceptionWithErrorBody use could be clearer imo - it's caught all over the code base but we only really want it to know what code to send to the client instead of a message
8. (Remove unused exceptions e.g. NoVisitedCriFoundException)
9. (Use Lombok for custom exception classes)

Bit of a menageries of thoughts sorry!

Joe-Edwards-GDS: Cool, thanks! I think you're right there could be some tech debt around error handling, feel free to raise a tech debt ticket if you identify specific changes (let me know though).

Thoughts:
1. We probably _don't_ want to handle `RuntimeException`s at the top level - a RTE should be a genuinely unexpected error, and so failing the lambda is probably correct.
2. Code review, but as above, yes they should only be for genuinely unexpected errors.
3. Yes ideally
4. Yeah, if we're only using the description then using error object is probably not that useful
5. We should probably have a consistent pattern for this - either all our exceptions should extend this, or not...
6. Sure
7. Maybe, I don't know how much it adds though

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -352,6 +370,39 @@ private void storeHmrcInheritedIdentity(
...
+                                "At least one of the existing (%s) or incoming (%s) VoTs from hmrc aren't expected",
+                                existingInheritedIdentityVot, incomingInheritedIdentityVot));
+            }
+
+            return indexOfExistingVot - indexOfIncomingVot < 0;
+        } catch (ParseException | IllegalArgumentException e) {
+            throw new CredentialParseException(
+                    String.format(
+                            "Encountered a parsing error while attempting to parse or compare credentials: %s",
+                            e.getMessage()));
```
Joe-Edwards-GDS: Maybe add another signature to `CredentialParseException` so you can just pass the exception as a cause, rather than inlining the message?

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -145,12 +147,14 @@ class InitialiseIpvSessionHandlerTest {
...
-    private static String inheritedIdentityJWT;
+    private static SignedJWT inheritedIdentityJWT;
+    private static String serialisedInheritedIdentityJWT;
 
     @BeforeAll
     static void setUpBeforeAll()
             throws InvalidKeySpecException, NoSuchAlgorithmException, JOSEException {
-        inheritedIdentityJWT = getInheritedIdentityJWT().serialize();
+        inheritedIdentityJWT = getInheritedIdentityJWT();
+        serialisedInheritedIdentityJWT = inheritedIdentityJWT.serialize();
```
Joe-Edwards-GDS: Is `inheritedIdentityJWT` actually used now - can we revert this?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -352,6 +370,39 @@ private void storeHmrcInheritedIdentity(
...
+                    HMRC_PROFILES_BY_STRENGTH.indexOf(incomingInheritedIdentityVot);
+
+            if (indexOfExistingVot == -1 || indexOfIncomingVot == -1) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "At least one of the existing (%s) or incoming (%s) VoTs from hmrc aren't expected",
+                                existingInheritedIdentityVot, incomingInheritedIdentityVot));
+            }
+
+            return indexOfExistingVot - indexOfIncomingVot < 0;
```
Joe-Edwards-GDS: `indexOfExistingVot > indexOfIncomingVot`?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -394,12 +394,11 @@ private boolean isHmrcInheritedIdentityWithStrongerVotPresent(
...
             }
 
-            return indexOfExistingVot - indexOfIncomingVot < 0;
+            return indexOfIncomingVot > indexOfExistingVot;
         } catch (ParseException | IllegalArgumentException e) {
             throw new CredentialParseException(
-                    String.format(
-                            "Encountered a parsing error while attempting to parse or compare credentials: %s",
-                            e.getMessage()));
+                    "Encountered a parsing error while attempting to parse or compare credentials: %s",
```
Joe-Edwards-GDS: You can remove the format string from the message now (the `%s`)

Joe-Edwards-GDS (commented):
One small comment left :)

Joe-Edwards-GDS (approved):

PR #1551, opened 2024-01-19
PYIC-4139: Handle case where retrieved VC store item has a null creden…

lwebb-iig:
…tial field

### What changed

- Adds null check against the credential field for retrieved VC store items
- Adds warning when VC found without credential field

### Why did it change

- A few VC store items appear to have null credentials, this was causing the replay lambda to error and not continue when finding a VC in this state

### Issue tracking

- [PYIC-4139](https://govukverify.atlassian.net/browse/PYIC-4139)

### Other considerations

[PYIC-4139]: https://govukverify.atlassian.net/browse/PYIC-4139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
+++ b/lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
@@ -92,12 +92,12 @@ public class ReplayCimitVcsHandler implements RequestStreamHandler {
             VcStoreItem vcStoreItem =
                     this.verifiableCredentialService.getVcStoreItem(
                             item.getUserId().get("S"), item.getCredentialIssuer().get("S"));
-            if (vcStoreItem.getCredential() != null) {
+            if (vcStoreItem != null) {
                 SignedJWT vc = SignedJWT.parse(vcStoreItem.getCredential());
                 ciMitService.submitVC(vc, null, null);
                 submittedVcs.add(vc.serialize());
             } else {
-                LOGGER.warn("Null credential in VC");
+                LOGGER.warn("VC not found");
             }
         }
         ciMitService.submitMitigatingVcList(submittedVcs, null, null);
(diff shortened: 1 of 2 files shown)


lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
```diff
@@ -92,9 +92,13 @@ private void handleBatch(List<ReplayItem> replayItems)
...
                             item.getUserId().get("S"), item.getCredentialIssuer().get("S"));
-            SignedJWT vc = SignedJWT.parse(vcStoreItem.getCredential());
-            ciMitService.submitVC(vc, null, null);
-            submittedVcs.add(vc.serialize());
+            if (vcStoreItem.getCredential() != null) {
+                SignedJWT vc = SignedJWT.parse(vcStoreItem.getCredential());
+                ciMitService.submitVC(vc, null, null);
+                submittedVcs.add(vc.serialize());
+            } else {
+                LOGGER.warn("Null credential in VC");
```
Joe-Edwards-GDS: Can we include the user id/cri here, in case we need to take any further action on these?

Joe-Edwards-GDS (approved):

lambdas/replay-cimit-vcs/src/main/java/uk/gov/di/ipv/core/replaycimitvcs/ReplayCimitVcsHandler.java
lwebb-iig: @Joe-Edwards-GDS are we alright to actually log that information? I assumed userId was off limits

Joe-Edwards-GDS: That's an interesting question - I think if it's not associated with any other info then it's not PII (it's basically a UUID), but it's probably safer to avoid if we don't need it (and we don't any more :)).

PR #1552, opened 2024-01-19
PYIC-4126 Use common CRI config to read component id

Joe-Edwards-GDS:
### What changed

Read the component ID from the new bundled CRI config

### Why did it change

This was an oversight in the original bundling of the CRI config. This only continued to work because we still had the old config values in SSM (but breaks for new deployments).

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -261,11 +261,8 @@ public class ConfigService {
     }
 
     public String getComponentId(String credentialIssuerId) {
-        String activeConnection = getActiveConnection(credentialIssuerId);
-        final String pathTemplate =
-                ConfigurationVariable.CREDENTIAL_ISSUERS.getPath()
-                        + "/%s/connections/%s/componentId";
-        return getSsmParameterWithOverride(pathTemplate, credentialIssuerId, activeConnection);
+        var criConfig = getOauthCriActiveConnectionConfig(credentialIssuerId);
+        return criConfig.getComponentId();
     }
 
     public String getAllowedSharedAttributes(String credentialIssuerId) {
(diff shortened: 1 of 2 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -261,11 +261,8 @@ public String getActiveConnection(String credentialIssuerId) {
...
 
     public String getComponentId(String credentialIssuerId) {
-        String activeConnection = getActiveConnection(credentialIssuerId);
-        final String pathTemplate =
-                ConfigurationVariable.CREDENTIAL_ISSUERS.getPath()
-                        + "/%s/connections/%s/componentId";
-        return getSsmParameterWithOverride(pathTemplate, credentialIssuerId, activeConnection);
+        var criConfig = getOauthCriActiveConnectionConfig(credentialIssuerId);
+        return criConfig.getComponentId();
     }
```
Joe-Edwards-GDS, line 266: We could consider removing this method entirely and inline this logic, but this is easier :)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
```diff
@@ -281,51 +281,18 @@ void getCriConfigShouldReturnACriConfig() {
...
-            when(ssmProvider.get("/test/core/credentialIssuers/address/activeConnection"))
-                    .thenReturn("stub");
-            when(ssmProvider.get(
-                            "/test/core/credentialIssuers/address/connections/stub/componentId"))
-                    .thenReturn(testComponentId);
-            assertEquals(testComponentId, configService.getComponentId(testCredentialIssuerId));
-        }
 
-        @Test
-        void shouldLookForFeatureSetOverrideOfComponentIdOnActiveConfiguration() {
```
Joe-Edwards-GDS, line 295: This is an interesting observation - it's now a bit trickier for feature sets to override specific parts of the CRI config, but it's not been an issue so far.

I can't think of a scenario where overriding just the component ID is actually useful (vs introducing a new connection config).

lwebb-iig (approved):
LGTM - I can't think of a scenario where we'd want to override the component id with a feature set either. Is this worth documenting anywhere?

PR #1553, opened 2024-01-19
PYIC-4089: Send f2f reset identity confirmation email

shivanshuit914:
## Proposed changes

- Send f2f reset identity confirmation email

### What changed

-  Sending f2f reset identity confirmation email to user when initiated buy user.

### Why did it change

- To notify user

### Issue tracking

- [PYIC-4089](https://govukverify.atlassian.net/browse/PYIC-4089)

[PYIC-4089]: https://govukverify.atlassian.net/browse/PYIC-4089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
```diff
@@ -112,7 +114,16 @@ public Map<String, Object> handleRequest(ProcessRequest event, Context context)
 
             // Make sure we do this before deleting the credentials!
             String userName = null;
-            if (isUserInitiated) {
+            CriResponseItem f2fRequest = criResponseService.getFaceToFaceRequest(userId);
+
+            if (isUserInitiated && f2fRequest != null) {
+                VcStoreItem credential =
+                        verifiableCredentialService.getVcStoreItem(
+                                clientOAuthSessionItem.getUserId(), FRAUD_CRI);
```
DanCorderIPV: Why limit this to the fraud VC for F2F?

lambdas/reset-identity/src/test/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandlerTest.java
```diff
@@ -9,22 +9,21 @@
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import uk.gov.di.ipv.core.library.auditing.AuditEvent;
-import uk.gov.di.ipv.core.library.domain.JourneyResponse;
-import uk.gov.di.ipv.core.library.domain.ProcessRequest;
+import uk.gov.di.ipv.core.library.domain.*;
```
DanCorderIPV: I don't think we allow `*` imports?

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -38,6 +38,15 @@ public EmailService(ConfigService configService, NotificationClient notification
 
     public void sendUserTriggeredIdentityResetConfirmation(
             String userEmailAddress, String fullName) {
+        sendUserTriggeredIdentityResetConfirmation(
```
DanCorderIPV: I wouldn't combine these like this. At the moment they have the same personalisation, but there's no guarantee things will stay that way. I'd copy `sendUserTriggeredIdentityResetConfirmation` and just change the template parameter.

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -46,14 +55,20 @@ public void sendUserTriggeredIdentityResetConfirmation(
                         "Attempting to send user triggered identity reset confirmation email"));
         // This template ID can vary between production and the other environments, so it can't be
         // hardcoded
```
DanCorderIPV: If you do leave things like they are now, then this comment doesn't make sense any more as the template IDs are being hardcoded in the other methods.

lambdas/reset-identity/src/test/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandlerTest.java
DanCorderIPV: and in other files too

DanCorderIPV (changes requested):

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
shivanshuit914: No particular reason thought we can get user full name from fraud CRI as we have verified fraud score, I was thinking to use F2F one but in this journey F2F will be pending so we can't get users details from evidence. Do you suggest any other CRI that we should use here ?

DanCorderIPV: The code that's already there will just check all VCs to find a name I think? If the fraud VC isn't special then can you just re-use what's already there?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
```diff
@@ -26,7 +26,9 @@ public enum ConfigurationVariable {
     RETURN_CODES_NON_CI_BREACHING_P0("self/returnCodes/nonCiBreachingP0"),
     GOV_UK_NOTIFY_API_KEY("self/gov-uk-notify/api-key"),
     GOV_UK_NOTIFY_TEMPLATE_ID_USER_TRIGGERED_IDENTITY_RESET_CONFIRMATION(
-            "self/govUkNotify/emailTemplates/UserTriggeredIdentityResetConfirmation");
+            "self/govUkNotify/emailTemplates/UserTriggeredIdentityResetConfirmation"),
+    GOV_UK_NOTIFY_TEMPLATE_ID_F2F_USER_TRIGGERED_IDENTITY_RESET_CONFIRMATION(
+            "self/govUkNotify/emailTemplates/FaceToFaceIdentityResetConfirmation");
```
DanCorderIPV: Also could you change this to `self/govUkNotify/emailTemplates/UserTriggeredIdentityResetConfirmationF2f` to match PR 4090 please?

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
shivanshuit914: yes 👍

shivanshuit914: I tried to use `findIdentityClaim` and found `isEvidenceVc` return false for all VcItems. may be because verification is still pending.

libs/email-service/src/main/java/uk/gov/di/ipv/core/library/service/EmailService.java
```diff
@@ -54,6 +54,24 @@ public void sendUserTriggeredIdentityResetConfirmation(
...
     }
 
+    public void sendUserTriggeredF2FIdentityResetConfirmation(
+            String userEmailAddress, String fullName) {
+        Map<String, Object> templateParameters = new HashMap<>();
+        templateParameters.put("fullName", fullName);
+
+        LOGGER.info(
+                LogHelper.buildLogMessage(
+                        "Attempting to send user triggered identity reset confirmation email"));
```
DanCorderIPV: Please update to include F2F

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
DanCorderIPV: I've chatted to Joe about this as using the Fraud VC seems fragile to future journey map changes.
Could you please instead add an overload of `UserIdentityService.findIdentityClaim()` that takes a flag to indicate whether to only use evidenceVCs or all VCs (and default the flag to the current behaviour).
Then you can update the old `ResetIdentityHandler.getUserName()` to `ResetIdentityHandler.getUnconfirmedUserName()` and use that for both emails.

DanCorderIPV (changes requested):

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
```diff
@@ -112,7 +114,16 @@ public Map<String, Object> handleRequest(ProcessRequest event, Context context)
 
             // Make sure we do this before deleting the credentials!
             String userName = null;
-            if (isUserInitiated) {
+            CriResponseItem f2fRequest = criResponseService.getFaceToFaceRequest(userId);
```
blakeyp, line 116: Just thinking do we _need_ to retrieve the pending F2F request here? Could we use `lambdaInput` in the journey map with a specific state for this route (from the F2F pending page)? something like
```
RESET_IDENTITY_USER_INITIATED_F2F:
  response:
    type: process
    lambda: reset-identity
    lambdaInput:
      isUserInitiatedF2f: true
  events:
    next:
      ...
```

shivanshuit914: Yes, need to retrieve the pending F2F request to make sure we sending correct email f2f or without f2f.

shivanshuit914: I tried this, it requires we need to add both lambda input all places, its not optional  like

```
RESET_IDENTITY_USER_INITIATED_F2F:
  response:
    type: process
    lambda: reset-identity
    lambdaInput:
      isUserInitiatedF2f: true
      isUserInitiated: false
```

blakeyp: Hmm ok that's not ideal, let's stick with the current approach

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -160,13 +160,30 @@ public UserIdentity generateUserIdentity(
...
+
+    private Optional<IdentityClaim> findIdentityClaimFromVcStoreItems(
+            List<VcStoreItem> vcStoreItems, boolean checkEvidence)
+            throws HttpResponseExceptionWithErrorBody, CredentialParseException {
         List<IdentityClaim> identityClaims = new ArrayList<>();
         for (VcStoreItem vcStoreItem : vcStoreItems) {
             try {
-                if (isEvidenceVc(vcStoreItem)
-                        && VcHelper.isSuccessfulVc(SignedJWT.parse(vcStoreItem.getCredential()))) {
+                if ((checkEvidence
```
DanCorderIPV: This seems simpler?
```
if ((!checkEvidence || isEvidenceVc(vcStoreItem))
    && VcHelper.isSuccessfulVc(SignedJWT.parse(vcStoreItem.getCredential()))) {
```

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
```diff
@@ -166,10 +186,10 @@ private void sendIpvVcResetAuditEvent(
 
     // Try to get the user's name from their VCs. It's not the end of the world if this fails so
     // just return null in that case.
-    private String getUserName(List<VcStoreItem> credentials) {
+    private String getUserName(List<VcStoreItem> credentials, boolean checkEvidence) {
```
DanCorderIPV: As before, please rename this to `getUnconfirmedUserName()` and remove the `checkEvidence` flag from here too. This method should always pass `false` to `findIdentityClaim()`

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
```diff
@@ -132,6 +142,16 @@ public Map<String, Object> handleRequest(ProcessRequest event, Context context)
                         ipvSessionItem.getEmailAddress(), userName);
             }
 
+            if (isUserInitiated && f2fRequest != null) {
+                sendIpvVcResetAuditEvent(event, userId, govukSigninJourneyId);
+
+                // Create a new email service for each request so that we don't risk using stale
+                // configuration.
+                final EmailService emailService = emailServiceFactory.getEmailService();
+                emailService.sendUserTriggeredF2FIdentityResetConfirmation(
```
DanCorderIPV: This is the only different line between these two block, please combine them

lambdas/reset-identity/src/main/java/uk/gov/di/ipv/core/resetidentity/ResetIdentityHandler.java
```diff
@@ -112,17 +113,26 @@ public Map<String, Object> handleRequest(ProcessRequest event, Context context)
...
+
+            if (isUserInitiated && f2fRequest != null) {
                 List<VcStoreItem> credentials =
                         verifiableCredentialService.getVcStoreItems(
                                 clientOAuthSessionItem.getUserId());
-                userName = getUserName(credentials);
+                userName = getUserName(credentials, false);
+            }
+
+            if (isUserInitiated && f2fRequest == null) {
```
DanCorderIPV: Once you update `getUserName()` there will be no difference here between F2F and non-F2F so these blocks can be combined.

DanCorderIPV (changes requested):

DanCorderIPV (approved):

PR #1563, opened 2024-01-22
PYIC-4459 POC for technical error journey split

Joe-Edwards-GDS:
Added a separate 'technical error' journey to demonstrate separation of journey map. This does not touch the main entry point, but shows how to traverse between journeys.


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -323,10 +320,15 @@ PYI_CRI_ESCAPE_NO_F2F:
 
 ERROR:
   response:
-    type: error
-    pageId: pyi-technical
-    statusCode: 500
-  parent: END_JOURNEY
+    type: journey
+    journeyType: TECHNICAL_ERROR
+    initialState: ERROR
```
Joe-Edwards-GDS: We could instead make the journey transition part of the `events` configuration, so rather than defining this explicit state for the journey transition, we can go 'directly'. e.g.:
```
SOME_STATE:
  response:
    type: page
    pageId: some-page
  events:
    error:
      targetState: ERROR
```
Reduces the number of states, but adds noise into the events, and will need a bit more refactoring of the state machine.

I might give it a try and see how it compares...

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -123,6 +127,14 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
                 sendMitigationStartAuditEvent(ipvSessionId, ipAddress, clientOAuthSessionItem);
             }
 
+            // If we have a new journey response, follow it
+            while (stepResponse instanceof JourneyStepResponse journeyResponse) {
+                ipvSessionItem.setJourneyType(journeyResponse.getJourneyType());
+                ipvSessionItem.setUserState(journeyResponse.getInitialState());
+                ipvSessionService.updateIpvSession(ipvSessionItem);
+                stepResponse = executeJourneyEvent(NEXT_EVENT, ipvSessionItem);
```
Joe-Edwards-GDS: We could also return this as an event and get the step function to re-run this lambda, it just feels needlessly inefficient.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/JourneyStepResponse.java
Joe-Edwards-GDS, line 1: This was never used, so I've repurposed...

lambdas/process-journey-event/src/main/resources/statemachine/test/technical-error.yaml
Joe-Edwards-GDS, line 1: (This is a bit annoying - because `IpvJourneyTypes` is an enum, we have to use the same types in tests)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: Agreed - I think just dive back in here. Although it means that all initial states in a journey will have to handle a `next` event. One option could be to have the JourneyStepResponse define the event to call on the initial state, which might make things a bit more flexible - effectively being able to drop in to journeys at arbitary points? I've not thought very hard about this so maybe it's not that easy.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -154,8 +166,10 @@ private StepResponse executeJourneyEvent(String journeyEvent, IpvSessionItem ipv
                                 ipvSessionItem.getJourneyType()));
             }
             LOGGER.info(
-                    "Found state machine for journey type: {}",
-                    ipvSessionItem.getJourneyType().getValue());
+                    LogHelper.buildLogMessage(
+                            String.format(
+                                    "Found state machine for journey type: %s",
+                                    ipvSessionItem.getJourneyType().name())));
```
Wynndow: This log message bugs me when I'm looking at journeys in the logs, as it gets logged so oftern. Maybe now it'll actually be useful 😄

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/JourneyStepResponse.java
Wynndow: This confused me quite a lot. I'm sure it used to be used, but got phased out in favour of other step responses. I think the use of the word journey is over used - especially as we (possibly needlessly) prepend all our journey events with `/journey/`.
I'd be tempted to rename it to something like `JourneyTypeTransitionStepResponse`. Or maybe just `JourneyTransitionStepResponse`?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: This was what I was thinking about when looking through this - would be interesting to see how this looks. Saves a potential round trip through the process-journey-event lambda too

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/technical-error.yaml
```diff
@@ -0,0 +1,48 @@
...
+    pageId: pyi-technical
+    statusCode: 500
+  events:
+    next:
+      targetState: END
+
+END:
+  response:
+    type: process
+    lambda: build-client-oauth-response
```
Wynndow: Interseting that we now have multiple `END` states. I don't supposed there will be too much repetition like this, but I wonder if there could be a journey map for shared states like this. Potentially some "page" states too.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/states/State.java
```diff
@@ -14,8 +14,4 @@
 public interface State {
     State transition(String eventName, String startState, JourneyContext journeyContext)
             throws UnknownEventException, UnknownStateException;
-
-    String getName();
-
-    void setName(String name);
```
Joe-Edwards-GDS, line 20: (these are actually never used polymorphically - they're only used when we know the subclass we have)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: Done :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: As discussed, there are some benefits to separating the two - e.g. ability to use feature flags etc.

The hardcoded `next` event is a bit unsatisfactory, but not much worse than what we have currently.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/JourneyStepResponse.java
Joe-Edwards-GDS: Gone :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Downgraded to `debug` (not sure if that does anything though)

PR #1564, opened 2024-01-22
PYIC-4540: Set VOT as enum on ipv session item

Wynndow:
### What changed

Set VOT as enum on ipv session item

### Why did it change

This updates the vot field in the IpvSessionItem to be a Vot enum. This will make a bunch of code cleaner, and we can use an annotation to do the conversion for storing in Dynamo.

### Issue tracking

- [PYIC-4540](https://govukverify.atlassian.net/browse/PYIC-4540)

[PYIC-4540]: https://govukverify.atlassian.net/browse/PYIC-4540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -2,8 +2,6 @@ package uk.gov.di.ipv.core.checkexistingidentity;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
-import com.nimbusds.jose.shaded.json.JSONArray;
-import com.nimbusds.jose.shaded.json.JSONObject;
 import com.nimbusds.jwt.SignedJWT;
 import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
@@ -52,7 +50,6 @@ import uk.gov.di.ipv.core.library.verifiablecredential.helpers.VcHelper;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
 
 import java.text.ParseException;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
```diff
@@ -17,7 +18,7 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class AuditExtensionsUserIdentity implements AuditExtensions {
-    @JsonProperty String levelOfConfidence;
+    @JsonProperty Vot levelOfConfidence;
```
Joe-Edwards-GDS: (I assume this 'just works' for serialization)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -28,7 +30,7 @@ public class IpvSessionItem implements DynamodbItem {
     private AccessTokenMetadata accessTokenMetadata;
     private String errorCode;
     private String errorDescription;
-    private String vot;
+    private @DynamoDBTypeConvertedEnum Vot vot;
```
Joe-Edwards-GDS: I know I told you to use it, but do we actually need this annotation?

I've noticed that `journeyType` is also enum-valued and seems to 'just work', so maybe it's not needed!

Joe-Edwards-GDS (commented):
This looks very reasonable to me, and I can't think of any reason we'd have any issues with enum parsing from the records.

If you wanted to be extra careful you could add a comment to the `Vot.java` saying that you need to wary of removing values that might be in use by active sessions.

This will (again :( ) clash with @rpayal 's open PR

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsUserIdentity.java
Wynndow: Yep - I've checked the audit event logs and it's happy.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Wynndow: I did sort of wonder - while playing around looking at how the enum was serialized. I'll try taking it out and see what happens.

rpayal:
> This looks very reasonable to me, and I can't think of any reason we'd have any issues with enum parsing from the records.
> 
> If you wanted to be extra careful you could add a comment to the `Vot.java` saying that you need to wary of removing values that might be in use by active sessions.
> 
> This will (again :( ) clash with @rpayal 's open PR

NP, yeah following this PR, will merge to main and resolve conflicts tomorrow.

Wynndow:
> > This looks very reasonable to me, and I can't think of any reason we'd have any issues with enum parsing from the records.
> > If you wanted to be extra careful you could add a comment to the `Vot.java` saying that you need to wary of removing values that might be in use by active sessions.
> > This will (again :( ) clash with @rpayal 's open PR
> 
> NP, yeah following this PR, will merge to main and resolve conflicts tomorrow.

I'm holding off on merging this till you're done Rohit - there's no rush on this at all and it'll probably be an easier rebase for me

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Wynndow: Yep, it's happy. Don't need the convertor.

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):
You can set up `pre-commit` to catch spotless errors when you commit - but it does make committing slightly slower.

PR #1567, opened 2024-01-24
PYIC-4571: Archive vcs boilerplate

lwebb-iig:
Adds boilerplate for the archive vcs lambda

### What changed

- Template and empty lambda added 

### Why did it change

- Pre-requisite for https://govukverify.atlassian.net/browse/PYIC-4555

### Issue tracking

- [PYIC-4571](https://govukverify.atlassian.net/browse/PYIC-4571)

### Other considerations

[PYIC-4571]: https://govukverify.atlassian.net/browse/PYIC-4571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/archive-vcs/src/main/resources/IpvLambdaJsonLayout.json
```diff
@@ -0,0 +1,88 @@
+{
```
MikeCollingwood, line 1: Is this identical for the other lambda replaying cimit ci's?

lambdas/archive-vcs/src/test/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandlerTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.di.ipv.core.library.service.ConfigService;
+import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@ExtendWith(MockitoExtension.class)
+class ReplayCimitVcsHandlerTest {
```
MikeCollingwood:
```suggestion
class ArchiveVcsHandlerTest {
```

MikeCollingwood (dismissed):
Think this can be merged after a small change

lambdas/archive-vcs/src/main/resources/IpvLambdaJsonLayout.json
lwebb-iig: @MikeCollingwood I believe these are identical/very similar for all of our lambdas and are put alongside the lambda code for the purposes of bundling the lambda artifacts for AWS

Joe-Edwards-GDS (approved):
All seems reasonable so far, although idk if there's any value in merging until you've got a bit more meat to it.

A couple of thoughts:
- Is 'ArchiveVcs' the right name? I think `RevokeVcs` might reflect the purpose better - but it's not the end of the world if it's a pain to change.
- Should we have a readme, so future devs know what this is for? (In general, rather than this specific case)

lwebb-iig:
> * Is 'ArchiveVcs' the right name? I think `RevokeVcs` might reflect the purpose better - but it's not the end of the world if it's a pain to change.

Revoke VCs makes sense - I think I went with archive as we had the concept of the new "archive" table but revoke probably makes more sense (and obvious that we're deleting things from the main table). Easy to change in a future PR (cloud-formation should gracefully handle the update).

> * Should we have a readme, so future devs know what this is for? (In general, rather than this specific case)

Yep, good idea - I think this was more of a starting point / branch that @MikeCollingwood & I could work from, but I'll bring this forward & add this to the list of tasks we came up with for this work.

PR #1568, opened 2024-01-24
Pyic-4123: Add CIC CRI consumer contract tests

shivanshuit914:
## Proposed changes

- Add CIC CRI consumer contract tests

### What changed

- Added CIC CRI consumer contract tests

### Why did it change

 - To build up contract test coverage for CIC CRI service

### Issue tracking

- [PYIC-4123](https://govukverify.atlassian.net/browse/PYIC-4123)

[PYIC-4123]: https://govukverify.atlassian.net/browse/PYIC-4123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
```diff
@@ -0,0 +1,428 @@
...
+                .method("POST")
+                .headers("x-api-key", PRIVATE_API_KEY, "Authorization", "Bearer dummyAccessToken")
+                .willRespondWith()
+                .status(200)
+                .body(
+                        new PactJwtIgnoreSignatureBodyBuilder(
+                                VALID_VC_HEADER, VALID_VC_BODY, VALID_VC_SIGNATURE))
+                .toPact();
+    }
+
```
RogVanGDS, line 352: Could we possibly group the PACT method with it's corresponding test implementation? I think it's slightly easier to follow the file 🙂

RogVanGDS:
Do we make a request to the /authorize endpoint for CIC? I think we should be including PACT tests for it if we do

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
shivanshuit914: Done thanks

RogVanGDS:
I'm assuming based on this that we can't get back a failed VC from CIC, which feels correct, but just wanted to double check

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
```diff
@@ -0,0 +1,503 @@
...
+                        CriApiException.class,
+                        () ->
+                                underTest.fetchVerifiableCredential(
+                                        new BearerAccessToken("dummyAccessToken"),
+                                        getCallbackRequest("dummyAuthCode", credentialIssuerConfig),
+                                        getCriOAuthSessionItem()));
+        // Assert
+        assertEquals(
+                "Failed to get credential from issuer", exception.getErrorResponse().getMessage());
+        assertEquals(500, exception.getHttpStatusCode());
```
blakeyp: I don't really follow this scenario - the method is `inValidRequestReturns400` but this is asserting a 500 response. Shouldn't this be a test to assert we get a 401 back if we call /credential with an invalid access token?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    @Pact(provider = "CicCriProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact validRequestReturnsValidAccessToken(PactDslWithProvider builder) {
```
blakeyp, line 145: Could we have the token endpoint tests before the credential ones, in line with the other contract tests?

blakeyp (approved):

blakeyp (changes requested):
clicked approve by mistake

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
shivanshuit914: Yes so we throw the exception in core lambda code when its 400 from CRI and that results in 500 error code.

blakeyp: I see, but I don't think we really expect a scenario where this endpoint returns us a 400 with an 'invalid' VC? the existing contract tests seem to just be checking the invalid access token scenario (which I think we're missing here?)

shivanshuit914: hm I think the name of the test is bit misleading, I am trying to test when we send cic cri request without birhtday or name it should return 400. may be I need to rename test to something like `fetchVerifiableCredential_whenCalledAgainstCicCri_withInvalidRequest_throwsAnException`.

DanCorderIPV: Should the CRI be able to send us a VC without a birthday?
- If they can then I guess it's not an invalid VC and we should have a test to cover it (but why would they also send a 400 return code if it's valid?)
- If they can't then we're testing something that should never happen - how would the CIC team write a test to generate a broken VC if their code doesn't allow it?

As Patrick says, if they do send us a 400 response (bad request) then they probably won't send us a VC at all because they couldn't understand the request.
Why do you expect them to send a 400 in response to the request in the pact (which looks fine)?

shivanshuit914: Ok, yes makes sense. we never going to have that scenario. I have removed this test :)

blakeyp: Thanks @shivanshuit914. I think we're still missing a test case we do want, for invalid access token on the `/credential` endpoint - something like this: https://github.com/govuk-one-login/ipv-core-back/blob/53c001d5fbe63dde0f385389148480e41a99743b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/passportCri/ContractTest.java#L326 ?

shivanshuit914: Yes, added it thanks :)

blakeyp (approved):

PR #1574, opened 2024-01-25
BAU: Update baseline secrets

Wynndow:
### What changed

Update baseline secrets

### Why did it change

A new test fixture was [added recently](https://github.com/govuk-one-login/ipv-core-back/commit/f460b83c1e9c2b883ff1872625d73931231eac8d#diff-38e3b85391f40008ee814e269794078e9f5fb090b2cc79f001f69e719dd63e70R55).

This needs adding to the baseline secrets to avoid it getting flagged in every PR.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -76,6 +76,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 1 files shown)


TalatAydinGDS (approved):

Joe-Edwards-GDS (approved):
The output from detect-secrets implies we can also add `// pragram: allowlist secret`, which might be a better way of ignoring a lot of these (so we don't have to update every time the line number changes).

Hopefully @TalatAydinGDS 's changes to the test fixtures will reduce the volume of these anyway.

Wynndow:
> The output from detect-secrets implies we can also add `// pragram: allowlist secret`, which might be a better way of ignoring a lot of these (so we don't have to update every time the line number changes).
> 
> Hopefully @TalatAydinGDS 's changes to the test fixtures will reduce the volume of these anyway.

That's what I originaly tried - but that also auto updates the baseline file 😅

PR #1575, opened 2024-01-25
PYIC-4601: Route users depending on reuse type 

Wynndow:
### What changed

Send users on the right journey depending on what type of vot they've achieved, and if they're currently migrating or not.

### Why did it change

[PYIC-4601: Route users depending on reuse type](https://github.com/govuk-one-login/ipv-core-back/commit/0565453d1fe86b44d16b49d02fdb1a920eccfa32)

This updates the check-existing-identity lambda to send users on the
correct journey, depending on if they are reusing an identity and if
they are in the process of migration or not.

Users reusing an operational profile that have already migrated (in a
previous session) should see the reuse page but with some extra context.
Those that are migrating in the current session should just go straigh
back to the service.

This has required some extra CRI_TICF states to achieve, due to the new
follow on routes that we didn't support before. Namely ending
immediately and showing the reuse page with context.

This has added a new CRI_TICF parent state to the joureny map to dry up
some of the existing states.

[PYIC-4601: Rename next event to ipv-gpg45-medium](https://github.com/govuk-one-login/ipv-core-back/commit/cfb545f57822b3bd027730daf6c6dd98ec9e4d20)

The next event is a little lack lustre in its description of what's
going on.

### Issue tracking

- [PYIC-4601](https://govukverify.atlassian.net/browse/PYIC-4601)

[PYIC-4601]: https://govukverify.atlassian.net/browse/PYIC-4601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -73,7 +73,7 @@ import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_ERROR_P
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_F2F_FAIL_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_FAIL_WITH_CI_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_IN_MIGRATION_REUSE_PATH;
-import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_NEXT_PATH;
+import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_IPV_GPG45_MEDIUM_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_OPERATIONAL_PROFILE_REUSE_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_PENDING_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_RESET_IDENTITY_PATH;
@@ -93,8 +93,8 @@ public class CheckExistingIdentityHandler
             new JourneyResponse(JOURNEY_IN_MIGRATION_REUSE_PATH).toObjectMap();
     private static final Map<String, Object> JOURNEY_PENDING =
             new JourneyResponse(JOURNEY_PENDING_PATH).toObjectMap();
-    private static final Map<String, Object> JOURNEY_NEXT =
-            new JourneyResponse(JOURNEY_NEXT_PATH).toObjectMap();
+    private static final Map<String, Object> JOURNEY_IPV_GPG45_MEDIUM =
+            new JourneyResponse(JOURNEY_IPV_GPG45_MEDIUM_PATH).toObjectMap();
     private static final Map<String, Object> JOURNEY_F2F_FAIL =
             new JourneyResponse(JOURNEY_F2F_FAIL_PATH).toObjectMap();
     private static final Map<String, Object> JOURNEY_RESET_IDENTITY =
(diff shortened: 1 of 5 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -365,6 +375,15 @@ private Map<String, Object> buildReuseResponse(AuditEventUser auditEventUser)
                         auditEventUser));
 
         LOGGER.info(LogHelper.buildLogMessage("Returning reuse journey"));
+
+        if (attainedVot.getProfileType() == OPERATIONAL_HMRC) {
+            // the only VC we should possibly have collected this session at this point is a
+            // migration VC
+            return vcReceivedThisSession.isEmpty()
```
Wynndow: Is this too lazy a check? At this stage in the journey the users should have only ever picked up an inherited identity VC, so this works. But should we be doing further scrutiny on it, to make sure it actually IS an inherited identity VC?

~Ignore me, this makes no sense and needs fixing.~

Actualy, ignore that. This does work. But the reuse page doesn't work for a operational reuse profile. We don't have enough details about the user for it to build correctly.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -96,6 +105,8 @@ CHECK_EXISTING_IDENTITY:
   events:
     next:
```
Wynndow, line 106: Maintaining the old `next` event out of a fear of a race condition. If the check-existing-identity lambda is deployed before the process-journey-event lambda we could get some unknown event errors. Maybe this isn't something to worry about too much as the timing might be very small.

Joe-Edwards-GDS: Yeah, I think it'd be a small enough window (<1s) that we could get away without it, but if we can easily handle it then we may as well.

TBD exactly how the blue/green deployments will work for core-back, but there's a chance it might avoid this anyway in future (by pinning lambda versions to each step function version).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
```diff
@@ -371,88 +397,82 @@ CRI_TICF_BEFORE_SUCCESS:
...
-      targetState: PYI_NO_MATCH
-    fail-with-ci:
-      targetState: PYI_NO_MATCH
-    error:
-      targetState: ERROR
+
+CRI_TICF_BEFORE_OPERATIONAL_PROFILE_REUSE:
+  response:
+    type: process
+    lambda: call-ticf-cri
```
Joe-Edwards-GDS, line 420: I briefly tried making `response` part of the parent as well, but it needs a bit of fiddling to get it to work (including in the journey map). Maybe something to look at after the journey split

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -1002,4 +1036,17 @@ private CriResponseItem createCriErrorResponseStoreItem(Instant dateCreated) {
         criResponseItem.setStatus(CriResponseService.STATUS_ERROR);
         return criResponseItem;
     }
+
+    private static SignedJWT createOperationalProfileVc(Vot vot) throws Exception {
```
Joe-Edwards-GDS, line 1040: Do you think this should go into TestFixtures (alongside other JWT creation methods)?

Might also overlap with @TalatAydinGDS 's work

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -365,6 +375,15 @@ private Map<String, Object> buildReuseResponse(AuditEventUser auditEventUser)
                         auditEventUser));
 
         LOGGER.info(LogHelper.buildLogMessage("Returning reuse journey"));
+
+        if (attainedVot.getProfileType() == OPERATIONAL_HMRC) {
+            // the only VC we should possibly have collected this session at this point is a
+            // migration VC
```
Joe-Edwards-GDS, line 381: This assumption makes me a _little_ nervous, but happy for now.

If we merge session initialisation and check existing identity, then we might be able to do this directly instead :)

Joe-Edwards-GDS (approved):

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
Wynndow: I don’t think so. Maybe a more fully fledged version but these are very much MVP’s to test the functionality

PR #1578, opened 2024-01-26
PYIC-4572: lambdas to revoke and restore vcs

MikeCollingwood:
### What changed

- Added revoking and restoring lambdas

### Why did it change

- We need to archive vcs as we revoke them as part of [PYIC-4555](https://govukverify.atlassian.net/jira/software/c/projects/PYIC/boards/431?selectedIssue=PYIC-4555)

### Issue tracking

- [PYIC-4572](https://govukverify.atlassian.net/browse/PYIC-4572)

### Testing

- Having tested this in the dev env:
Screenshot 2024-01-30 at 13 00 51
Screenshot 2024-01-30 at 13 01 00
Screenshot 2024-01-30 at 13 02 40
Screenshot 2024-01-30 at 13 04 48

[PYIC-4555]: https://govukverify.atlassian.net/browse/PYIC-4555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-4572]: https://govukverify.atlassian.net/browse/PYIC-4572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/archive-vcs/src/main/java/README.md
Joe-Edwards-GDS, line 1: This is probably the wrong place for this README - I think t should probably just be in `/lambdas/archive-vcs`?

lambdas/archive-vcs/src/main/java/README.md
```diff
@@ -0,0 +1,12 @@
+# ArchiveVcsHandler
+
+To run this lambda, create an appropriate payload including userIds with kbv vcs in the the appropriate environment.
```
Joe-Edwards-GDS: Might be useful to say what this actually _does_, but perhaps just link to the ticket to avoid duplication.

We should also be a little bit careful about what information we put here, since it's a public repo :)

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -2,42 +2,100 @@
...
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            List<String> userIds = mapper.readValue(inputStream, RevokeRequest.class).getUserIds();
+            LOGGER.info("Revoking {} VCs", userIds.size());
+
+            List<List<String>> batchedRequest = ListHelper.getBatches(userIds, 100);
+            for (int i = 0; i < batchedRequest.size(); i++) {
+                LOGGER.info("Processing batch {} of {}", i, batchedRequest.size());
+                revoke(batchedRequest.get(i));
+            }
```
Joe-Edwards-GDS: I'm not sure there's a massive benefit in batching these here, unless we're actually going to do batch requests to dynamodb (which is possible, but probably not worth it unless it's unusably slow).

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -2,42 +2,100 @@
...
+            LOGGER.info("Completed revoking {} VCs", userIds.size());
+        } catch (IOException e) {
+            throw new FailedVcRevokeException("Failed to map the revoke request", e);
+        }
+    }
+
+    private void revoke(List<String> userIds) {
+        for (String userId : userIds) revoke(userId);
```
Joe-Edwards-GDS: We probably want to move the info logs here, as well as maybe wrapping in a try/catch (so an individual failure doesn't stop the entire run)?

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -2,42 +2,100 @@
...
+            LOGGER.info("Revoking {} VCs", userIds.size());
+
+            List<List<String>> batchedRequest = ListHelper.getBatches(userIds, 100);
+            for (int i = 0; i < batchedRequest.size(); i++) {
+                LOGGER.info("Processing batch {} of {}", i, batchedRequest.size());
+                revoke(batchedRequest.get(i));
+            }
+            LOGGER.info("Completed revoking {} VCs", userIds.size());
+        } catch (IOException e) {
+            throw new FailedVcRevokeException("Failed to map the revoke request", e);
```
Joe-Edwards-GDS: (there's probably limited value in rethrowing here since it's never caught - it'll just fail the lambda anyway, but doesn't matter very much)

lambdas/archive-vcs/src/test/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandlerTest.java
```diff
@@ -5,21 +5,54 @@
...
+import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
 import uk.gov.di.ipv.core.library.service.ConfigService;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
 
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import java.io.IOException;
+import java.io.InputStream;
+import java.time.Instant;
+
+import static org.mockito.Mockito.*;
```
MikeCollingwood: wildcard import

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -2,42 +2,101 @@
...
 import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.List;
 
 @ExcludeFromGeneratedCoverageReport
 @SuppressWarnings("unused") // Temporarily disable to pass sonarqube
 public class ArchiveVcsHandler implements RequestStreamHandler {
```
MikeCollingwood: Probably do want to rename this to RevokeVcsHandler

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/domain/RevokeRequest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RevokeRequest {
+    private List<String> userIds;
```
Joe-Edwards-GDS: You might want to make these `userId` + `criId` pairs to make it a bit more flexible in future?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        LOGGER.info("Revoking {} VCs", userIdCriIdPairs.size());
+
+        revoke(userIdCriIdPairs);
+        LOGGER.info("Finished attempt to revoke {} VCs", userIdCriIdPairs.size());
+    }
+
+    private void revoke(List<UserIdCriIdPair> userIdCriIdPairs) {
+        for (int i = 0; i < userIdCriIdPairs.size(); i++) {
+            var userIdCriIdPair = userIdCriIdPairs.get(i);
+            LOGGER.info("Processing VC {} / {}", i, userIdCriIdPairs.size());
```
Joe-Edwards-GDS: (`i` is zero-indexed, so this will start off with `Processing VC 0 / 1234`)

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+
+            try {
+                revoke(userIdCriIdPair);
+                LOGGER.info("Successfully revoked VC.");
```
Joe-Edwards-GDS: Maybe include the index here as well (and below)?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            } catch (Exception e) {
+                LOGGER.error("Failed to revoke VC.");
```
Joe-Edwards-GDS: Can we log the exception as well? This won't give us any information to debug why it failed

lambdas/revoke-vcs/README.md
```diff
@@ -0,0 +1,14 @@
+# RevokeVcsHandler
+
+What this lambda is for and does: https://govukverify.atlassian.net/browse/PYIC-4555
+
+To run this lambda, create an appropriate payload including userIds with kbv vcs in the the appropriate environment.
```
Joe-Edwards-GDS: Here (and elsewhere) there are a few mentions of KBV specifically that we can probably make generic

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            dataStore.create(vc);
+
+            // Submit audit event
+            // ...
+
+            // Delete VC from the main table
+            // ...
+
+        } else {
+            LOGGER.error("VC cannot be found");
```
Joe-Edwards-GDS: Note that the logic above will still log a success in this case - that might be fine, but I think we want to make it easy to count all the different cases.

(e.g. it might be preferable to log _either_ 'successfully deleted', 'cannot be found', or 'unexpected error occurred' for each record)

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
MikeCollingwood: I was concerned the exception might include a userId...

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
MikeCollingwood: The audit event piece of work does more with this because we then send an audit event

MikeCollingwood: This log is replaced with an exception that is caught and then we log the progress and send an audit event

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
Joe-Edwards-GDS: Hmm, that's true, but I think it's reasonable to see unexpected details otherwise we have no ability to debug. If there are particular cases you're worried about, then we could catch them explicitly.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsReceivedVcEvidence.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.di.ipv.core.library.auditing.extension;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@Getter
+public class AuditExtensionsReceivedVcEvidence extends AuditExtensionsVcEvidence {
```
MikeCollingwood: AuditExtensionsReceivedVcEvidence (renamed from AuditExtensionsVcEvidence) extends the new AuditExtensionsVcEvidence because it's specific to received vcs. The new AuditExtensionsVcEvidence can then be used for the extension to the IPV_VC_REVOKED audit event

MikeCollingwood:
Testing for the case that a VC doesn't exist in the db:

Screenshot 2024-01-30 at 13 07 01
Screenshot 2024-01-30 at 13 07 43

MikeCollingwood:
Then crudely replacing the send audit event method with a thrown exception:
`throw new SqsException("sendVcRevokedAuditEvent example exception"); // auditService.sendAuditEvent(auditEvent);`
We can test the following cases:

1. There is a VC but sending audit event fails
Screenshot 2024-01-30 at 13 19 14

2. There is not a VC and sending audit event fails
Screenshot 2024-01-30 at 13 17 07

MikeCollingwood:
Q: Do we want to create a service in which facilitates lambdas like these? The restore vcs lambda will look quite a lot like the revoke lambda - so maybe it's worth creating a common service?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,204 @@
...
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "Successfully revoked VC (%s / %s)", i + 1, numberOfVcs)));
+            } catch (Exception e) {
+                LOGGER.error(
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "Unexpected error occurred (%s / %s): %s",
+                                        i + 1, numberOfVcs, e.getMessage())));
```
Joe-Edwards-GDS: Can you use `buildErrorMessage` here, and pass the exception directly?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -0,0 +1,204 @@
...
+                new AuditEvent(
+                        AuditEventTypes.IPV_VC_REVOKED,
+                        configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                        auditEventUser,
+                        auditExtensions);
+        LOGGER.debug(
+                LogHelper.buildLogMessage(
+                        String.format(
+                                "Sending audit event IPV_VC_REVOKED (%s / %s)",
+                                i + 1, numberOfVcs)));
```
Joe-Edwards-GDS: We can probably remove this once you're happy. If you need, I think there's a dummy lambda that logs audit events in build (and we can get the QAs to help test it properly in staging).

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionErrorMessage.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.di.ipv.core.library.auditing.extension;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@Getter
+public class AuditExtensionErrorMessage implements AuditExtensions {
```
Joe-Edwards-GDS: I'm not sure about including the error message here - the audit log is to track what has happened from an audit perspective (so this event says 'we tried and failed to revoke a VC'), and if we include arbitrary exception messages we start having to make sure it doesn't contain anything bad.

I think we probably _should_ include the CRI ID though. This is already a known field type as `extensions.cri` if you want to reuse it.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsReceivedVcEvidence.java
Joe-Edwards-GDS: Is it specific just because of the `successful` field?

One other option would be to make the `successful` property a `Boolean` (capital B) and make it `@JsonInclude(NON_NULL)` - that way it'll be omitted if set to null.

See `AuditExtensionsReproveIdentity` for an example

Joe-Edwards-GDS (commented):
Looks basically there I think :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -72,6 +73,19 @@ public void create(T item) {
         table.putItem(item);
     }
 
+    public void createIfEmpty(T item) {
```
Joe-Edwards-GDS: Minor, but can we align this with the dynamodb terminology - `createIfNotExists`

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/VerifiableCredentialService.java
```diff
@@ -58,6 +60,21 @@ public void persistUserCredentials(
         }
     }
 
+    public void persistUserCredentialsIfEmpty(
```
Joe-Edwards-GDS: Ditto

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/OverwriteAvoidedException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.di.ipv.core.library.exceptions;
+
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+public class OverwriteAvoidedException extends Exception {
```
Joe-Edwards-GDS: I'd prefer `CredentialAlreadyExistsException` or something similar - feels like a more accurate description of the 'exception'

Joe-Edwards-GDS (commented):
I haven't gone through in fine detail but looks sensible to me, I'll ask security if they want to have a look, and we can run some tests (if you haven't already!)

Joe-Edwards-GDS (dismissed):

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -87,21 +90,25 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
                             String.format(
                                     "Finished attempt to revoke %s VCs.",
                                     userIdCriIdPairs.size())));
+
+            objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
+            outputStream.write(objectMapper.writeValueAsString(results).getBytes());
         } catch (SqsException e) {
             LOGGER.error(
                     LogHelper.buildErrorMessage(
                             "Stopped revoking VCs because of failure to send audit event.", e));
```
Joe-Edwards-GDS, line 103: Do we want to return anything in this case, or perhaps rethrow the exception?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/domain/RevokeVcsResult.java
Joe-Edwards-GDS, line 1: It might be terser to make the structure more like:
```
{
  "successes": [{ "userId": "<blah>", "criId": "kbv" }, ...],
  "failures": [{ "userId": "<blah>", "criId": "kbv", "error": "VC missing" }, ...]
}
```
but not that important

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -87,21 +90,25 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
-            revoke(userIdCriIdPairs);
+            var results = revoke(userIdCriIdPairs);
             LOGGER.info(
                     LogHelper.buildLogMessage(
                             String.format(
                                     "Finished attempt to revoke %s VCs.",
                                     userIdCriIdPairs.size())));
+
+            objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
+            outputStream.write(objectMapper.writeValueAsString(results).getBytes());
```
Joe-Edwards-GDS: I think you might be able to write directly to the output stream - try `objectMapper.writeValue(outputStream, results)`

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/domain/RevokeVcsResult.java
MikeCollingwood: That's fair but I think it's easier to spot content when expanded down - for going through errors and their reasons

Joe-Edwards-GDS: Bear in mind there's going to be 5000 of them, but sure

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -39,6 +41,8 @@
 @SuppressWarnings("unused") // Temporarily disable to pass sonarqube
 public class RevokeVcsHandler implements RequestStreamHandler {
     private static final Logger LOGGER = LogManager.getLogger();
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    private final RevokeVcsResult result = new RevokeVcsResult();
```
Joe-Edwards-GDS: Be very careful using class variables like this - they might persist between invocations of the lambda. It'd be much better to initialise this in the handler method.

MikeCollingwood: oh! ok, will do that thank you

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/domain/RevokeVcsResult.java
MikeCollingwood: Think it'd be easier to scroll (after a ctrl+f to find failures) through them vertically than horizontally but either way can convert between the two (locally ofc)

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -86,6 +86,7 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
                 new ObjectMapper()
                         .readValue(inputStream, VcsActionRequest.class)
                         .getUserIdCriIdPairs();
+        result = new RevokeVcsResult();
```
Joe-Edwards-GDS: I'd still prefer this to be a new variable in the handler method, but this _should_ be fine...

(It'll only be bad if anything above it throws...)

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1579, opened 2024-01-26
PYIC-4294: Clean up core-back test fixtures

TalatAydinGDS:
### What changed

VCs in unit tests have been updated. Detailed checks were performed for each CRI through 7-8 different sampling instances, comparing VC details with the current VC example, and subsequently updating the unit tests accordingly."

### Why did it change

  We should refresh them with a curated list of good examples from existing CRIs. This will give is greated confidence that our code will handle the VCs it might encounter.

### Issue tracking

- [PYIC-4294](https://govukverify.atlassian.net/browse/PYIC-4294)

### Other considerations

[PYIC-4294]: https://govukverify.atlassian.net/browse/PYIC-4294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TalatAydinGDS:
Currently, only VCs have been updated based on the latest content changes. In my opinion, once this task is completed https://govukverify.atlassian.net/browse/PYIC-3908, it would be possible to create samples using the generated domain objects, and these samples could be utilised within unit tests. This way, any future changes to VCs would be dynamically incorporated directly into the domain objects. It would suffice to only update the data sent to the domain objects. I believe that as the schemas of domain objects are updated in the future, it will be necessary to update contract tests and potentially find a comprehensive solution for all teams.

lwebb-iig (dismissed):

Joe-Edwards-GDS (approved):
I think it might be marginally better to use unique test details in the VCs to avoid exposing too much about our test environments, but it's not actually sensitive, and we can always tidy up with the next ticket if we wish.

PR #1580, opened 2024-01-26
PYIC-4339 Refactor to reorder vc correlation check after attaining vot.

rpayal:
## Proposed changes
Check  for correlation in CheckExistingIdentityHandler and reorder, so that it is called after matching a GPG45 profile, and is considered a failure to meet the profile.

### What changed
- Refactor to reorder vc correlation check after attaining vot.
- Moved forced reset identity along with reprove identity block
- 

### Issue tracking

- [PYIC-4339](https://govukverify.atlassian.net/browse/PYIC-4339)

### Other considerations

[PYIC-4339]: https://govukverify.atlassian.net/browse/PYIC-4339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -317,6 +317,9 @@ public class CheckExistingIdentityHandler
                 gpg45ProfileEvaluator.parseCredentials(
                         userIdentityService.getIdentityCredentials(vcStoreItems));
 
+        // Check for credentials correlation failure
+        var areGpg45VcsCorrelated = userIdentityService.areVCsCorrelated(vcStoreItems);
+
         // Check for attained vot from vtr
         var strongestAttainedVotFromVtr =
                 getStrongestAttainedVotForVtr(
@@ -324,7 +327,7 @@ public class CheckExistingIdentityHandler
                         credentials,
                         vcStoreItems,
                         auditEventUser,
-                        isF2FComplete);
+                        areGpg45VcsCorrelated);
 
         // vot achieved for vtr
         if (strongestAttainedVotFromVtr.isPresent()) {
@@ -339,6 +342,9 @@ public class CheckExistingIdentityHandler
                             auditEventUser));
         }
 
+        if (!areGpg45VcsCorrelated && isF2FComplete) {
+            sendAuditEvent(AuditEventTypes.IPV_F2F_CORRELATION_FAIL, auditEventUser);
+        }
         return Optional.empty();
     }
 
@@ -400,9 +406,8 @@ public class CheckExistingIdentityHandler
             List<SignedJWT> credentials,
             List<VcStoreItem> vcStoreItems,
             AuditEventUser auditEventUser,
-            boolean isF2FComplete)
-            throws UnknownEvidenceTypeException, ParseException, SqsException,
-                    HttpResponseExceptionWithErrorBody, CredentialParseException {
+            boolean areGpg45VcsCorrelated)
+            throws UnknownEvidenceTypeException, ParseException, SqsException {
 
         var requestedVotsByStrength =
                 SUPPORTED_VOTS_BY_STRENGTH.stream()
@@ -410,15 +415,16 @@ public class CheckExistingIdentityHandler
                         .toList();
 
         for (var requestedVot : requestedVotsByStrength) {
-            var requestedVotAttained =
-                    requestedVot.getProfileType().equals(ProfileType.GPG45)
-                            ? achievedWithGpg45Profile(
-                                    requestedVot,
-                                    credentials,
-                                    vcStoreItems,
-                                    auditEventUser,
-                                    isF2FComplete)
-                            : hasOperationalProfileVc(requestedVot, credentials);
+            boolean requestedVotAttained = false;
+            if (requestedVot.getProfileType().equals(ProfileType.GPG45)) {
+                if (areGpg45VcsCorrelated) {
+                    requestedVotAttained =
+                            achievedWithGpg45Profile(
+                                    requestedVot, credentials, vcStoreItems, auditEventUser);
+                }
+            } else {
+                requestedVotAttained = hasOperationalProfileVc(requestedVot, credentials);
+            }
 
             if (requestedVotAttained) {
                 return Optional.of(requestedVot);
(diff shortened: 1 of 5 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -226,36 +227,37 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
                 ipvSessionItem.setVot(attainedVot.name());
                 ipvSessionService.updateIpvSession(ipvSessionItem);
-
+                if (attainedVot.getProfileType().equals(ProfileType.GPG45)) {
+                    if (!userIdentityService.areVCsCorrelated(vcStoreItems)) {
+                        return isF2FComplete
+                                ? buildF2FNotCorrelatedResponse(auditEventUser)
+                                : buildNotCorrelatedResponse(auditEventUser);
+                    }
+                }
```
Joe-Edwards-GDS: I think we actually need to move the correlation check within `achievedWithGpg45Profile`. Logically, correlation is actually part of forming a valid profile, and if a user has an uncorrelated GPG45 profile, they should still be able to achieve an operational profile.

It shouldn't be too hard to return the right journey event in this case, but you might have to do some refactoring to get the right audit events.

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -985,6 +1060,45 @@ void shouldReturnJourneyReuseResponseIfCheckRequiresAdditionalEvidenceResponseFa
         assertEquals(Vot.P2.name(), ipvSessionItem.getVot());
     }
 
+    //    @Test
```
Joe-Edwards-GDS: Do we need to do something with this?

rpayal: done....

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -210,50 +212,22 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
-                        : buildNotCorrelatedResponse(auditEventUser);
-            }
-
-            // Force reset
-            if (configService.enabled(RESET_IDENTITY.getName())) {
-                return buildForceResetResponse();
+            Map<String, Object> response =
+                    checkForCIScoringFailure(
+                            ipAddress, clientOAuthSessionItem, govukSigninJourneyId);
+            if (!response.isEmpty()) {
```
Joe-Edwards-GDS: Not critical, but I think I'd prefer this to return `Optional`, to better signify the intent (what if `{}` was a valid response?)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -283,6 +257,70 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
+            Vot attainedVot = strongestAttainedVotFromVtr.get();
+            ipvSessionItem.setVot(attainedVot.name());
+            ipvSessionService.updateIpvSession(ipvSessionItem);
+            if (attainedVot.getProfileType().equals(ProfileType.GPG45)
+                    && (!userIdentityService.areVCsCorrelated(vcStoreItems))) {
+                // Matched - Credential correlation failure
+                return isF2FComplete
+                        ? buildF2FNotCorrelatedResponse(auditEventUser)
+                        : buildNotCorrelatedResponse(auditEventUser);
+            }
```
Joe-Edwards-GDS: Sorry Rohit, I think this is still not quite in the right place... I think we need to include a correlation check in `achievedWithGpg45Profile`. In particular, if a user has a complete but uncorrelated set of VCs, they should still be able to use an operational profile.

This will make it hard to distinguish `buildF2FNotCorrelatedResponse` from `buildF2FNoMatchResponse`... They both result in the `f2f-fail` journey event, but record different audit events. Hopefully you can figure it out!

We might also mix up `buildNotCorrelatedResponse` and `buildNoMatchResponse`, but those record the same event, so not such a big problem.

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -448,6 +447,84 @@ void shouldMatchStrongestVotRegardlessOfVtrOrder() throws Exception {
...
+
+        InOrder inOrder = inOrder(ipvSessionItem, ipvSessionService);
+        inOrder.verify(ipvSessionItem).setVot(Vot.P2.name());
+        inOrder.verify(ipvSessionService).updateIpvSession(ipvSessionItem);
+        inOrder.verify(ipvSessionItem, never()).setVot(any());
+        assertEquals(Vot.P2.name(), ipvSessionItem.getVot());
+    }
+
+    @Test
+    void shouldMatchStrongestVotRegardlessOfVtrOrderButVCsDoNotCorrelate() throws Exception {
```
Joe-Edwards-GDS: (As above, I think the missing test case here is something like `shouldMatchOperationalProfileIfGPG45VCsDoNotCorrelate`)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: This was just refactoring to resolve sonar issue... yeah working on that.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -460,9 +431,19 @@ private boolean achievedWithGpg45Profile(
...
+            AuditEventUser auditEventUser,
+            boolean isF2FComplete)
+            throws UnknownEvidenceTypeException, ParseException, SqsException,
+                    HttpResponseExceptionWithErrorBody, CredentialParseException {
+        // Check for credential correlation failure
+        if (!userIdentityService.areVCsCorrelated(vcStoreItems)) {
+            if (isF2FComplete) {
+                sendAuditEvent(AuditEventTypes.IPV_F2F_CORRELATION_FAIL, auditEventUser);
+            } else {
+                sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_RESET, auditEventUser);
```
Joe-Edwards-GDS: I think this is still not _quite_ right - you might still match an operational profile after this (in which case we don't want these events).

I believe we should record:
- IPV_F2F_CORRELATION_FAIL if f2f is complete, your VCs don't correlate, and you don't match _any_ profile
- IPV_F2F_PROFILE_NOT_MET_FAIL if f2f is complete, your VCs do correlate, but you don't match any profile
- IPV_IDENTITY_REUSE_RESET if it's not f2f, and you don't match any profile

I can think of two approaches:
- Pull the correlation check back out to the top level, but store in a variable `areGpg45VcsCorrelated`
- Pass that variable into `checkProfileMatch` and skip GPG45 profiles if correlation fails
- use the combination of `isF2FComplete` and `areGpg45VcsCorrelated` to decide which audit and journey events to log afterwards

or

- Keep the correlation check here, and re-check correlation if you fail to match a profile with f2f complete - has some duplication, but might be easier to implement/understand

rpayal: With above change I tried to keep all the audit events as it is.
Here we can only log audit event IPV_F2F_CORRELATION_FAIL if f2f is complete.

With above statement - use the combination of isF2FComplete and areGpg45VcsCorrelated to decide which audit and journey events to log afterwards.
R u suggesting doing this in achievedWithGpg45Profile().

Joe-Edwards-GDS (approved):
Thanks Rohit!

PR #1583, opened 2024-01-29
PYIC-4614: Add revoke and failure audit events

MikeCollingwood:
### Issue tracking

### Other considerations


lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -71,32 +87,80 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
             // Submit audit event
-            // ...
+            sendVcRevokedAuditEvent(userId, vcStoreItem);
 
             // Delete VC from the main table
             // ...
 
         } else {
-            LOGGER.error("VC {} cannot be found", userId);
+            throw new Exception("VcStoreItem cannot be found");
```
MikeCollingwood: Could make a new exception type here but it's immediately caught

lambdas/archive-vcs/src/main/java/uk/gov/di/ipv/core/archivevcs/ArchiveVcsHandler.java
```diff
@@ -71,32 +87,80 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
+                                i, userIdsSize)));
+        try {
+            auditService.sendAuditEvent(auditEvent);
+        } catch (SqsException sqsException) {
+            LOGGER.error(
+                    LogHelper.buildLogMessage(
+                            String.format(
+                                    "Failed to send audit event at index %s / %s.",
+                                    i, userIdsSize)));
+            throw new RuntimeException();
```
MikeCollingwood: In the case that sending audit events doesn't work, we will log the progress through the list (no userId but it can be worked out from the input list), then throw an unhandled exception because we really want to sort out the audit event issue before carrying on

PR #1593, opened 2024-02-02
fix: spotless error

shivanshuit914:
## Proposed changes

- fix spotless error

### What changed

- fixed spotless error

### Why did it change

- To pass build

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
@@ -63,7 +63,7 @@ public class UserIdentity {
             @JsonProperty(value = DRIVING_PERMIT_CLAIM_NAME) JsonNode drivingPermitClaim,
             @JsonProperty(value = NINO_CLAIM_NAME) JsonNode ninoClaim,
             @JsonProperty(value = "sub", required = true) String sub,
-            @JsonProperty(value = "vot", required = true) Vot vot,
+            @JsonProperty(value = VOT_CLAIM_NAME, required = true) Vot vot,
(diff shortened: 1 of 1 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -19,7 +19,6 @@
 import static uk.gov.di.ipv.core.library.domain.VocabConstants.PASSPORT_CLAIM_NAME;
 import static uk.gov.di.ipv.core.library.domain.VocabConstants.RETURN_CODE_NAME;
 import static uk.gov.di.ipv.core.library.domain.VocabConstants.VCS_CLAIM_NAME;
-import static uk.gov.di.ipv.core.library.domain.VocabConstants.VOT_CLAIM_NAME;
```
rpayal: OR instead of removing this import can use it at line no. 65.

            @JsonProperty(value = VOT_CLAIM_NAME, required = true) Vot vot,

shivanshuit914: hm, I used spotless apply it did it automatically :)

shivanshuit914: I think `VOT_CLAIM_NAME` was not used and that's why spotless complaining.

rpayal: missed out during merge ....you can add now.

Also not sure why pre-commit didn't picked it during build.

rpayal (approved):

PR #1596, opened 2024-02-05
BAU: Refactor check existing identity to get a journey response

Joe-Edwards-GDS:
A bit of tidying to separate calculating the journey response from processing the request - to make it easier to potentially move this logic around in future.


MikeCollingwood (approved):
This LGTM! Makes sense using JourneyResponse like this, instead of the maps. The getJourneyResponse method also extracts the logic from the handleRequest method which is good because we can limit it's function to dealing with lambda'y things (+ this is consistent with other changes)

PR #1597, opened 2024-02-05
PYIC-4459 journey split

Joe-Edwards-GDS:
### What changed

Split the journey map into separate types:
```
    // Main entry point
    IPV_CORE_MAIN_JOURNEY("ipv-core-main-journey"),

    // Successful journeys
    REUSE_EXISTING_IDENTITY("reuse-existing-identity"),
    NEW_P2_IDENTITY("new-p2-identity"),

    // Unsuccessful journeys
    INELIGIBLE("ineligible"),
    FAILED("failed"),
    TECHNICAL_ERROR("technical-error"),
    SESSION_TIMEOUT("session-timeout"),

    // F2F return journeys
    F2F_PENDING("f2f-pending"),
    F2F_FAILED("f2f-failed"),

    // Operational profile journeys
    OPERATIONAL_PROFILE_MIGRATION("operational-profile-migration"),
    OPERATIONAL_PROFILE_REUSE("operational-profile-reuse");
```
Together with enhancements to the journey engine to allow for switching to another journey type

### Why did it change

- The journey map is reaching a significant level of complexity, even when visualised (particularly when including all the failure cases), making it harder to reason about when adding cross-cutting functionality like TICF CRI.
- We are starting to introduce logically separate journeys - e.g. migration, uplift, other profiles that only link at the start and end, and do not need to be interleaved.
- We would like to be able to easily report on the categories of user journey through IPV Core (e.g. to identify which users are being asked to prove a new P2 identity).

### Issue tracking

- [PYIC-4459](https://govukverify.atlassian.net/browse/PYIC-4459)

### Other considerations

We need corresponding changes to the journey map visualisation and documentation

[PYIC-4459]: https://govukverify.atlassian.net/browse/PYIC-4459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS, line 1: I've chickened out of combining `InitialiseSessionHandler` and `CheckExistingIdentityHandler` for now, but happy to hear opinions on whether it is a good idea or not :)

Pros:
- We can remove this wrapper journey (which only contains a single step to select the 'real' journey type)
- We can use any information in the orchestration JAR during the `CheckExistingIdentity` process

Cons:
- Requires a lot more refactoring and release preparation that provides limited user benefit
- We would need to encode more information about the journey maps into code (as opposed to YAML), so that the check identity service returns journey type + entry point, rather than just a journey event
  - (this might not _necessarily_ be a bad thing)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -139,42 +139,41 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
     @Tracing
     private StepResponse executeJourneyEvent(String journeyEvent, IpvSessionItem ipvSessionItem)
             throws JourneyEngineException {
-        String currentUserState = ipvSessionItem.getUserState();
         if (sessionIsNewlyExpired(ipvSessionItem)) {
-            updateUserSessionForTimeout(currentUserState, ipvSessionItem);
-            return new PageStepResponse(PYIC_TIMEOUT_UNRECOVERABLE_ID, "", null);
+            updateUserSessionForTimeout(ipvSessionItem.getUserState(), ipvSessionItem);
+            journeyEvent = NEXT_EVENT;
```
MikeCollingwood, line 144: Why would a newly expired session lead to a next event?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -139,42 +139,41 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
...
 
-            BasicState newState =
-                    (BasicState)
-                            stateMachine.transition(
-                                    ipvSessionItem.getUserState(),
-                                    journeyEvent,
-                                    JourneyContext.withFeatureSet(configService.getFeatureSet()));
+            var basicState = (BasicState) newState;
 
             updateUserState(
```
MikeCollingwood, line 166: Could be an opportunity to bring this file up-to-date with LogHelper.buildLogMessage

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -226,20 +247,13 @@ private void clearOauthSessionIfExists(IpvSessionItem ipvSessionItem) {
...
-                        .with("to", CORE_SESSION_TIMEOUT_STATE);
-        LOGGER.info(message);
+        ipvSessionItem.setJourneyType(IpvJourneyTypes.SESSION_TIMEOUT);
+        updateUserState(oldState, CORE_SESSION_TIMEOUT_STATE, "timeout", ipvSessionItem);
     }
 
     @Tracing
     private boolean sessionIsNewlyExpired(IpvSessionItem ipvSessionItem) {
-        return (!CORE_SESSION_TIMEOUT_STATE.equals(ipvSessionItem.getUserState()))
+        return (!IpvJourneyTypes.SESSION_TIMEOUT.equals(ipvSessionItem.getJourneyType()))
```
MikeCollingwood, line 256: Would this have worked unchanged? I prefer this version but curious

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/states/JourneyChangeState.java
```diff
@@ -0,0 +1,21 @@
...
+@Data
+@AllArgsConstructor
+public class JourneyChangeState implements State {
+    private IpvJourneyTypes journeyType;
+    private String initialState;
+
+    @Override
+    public State transition(String eventName, String startState, JourneyContext journeyContext)
+            throws UnknownEventException, UnknownStateException {
+        throw new IllegalStateException("Cannot transition from JourneyChangeState");
```
MikeCollingwood, line 19: Might help to indicate that it can't transition because this is the end-of-the-line for this state machine

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/stepresponses/JourneyStepResponse.java
MikeCollingwood, line 1: Got it, state machine step response can't be a journey step response. Got confused with that and journeyResponse

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
MikeCollingwood: Think it definitely makes sense to keep it out of this PR for readability at least. I think removing the wrapper neatens things up, so I think it's still a good idea to do. It's probably wanted soon as well since these are fundamental changes to the journey map and wouldn't want to drag out

MikeCollingwood: I've not fully understood why the check identity service has to return an entry point (to the map I assume) as well as the journey map type?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: This is checking whether you are (anywhere) within the 'timeout' journey, rather than a specific state. So e.g. if we added more screens in the timeout journey it would still work.

It's also nicer to rely on the enum, rather than a hardcoded string (that needs to match the YAML).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: So e.g. in the `new-p2-identity` case, we'd need to say that the JourneyType is `NEW_P2_IDENTITY`, but _also_ the entry point - i.e. to distinguish `START`/`START_WITH_RESET`/`ENHANCED_VERIFICATION` (i.e. we'd have to include all those values in code)

blakeyp: I don't feel strongly although CheckExistingIdentity is doing quite a lot already so combining them could get a bit bloated

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/session-timeout.yaml
```diff
@@ -0,0 +1,21 @@
...
+  events:
+    next:
+      targetState: TIMEOUT_UNRECOVERABLE_PAGE
+
+# Journey states
+
+TIMEOUT_UNRECOVERABLE_PAGE:
+  response:
+    type: page
+    pageId: pyi-timeout-unrecoverable
```
blakeyp, line 13: FYI @RogVanGDS is working on [PYIC-4432](https://govukverify.atlassian.net/browse/PYIC-4432) to make the pyi-timeout page dynamic so there'll probably be a clash here

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
blakeyp: It looks like the page response is being handled in the journey map now - session-timeout.yaml

blakeyp (dismissed):
Looks good to me

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -139,42 +139,41 @@ public Map<String, Object> handleRequest(Map<String, String> input, Context cont
     @Tracing
     private StepResponse executeJourneyEvent(String journeyEvent, IpvSessionItem ipvSessionItem)
             throws JourneyEngineException {
-        String currentUserState = ipvSessionItem.getUserState();
         if (sessionIsNewlyExpired(ipvSessionItem)) {
-            updateUserSessionForTimeout(currentUserState, ipvSessionItem);
-            return new PageStepResponse(PYIC_TIMEOUT_UNRECOVERABLE_ID, "", null);
+            updateUserSessionForTimeout(ipvSessionItem.getUserState(), ipvSessionItem);
```
Wynndow, line 143: Can `updateUserSessionForTimeout` be updated to just require the session item, and then extract the current user state within the method?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -202,6 +201,28 @@ private StepResponse executeJourneyEvent(String journeyEvent, IpvSessionItem ipv
...
+    private State executeStateTransition(IpvSessionItem ipvSessionItem, String journeyEvent)
+            throws StateMachineNotFoundException, UnknownEventException, UnknownStateException {
+        StateMachine stateMachine = stateMachines.get(ipvSessionItem.getJourneyType());
+        if (stateMachine == null) {
+            throw new StateMachineNotFoundException(
+                    String.format(
+                            "State machine not found for journey type: '%s'",
+                            ipvSessionItem.getJourneyType()));
+        }
+        LOGGER.debug(
```
Wynndow, line 214: 👌

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: I think I come down on the side of not combining them for now - this doesn't feel like a large burden if it avoids the changes you mention. Although interested to understand exactly how having access to the JAR in the check existing identity handler would be useful, as that could be a game changer.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -0,0 +1,748 @@
...
+          targetState: PYI_CRI_ESCAPE_NO_F2F
+    next:
+      targetState: EVALUATE_GPG45_SCORES
+    enhanced-verification:
+      targetState: MITIGATION_02_OPTIONS_WITH_F2F
+      checkIfDisabled:
+        f2f:
+          targetState: MITIGATION_02_OPTIONS
+
+# DCMAW journey (J1)
```
Wynndow, line 325: These journey suffix's (`_J1`, `_J2` etc), were only introduced as a way to keep track of the different journeys when all the states were in the same map. Wondering if there's an opportunity to split them out here, and make this map smaller. Haven't thought about it loads so there might be good reasons not to. Possibly a next step?

Wynndow (dismissed):
This all seems to make sense. Just a few comments.

Something else that occured to me though - in the initialise session lambda we currently set the initial journey type to a value defined in config. I think we could probably safely remove that and just use a hard coded value?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: @Wynndow things like:
- We currently store the VTR in DynamoDB in InitialiseIpvSession and use it in CheckExistingIdentity, if they were combined we wouldn't need to
- We track whether a user is performing a migration indirectly by checking the vcReceivedThisSession array in the session, if they were combined we could check the JAR for an inherited identity
- We track if the `reprove_identity` flag is set and store it in the session, if they were combined we could just read it off the JAR

Maybe we should just be storing the entire JAR in the session instead 🤔

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Joe-Edwards-GDS: Yeah, I think the main issue is that they are not all cleanly separated, e.g. J2 and J3 (passport and DL web) join up again after the web CRI - would you combine them, or add a third part of the journey for Address/Fraud/KBV?

It felt easier to only consider 'overall' journeys 🤷

Wynndow: Possibly you could define the address/fraud/kbv part of the journey in each of the separate files (so duplication), but they could potentially all reference the same state. So an abstract idea of a state with multiple pointers to it. This is probably a terrible idea though.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Wynndow: Might make sense to stick it in the client oauth session

Wynndow (dismissed):

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Joe-Edwards-GDS, line 1: (Is this test ever used?)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ipv-core-main-journey.yaml
Joe-Edwards-GDS: I'm going to park this and we can come back later if those rules get more complex :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -101,14 +100,14 @@ public IpvSessionItem generateIpvSession(
...
-        String journeyType = this.configService.getSsmParameter(ConfigurationVariable.JOURNEY_TYPE);
-        ipvSessionItem.setJourneyType(IpvJourneyTypes.valueOf(journeyType));
-
-        if (errorObject != null) {
+        if (errorObject == null) {
+            ipvSessionItem.setJourneyType(IpvJourneyTypes.INITIAL_JOURNEY_SELECTION);
+            ipvSessionItem.setUserState(START_STATE);
+        } else {
+            ipvSessionItem.setJourneyType(IpvJourneyTypes.TECHNICAL_ERROR);
+            ipvSessionItem.setUserState(ERROR_STATE);
```
Joe-Edwards-GDS, line 110: This means the initial journey is no longer controlled via config, but I'm not convinced we ever used that (or would have wanted to)

Joe-Edwards-GDS, line 110: I couldn't find `FAILED_CLIENT_JAR` anywhere, so I've routed this to the technical error state. I'm assuming it would have done that anyway, but via a 'unknown state'...

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Wynndow: I don't think so. There was an idea to add a suite of integration tests back in the day, to check that we can integrate with AWS correctly, and this is as far as it got. With the advent of the dev envs this is probably fairly redundant.

Wynndow (approved):

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Joe-Edwards-GDS: Hmm, some day we should just delete them

PR #1602, opened 2024-02-06
PYIC-4248: New IPV_INHERITED_IDENTITY_VC_RECEIVED TxMA Event - ipv inherited identity vc received

sammyiyke:
### What changed

- Adds audit event `IPV_INHERITED_IDENTITY_VC_RECEIVED` , signalling the start of a user migration from HMRC. Event information includes the user details and the `VoT`.
- Adds audit event `IPV_JOURNEY_START`, triggered when CORE receives a request from orchestration and the session is initialised. Information sent in this event includes the `VTR` in the extension data
- Adds new test `shouldHandleParseExceptionFromSendingAuditEvent` for handling bad vots
- Removes `shouldHandleParseExceptionFromCheckingInheritedIdentityVotStrength` as its redundant and already being checked by `shouldHandleParseExceptionFromSendingAuditEvent`
- Adds new `AuditExtensionException` to handle exceptions unique to Audit Events

### Why did it change

[PYIC-4248 New IPV_INHERITED_IDENTITY_VC_RECEIVED TxMA Event](https://govukverify.atlassian.net/browse/PYIC-4248)

To capture and report detailed information about the start of a user's migration process, including specifics such as the Vector of Trust (VoT), Vector of Trust Request (VTR), and evidence provided, to facilitate insightful analysis and metric tracking.

### Issue tracking

- [PYIC-4248](https://govukverify.atlassian.net/browse/PYIC-4248)

[PYIC-4248]: https://govukverify.atlassian.net/browse/PYIC-4248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -516,6 +516,12 @@ void shouldValidateAndStoreAnyInheritedIdentityWhenNoExistingIdentity() throws E
         assertEquals(
                 ipvSessionItemCaptor.getValue().getVcReceivedThisSession(),
                 List.of(serialisedInheritedIdentityJWT));
+
+        //        ArgumentCaptor<AuditEvent> auditEventCaptor =
+        // ArgumentCaptor.forClass(AuditEvent.class);
+        //        verify(mockAuditService).sendAuditEvent(auditEventCaptor.capture());
+        //        assertEquals(AuditEventTypes.IPV_INHERITED_IDENTITY_VC_RECEIVED,
+        // auditEventCaptor.getAllValues().get(1).getEventName());
```
Joe-Edwards-GDS: Does this need updating?

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidence.java
```diff
@@ -23,13 +25,23 @@ public class AuditExtensionsVcEvidence implements AuditExtensions {
...
     public AuditExtensionsVcEvidence(
             @JsonProperty(value = "iss", required = false) String iss,
             @JsonProperty(value = "evidence", required = false) String evidence,
-            @JsonProperty(value = "successful", required = false) Boolean successful)
-            throws JsonProcessingException {
-        this.iss = iss;
-        this.evidence = evidence == null ? null : new ObjectMapper().readTree(evidence);
-        this.successful = successful;
+            @JsonProperty(value = "successful", required = false) Boolean successful,
+            @JsonProperty(value = "vot", required = false) Vot vot)
```
Joe-Edwards-GDS: If you wanted to avoid adding `null` to all the existing cases then you could consider adding an an extra constructor that delegates to this one. Something like:
```
public AuditExtensionsVcEvidence(
  String iss,
  String evidence,
  Boolean successful) {
  this(iss, evidence, successful, null);
}
```

Alternatively, I wonder if it's worth pulling out a helper/factory method for generating one of these from a SignedJWT - we seem to have several parallel implementations. (Perhaps moving `getExtensionsForAudit` into a helper on `VcHelper`?)

sammyiyke: Yea, didnt think of that. This has been implemented and the existing cases have been amended

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -21,10 +20,7 @@
 import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 import uk.gov.di.ipv.core.library.domain.ProcessRequest;
 import uk.gov.di.ipv.core.library.enums.Vot;
-import uk.gov.di.ipv.core.library.exceptions.ConfigException;
-import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
-import uk.gov.di.ipv.core.library.exceptions.SqsException;
-import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
+import uk.gov.di.ipv.core.library.exceptions.*;
```
Joe-Edwards-GDS: We've got a general style rule to avoid wildcard exceptions (although I'm not sure if this is documented anywhere...) you can disable it in IntelliJ by increasing 'Class count to use import with '*'' (under Editor > Code Style > Java > Imports)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -411,6 +415,29 @@ private boolean isHmrcInheritedIdentityWithStrongerVotPresent(
...
+    private void sendInheritedIdentityReceivedAuditEvent(
+            SignedJWT signedInheritedIdentityJWT, AuditEventUser auditEventUser)
+            throws SqsException, ParseException, AuditExtensionException, CredentialParseException {
+        try {
+            Vot vot = userIdentityService.getVot(signedInheritedIdentityJWT);
+            var claimsSet = signedInheritedIdentityJWT.getJWTClaimsSet();
+            var vc = (JSONObject) claimsSet.getClaim(VC_CLAIM);
+            var evidence = vc.getAsString(EVIDENCE);
+            AuditExtensionsVcEvidence auditExtensions =
+                    new AuditExtensionsVcEvidence(claimsSet.getIssuer(), evidence, null, vot);
```
Joe-Edwards-GDS: The only thing we could consider is whether it's worth using a shared helper for this logic - we've got at least three copies of it now:
- `AuditCriResponseHelper#getExtensionsForAudit`
- `CriStoringService#getAuditExtensions`
- This method

I wonder if we should: rename `AuditCriResponseHelper` to `AuditExtensionsHelper` and move into the common `audit-service` lib, and then update all usages to use that.

For extra credit, we could consider moving some of the methods from `VcHelper` there. You'll need to rebase/merge to see those changes though.

Joe-Edwards-GDS (commented):
I think you've got some merge conflicts to resolve now, sorry!

I've also suggested another refactoring after we do that, but it's not strictly required.

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandlerTest.java
```diff
@@ -294,7 +294,7 @@ void handleRequestShouldReturnJourneyErrorResponseIfCiStoringServiceThrows() thr
                 List.of(
                         new SqsException("Oops"),
                         new ParseException("Oops", 1),
-                        new JsonProcessingException("") {},
+                        new AuditExtensionException("") {},
```
Wynndow: I think you can drop the `{}` from this line. I don't think it should have been there in the first place. This is creating [an anonymous class](https://docs.oracle.com/javase/tutorial/java/javaOO/anonymousclasses.html), which we don't really need.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -74,7 +74,7 @@ public class InitialiseIpvSessionHandler
     private static final String REQUEST_EMAIL_ADDRESS_KEY = "email_address";
     private static final String REQUEST_VTR_KEY = "vtr";
     private static final List<Vot> HMRC_PROFILES_BY_STRENGTH = List.of(Vot.PCL250, Vot.PCL200);
-
+    private static final String EVIDENCE = "evidence";
```
Wynndow: We have a class, `VerifiableCredentialConstants`, that contains a bunch of constants like this we can use. `evidence` is already in there so probably worth pulling that it, rather than redefining here.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -223,4 +223,13 @@ private static Integer getAge(String dobValue) {
             return null;
         }
     }
+
+    public static Vot getVcVot(SignedJWT vc) {
+        try {
+            return Vot.valueOf(vc.getJWTClaimsSet().getStringClaim(VOT_CLAIM_NAME));
+        } catch (Exception e) {
```
Wynndow: Rather than catching all `Exception`s here, it's better to be more specific and just catch the `ParseException` that `getJWTClaimsSet` can thrown.

Also, `valueOf` will throw an `IllegalArgumentExcpetion` if there is a value for `vot` but it doesn't match with any defined enum. We shouldn't really get into that position, but we should potentially handle it differently.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -223,4 +223,13 @@ private static Integer getAge(String dobValue) {
...
+            LOGGER.error("Failed to get VOT from VC");
```
Wynndow: I'm not sure this should be of error level. The vast majority of VCs won't have a `vot` value. Might be better as `info` or possibly `debug`

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandlerTest.java
sammyiyke: Reviewed, I wasnt sure what it was doing there in the first place and left it just in case

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -29,15 +29,13 @@
...
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsVcEvidence;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
 import uk.gov.di.ipv.core.library.config.CoreFeatureFlag;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.enums.Vot;
-import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
-import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
-import uk.gov.di.ipv.core.library.exceptions.SqsException;
-import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
+import uk.gov.di.ipv.core.library.exceptions.*;
```
Wynndow: Another wildcard here to remove

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -257,7 +255,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                             "Jar validation failed.", e.getErrorObject().getDescription()));
             return ApiGatewayResponseGenerator.proxyJsonResponse(
                     HttpStatus.SC_BAD_REQUEST, ErrorResponse.INVALID_SESSION_REQUEST);
-        } catch (SqsException e) {
+        } catch (SqsException | AuditExtensionException e) {
```
Wynndow: I actually think that we might want to catch the `AuditExtensionException` in the `validateAndStoreHMRCInheritedIdentity` method and rethrow it as a `RecoverableJarValidationException` in the same way we do with the `UnrecognisedVotException`.
Would mean we don't catch it here.

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -660,16 +706,7 @@ void shouldHandleParseExceptionFromCheckingInheritedIdentityVotStrength() throws
...
-        when(mockVerifiableCredentialService.getVcStoreItem("test-user-id", HMRC_MIGRATION_CRI))
-                .thenReturn(
-                        new VcStoreItem(
-                                "test-user-id",
-                                HMRC_MIGRATION_CRI,
-                                existingInheritedIdentityJwt.serialize(),
-                                Instant.now(),
-                                Instant.now()));
-        doThrow(new ParseException("", 0) {}).when(mockUserIdentityService).getVot(any());
+        //        doThrow(new ParseException("", 0) {}).when(mockUserIdentityService).getVot(any());
```
Wynndow: I think this can be dropped

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -638,7 +684,7 @@ void shouldValidateAndStoreAnyInheritedIdentityWhenNoInheritedVcExist() throws E
     }
 
     @Test
-    void shouldHandleParseExceptionFromCheckingInheritedIdentityVotStrength() throws Exception {
+    void shouldHandleParseExceptionFromSendingAuditEvent() throws Exception {
```
Wynndow: Think the name of this could be something else now as we're no longer throwing a ParseException

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -114,10 +116,19 @@ public SQSBatchResponse handleRequest(SQSEvent event, Context context) {
...
                 LOGGER.error(
                         LogHelper.buildErrorMessage("Failed to process VC response message.", e));
                 failedRecords.add(new SQSBatchResponse.BatchItemFailure(message.getMessageId()));
+            } catch (AuditExtensionException e) {
+                LOGGER.error(
+                        new StringMapMessage()
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        "Failed to process/send audit event.")
+                                .with(LOG_ERROR_DESCRIPTION.getFieldName(), e.getMessage()));
```
Wynndow:
```suggestion
                LOGGER.error(LogHelper.buildErrorMesssage("Failed to process/send audit event.", e.getMessage()));
```

I think we can use the above which is functionaly equivalent and a bit less verbose.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidence.java
```diff
@@ -35,13 +41,29 @@ public AuditExtensionsVcEvidence(
...
-            throws JsonProcessingException {
-        this.iss = iss;
-        this.evidence = evidence == null ? null : new ObjectMapper().readTree(evidence);
-        this.successful = successful;
-        this.isUkIssued = isUkIssued;
-        this.age = age;
+            throws AuditExtensionException {
+        try {
+            this.iss = iss;
+            this.evidence = evidence == null ? null : new ObjectMapper().readTree(evidence);
```
Wynndow: Rather than creating a new ObjectMapper instance for every invocation of this constructor I think it would be more efficient to just create a static final field instance and reuse it.

libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidenceTest.java
```diff
@@ -1,16 +1,41 @@
 package uk.gov.di.ipv.core.library.auditing.extension;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.ipv.core.library.exceptions.AuditExtensionException;
 
-import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.*;
```
Wynndow: Another sneaky wildcard import here.

Wynndow (commented):
Just a few more comment, nothing big though. After this I think it's good to go.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidence.java
sammyiyke: Not sure how you want this implemented. But I do agree that creating a new ObjectMapper instance every time the constructor is called can be inefficient. I have modified the code to use a static ObjectMapper instance instead that gets reused every time the constructor is called.

`private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();`

`this.evidence = evidence == null ? null : OBJECT_MAPPER.readTree(evidence);`

Wynndow (commented):
All looks good. Be worth squashing down the commits. Will give it a tick after that.

PR #1628, opened 2024-02-14
PYIC-4248: IPV_INHERITED_IDENTITY_VC audit event

Wynndow:
### What changed

- Adds audit event `IPV_INHERITED_IDENTITY_VC_RECEIVED` , signalling the start of a user migration from HMRC. Event information includes the user details and the `VoT`.
- Adds audit event `IPV_JOURNEY_START`, triggered when CORE receives a request from orchestration and the session is initialised. Information sent in this event includes the `VTR` in the extension data
- Adds new test `shouldHandleParseExceptionFromSendingAuditEvent` for handling bad vots
- Removes `shouldHandleParseExceptionFromCheckingInheritedIdentityVotStrength` as its redundant and already being checked by `shouldHandleParseExceptionFromSendingAuditEvent`
- Adds new `AuditExtensionException` to handle exceptions unique to Audit Events

### Why did it change

[PYIC-4248 New IPV_INHERITED_IDENTITY_VC_RECEIVED TxMA Event](https://govukverify.atlassian.net/browse/PYIC-4248)

To capture and report detailed information about the start of a user's migration process, including specifics such as the Vector of Trust (VoT), Vector of Trust Request (VTR), and evidence provided, to facilitate insightful analysis and metric tracking.

### Issue tracking

- [PYIC-4248](https://govukverify.atlassian.net/browse/PYIC-4248)

[PYIC-4248]: https://govukverify.atlassian.net/browse/PYIC-4248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/restore-vcs/src/main/java/uk/gov/di/ipv/core/restorevcs/RestoreVcsHandler.java
```diff
@@ -178,8 +180,10 @@ private void sendVcRestoredAuditEvent(String userId, VcStoreItem vcStoreItem)
                         jwtClaimsSet.getIssuer(),
```
sammyiyke: I think this is another case where we can use the `getExtensionsForAudit()` from the `library.auditing.helpers`

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
```diff
@@ -175,6 +177,7 @@ private void sendVcRevokedAuditEvent(String userId, VcStoreItem vcStoreItem)
                         jwtClaimsSet.getIssuer(),
                         evidence,
                         null,
+                        VcHelper.getVcVot(signedCredential),
```
sammyiyke: Same with this too

sammyiyke (dismissed):

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -223,4 +224,13 @@ private static Integer getAge(String dobValue) {
             return null;
         }
     }
+
+    public static Vot getVcVot(SignedJWT vc) throws UnrecognisedVotException {
+        try {
+            String vot = vc.getJWTClaimsSet().getStringClaim(VOT_CLAIM_NAME);
+            return vot == null ? null : Vot.valueOf(vot);
+        } catch (ParseException | IllegalArgumentException e) {
+            throw new UnrecognisedVotException("Invalid VOT found for this VC");
```
MichaelCollingwood, line 233: Could it be worth including the Vot in this error message?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -223,4 +224,13 @@ private static Integer getAge(String dobValue) {
             return null;
         }
     }
+
+    public static Vot getVcVot(SignedJWT vc) throws UnrecognisedVotException {
+        try {
+            String vot = vc.getJWTClaimsSet().getStringClaim(VOT_CLAIM_NAME);
+            return vot == null ? null : Vot.valueOf(vot);
```
MichaelCollingwood, line 231: Do we ever expect a null vot? If not, or we want to consider it a problem, then we can just pass it straight to valueOf and include NullPointer exception to the error handling. Happy to be wrong though

libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/helpers/AuditExtensionsHelperTest.java
```diff
@@ -47,8 +46,7 @@ void shouldGetVerifiableCredentialExtensionsForAudit()
     }
 
     @Test
-    void shouldGetVerifiableCredentialExtensionsForAuditWithICAOCode()
-            throws ParseException, JsonProcessingException {
+    void shouldGetVerifiableCredentialExtensionsForAuditWithICAOCode() throws Exception {
```
MichaelCollingwood, line 49: Does sonarcloud complain about inspecific exception signatures?

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidence.java
```diff
@@ -35,13 +43,29 @@ public AuditExtensionsVcEvidence(
...
+            throws AuditExtensionException {
+        try {
+            this.iss = iss;
+            this.evidence = evidence == null ? null : OBJECT_MAPPER.readTree(evidence);
+            this.successful = successful;
+            this.vot = vot;
+            this.isUkIssued = isUkIssued;
+            this.age = age;
+        } catch (JsonProcessingException e) {
+            throw new AuditExtensionException(e.getMessage());
```
MichaelCollingwood: Not sure, would it be more complete to pass the entire error?

`AuditExtensionException(e);`

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -411,6 +421,26 @@ private boolean isHmrcInheritedIdentityWithStrongerVotPresent(
...
+                    getExtensionsForAudit(signedInheritedIdentityJWT, null);
+            auditService.sendAuditEvent(
+                    new AuditEvent(
+                            AuditEventTypes.IPV_INHERITED_IDENTITY_VC_RECEIVED,
+                            configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser,
+                            auditExtensions));
+        } catch (ParseException | IllegalArgumentException e) {
+            throw new CredentialParseException(
+                    "Encountered a parsing error while attempting to parse or compare credentials",
```
MichaelCollingwood, line 438: Think this error message wants an update. e.g. we're not comparing credentials anymore

MichaelCollingwood (commented):
Looks good but had a few comments

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
sammyiyke: Yea, there are cases where the vot would be null, I might be wrong but I think it would be null in most cases except for inherited identities.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Wynndow: Seems reasonable

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Wynndow: Yeah, we do expect to get null's here. Rather do the check than throw an NPE and handle it.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionsVcEvidence.java
sammyiyke: Sounds good. But before I implement, just curious, would the error not potentially contain user identifiable information?

Wynndow: That's a good question! But the evidence block of the VC shouldn't contain information about the subject. That's in the credentialSubject section.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
sammyiyke: On this, since the `ParseException` would happen in `getStringClaim` || `getJWTClaimsSet` before its able to get a vot this would mean the vot value would not be available in some cases thereby -- correct me if I am wrong (Java newbie) -- transferring the problem into the catch block. 

Potential solution could be having to separate catch blocks. And since it's a very unlikely error, is it still worth doing that?

Wynndow: Ah yeah, that's a good point. We may not actually have a vot. 

I think you're right. This is very unliely, and if we ever did need to investigate we could fetch the VC in question and work out what happened.

MikeCollingwood: I think in that case we may want to change the rethrown exception?

~~We could override the valueOf method on the enum.~~ Turns out you can't! 

But we can create a new method on the enum, e.g:

```
public enum Vot {
    ...
    parseValue(String value) {
        try {
            this.valueOf(value)
            throw new UnrecognisedVotException(String.format("Invalid VOT found for this VC: {}", value))
        }
    }
}
```

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
MikeCollingwood: That makes sense 👍

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelperTest.java
```diff
@@ -194,6 +195,11 @@ void shouldCheckIsItOperationalVC() throws ParseException {
         assertFalse(VcHelper.isOperationalProfileVc(SignedJWT.parse(M1A_PASSPORT_VC)));
     }
 
+    @Test
+    void shouldGetVcVot() throws Exception {
```
MikeCollingwood, line 178: Maybe do unhappy paths test as well, e.g. null or "not-a-vot"

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -223,4 +224,13 @@ private static Integer getAge(String dobValue) {
             return null;
         }
     }
+
+    public static Vot getVcVot(SignedJWT vc) throws UnrecognisedVotException {
```
Joe-Edwards-GDS, line 228: Should this replace usages of `UserIdentityService#getVot` - it seems basically identical except that it catches its own exceptions?

Joe-Edwards-GDS (dismissed):
Happy for you to get this merged, to avoid any more conflicts.

If you're doing any more markups, happy to see them in a separate PR :)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Wynndow: We talked about this, and it's definitley a candidate for fixing up. The use cases are slightly different. This is used when we know and expect for a VC to potentially not contain a vot. The other is used when we're explicitly expecting a vot to exist. I suppose this version could be used and then null checks done on the result.

Joe-Edwards-GDS (approved):
As above

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: Yeah, I think that would be a better split of responsibility.

When we introduce proper VC model classes this problem might go away, as we can deserialize directly to a `Vot` :)

PR #1636, opened 2024-02-15
BAU: Add Snapstart scanning tool to core-back

Wynndow:
### What changed

Add snapstart scanning tool to core-back

### Why did it change

AWS offer a scanning tool to help ensure that lambdas don't fall foul of assuming uniqueness that could be baked into your snapstart images: https://docs.aws.amazon.com/lambda/latest/dg/snapstart-uniqueness.html#snapstart-scanning

It's a spotbugs plugin. This adds spotbugs as a tool, along with the AWS plugin. It also restricts the detectors used to just those from the AWS plugin to avoid all the noise generated by the default spotbugs detectors.

Using the defaults is an option but we already have a staic analysis tool and we don't want to overload ourselves.

The results of running the scanning tool show no current issues. It might be useful as a check for future work though.

This also doesn't cover off all things needed to check for snapstart, such as network connections, config getting baked in etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,7 @@ ext {
 		awsLambdaJavaCore:'1.2.1',
 		awsLambdaJavaEvents:'3.11.0',
 		awsLambdaJavaLog4j2:'1.5.1',
+		awsLambdaSnapstartJavaRules:'0.2.0',
 		dynamodbEnhanced:'2.17.89',
 		gson:'2.8.9',
 		jacksonDataformatYaml:'2.13.2',
@@ -51,7 +52,8 @@ ext {
 		notificationsJavaClient: '4.1.1-RELEASE',
 		powertoolsParameters:'1.16.1',
 		powertoolsLogging:'1.16.1',
-		powertoolsTracing:'1.16.1'
+		powertoolsTracing:'1.16.1',
+		spotbugs:'4.8.3'
 	]
 }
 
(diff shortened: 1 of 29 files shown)


Joe-Edwards-GDS:
Thanks Chris - definitely nice to have tested this, but I think on balance I'm leaning away from making this a permanent part of our pipeline:
- It's another thing we'll have to maintain
- We don't need it for general static analysis
- I'm optimistic that introducing new issues _should_ be relatively rare/quickly spotted

but happy to be swayed if you'd like it to stay!

It might also be worth a chat with @chrisclayson , as I know they've implemented snapstart and I'm wondering if they've tried a similar tool (or if they've had any particular issues it might have helped with).

Wynndow:
> Thanks Chris - definitely nice to have tested this, but I think on balance I'm leaning away from making this a permanent part of our pipeline:
> 
>     * It's another thing we'll have to maintain
> 
>     * We don't need it for general static analysis
> 
>     * I'm optimistic that introducing new issues _should_ be relatively rare/quickly spotted
> 
> 
> but happy to be swayed if you'd like it to stay!
> 
> It might also be worth a chat with @chrisclayson , as I know they've implemented snapstart and I'm wondering if they've tried a similar tool (or if they've had any particular issues it might have helped with).

Yep, no worries, as I expected.

Wil close this out but leave the branch around for a bit.

Wynndow:
Closing due to the above

PR #1637, opened 2024-02-15
PYIC-4982: Defer setting of signing key ID

Wynndow:
### What changed

Defer setting of signing key ID

### Why did it change

The process-cri-callback lambda uses the CriApiService to generated token requests for CRIs.

These token requests are signed, and we use a custom KMS signer to do this. The signer needs a KMS key Id configuring on it.

We were doing this in the initialisation phase of the lambda. Moving to snapstart would meant that value was permanently baked in to the lambda. This would mean we couldn't switch signing keys easily if required.

This change defers the setting of the key id to the runtime. We use a KMS signer when running in a live enviroment, and a normal ES245 signer in the tests, which is why we need to check if the signer is of a particular type before trying to set the KMS key ID on it.

### Issue tracking

- [PYIC-4982](https://govukverify.atlassian.net/browse/PYIC-4982)

[PYIC-4982]: https://govukverify.atlassian.net/browse/PYIC-4982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -137,6 +138,11 @@ public HTTPRequest buildFetchAccessTokenRequest(
                                                             ConfigurationVariable.JWT_TTL_SECONDS)))
                                     .toEpochSecond(),
                             secureTokenHelper.generate());
+
+            if (signer instanceof KmsEs256Signer kmsEs256Signer) {
+                kmsEs256Signer.setKeyId(configService.getSigningKeyId());
+            }
```
Joe-Edwards-GDS: I think this is fine as-is, but if it's leaving you a bit unsatisfied, then a more robust approach might be to use a factory instead 🏭

i.e. create a `KmsSignerFactory` (which can cache the AWS KMS client), and then here we call: `kmsSignerFactory.getSigner(configService.getSigningKeyId())` to dynamically get an instance using the appropriate signing key.

(Alternatively, we could construct a new one every time - but that's much harder to test)

Joe-Edwards-GDS (dismissed):

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
Wynndow: Do you mean as an alternative to passing the signer in from the handler, and instead construct it in this class using a factory?
I'm not quite seeing how it helps with the need to use a different type of signer for the tests - hence why injecting it like we do is useful?

Joe-Edwards-GDS: The thing I don't like about this approach is that anybody using the `KmsEs256Signer` has to remember to call `setKeyId` before they can use it - which is a bit surprising.

If you inject a factory instead, then you can still inject a different version for the tests, but it also forces you to provide the key id in order to get a signer.

Wynndow: Ah ok, I'm with you. Will fix it up.

Joe-Edwards-GDS (approved):
Thanks for doing this, I think it's cleaner result!

PR #1642, opened 2024-02-16
PYIC-4534 Journey map diagram split

Joe-Edwards-GDS:
- Updates the journey map diagram to support the split journeys
- Adds ability to click on a page or journey transition node a second time to follow the link immediately
- Adds links in the header to switch to a specific journey

image

image

Future updates include:
- Sourcing the configuration options from all journeys, rather than just the selected one
- Adding a view that combines all the journeys together


journey-map/public/render.mjs
```diff
@@ -126,20 +127,39 @@ const renderTransitions = (journeyMap, formData) => {
...
             eventsByTarget[target].push(eventName);
+
+            if (!journeyMap[target]) {
+                if (targetJourney) {
+                    journeyMap[target] = {
+                        response: {
+                            type: 'journeyTransition',
+                            targetJourney,
+                            targetState,
+                        }
```
Joe-Edwards-GDS, line 154: We add a fictional `journeyTransition` node when we encounter a journey transition

Wynndow (approved):
After spinning this up it makes me thing we should rename the "ipv-core-main-journey" journey to something less now weird.

PR #1644, opened 2024-02-16
PYIC 4961 - Automatic Removal of Weaker Operational Profiles

sammyiyke:
### What changed

- Adds automatic removal of weaker operational profiles during new IPV journey
- Adds automatic removal of weaker operational profiles upon return from F2F
- Adds tests for automatic removal of weaker operational profiles during new IPV journey
- Adds tests for automatic removal of weaker operational profiles upon return from F2F
- Fixes bug with cases where the evidence is empty for a VC

### Why did it change

[PYIC 4255 - Automated Removal of Weaker Operational Profiles in VC Store upon Awarding Superior GPG45 Profiles](https://govukverify.atlassian.net/browse/PYIC-4255)

To ensure that CORE maintains the most relevant and highest strength identity profiles for users, thereby enhancing the security and integrity of the identity verification process.

### Issue tracking

- [PYIC-4255](https://govukverify.atlassian.net/browse/PYIC-4255)

[PYIC-4255]: https://govukverify.atlassian.net/browse/PYIC-4255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -451,11 +452,26 @@ private boolean achievedWithGpg45Profile(
                         ? gpg45ProfileEvaluator.getFirstMatchingProfile(
                                 gpg45Scores, requestedVot.getSupportedGpg45Profiles())
                         : Optional.empty();
+        List<SignedJWT> gpg45Credentials = new ArrayList<>();
+        for (SignedJWT credential : credentials) {
+            if (!VcHelper.isOperationalProfileVc(credential)) {
+                gpg45Credentials.add(credential);
+            }
+        }
```
Joe-Edwards-GDS, line 469: We've got a helper for this already - `VcHelper.filterVCBasedOnProfileType`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -451,11 +452,26 @@ private boolean achievedWithGpg45Profile(
...
 
         // Successful match
         if (matchedGpg45Profile.isPresent()) {
+            // remove weaker operational profile
+            if (configService.enabled(INHERITED_IDENTITY.getName())) {
+                for (SignedJWT credential : credentials) {
+                    if (VcHelper.isOperationalProfileVc(credential)) {
+                        verifiableCredentialService.deleteVcStoreItem(
+                                auditEventUser.getUserId(), HMRC_MIGRATION_CRI);
```
Joe-Edwards-GDS, line 478: This is a bit weird - when checking the VC, we're inspecting the VC and using the presence of the `vot` field, but when deleting we're hardcoding `HMRC_MIGRATION_CRI`.

Perhaps the easiest would be to iterate through `vcStoreItems` and delete the `HMRC_MIGRATION_CRI` entry if there is one?

(More generally I think we need to start being more consistent about whether we're using 'presence of VOT' or 'issuer = HMRC Migration' to identify operational profiles - at the moment they're equivalent, but they may not be in future)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -129,13 +132,33 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
             }
 
-            return checkForMatchingGpg45Profile(
+            List<SignedJWT> gpg45Credentials = new ArrayList<>();
+
+            for (SignedJWT credential : credentials) {
+                if (!VcHelper.isOperationalProfileVc(credential)) {
+                    gpg45Credentials.add(credential);
+                }
+            }
```
Joe-Edwards-GDS: (ditto with the helper)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -129,13 +132,33 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
                             ipvSessionItem,
                             clientOAuthSessionItem,
-                            credentials,
-                            ipAddress)
-                    .toObjectMap();
+                            gpg45Credentials,
+                            ipAddress);
+
+            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)) {
+                if (checkForMatchingGpg45Profile.equals(JOURNEY_MET)) {
```
Joe-Edwards-GDS: This condition is a bit weird - I'd prefer to apply this 'directly' when matching a profile (rather than checking the journey response).

Can we make that method return the matched profile (or just true/false) and move the routing-specific logic here? - Then this check can be `if (metProfile) { ... }` or similar

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -129,13 +132,33 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
+                    for (SignedJWT credential : credentials) {
+                        if (VcHelper.isOperationalProfileVc(credential)) {
+                            verifiableCredentialService.deleteVcStoreItem(
+                                    userId, HMRC_MIGRATION_CRI);
+                        }
```
Joe-Edwards-GDS: As above

Joe-Edwards-GDS (commented):
A couple of comments for you to think about :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sammyiyke: True, but this takes `VcStoreItem` instead of the `SignedJWT` we are trying to filter at this stage. 

This would mean creating another instance of `VcHelper.filterVCBasedOnProfileType` to handle `SignedJWT`.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
Joe-Edwards-GDS: In this case I think we could filter the VC list _before_ parsing into `SignedJWT`s? - I don't think there's any need to handle non-GPG45 VCs here.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Ugh, yeah the difference between `SignedJWT` and `VCStoreItem` is awkward. I'm hoping we can fix this as part of PYIC-3908, but we might have to do some groundwork first.

If you can't see a neat fix then perhaps this is ok for now.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
sammyiyke: I don't think `VcHelper.filterVCBasedOnProfileType` filters operational profiles. It currently returns either `OPERATIONAL_CRIS` or none `OPERATIONAL_CRIS`. It does not filter out the operational profiles matching `HMRC_MIGRATION_CRI`.

Joe-Edwards-GDS: Yeah, this is what I meant in the other comment... At the moment we have two different ways of identifying operational profiles:
- A VC with a `vot` matching an operational profile
- A VC from `HMRC_MIGRATION_CRI`

At the moment these are exactly equivalent, but it'd be nice to be consistent (or be clear about when we should use one vs. the other).

In this case, could you do `VcHelper.filterVCBasedOnProfileType(vcStoreItems, ProfileType.GPG45)`? I think that better reflects what you're trying to do - i.e. find all the GPG45 VCs.

sammyiyke: Yes, that's exactly what I did here with:

```
List<VcStoreItem> vcStoreItems =
                    VcHelper.filterVCBasedOnProfileType(
                            verifiableCredentialService.getVcStoreItems(userId), ProfileType.GPG45);
```

but it is still including the `HMRC_MIGRATION_CRI` vc in the returned list. 

Still investigating as the is not supposed to be the expected result. Maybe I am wrong.

PR #1655, opened 2024-02-19
PYIC-4896: Modify tests to generate their own VCs

lwebb-iig:
Replace the use of base64 encoded strings with generative code for VCs so that VCs created for tests are presented in a human-readable format.

### What changed

- Created TestVc class as a mock class and generator for VCs
- Created VcFixtures interface to use above class and generate VCs identical to those in TestFixtures
- Replaces TestFixture values with VcFixture generators across test files

### Why did it change

Unit tests difficult to understand or make sense of using hardcoded base64 encoded strings

### Issue tracking

- [PYIC-4896](https://govukverify.atlassian.net/browse/PYIC-4896)

### Other considerations

[PYIC-4896]: https://govukverify.atlassian.net/browse/PYIC-4896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/build-proven-user-identity-details/src/test/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandlerTest.java
```diff
@@ -103,6 +112,20 @@ void setUp() {
         when(mockIpvSessionItem.getVot()).thenReturn(Vot.P2);
     }
 
+    @BeforeAll
+    static void setVcs() throws Exception {
```
Wynndow: Rather than setting the values in this before all method, could you just set the values when you declare the fields?
Alternatively, just use the imported methods directly in the tests?

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriServiceTest.java
```diff
@@ -50,45 +42,51 @@
...
     @Mock private HttpResponse<String> mockHttpResponse;
     @Captor private ArgumentCaptor<HttpRequest> requestCaptor;
     @InjectMocks private TicfCriService ticfCriService;
 
+    @BeforeAll
+    static void setVcs() throws Exception {
+        VC_ADDRESS = vcAddressOne();
+        String M1A_ADDRESS_VC = vcAddressM1a();
+        String M1A_EXPERIAN_FRAUD_VC = vcExperianFraudM1a();
+        String M1B_DCMAW_VC = vcDcmawM1b();
```
Wynndow: As these aren't constants, and are only being used in this method, do you need to create these vars? You could just construct the objects with direct calls to the imported method? And could you do that when declaring the fields, rather than in a before all block?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandlerTest.java
```diff
@@ -58,6 +61,25 @@ class ProcessCriCallbackHandlerTest {
...
+                new ProcessCriCallbackHandler(
+                        mockConfigService,
+                        mockIpvSessionService,
+                        mockCriOAuthSessionService,
+                        mockVerifiableCredentialJwtValidator,
+                        mockClientOAuthSessionDetailsService,
+                        mockCriApiService,
+                        mockCriStoringService,
+                        mockCriCheckingService);
+    }
```
Wynndow: Is this setUp method needed? The `@InjectMocks` annotation on the `processCriCallbackHandler` field should already be doing exactly this for us?

libs/common-services/build.gradle
```diff
@@ -25,6 +25,8 @@ dependencies {
 			"uk.org.webcompere:system-stubs-jupiter:2.1.3",
 			"org.hamcrest:hamcrest:2.2"
 	testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+	testCompileOnly("org.projectlombok:lombok:$rootProject.ext.dependencyVersions.lombok")
+	testAnnotationProcessor("org.projectlombok:lombok:$rootProject.ext.dependencyVersions.lombok")
```
Wynndow, line 29: Are these needed? We already have lombok declared in the `compileOnly` and `annotationProcessor` configurations, and I'd expect everything declared in those to be available in the `testt` versions too.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/TestFixtures.java
```diff
@@ -49,96 +49,6 @@ public interface TestFixtures {
 
     // As this is public test data you can generate VC signed JWTs using https://jwt.io/. Use the EC
     // public/private keys above to sign test VCs
-    String VC_PASSPORT_NON_DCMAW_SUCCESSFUL =
```
Wynndow, line 52: Nice..

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                            .identityFraudScore(1)
+                            .build());
+
+    List<TestVc.TestEvidence> FRAUD_FAILED_EVIDENCE =
+            List.of(
+                    TestVc.TestEvidence.builder()
+                            .checkDetails(null)
+                            .txn("RB000103490087")
+                            .identityFraudScore(0)
+                            .build());
```
Wynndow, line 50: Would this typically come with a CI? I'm not sure, but I suspect so

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+
+    List<TestVc.TestEvidence> FRAUD_EVIDENCE_SCORE_ONE =
```
Wynndow, line 52: This has a score of 2 - might be worth renaming? `FRAUD_EVIDENCE` has a score of 1 but doesn't have any check details. Could that be updated?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                            .build());
+
+    List<TestVc.TestEvidence> FRAUD_EVIDENCE_CRI_STUB_CHECK =
+            List.of(
+                    TestVc.TestEvidence.builder()
+                            .checkDetails(null)
+                            .type(CRI_STUB_CHECK_EVIDENCE_TYPE)
+                            .txn("some-uuid")
+                            .identityFraudScore(1)
+                            .build());
```
Wynndow, line 85: Is this needed as well as the other fraud score 1 evidence block above, or can we reduce to just using one of them?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                            .txn("cbf56c46-f33a-49de-8be0-c6318d8eecfc")
+                            .strengthScore(3)
+                            .validityScore(2)
+                            .activityHistoryScore(1)
+                            .checkDetails(
+                                    List.of(
+                                            Map.of(
+                                                    "checkMethod", "data",
+                                                    "identityCheckPolicy", "published",
+                                                    "activityFrom", "1982-05-23")))
```
Wynndow, line 119: Interested how this differs from the DCMAW_EVIDENCE block above? It's not super clear from the names when you might use one over the other?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                    "issuedBy", "DVA",
+                    "issueNumber", 123456,
+                    "issueDate", "2018-04-19");
+
+    Map<String, Object> DRIVING_PERMIT_2 =
+            Map.of(
+                    "personalNumber", "PARKE710112PBFGA",
+                    "expiryDate", "2032-02-02",
+                    "issuedBy", "DVLA",
+                    "issueDate", "2005-02-02");
```
Wynndow, line 276: Maybe these could have clearer names to show that one is DVLA and one DVA?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+public interface VcFixtures {
+    List<TestVc.TestEvidence> SUCCESSFUL_EVIDENCE =
+            List.of(TestVc.TestEvidence.builder().strengthScore(4).validityScore(2).build());
+    List<TestVc.TestEvidence> UNSUCCESSFUL_EVIDENCE =
+            List.of(TestVc.TestEvidence.builder().strengthScore(4).validityScore(0).build());
+    List<TestVc.TestEvidence> D02_CI_EVIDENCE =
+            List.of(
+                    TestVc.TestEvidence.builder()
+                            .strengthScore(4)
+                            .validityScore(2)
```
Wynndow, line 31: With this CI, I think the validity here would be `0`

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_FAMILY_NAME;
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_GIVEN_NAME;
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VC_NAME_PARTS;
+import static uk.gov.di.ipv.core.library.domain.VerifiableCredentialConstants.VERIFIABLE_CREDENTIAL_TYPE;
+import static uk.gov.di.ipv.core.library.helpers.VerifiableCredentialGenerator.generateVerifiableCredential;
+
+public interface VcFixtures {
+    List<TestVc.TestEvidence> SUCCESSFUL_EVIDENCE =
+            List.of(TestVc.TestEvidence.builder().strengthScore(4).validityScore(2).build());
+    List<TestVc.TestEvidence> UNSUCCESSFUL_EVIDENCE =
```
Wynndow, line 25: I'm not sure this is valid - if it has zero validity score it should probably come with a CI?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+        return generateVerifiableCredential(
+                TestVc.builder()
+                        .credentialSubject(credentialSubject)
+                        .evidence(SUCCESSFUL_EVIDENCE)
+                        .build());
+    }
+
+    static String vcAddressOne() throws Exception {
+        TestVc.TestCredentialSubject credentialSubject =
+                TestVc.TestCredentialSubject.builder().address(List.of(ADDRESS_1)).build();
```
Wynndow: I initially thought this was going to be a VC from the address CRI, until I realised that it will default to being a passport VC. Wonder if a different method name might make it clearer?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+
+    static String vcAddressM1a() throws Exception {
+        TestVc.TestCredentialSubject credentialSubject =
+                TestVc.TestCredentialSubject.builder()
+                        .name(null)
+                        .address(List.of(ADDRESS_3))
+                        .passport(null)
+                        .build();
+        return generateVerifiableCredential(
+                TestVc.builder().credentialSubject(credentialSubject).evidence(null).build(),
```
Wynndow: Is it imporant to set evidence to null in these address VC?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+        TestVc.TestCredentialSubject credentialSubject =
+                TestVc.TestCredentialSubject.builder()
+                        .name(null)
+                        .address(MULTIPLE_ADDRESSES_VALID)
+                        .passport(null)
+                        .build();
+        return generateVerifiableCredential(
+                TestVc.builder().credentialSubject(credentialSubject).evidence(null).build(),
+                "urn:uuid:e6e2e324-5b66-4ad6-8338-83f9f837e345",
+                "https://review-a.integration.account.gov.uk",
```
Wynndow, line 393: Could these strings be constants defined somewhere?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                "https://review-f.staging.account.gov.uk",
+                Instant.ofEpochSecond(1704290386));
+    }
+
+    static String vcDrivingPermit() throws Exception {
+        TestVc.TestCredentialSubject credentialSubject =
+                TestVc.TestCredentialSubject.builder()
+                        .address(List.of(ADDRESS_4))
+                        .drivingPermit(List.of(DRIVING_PERMIT))
+                        .passport(null)
```
Wynndow, line 503: I'm wondering if defining passport with a default value in `TestCredentialSubject` might not be the best thing to do now. Forces you to set all these nulls, and it could easily be missed when creating new fixtures.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                TestVc.builder()
+                        .evidence(FRAUD_EVIDENCE_CRI_STUB_CHECK)
+                        .credentialSubject(credentialSubject)
+                        .build(),
+                "urn:uuid:7fadacac-0d61-4786-aca3-8ef7934cb092",
+                "https://review-f.staging.account.gov.uk",
+                Instant.ofEpochSecond(1704290386));
+    }
+
+    static String vcDrivingPermit() throws Exception {
```
Wynndow: These `vcDrivingPermit` fixtures are for DCMAW, rather than the driving licence CRI. Might be worth including that in the method name to be super clear.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                        .build(),
+                "urn:uuid:51dfa9ac-8624-4b93-aa8f-99ed772ff0ec",
+                "https://review-xx.account.gov.uk",
+                Instant.ofEpochSecond(1697097326));
+    }
+
+    static String vcTicf() throws Exception {
+        return generateVerifiableCredential(
+                TestVc.builder()
+                        .credentialSubject(null)
```
Wynndow, line 638: Should this have a subject?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                        .evidence(TICF_EVIDENCE)
+                        .type(
+                                new String[] {
+                                    VERIFIABLE_CREDENTIAL_TYPE, RISK_ASSESSMENT_CREDENTIAL_TYPE
+                                })
+                        .build(),
+                "urn:uuid:d1823066-2137-4380-b0ba-4b61947e08e6",
+                "https://tcif.stubs.account.gov.uk",
```
Wynndow:
```suggestion
                "https://ticf.stubs.account.gov.uk",
```

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                                                .verificationScore(2)
+                                                .build()))
+                        .credentialSubject(credentialSubject)
+                        .build(),
+                "test-subject",
+                "https://review-k.integration.account.gov.uk",
+                Instant.ofEpochSecond(1653403140));
+    }
+
+    static String vcDcmawM1b() throws Exception {
```
Wynndow: Is this signifcantly different to the driving permit VC above? Can they be reduced?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+        TestVc.TestCredentialSubject credentialSubject =
+                TestVc.TestCredentialSubject.builder()
+                        .name(List.of(MARY_WATSON_NAME))
+                        .passport(
+                                List.of(
+                                        Map.of(
+                                                "expiryDate",
+                                                "2030-01-01",
+                                                "documentNumber",
+                                                "824159121")))
```
Wynndow, line 718: I think this should have a `icaoIssuerCode`

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -0,0 +1,890 @@
...
+                                                "expiryDate",
+                                                "2030-07-13")))
+                        .build();
+        return generateVerifiableCredential(
+                TestVc.builder()
+                        .evidence(
+                                List.of(
+                                        TestVc.TestEvidence.builder()
+                                                .txn("some-uuid")
+                                                .type(CRI_STUB_CHECK_EVIDENCE_TYPE)
```
Wynndow, line 779: I'm not sure what this type should be, but I'm don't think it should be `CriStubCheck`?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
```diff
@@ -0,0 +1,94 @@
...
+                List.of(
+                        Map.of(
+                                VC_NAME_PARTS,
+                                List.of(
+                                        new NameParts("KENNETH", VC_GIVEN_NAME),
+                                        new NameParts("DECERQUEIRA", VC_FAMILY_NAME))));
+
+        @Builder.Default private List<BirthDate> birthDate = List.of(new BirthDate("1965-07-08"));
+
+        @Builder.Default
```
Wynndow, line 55: Mentioned above, but wondering if the default should be dropped as it adds noise to most of the fixtures.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
```diff
@@ -0,0 +1,94 @@
...
+                        Map.of(
+                                "documentNumber", "321654987",
+                                "icaoIssuerCode", "GBR",
+                                "expiryDate", "2030-01-01"));
+
+        private List<Object> address;
+        private List<Object> socialSecurityRecord;
+        private List<Object> drivingPermit;
+        private List<Object> residencePermit;
+        private List<Object> idCard;
```
Wynndow, line 67: I was wondering if these should all have their objects modeled as classes. That way it's easier to enforce that the correct attributes are being used.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
```diff
@@ -49,7 +52,63 @@ public static String generateVerifiableCredential(
...
+                new JWTClaimsSet.Builder()
+                        .claim(SUBJECT, "urn:uuid:811cefe0-7db6-48ad-ad89-0b93d2259980")
+                        .claim(ISSUER, "https://review-p.staging.account.gov.uk")
+                        .claim(NOT_BEFORE, now.getEpochSecond())
+                        .claim(VC_CLAIM, vcClaim)
+                        .build();
+        return signTestJWT(claimsSet);
+    }
+
+    public static String generateVerifiableCredential(
```
Wynndow, line 90: Could the other overloads of this method delegate to this method - the most flexible one?

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -82,27 +82,7 @@
...
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_INHERITED_IDENTITY_MIGRATION_WITH_NO_EVIDENCE;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_NINO_MISSING_SOCIAL_SECURITY_RECORD;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_NINO_SUCCESSFUL;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_NINO_UNSUCCESSFUL;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_PASSPORT_MISSING_BIRTH_DATE;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_PASSPORT_MISSING_NAME;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_PASSPORT_MISSING_PASSPORT;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_PASSPORT_NON_DCMAW_SUCCESSFUL;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_TICF;
+import static uk.gov.di.ipv.core.library.fixtures.VcFixtures.*;
```
Wynndow, line 85: Wildcard import here

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -1352,7 +1375,7 @@ void shouldReturnEmptyWhenMissingDrivingPermitProperty() throws Exception {
                         USER_ID_1, "test-sub", Vot.P2, emptyContraIndicators);
 
         // Assert
-        assertNull(credentials.getDrivingPermitClaim());
+        assertTrue(credentials.getDrivingPermitClaim().isNull());
```
Wynndow, line 1323: Could this have stayed as `assertNull`?

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelperTest.java
```diff
@@ -143,28 +205,24 @@ void shouldExtractAgeFromCredentialWithInvalidBirthDate() throws ParseException
...
-                VcHelper.checkIfDocUKIssuedForCredential(
-                        SignedJWT.parse(VC_PASSPORT_NON_DCMAW_SUCCESSFUL_WITH_ICAOCODE)));
-    }
-
-    @Test
-    void shouldCheckIfDocUKIssuedForCredentialWithMissingICAOCode() throws ParseException {
-        assertTrue(
+    void shouldCheckIfDocUKIssuedForCredential() throws ParseException {
+        assertEquals(
+                Boolean.TRUE,
```
Wynndow, line 166: Can this just stay as `assertTrue`?

Same for the changes below

Wynndow (commented):
Overall I think the approach is good, just a few questions on the test data itself. Also wondeirng about how the fixtures are brought into the tests themselves.

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelperTest.java
lwebb-iig: This was a suggested change from a static analysis warning

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
lwebb-iig: Not sure what you mean by this or what the required change would be?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
lwebb-iig: I think that's more the work of Mike's ticket https://govukverify.atlassian.net/browse/PYIC-5032

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
lwebb-iig: `assertNull` is not the same as `isNull` when called on a JsonNode (which is what the credentials. getDrivingPermitClaim() now returns

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
Wynndow: So I was thinking the method on line 70 would become:
```
    public static String generateVerifiableCredential(TestVc vcClaim) throws Exception {
        return generateVerifiableCredential(
                vcClaim,
                "urn:uuid:811cefe0-7db6-48ad-ad89-0b93d2259980",
                "https://review-p.staging.account.gov.uk",
                null,
                null,
                null);
    }
```

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
Wynndow: Oh intereseting. Fair

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelperTest.java
Wynndow: ...weird

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
Wynndow: Yeah that makes sense. Would be good to come back and update these fixtures after his is in

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -129,9 +129,9 @@ public static Integer extractAgeFromCredential(SignedJWT credential) throws Pars
         var jwtClaimsSet = credential.getJWTClaimsSet();
         var vc = (JSONObject) jwtClaimsSet.getClaim(VC_CLAIM);
         var credentialSubject = (JSONObject) vc.get(VC_CREDENTIAL_SUBJECT);
-        if (credentialSubject != null) {
+        if (credentialSubject != null && !credentialSubject.isEmpty()) {
             var birthDateArr = (JSONArray) credentialSubject.get(VC_BIRTH_DATE);
-            if (birthDateArr != null) {
+            if (birthDateArr != null && !birthDateArr.isEmpty()) {
```
Joe-Edwards-GDS, line 136: Does this happen? (Is it in our example VCs?)

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandlerTest.java
```diff
@@ -86,6 +85,17 @@ class CallTicfCriHandlerTest {
...
+    private static String M1A_EXPERIAN_FRAUD_VC;
+    private static String M1B_DCMAW_VC;
+
+    @BeforeAll
+    static void setVcs() throws Exception {
+        M1A_ADDRESS_VC = vcAddressM1a();
+        M1A_EXPERIAN_FRAUD_VC = vcExperianFraudM1a();
+        M1B_DCMAW_VC = vcDcmawM1b();
+        VC_IN_STORE = List.of(M1B_DCMAW_VC, M1A_ADDRESS_VC, M1A_EXPERIAN_FRAUD_VC);
+    }
```
Joe-Edwards-GDS: I still don't love these static initialisation blocks, but happy to leave it if you don't think it's easily fixable. Feels like we could either
- Inline the `vcAddressM1a()` etc. calls, so we create them on-demand
- Turn the exception into a runtime exception and have TestFixtures contain a static `M1A_ADDRESS_VC` again, that's just built directly.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
lwebb-iig: @Joe-Edwards-GDS It _can_ happen now when initializing a testVc object - It does now happen in some of the birth date VCs

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandlerTest.java
lwebb-iig: I think I need to add static timestamps to these VCs so we can create them inline - I think the VCs being generated at multiple points in the tests with differing timestamps was causing issues

Joe-Edwards-GDS: Ok fair enough. The second option might also minimise the changes needed to existing tests, so hopefully that's a positive :)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: I guess what I mean is, can it happen with production VCs? If it's happening in our test VCs only, then they're not representative :/

(Either way this doesn't seem like a bad change, it's just suspicious that it's not been noticed before)

libs/common-services/build.gradle
lwebb-iig: I get errors complaining about not being able to import relevant `lombok` dependencies if I remove these lines

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriServiceTest.java
```diff
@@ -50,44 +42,47 @@
...
     @Mock private HttpClient mockHttpClient;
     @Mock private VerifiableCredentialJwtValidator mockVerifiableCredentialJwtValidator;
     @Mock private HttpResponse<String> mockHttpResponse;
     @Captor private ArgumentCaptor<HttpRequest> requestCaptor;
     @InjectMocks private TicfCriService ticfCriService;
+    private static String VC_ADDRESS;
+
+    @BeforeAll
+    static void setVcs() throws Exception {
+        credentials = List.of(M1B_DCMAW_VC, M1A_EXPERIAN_FRAUD_VC, M1A_ADDRESS_VC);
```
Joe-Edwards-GDS: Does this still need to be done statically?

It's not massively important, but it feels like we're a little inconsistent. If you'd like to change it, it might also be easier to merge this and then fix up after :)

libs/common-services/build.gradle
Joe-Edwards-GDS: I'm guessing this is the first place we're using lombok directly in test code(?), and I think this is the correct fix.

(test code has a dependency on the output dependencies of main code, but not compile/annotation dependencies)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
Joe-Edwards-GDS: I don't know if that's _necessarily_ true if it's a non-breaching CI, but it's less common certainly.

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriServiceTest.java
lwebb-iig: Oops, no I missed this - I'll revert this to initialising inline with the `credentials` variable

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
lwebb-iig: Unsure - but theoretically we could map something from a production VC to an empty array, in which case we should probably catch this?

Joe-Edwards-GDS: Yeah, I think catching this case is fine - I'm just worried that our test cases aren't representative.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
Joe-Edwards-GDS: I think that probably makes sense since the UX is so weird if you want a different doc. If you want you could add a new static method for `defaultPassport()` or something that sets it for you.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
Joe-Edwards-GDS: I don't think they do (although it's not 100% clear from the RFC...)

Joe-Edwards-GDS (approved):
Looks like there are still quite a few outstanding comments from Chris that would be good to address, but we can merge this in the meantime and address in a follow-up PR.

PR #1672, opened 2024-02-20
PYIC-5037: Enable snapstart in int and prod

Wynndow:
**To be merged shortly after https://github.com/govuk-one-login/ipv-core-back/pull/1671, and when traffic is minimal. Currently slated for 7.30am on Thursday morning**

### What changed

Enable snapstart in int and prod

This PR inclues all the changes in https://github.com/govuk-one-login/ipv-core-back/pull/1671. This is to avoid a merge conflict when that PR is merged just before this one. As such only the last two commits should be reviewed. The first two should be reviewed as part of the other PR.

### Why did it change

This removes the logic around which envs to enable snapstart on and just
applies it globally. Int and prod are the last envs to start using it.

### Issue tracking

- [PYIC-5037](https://govukverify.atlassian.net/browse/PYIC-5037)

[PYIC-5037]: https://govukverify.atlassian.net/browse/PYIC-5037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -160,42 +160,42 @@
         "filename": "deploy/template.yaml",
         "hashed_secret": "b811ac90fe7fab03f6144a17aaebc38dcf3e007b",
         "is_verified": false,
-        "line_number": 167
+        "line_number": 162
       },
       {
         "type": "Secret Keyword",
         "filename": "deploy/template.yaml",
         "hashed_secret": "690de9fd42add772818ae392cb68a4f81d1511e3",
         "is_verified": false,
-        "line_number": 227
+        "line_number": 217
       },
       {
         "type": "Secret Keyword",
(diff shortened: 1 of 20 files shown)


Joe-Edwards-GDS (dismissed):
Not sure how I feel about this way of triggering a redeployment of each lambda, but I don't have a better idea right now

Wynndow:
> Not sure how I feel about this way of triggering a redeployment of each lambda, but I don't have a better idea right now

It's not so much a case of redeploying them, it's about creating a new version. And it seems to need a code change to do that. It's possible that running it through the pipeline would actually be ok without, but in my dev env it was very unhappy and refused to deploy without.

PR #1675, opened 2024-02-21
PYIC-4255 Automated Removal of Weaker Operational Profiles

sammyiyke:
### What changed

- Adds automatic removal of weaker operational profiles during new IPV journey
- Adds automatic removal of weaker operational profiles upon return from F2F
- Adds tests for automatic removal of weaker operational profiles during new IPV journey
- Adds tests for automatic removal of weaker operational profiles upon return from F2F
- Fixes bug with cases where the evidence is empty for a VC

### Why did it change

[PYIC 4255 - Automated Removal of Weaker Operational Profiles in VC Store upon Awarding Superior GPG45 Profiles](https://govukverify.atlassian.net/browse/PYIC-4255)

To ensure that CORE maintains the most relevant and highest strength identity profiles for users, thereby enhancing the security and integrity of the identity verification process.

### Issue tracking

- [PYIC-4255](https://govukverify.atlassian.net/browse/PYIC-4255)

[PYIC-4255]: https://govukverify.atlassian.net/browse/PYIC-4255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -458,11 +461,26 @@ private boolean achievedWithGpg45Profile(
                         ? gpg45ProfileEvaluator.getFirstMatchingProfile(
                                 gpg45Scores, requestedVot.getSupportedGpg45Profiles())
                         : Optional.empty();
+        List<SignedJWT> gpg45Credentials = new ArrayList<>();
+        for (SignedJWT credential : credentials) {
+            if (!VcHelper.isOperationalProfileVc(credential)) {
+                gpg45Credentials.add(credential);
+            }
+        }
```
Wynndow: I think this could be moved into the if block below - that way we don't do this processing if we don't need to.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -458,11 +461,26 @@ private boolean achievedWithGpg45Profile(
...
 
         // Successful match
         if (matchedGpg45Profile.isPresent()) {
+            // remove weaker operational profile
+            if (configService.enabled(INHERITED_IDENTITY.getName())) {
+                for (SignedJWT credential : credentials) {
+                    if (VcHelper.isOperationalProfileVc(credential)) {
+                        verifiableCredentialService.deleteVcStoreItem(
+                                auditEventUser.getUserId(), HMRC_MIGRATION_CRI);
+                    }
```
Wynndow: You could add a `break` in after the if block - we're only going to be deleting one VC so no need to continue looping through the rest of the credentials.

Another option would be to compare the `credentials` list to the `gpg45Credentials` list. If `credentials` is longer you would know that there is an operational profile present and so can go straight ahead with the deletion. This feels a bit indirect though and would cause issues if we add another profile type.

Alternatively, we could just run the deletion every time without checking...

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -458,11 +461,26 @@ private boolean achievedWithGpg45Profile(
...
+        for (SignedJWT credential : credentials) {
+            if (!VcHelper.isOperationalProfileVc(credential)) {
+                gpg45Credentials.add(credential);
+            }
+        }
 
         // Successful match
         if (matchedGpg45Profile.isPresent()) {
+            // remove weaker operational profile
+            if (configService.enabled(INHERITED_IDENTITY.getName())) {
```
Wynndow: Might be worth considering adding a check that the requested Vot is P2. If we add P1 this'll be important.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -120,7 +123,10 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
             String govukSigninJourneyId = clientOAuthSessionItem.getGovukSigninJourneyId();
             LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
 
-            List<VcStoreItem> vcStoreItems = verifiableCredentialService.getVcStoreItems(userId);
+            List<VcStoreItem> allVcStoreItems = verifiableCredentialService.getVcStoreItems(userId);
+            List<VcStoreItem> vcStoreItems =
```
Wynndow: MIght be more clear to call this something like gpg45VcStoreItems

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -129,13 +135,27 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
...
+                            ipAddress);
+
+            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)
+                    && hasMatchingGpg45Profile) {
+                for (VcStoreItem vcStoreItem : allVcStoreItems) {
+
+                    if (vcStoreItem.getCredentialIssuer().equals(HMRC_MIGRATION_CRI)) {
+                        verifiableCredentialService.deleteVcStoreItem(
+                                vcStoreItem.getUserId(), vcStoreItem.getCredentialIssuer());
+                    }
```
Wynndow: Could use `VcHelper.filterVCBasedOnProfileType(allVcStoreItems, ProfileType.OPERATIONAL_HMRC)` and then run the deletion if the result is non empty?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -166,7 +186,7 @@ private Map<String, Object> buildJourneyErrorResponse(ErrorResponse errorRespons
     }
 
     @Tracing
-    private JourneyResponse checkForMatchingGpg45Profile(
+    private boolean checkForMatchingGpg45Profile(
```
Wynndow: Might be worth renaming this method to something like `gpg45ProfileMatched`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -458,11 +461,26 @@ private boolean achievedWithGpg45Profile(
...
+                for (SignedJWT credential : credentials) {
+                    if (VcHelper.isOperationalProfileVc(credential)) {
+                        verifiableCredentialService.deleteVcStoreItem(
+                                auditEventUser.getUserId(), HMRC_MIGRATION_CRI);
```
Joe-Edwards-GDS: Can we make this a bit more precise about what you're actually doing here - I think we're deleting any _HMRC_ inherited identity, and we should make the VC selection match that:

```
private void deleteHmrcInheritedIdentityIfPresent(List<VcStoreItem> vcStoreItems) {
    for (var vcStoreItem : vcStoreItems) {
        if (HMRC_MIGRATION_CRI.equals(vcStoreItem.getCredentialIssuer())) {
            verifiableCredentialService.deleteVcStoreItem(
                    vcStoreItem.getUserId(),
                    vcStoreItem.getCredentialIssuer());
        }
    }
}
```
I think this ensures that the thing being deleted is exactly what we're looking for. It also means that it won't pick up _other_ operational profiles, which might not want to be deleted.

Pulling it out as a separate method also allows you to implictly explain what's going on.

Joe-Edwards-GDS: This also matches the implementation you've got in `EvaluateGpg45ScoresHandler` :)

Joe-Edwards-GDS (commented):
One comment to make the two implementations match, but otherwise looking good :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sammyiyke: I was thinking the same thing. But wasn't settled on the HMRC part yet. Which I wanted to discuss with you.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sammyiyke: Can you please explain why?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sammyiyke: Changed the implementation of this check and delete, would need another review

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/service/VerifiableCredentialServiceTest.java
```diff
@@ -39,15 +39,10 @@
...
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.EXPERIAN_FRAUD_CRI;
-import static uk.gov.di.ipv.core.library.domain.CriConstants.PASSPORT_CRI;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.EC_PRIVATE_KEY;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_EXPERIAN_FRAUD_SCORE_1;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_EXPERIAN_KBV_SCORE_2;
-import static uk.gov.di.ipv.core.library.fixtures.TestFixtures.VC_PASSPORT_NON_DCMAW_SUCCESSFUL;
+import static org.mockito.Mockito.*;
```
Joe-Edwards-GDS: (wildcard imports)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: We only want to remove the operational profiles if the user has attained a stronger vot, which P2 is. P1 however isn't, so in a future scenario where we handle P1s, we wouldn't want to remove the operational profiles.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
Joe-Edwards-GDS: I think in this case using the HMRC CRI might actually be better - we don't know whether other operational profiles will be weaker/stronger

Wynndow: Does checking the issuer is HMRC differentiate between the `OPERATIONAL_HMRC` profile type? If they started using other profile types it'd still be the same issuer?

Wynndow (approved):
🚀

PR #1678, opened 2024-02-21
PYIC-4727 Changes made to support multiple feature sets.

rpayal:
## Proposed changes
Support multiple feature sets concurrently

### What changed
Changes made to support multiple feature sets.
- While searching for SSM param will start searching with 1st one  (in featureSet list) once found will return that value.
- 

### Issue tracking

- [PYIC-4727](https://govukverify.atlassian.net/browse/PYIC-4727)

### Other considerations

[PYIC-4727]: https://govukverify.atlassian.net/browse/PYIC-4727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -131,48 +131,51 @@ public SSMProvider getSsmProvider() {
         return ssmProvider;
     }
 
-    public String getFeatureSet() {
-        return featureSet;
+    public List<String> getFeatureSet() {
+        return (featureSet != null && !featureSet.isEmpty())
+                ? featureSet
+                : Collections.singletonList(null);
```
Joe-Edwards-GDS: Why isn't this just empty list (or `null`) - that would be a lot less surprising

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -131,48 +131,51 @@ public SSMProvider getSsmProvider() {
...
         }
     }
 
     public String getEnvironmentVariable(EnvironmentVariable environmentVariable) {
         return System.getenv(environmentVariable.name());
     }
 
-    public String getSsmParameter(String path) {
-        return ssmProvider.get(path);
-    }
```
Joe-Edwards-GDS, line 152: Why did this move? It feels natural to keep overrides of the same method together

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -98,20 +99,25 @@ public static String getIpvSessionId(JourneyRequest request, boolean allowNull)
...
-    public static String getFeatureSet(APIGatewayProxyRequestEvent event) {
-        String featureSet = getFeatureSet(event.getHeaders());
-        LogHelper.attachFeatureSetToLogs(featureSet);
-        return featureSet;
+    public static List<String> getFeatureSet(Map<String, String> headers) {
+        String featureSetHeaderValue = RequestHelper.getHeaderByKey(headers, FEATURE_SET_HEADER);
+        return (featureSetHeaderValue != null)
+                ? Arrays.asList(featureSetHeaderValue.split(","))
+                : null;
     }
```
Joe-Edwards-GDS, line 121: These three methods are inconsistent now - the first _always_ attaches the feature set to logs, the second only does so if nonempty, and the last never does?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
```diff
@@ -41,10 +44,12 @@ public static String getIpAddress(Map<String, String> input) {
...
     }
 
-    public static String getFeatureSet(Map<String, String> input) {
+    public static List<String> getFeatureSet(Map<String, String> input) {
         String featureSet = input.get(FEATURE_SET);
-        LogHelper.attachFeatureSetToLogs(featureSet);
-        return featureSet;
+        LogHelper.attachFeatureSetToLogs(Collections.singletonList(featureSet));
+        return (featureSet != null)
+                ? Stream.of(featureSet.split(",")).map(String::trim).toList()
```
Joe-Edwards-GDS: This is different to how we parse the featureset on other requests - is there a reason?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -36,7 +36,7 @@ public class IpvSessionItem implements DynamodbItem {
     private String emailAddress;
     // Only for passing the featureSet to the external API lambdas at the end of the user journey.
     // Not for general use.
-    private String featureSet;
+    private List<String> featureSet;
```
Joe-Edwards-GDS: I assume we won't be using any feature sets in prod? This could potentially break active sessions that are using feature sets

Joe-Edwards-GDS (commented):
I think we could have considered keeping `featureSet` as a string everywhere and just parse as a CSV in the config service (since nothing else really needs to know), and it'd probably be a much smaller change.

This feels like a more 'proper' solution though, so don't mind too much.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
rpayal: Done-reverted

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -60,4 +64,13 @@ public void addVcReceivedThisSession(String vc) {
         }
         vcReceivedThisSession.add(vc);
     }
+
+    public List<String> getFeatureSet() {
+        return Stream.of(featureSet.split(",")).map(String::trim).toList();
```
Joe-Edwards-GDS: Can `featureSet` be null here? It can be `null` when being set...

Will we ever set it with whitespace - do we need the `trim`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -60,4 +64,13 @@ public void addVcReceivedThisSession(String vc) {
...
+    }
+
+    public void setFeatureSet(List<String> featureSet) {
+        this.featureSet =
+                (featureSet != null) ? featureSet.stream().collect(Collectors.joining(",")) : null;
```
Joe-Edwards-GDS: (If you like, you can do this slightly more concisely with `String.join(",", featureSet)`)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -131,48 +131,49 @@ public SSMProvider getSsmProvider() {
...
-        if (featureSet == null || featureSet.isBlank()) {
+    public void setFeatureSet(List<String> featureSet) {
+        if (featureSet == null || featureSet.isEmpty()) {
             this.featureSet = null;
         } else {
-            this.featureSet = featureSet;
+            this.featureSet =
+                    featureSet.stream()
+                            .filter(fs -> (fs != null && !fs.isEmpty() && !fs.isBlank()))
+                            .toList();
```
Joe-Edwards-GDS: It feels as though we're duplicating a lot of empty/null checks all over the place...

Can we _either_:
- Do all the validation/parsing when reading from the request object (in `RequestHelper` etc.), so by the time we get to here we don't need to re-run it
- Keep it as a `String` elsewhere in the code, and only parse into a list (and do all the validation/parsing) here, once we need to

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -131,48 +131,49 @@ public SSMProvider getSsmProvider() {
         return ssmProvider;
     }
 
-    public String getFeatureSet() {
-        return featureSet;
+    public List<String> getFeatureSet() {
+        return (featureSet != null && !featureSet.isEmpty()) ? featureSet : null;
```
Joe-Edwards-GDS: Do we actually need to check for empty here - what will break if it is?

(If anything, having an empty list works better than null, since we don't need to add null checks everywhere!)

rpayal: To handle existing test - shouldNormaliseNullAndEmptyFeatureSetsToNull

rpayal:
Tested in dev env. Also ran regre test pack against dev env.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -60,4 +64,15 @@ public void addVcReceivedThisSession(String vc) {
         }
         vcReceivedThisSession.add(vc);
     }
+
+    public List<String> getFeatureSet() {
+        return (featureSet != null)
+                ? Stream.of(featureSet.split(",")).map(String::trim).toList()
+                : null;
```
Joe-Edwards-GDS: (do we also need to update this to empty list?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: (still applies)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
```diff
@@ -405,15 +370,29 @@ void shouldReturnListOfClientRedirectUrls(String testDataSet, String featureSet)
...
+                    configService.enabled(TestFeatureFlag.TEST_FEATURE));
+        } else {
+            if (fsList.get(0).equals("FS07")) {
+                assertEquals(
+                        Boolean.parseBoolean(testConfiguration.getExpectedValue(fsList.get(0))),
+                        configService.enabled(TestFeatureFlag.TEST_FEATURE));
+            } else {
+                assertEquals(
+                        Boolean.parseBoolean(testConfiguration.getExpectedValue(fsList.get(1))),
+                        configService.enabled(TestFeatureFlag.TEST_FEATURE));
```
Joe-Edwards-GDS: I think it's pretty unclear having conditional assertions in the test - perhaps this should just be 3 different tests (with no parameterisation)? That way you can write the assertions you want, and give them meaningful names.

If you want to share the setup, you could pull out a method, but tbh I'm not that fussed.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
```diff
@@ -405,15 +370,29 @@ void shouldReturnListOfClientRedirectUrls(String testDataSet, String featureSet)
     }
 
     @ParameterizedTest
-    @CsvSource({"FEATURE_FLAGS,", "FEATURE_FLAGS,FS07"})
+    @CsvSource({"FEATURE_FLAGS,", "FEATURE_FLAGS,'FS07,DS01'", "FEATURE_FLAGS,'DS01,FS07'"})
     void shouldGetNamedFeatureFlag(String testDataSet, String featureSet) {
         environmentVariables.set("ENVIRONMENT", "test");
-        configService.setFeatureSet(featureSet);
+        List<String> fsList =
+                featureSet != null ? List.of(featureSet.split(",")) : Collections.emptyList();
```
Joe-Edwards-GDS: Maybe move this logic to `getFeatureSet` and use it?

...perhaps we need a helper somewhere that parses a featureSet string into a list (which you can use in the main code too).

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
```diff
@@ -260,7 +261,7 @@ private CriCallbackRequest buildCriCallbackRequest(Request request) {
                 .error(requestBody.get("error"))
                 .errorDescription(requestBody.get("errorDescription"))
                 .ipAddress(request.headers(IP_ADDRESS))
-                .featureSet(request.headers(FEATURE_SET))
+                .featureSet(Arrays.asList(request.headers(FEATURE_SET).split(",")))
```
Joe-Edwards-GDS, line 264: This is using `Arrays.asList` but the others are using `Stream.of(..).toList()`?

This seems cleaner (and avoids an unnecessary stream), so maybe prefer that? If you create a helper for this then you can make sure it's uniform everywhere :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: Its now a private method so moved down.

Joe-Edwards-GDS: Ok sure - I think it's a bit suspicious having one override be public and one be private. I wonder if we should just inline the private one - it's only used in one place.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
rpayal: Done - refactor test methods,  created different test methods.

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: While refactoring should've done inline. - Done

Joe-Edwards-GDS (approved):
Thanks Rohit - I think the QAs will really appreciate this!

PR #1685, opened 2024-02-22
PYIC-5001: Reprove Identity Flag + Breaching CIs Should Fail/Use Mitigation Route

sammyiyke:
### What changed

[PYIC 5001](https://govukverify.atlassian.net/browse/PYIC-5001) Reprove identity flag + breaching CIs should fail/use mitigation route

- Adds new journey for reset identity for user with CI and no mitigation
- Fixes reset identity + CI mitigation bug 

### Why did it change
If a user has breaching CIs and the reprove identity flag, the reprove flag triggers a reset, but then begins a normal identity proving journey, which will fail after the first CRI.  Instead, we still need to respect CIs after the reset.

### Issue tracking

- [PYIC 5001](https://govukverify.atlassian.net/browse/PYIC-5001)

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -68,6 +68,7 @@ import static uk.gov.di.ipv.core.library.helpers.RequestHelper.getIpAddress;
 import static uk.gov.di.ipv.core.library.helpers.RequestHelper.getIpvSessionId;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_ERROR_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_F2F_FAIL_PATH;
+import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_FAIL_WITH_CI_AND_FORCED_RESET_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_FAIL_WITH_CI_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_IN_MIGRATION_REUSE_PATH;
 import static uk.gov.di.ipv.core.library.journeyuris.JourneyUris.JOURNEY_IPV_GPG45_MEDIUM_PATH;
@@ -99,6 +100,8 @@ public class CheckExistingIdentityHandler
             new JourneyResponse(JOURNEY_RESET_GPG45_IDENTITY_PATH);
     private static final JourneyResponse JOURNEY_FAIL_WITH_CI =
             new JourneyResponse(JOURNEY_FAIL_WITH_CI_PATH);
+    private static final JourneyResponse JOURNEY_FAIL_WITH_CI_AND_FORCED_RESET =
+            new JourneyResponse(JOURNEY_FAIL_WITH_CI_AND_FORCED_RESET_PATH);
     public static final List<Vot> SUPPORTED_VOTS_BY_STRENGTH =
             List.of(Vot.P2, Vot.PCL250, Vot.PCL200);
 
@@ -199,15 +202,6 @@ public class CheckExistingIdentityHandler
             // Clear TICF VCs
             verifiableCredentialService.deleteVcStoreItem(userId, TICF_CRI);
 
-            // Reset identity if reprove is true.
-            // or
-            // Force reset
-            Boolean reproveIdentity = clientOAuthSessionItem.getReproveIdentity();
-            if ((reproveIdentity != null && reproveIdentity)
-                    || configService.enabled(RESET_IDENTITY.getName())) {
-                return buildForceResetResponse();
-            }
-
             AuditEventUser auditEventUser =
                     new AuditEventUser(userId, ipvSessionId, govukSigninJourneyId, ipAddress);
 
@@ -221,10 +215,15 @@ public class CheckExistingIdentityHandler
             final boolean isF2FIncomplete = !Objects.isNull(f2fRequest) && !hasF2fVc;
             final boolean isF2FComplete = !Objects.isNull(f2fRequest) && hasF2fVc;
 
-            var ciScoringCheckResponse =
+            Optional<JourneyResponse> ciScoringCheckResponse =
                     checkForCIScoringFailure(
                             ipAddress, clientOAuthSessionItem, govukSigninJourneyId);
 
+            Boolean reproveIdentity = clientOAuthSessionItem.getReproveIdentity();
+            if ((reproveIdentity != null && reproveIdentity)
+                    || configService.enabled(RESET_IDENTITY.getName())) {
+                return buildForceResetResponse(ciScoringCheckResponse.orElse(null));
+            }
             if (ciScoringCheckResponse.isPresent()) {
                 return isF2FIncomplete
                         ? buildF2FIncompleteResponse(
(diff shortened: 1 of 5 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -219,10 +213,15 @@ private JourneyResponse getJourneyResponse(
             final boolean isF2FIncomplete = !Objects.isNull(f2fRequest) && !hasF2fVc;
             final boolean isF2FComplete = !Objects.isNull(f2fRequest) && hasF2fVc;
 
-            var ciScoringCheckResponse =
+            Optional<JourneyResponse> ciScoringCheckResponse =
```
Joe-Edwards-GDS: Interesting, I prefer using `var` for local variables - the IDE and compiler still know the type, but it's less verbose and makes refactoring easier!

I don't think it's a big deal though (some day I might write a style guide, but today is not that day)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -219,10 +213,15 @@ private JourneyResponse getJourneyResponse(
...
                     checkForCIScoringFailure(
                             ipAddress, clientOAuthSessionItem, govukSigninJourneyId);
 
+            Boolean reproveIdentity = clientOAuthSessionItem.getReproveIdentity();
+            if ((reproveIdentity != null && reproveIdentity)
+                    || configService.enabled(RESET_IDENTITY.getName())) {
+                return buildForceResetResponse(ciScoringCheckResponse.orElse(null));
```
Joe-Edwards-GDS, line 225: Can we just pass the optional to this method, rather than falling back to nullable types?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -274,9 +273,20 @@ private JourneyResponse getJourneyResponse(
...
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "resetIdentity flag is enabled, reset users identity."));
+            if (JOURNEY_FAIL_WITH_CI.equals(ciScoringCheckResponse)) {
+                //                forces a reset of the user's identity if the CI score is breached
+                // and no mitigation is possible
+                return JOURNEY_FAIL_WITH_CI_AND_FORCED_RESET;
+            }
+            //            sends the user on mitigation journey if the CI score is breached and
+            // mitigation is possible
```
Joe-Edwards-GDS: Strange indentation in the comment?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -274,9 +273,20 @@ private JourneyResponse getJourneyResponse(
...
 
     @Tracing
-    private JourneyResponse buildForceResetResponse() {
-        LOGGER.info(
-                LogHelper.buildLogMessage("resetIdentity flag is enabled, reset users identity."));
+    private JourneyResponse buildForceResetResponse(JourneyResponse ciScoringCheckResponse) {
+        if (ciScoringCheckResponse != null) {
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "resetIdentity flag is enabled, reset users identity."));
```
Joe-Edwards-GDS: Should this be outside the `if`?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -16,13 +16,46 @@ FAILED_BAV:
...
+    type: process
+    lambda: reset-identity
+  events:
+    next:
+      targetState: NO_MATCH_PAGE
+
+CRI_TICF_BEFORE_RESET_VCS:
+  response:
+      type: process
+      lambda: call-ticf-cri
```
Joe-Edwards-GDS: You might want to add:
```
    lambdaInput:
      isUserInitiated: false
      deleteOnlyGPG45VCs: false
```
to make it explicit?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -16,13 +16,46 @@ FAILED_BAV:
         ticfCriBeta:
           targetState: CRI_TICF_BEFORE_NO_MATCH_BAV
 
+FAILED_FORCED_RESET:
+  events:
+    next:
+      targetState: RESET_VCS
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF_BEFORE_RESET_VCS
```
Joe-Edwards-GDS: In the other cases we've got the reset step in `initial-journey-selection`, which I think might be slightly neater because you can just route to FAILED/FAILED as normal afterwards?

In either case, I think it's fine to call TICF after the reset, which should make it cleaner (as you won't need a new TICF state).

Joe-Edwards-GDS (commented):
Happy for you to take this approach since you've gone down this route :)

A few comments and obviously we'll need some tests.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
sammyiyke: I'm not sure I fully understand what you are asking for exactly :).

Joe-Edwards-GDS: No worries, I think that we could:
- Move the reset state into `initial-journey-selection` (so it goes alongside the existing `RESET_IDENTITY` and `RESET_GPG45_IDENTITY` states)
- Route `fail-with-ci-and-forced-reset` to that state
- Route from there to `FAILED`/`FAILED`
- (then we don't need this new TICF state)

Does that make sense?

sammyiyke: Just push a change to address this. Please review and let me know if the new implementation is what you had in mind

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -34,6 +34,9 @@ CHECK_EXISTING_IDENTITY:
     fail-with-ci:
       targetJourney: FAILED
       targetState: FAILED
+    fail-with-ci-and-forced-reset:
+      targetJourney: FAILED
```
Joe-Edwards-GDS: We don't need the `targetJourney` any more (since it's in the same journey)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -67,3 +70,14 @@ RESET_GPG45_IDENTITY:
     next:
       targetJourney: NEW_P2_IDENTITY
       targetState: START
+
+RESET_VCS:
```
Joe-Edwards-GDS: Could we call this `RESET_FAILED_IDENTITY` so it's a bit more consistent with the other `RESET_xx` states?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -67,3 +70,14 @@ RESET_GPG45_IDENTITY:
...
+  response:
+    type: process
+    lambda: reset-identity
+    lambdaInput:
+      isUserInitiated: false
+      deleteOnlyGPG45VCs: false
+  events:
+    next:
+      targetState: NO_MATCH_PAGE
```
Joe-Edwards-GDS: This needs to be:
```
targetJourney: FAILED
targetState: FAILED
```

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -23,6 +23,22 @@ FAILED_NO_TICF:
 
 # Journey states
 
+CRI_TICF_BEFORE_RESET_VCS:
```
Joe-Edwards-GDS: I think we can delete this now :)

Joe-Edwards-GDS (commented):
Yep, this is what I meant - just a few bits to tidy up :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -34,6 +34,8 @@ CHECK_EXISTING_IDENTITY:
     fail-with-ci:
       targetJourney: FAILED
       targetState: FAILED
+    fail-with-ci-and-forced-reset:
+      targetState: RESET_VCS
```
Joe-Edwards-GDS: This needs to be `RESET_FAILED_IDENTITY` now

We could consider adding a test to `JourneyMapTest` to catch this sort of thing, but idk if it's worth it.

Joe-Edwards-GDS (commented):
One small fix (and another optional existing comment), but I think it should be almost there

Joe-Edwards-GDS (approved):

PR #1695, opened 2024-02-26
PYIC-5032: Create common vc class

MikeCollingwood:
### Issue tracking

### Other considerations


lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -198,19 +186,4 @@ private Map<String, Object> callTicfCri(IpvSessionItem ipvSessionItem, ProcessRe
         LOGGER.info(LogHelper.buildLogMessage("CI score not breaching threshold"));
         return JOURNEY_NEXT;
     }
-
-    @Tracing
-    private List<String> getVcToSendToTicf(String userId, IpvSessionItem ipvSessionItem) {
-        List<String> vcInStore =
-                userIdentityService.getIdentityCredentials(
```
MikeCollingwood, line 207: This getIdentityCredentials method was the only reason why getVcToSendToTicf didn't just immediately return ipvSessionItem.getVcReceivedThisSession(). getIdentityCredentials would filter out the TICF VC, though the TICF VC is removed at the start of a journey so won't exist until after this handler is called.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -415,8 +411,7 @@ private JourneyResponse buildErrorResponse(ErrorResponse errorResponse, Exceptio
     @Tracing
     private Optional<Vot> getStrongestAttainedVotForVtr(
             List<String> vtr,
-            List<SignedJWT> credentials,
-            List<VcStoreItem> vcStoreItems,
+            List<VerifiableCredential> vcs,
```
MikeCollingwood, line 414: Need to check this method again.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -351,30 +348,34 @@ private SignedJWT validateHmrcInheritedIdentity(
...
             JWTClaimsSet claimsSet,
             IpvSessionItem ipvSessionItem,
-            SignedJWT signedInheritedIdentityJWT)
+            VerifiableCredential inheritedIdentityVc)
             throws RecoverableJarValidationException, ParseException {
         try {
-            verifiableCredentialService.persistUserCredentials(
-                    signedInheritedIdentityJWT, HMRC_MIGRATION_CRI, userId);
-            ipvSessionItem.addVcReceivedThisSession(signedInheritedIdentityJWT.serialize());
+            verifiableCredentialService.persistUserCredentials(inheritedIdentityVc, userId);
```
MikeCollingwood: No longer needs to pass the criId since it's baked into the VerifiableCredential

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -225,12 +225,11 @@ public JourneyResponse checkVcResponse(
             return ciMitUtilityService.getCiMitigationJourneyStep(cis).orElse(JOURNEY_FAIL_WITH_CI);
         }
 
-        if (!userIdentityService.areVCsCorrelated(
-                verifiableCredentialService.getVcStoreItems(clientOAuthSessionItem.getUserId()))) {
+        if (!userIdentityService.areVcsCorrelated(vcs)) { // can I just use the passed-in vcs here?
```
MikeCollingwood, line 228: Need to change this because of course we're checking if any of their VCs are uncorrelateed

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialJwtValidator.java
```diff
@@ -58,37 +60,46 @@ public VerifiableCredentialJwtValidator(
...
-    public void validate(
-            SignedJWT verifiableCredential, ECKey signingKey, String componentId, String userId)
-            throws VerifiableCredentialException {
-        validateSignatureAndClaims(verifiableCredential, signingKey, componentId, userId);
-        validateCiCodes(verifiableCredential);
-        LOGGER.info(LogHelper.buildLogMessage("Verifiable Credential validated."));
+        return validVcs;
     }
 
-    public void validateSignatureAndClaims(
```
MikeCollingwood, line 85: Was only used for CiMit VCs which can be validated with the full method which validates ci codes are valid

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -325,19 +320,16 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
             ClientOAuthSessionItem clientOAuthSessionItem,
             AuditEventUser auditEventUser,
-            List<VcStoreItem> vcStoreItems,
+            List<VerifiableCredential> vcs,
             boolean areGpg45VcsCorrelated)
             throws ParseException, UnknownEvidenceTypeException, SqsException {
-        List<SignedJWT> credentials =
-                gpg45ProfileEvaluator.parseCredentials(
-                        userIdentityService.getIdentityCredentials(vcStoreItems));
+        var identityVcs = userIdentityService.getIdentityCredentials(vcs);
```
MikeCollingwood: getIdentityCredentials filters out TICF VC but we remove TICF VC earlier in the CheckExistingIdentityHandler handlerRequest method

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -375,24 +384,10 @@ private EvidenceRequest getEvidenceRequestForF2F(List<SignedJWT> credentials)
...
-        for (String credential : credentials) {
-            try {
-                signedJWTs.add(SignedJWT.parse(credential));
-            } catch (ParseException e) {
-                throw new HttpResponseExceptionWithErrorBody(
-                        500, ErrorResponse.FAILED_TO_PARSE_ISSUED_CREDENTIALS);
-            }
-        }
-        return signedJWTs;
+    private List<VerifiableCredential> getVcs(String userId) throws CredentialParseException {
```
Joe-Edwards-GDS: Maybe `getGpg45Vcs`?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -255,66 +230,61 @@ public boolean areVCsCorrelated(List<VcStoreItem> vcStoreItems)
         return true;
     }
 
-    private List<VcStoreItem> excludeTicfVC(List<VcStoreItem> vcStoreItems) {
-        return vcStoreItems.stream()
-                .filter(vcStoreItem -> !vcStoreItem.getCredentialIssuer().equals(TICF_CRI))
+    public List<VerifiableCredential> getIdentityCredentials(List<VerifiableCredential> vcs) {
+        return vcs.stream()
+                .filter(vcStoreItem -> !vcStoreItem.getCriId().equals(TICF_CRI))
                 .toList();
```
Joe-Edwards-GDS: This seems like a fishy place for this method, I wonder if we can get rid of it entirely:
- We shouldn't need it in `CheckGpg45ScoreHandler` as we shouldn't have a TICF VC at that point
- We do still need to do _something_ in `BuildCriOauthRequestHandler`, but maybe it should be a special case there?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -522,13 +481,13 @@ private IdentityClaim getIdentityClaim(String credential)
...
-        return Vot.valueOf(credential.getJWTClaimsSet().getStringClaim(VOT_CLAIM_NAME));
+    public Vot getVot(VerifiableCredential vc) throws IllegalArgumentException, ParseException {
+        return Vot.valueOf(vc.getClaimsSet().getStringClaim(VOT_CLAIM_NAME));
     }
 
-    private Optional<JsonNode> generateAddressClaim(List<VcStoreItem> vcStoreItems)
+    private Optional<JsonNode> generateAddressClaim(List<VerifiableCredential> vcs)
             throws HttpResponseExceptionWithErrorBody {
-        var addressStoreItem = findStoreItem(ADDRESS_CRI, vcStoreItems);
+        var addressStoreItem = findVc(ADDRESS_CRI, vcs);
```
Joe-Edwards-GDS: Minor - this should probably be `addressVc` (similarly with `ninoStoreItem` below)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -93,78 +90,57 @@ public class UserIdentityService {
...
     }
 
-    public Optional<IdentityClaim> findIdentityClaim(List<VcStoreItem> vcStoreItems)
+    public Optional<IdentityClaim> findIdentityClaim(List<VerifiableCredential> vcs)
             throws HttpResponseExceptionWithErrorBody, CredentialParseException {
-        return findIdentityClaimFromVcStoreItems(vcStoreItems, true);
+        return findIdentityClaimFromVcStoreItems(vcs, true);
     }
 
     private Optional<IdentityClaim> findIdentityClaimFromVcStoreItems(
```
Joe-Edwards-GDS: This can probably just be `findIdentityClaim`, and replace the existing public version that just passes through to this?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -222,11 +197,11 @@ public boolean isVcSuccessful(List<VcStatusDto> currentVcStatuses, String criIss
                 .getIsSuccessfulVc();
     }
 
-    public boolean areVCsCorrelated(List<VcStoreItem> vcStoreItems)
+    public boolean areVcsCorrelated(List<VerifiableCredential> vcs)
             throws HttpResponseExceptionWithErrorBody, CredentialParseException {
-        List<VcStoreItem> successfulGPG45VCStoreItems =
+        var successfulGPG45VCStoreItems =
```
Joe-Edwards-GDS: `successfulVcs`

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -450,20 +420,8 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
-                if (VcHelper.isSuccessfulVc(SignedJWT.parse(vcStoreItem.getCredential()))) {
-                    successfulVCStoreItems.add(vcStoreItem);
-                }
-            } catch (ParseException e) {
-                throw new CredentialParseException(
-                        "Encountered a parsing error while attempting to parse successful VC Store items.");
-            }
-        }
-        return successfulVCStoreItems;
+    private List<VerifiableCredential> getSuccessfulVCStoreItems(List<VerifiableCredential> vcs) {
```
Joe-Edwards-GDS: `getSuccessfulVcs`?

I also wonder if it should go into `VcHelper` alongside `isSuccessfulVc`, or just be inlined :shrug:

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Joe-Edwards-GDS: I think that's not _totally_ true - this was also so that the VCs we were sending were 'properly' from the store (ideally we'd have just stored some ids in the session item). That's not particularly compelling, so fine to leave this if you like.

It's worth knowing that in future we might flip this again... Once we introduce encrypted VC storage, our original store might become the record of 'VCs received this session' after all!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -37,7 +38,7 @@ public class IpvSessionItem implements DynamodbItem {
     // Only for passing the featureSet to the external API lambdas at the end of the user journey.
     // Not for general use.
     private String featureSet;
-    private List<String> vcReceivedThisSession;
+    private List<VerifiableCredential> vcReceivedThisSession;
```
Joe-Edwards-GDS: I don't think we should change this type - it makes sense that this is an array of strings when serialized.

Do we ever need to parse these credentials if we're just passing them directly to the TICF CRI (as strings)? I think it makes sense to update the type on `addVcReceivedThisSession`, but that can call `getVcString()` before adding to the session.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -446,33 +438,30 @@ private Optional<Vot> getStrongestAttainedVotForVtr(
...
-            List<VcStoreItem> vcStoreItems,
-            AuditEventUser auditEventUser)
+            Vot requestedVot, List<VerifiableCredential> vcs, AuditEventUser auditEventUser)
             throws UnknownEvidenceTypeException, ParseException, SqsException {
 
-        Gpg45Scores gpg45Scores = gpg45ProfileEvaluator.buildScore(credentials);
+        Gpg45Scores gpg45Scores = gpg45ProfileEvaluator.buildScore(vcs);
         Optional<Gpg45Profile> matchedGpg45Profile =
-                !userIdentityService.checkRequiresAdditionalEvidence(vcStoreItems)
+                !userIdentityService.checkRequiresAdditionalEvidence(vcs) // hmm why
```
Joe-Edwards-GDS: why indeed...

This is a bodge to prevent matching M1B with a bank account VC. I'm not 100% sure this is the best place for it, but no need to change it now.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -278,34 +284,59 @@ public JourneyResponse getJourneyResponse(CriCallbackRequest callbackRequest)
...
+            CriOAuthSessionItem criOAuthSessionItem,
+            IpvSessionItem ipvSessionItem)
+            throws VerifiableCredentialException, JsonProcessingException, SqsException,
+                    InvalidCriCallbackRequestException, ParseException, CiPutException,
+                    CiPostMitigationsException, AuditExtensionException, UnrecognisedVotException {
         if (VerifiableCredentialStatus.PENDING.equals(vcResponse.getCredentialStatus())) {
             criCheckingService.validatePendingVcResponse(vcResponse, clientOAuthSessionItem);
             criStoringService.storeCriResponse(callbackRequest, clientOAuthSessionItem);
+
+            return new ArrayList<>();
```
Joe-Edwards-GDS: Prefer `Collections.emptyList()` (or `List.of()`) unless you particularly need it to be mutable

lambdas/restore-vcs/src/main/java/uk/gov/di/ipv/core/restorevcs/RestoreVcsHandler.java
```diff
@@ -44,17 +48,20 @@ public class RestoreVcsHandler implements RequestStreamHandler {
     private static final Logger LOGGER = LogManager.getLogger();
     private final ConfigService configService;
     private final VerifiableCredentialService verifiableCredentialService;
+    private final DataStore<VcStoreItem> vcDataStore;
```
Joe-Edwards-GDS, line 48: If you're doing it 'directly' with a data store, do you still need the `VerifiableCredentialService`?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
Joe-Edwards-GDS, line 1: Do you think there's any need to re-test after this change?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Getter
+public class VerifiableCredential {
+    private final String userId;
+    private final String criId;
+    private final String vcString;
+    private final JWTClaimsSet claimsSet;
+
+    // Only construct after validation
+    public VerifiableCredential(String userId, String criId, SignedJWT jwt)
```
Joe-Edwards-GDS: If you want to make this more explicit, you could make the constructor private and add another static factory method `fromValidJwt` (or similar).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -0,0 +1,42 @@
...
+import com.nimbusds.jwt.SignedJWT;
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
+import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
+
+import java.text.ParseException;
+
+@Getter
+public class VerifiableCredential {
+    private final String userId;
```
Joe-Edwards-GDS, line 15: Do you think this is a useful field to have? It looks like we only really use it in a couple of places where we could source the user ID elsewhere.

Note that this is _almost_ always the same as `claimsSet.getSubject()` - but won't be for inherited identity VCs.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -225,9 +209,9 @@ private static Integer getAge(String dobValue) {
         }
     }
 
-    public static Vot getVcVot(SignedJWT vc) throws UnrecognisedVotException {
+    public static Vot getVcVot(VerifiableCredential vc) throws UnrecognisedVotException {
```
Joe-Edwards-GDS, line 223: (Not related to your change, but I wonder if we should always use this instead of the version in `UserIdentityService` - we'll need to handle `null` instead of just relying on an exception...)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialJwtValidator.java
Joe-Edwards-GDS: I assume you've tested this(?) - Just making sure that it doesn't fall over when receiving the CIMIT VC format

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -351,30 +348,34 @@ private SignedJWT validateHmrcInheritedIdentity(
         }
 
         // Validate VC in claim is valid
-        var signedInheritedIdentityJWT = SignedJWT.parse(inheritedIdentityJwtList.get(0));
+        var inheritedIdentityJwt = SignedJWT.parse(inheritedIdentityJwtList.get(0));
         var inheritedIdentityCriConfig = configService.getCriConfig(HMRC_MIGRATION_CRI);
 
         // The HMRC inherited identity VC will contain an HMRC-specific pairwise identifier
         // rather than our internal user id, so we cannot validate it against the OAuth user id.
         // Instead, SPOT will validate this when generating an identity bundle.
```
Joe-Edwards-GDS, line 354: This logic isn't quite going to work now - you're baking the `userId` into the `VerifiableCredential` that you build, but now it's not the right one.

I think we could do one (or more) of:
- Remove `userId` from `VerifiableCredential`, so it's not an issue
- Remove the special casing here, and pass a more explicit flag, e.g. `skipSubjectCheck`
  - This feels like the 'right' solution, since the current approach is obviously a hack (that I introduced...)
- Move this logic inside `VerifiableCredentialJwtValidator`, so that it knows about the inherited id special case
  - I don't like this!

Joe-Edwards-GDS (commented):
This is making me very happy - loads of stuff looking a lot cleaner already!

Lots of comments, but I think that's somewhat inevitable given the size of the change, and most of them are tiny.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood: I think the second solution is best because it makes more sense. I think the userId is useful for now because it refers to how we store it, and with the first change, it can be consistently used

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
MikeCollingwood: It feels quite natural to pair with criId since that's how we store it in the Vc Store. And since it's an object that's past validation etc, we should be able to effortlessly convert to VcStoreItem without having to pass anything new in?

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
MikeCollingwood: I think it'd be good to do a simple test of the happy case, since the handler logic hasn't changed, just one component

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
MikeCollingwood: That's good I'll do that!

lambdas/revoke-vcs/src/main/java/uk/gov/di/ipv/core/revokevcs/RevokeVcsHandler.java
Joe-Edwards-GDS: Sounds good - can you make sure that happens? We'll probably just want to do another test when we finally get approval

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
MikeCollingwood: I can make that change if you want? I think the null case is already handled because it's only used here:
```java
        try {
            var hmrcMigrationVc = verifiableCredentialService.getVc(userId, HMRC_MIGRATION_CRI);
            if (hmrcMigrationVc == null) {
                return false;
            }

            var existingInheritedIdentityVot = userIdentityService.getVot(hmrcMigrationVc);
            var incomingInheritedIdentityVot =
                    userIdentityService.getVot(incomingInheritedIdentity);

            var indexOfExistingVot =
                    HMRC_PROFILES_BY_STRENGTH.indexOf(existingInheritedIdentityVot);
            var indexOfIncomingVot =
                    HMRC_PROFILES_BY_STRENGTH.indexOf(incomingInheritedIdentityVot);

            if (indexOfExistingVot == -1 || indexOfIncomingVot == -1) {
                throw new IllegalArgumentException(
                        String.format(
                                "At least one of the existing (%s) or incoming (%s) VoTs from hmrc aren't expected",
                                existingInheritedIdentityVot, incomingInheritedIdentityVot));
            }

            return indexOfIncomingVot > indexOfExistingVot;
                    "Encountered a parsing error while attempting to parse credentials", e);
        }
```

Where the nullpointerexception is caught

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
MikeCollingwood: I've removed getIdentityCredentials and in BuildCriOauthRequestHandler, I've modified this method to do it directly, since it's the only remaining use-case: 
```java
    private List<VerifiableCredential> getGpg45Vcs(String userId) throws CredentialParseException {
        return VcHelper.filterVCBasedOnProfileType(
                        verifiableCredentialService.getVcs(userId), ProfileType.GPG45)
                .stream()
                .filter(vc -> !vc.getCriId().equals(TICF_CRI))
                .toList();
    }
```

Joe-Edwards-GDS: I think you might need to be a little careful - for F2F we collect the TICF VC _before_ sending the user to the F2F CRI, so we might actually have a TICF VC in `BuildCriOauthRequestHandler`

MikeCollingwood: Ah yep, sorry I didn't explain that very well. It's still filtering out the TICF VC in BuildCriOauthRequestHandler - in the method I've referenced

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: I feel like catching a `NullPointerException` is always a bit of a code smell!

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialJwtValidator.java
MikeCollingwood: Annoyingly, this isn't allowed - the CiMit evidence is a different shape. 

Ideally, we'd validate every different type of VC with the same method. For this we should:
- put validateCiCodes in validateClaimsSet
- pass validateClaimsSet an indicator of what type of VC it's validating
- let the indicator inform the specification of the generic claimsSet attribute (?)
I'll look into this for the next PR

Joe-Edwards-GDS: Ah, shame. We can always revisit in part 2, if we end up parsing into distinct classes at this point.

`if (vc instanceof IdentityCheckCredential) { ... }` or whatever

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -49,38 +50,50 @@ DefaultJWTClaimsVerifier<SimpleSecurityContext> createVerifier(
...
         var validVcs = new ArrayList<VerifiableCredential>();
         for (String vc : vcs) {
-            validVcs.add(validate(vc, criId, signingKey, componentId, userId));
+            validVcs.add(validate(userId, criId, vc, vcType, signingKey, componentId, false));
         }
-
         return validVcs;
     }
 
     public VerifiableCredential validate(
```
Joe-Edwards-GDS: You could consider making this one private and adding a new public one `validateWithNoSubjectCheck`.

It's not necessary, but avoids having a boolean flag on the public interface (which can be hard to read).

Joe-Edwards-GDS (commented):
I think the prod code is looking good, hope the tests aren't driving you too mad!

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandlerTest.java
```diff
@@ -293,13 +302,19 @@ void shouldReceive200ResponseCodeAndReturnCredentialIssuerResponseWithoutRespons
                 .thenReturn(addressOauthCriConfig.getComponentId());
         when(mockIpvSessionService.getIpvSession(SESSION_ID)).thenReturn(mockIpvSessionItem);
         mockVcHelper.when(() -> VcHelper.isSuccessfulVc(any())).thenReturn(true, true);
-        when(userIdentityService.getIdentityCredentials(any()))
+        when(VcHelper.filterVCBasedOnProfileType(any(), any()))
```
MikeCollingwood, line 305: Since getGpg45Vcs doesn't need to filter out the TICF VC here anymore

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandlerTest.java
```diff
@@ -293,13 +302,19 @@ void shouldReceive200ResponseCodeAndReturnCredentialIssuerResponseWithoutRespons
...
                 .thenReturn(
                         List.of(
                                 generateVerifiableCredential(
```
MikeCollingwood, line 308: This generates the new VerifiableCredential class object, so wants to be passed a userId and criId

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -179,19 +175,15 @@ private NameAndDateOfBirth getProvenIdentityNameAndDateOfBirth(
...
     @Tracing
     private List<Address> getProvenIdentityAddresses(
-            List<VcStoreItem> credentialIssuerItems, List<VcStatusDto> currentVcStatuses)
+            List<VerifiableCredential> vcs, List<VcStatusDto> currentVcStatuses)
             throws ParseException, JsonProcessingException, ProvenUserIdentityDetailsException,
                     NoVcStatusForIssuerException {
-        for (VcStoreItem item : credentialIssuerItems) {
-            if (item.getCredentialIssuer().equals(ADDRESS_CRI)
+        for (var vc : vcs) {
+            if (vc.getCriId().equals(ADDRESS_CRI)
```
MikeCollingwood, line 182: Ah, I see that it the criId attribute is null then this would fail. It's the same as the logic before but might want a change if we prefer dealing in a more controlled manner

lambdas/build-proven-user-identity-details/src/test/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandlerTest.java
```diff
@@ -103,27 +100,22 @@ void setUp() {
 
     @Test
     void shouldReceive200ResponseCodeProvenUserIdentityDetails() throws Exception {
-        when(mockUserIdentityService.isVcSuccessful(any(), any())).thenCallRealMethod();
```
MikeCollingwood, line 106: `thenCallRealMethod` has been unreliable in testing. It seems like the interference of mocking the VcHelper (required for the new logic where we have to mock the filterVCBasedOnProfileType method) affects the VcHelper calls by the userIdentityService as it calls the real method

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -65,22 +65,26 @@ public class BuildUserIdentityHandler
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
     private final CiMitService ciMitService;
     private final CiMitUtilityService ciMitUtilityService;
+    private final VerifiableCredentialService verifiableCredentialService;
 
+    @ExcludeFromGeneratedCoverageReport // This constructor is only used in tests
```
MikeCollingwood: Required because sonarcloud takes issue with there being 8 parameters to this constructor. At the moment, 
- there is no method of reducing the number of parameters without (<<< EDIT) significant rework
- this constructor is only used in tests

So I have excluded it from the coverage report

lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -21,16 +21,7 @@
...
-import uk.gov.di.ipv.core.library.domain.BirthDate;
-import uk.gov.di.ipv.core.library.domain.ContraIndicatorConfig;
-import uk.gov.di.ipv.core.library.domain.ContraIndicators;
-import uk.gov.di.ipv.core.library.domain.ErrorResponse;
-import uk.gov.di.ipv.core.library.domain.IdentityClaim;
-import uk.gov.di.ipv.core.library.domain.Name;
-import uk.gov.di.ipv.core.library.domain.NameParts;
-import uk.gov.di.ipv.core.library.domain.ReturnCode;
-import uk.gov.di.ipv.core.library.domain.UserIdentity;
+import uk.gov.di.ipv.core.library.domain.*;
```
MikeCollingwood: Wildcard import

lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandlerTest.java
```diff
@@ -83,104 +86,38 @@ class CallTicfCriHandlerTest {
...
-
-        InOrder inOrder = inOrder(mockIpvSessionService);
-        inOrder.verify(mockIpvSessionService).updateIpvSession(spyIpvSessionItem);
-        inOrder.verifyNoMoreInteractions();
-
-        assertEquals("/journey/next", lambdaResult.get("journey"));
-    }
-
-    @Test
-    void handleRequestShouldOnlySendVcsReceivedInCurrentSession() throws Exception {
```
MikeCollingwood, line 129: Since removing logic between VcsReceivedThisSession and the "vcs to send", testing this behaviour is redundant as tests the same as above

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -815,46 +760,13 @@ void shouldResetIdentityIfDataDoesNotCorrelateAndNotF2F() {
         assertNull(ipvSessionItem.getVot());
     }
 
-    @Test
-    void shouldReturn500IfFailedToParseCredentials() throws Exception {
```
MikeCollingwood, line 820: Doesn't need to parse credentials since they are VerifiableCredentials and are already valid (this change is repeated in a number of places)

lambdas/build-proven-user-identity-details/src/test/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandlerTest.java
Joe-Edwards-GDS: Makes sense - I think it's a bit of a smell anyway, having a partially mocked class.

lambdas/build-proven-user-identity-details/src/test/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandlerTest.java
```diff
@@ -103,27 +100,22 @@ void setUp() {
...
+        when(mockUserIdentityService.isVcSuccessful(any(), any())).thenReturn(true);
         when(mockUserIdentityService.findIdentityClaim(any())).thenCallRealMethod();
         when(mockIpvSessionService.getIpvSession(SESSION_ID)).thenReturn(mockIpvSessionItem);
-        when(mockVerifiableCredentialService.getVcStoreItems(TEST_USER_ID))
+        when(VcHelper.isSuccessfulVc(any()))
+                .thenReturn(true)
+                .thenReturn(true)
+                .thenReturn(true)
+                .thenReturn(true)
+                .thenReturn(true);
```
Joe-Edwards-GDS: Do you actually need to specify this repeatedly? I think by default the last one applies to all subsequent calls

If you need to specify different return values for repeated calls, you can also do `.thenReturn(true, true, true...)` if you prefer.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Joe-Edwards-GDS: Those are two very different reasons :)

If it's only used in tests, then it should be covered by tests - although sometimes it won't be (if it's used implicitly by the mock injection framework).

If Sonar is complaining about the code style and we want to ignore it, then we can add an annotation to resolve that specific issue with `@SuppressWarnings` and a reference to the issue. (Search for examples) - that way we'll still pick up other issues, rather than just blanket excluding it.

lambdas/process-async-cri-credential/src/test/java/uk/gov/di/ipv/core/processasynccricredential/pact/f2fCri/ContractTest.java
```diff
@@ -772,47 +826,58 @@ void testF2fMessageReturnsIssuedEeaCardCredential(
...
+                                                "SPEC12031",
+                                                idCardNode.get(DOCUMENT_NUMBER).asText());
+
+                                        assertEquals(
+                                                "1970-01-01", birthDateNode.get(VALUE).asText());
+                                    } catch (JsonProcessingException e) {
+                                        throw new RuntimeException(e);
+                                    }
+                                });
+            } catch (JsonProcessingException | ParseException | VerifiableCredentialException e) {
```
Joe-Edwards-GDS, line 880: cc @DanCorderIPV

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -207,7 +205,7 @@ public VerifiableCredentialResponse fetchVerifiableCredential(
                 var verifiableCredentialResponse =
                         VerifiableCredentialResponse.builder()
                                 .verifiableCredentials(
-                                        Collections.singletonList(response.getContent()))
+                                        Collections.singletonList(response.getContent().trim()))
```
Joe-Edwards-GDS: Are you sure this is correct? It doesn't necessarily do any harm, but it looks a bit suspicious.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -280,93 +290,115 @@ public interface VcFixtures {
                     "personalNumber", "MORGA753116SM9IJ",
                     "expiryDate", "2042-10-01");
 
-    String PASSPORT_NON_DCMAW_SUCCESSFUL_VC =
+    VerifiableCredential PASSPORT_NON_DCMAW_SUCCESSFUL_VC =
             generateVerifiableCredential(
+                    "urn:uuid:e6e2e324-5b66-4ad6-8338-83f9f837e345",
```
Joe-Edwards-GDS: What is this magic string? Should it be a constant, or handled internally to `generateVerifiableCredential`?

MikeCollingwood:
NB: Ran E2E tests and they all pass

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
```diff
@@ -132,7 +144,10 @@ private static String signTestJWT(JWTClaimsSet claimsSet) throws Exception {
         SignedJWT signedJWT = new SignedJWT(jwsHeader, claimsSet);
         signedJWT.sign(ecdsaSigner);
 
-        return signedJWT.serialize();
+        // Parsing and serialising here allows us to cast like before.
+        // The casting will be replaced in PYIC-5171 and so should this.
```
Joe-Edwards-GDS, line 138: Can you elaborate on this? I'm not sure I follow

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -121,12 +121,19 @@ public UserIdentity generateUserIdentity(
...
+        return findIdentityClaimPrivate(vcs, true);
     }
 
     public Optional<IdentityClaim> findIdentityClaim(
             List<VerifiableCredential> vcs, boolean checkEvidence)
             throws HttpResponseExceptionWithErrorBody, CredentialParseException {
+        return findIdentityClaimPrivate(vcs, checkEvidence);
+    }
+
+    // Need to have a private method for tests ?
```
Joe-Edwards-GDS: ?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -444,8 +451,7 @@ private <T> T getJsonProperty(JsonNode jsonNode, String propertyName, Collection
 
     private IdentityClaim getIdentityClaim(VerifiableCredential vc)
             throws HttpResponseExceptionWithErrorBody, CredentialParseException {
-        // TODO: PYIC-3908 Don't reparse here
```
Joe-Edwards-GDS: I think these comments kinda still apply? We might need to link to a different ticket if we end up deferring

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/dto/VerifiableCredentialResponseDto.java
```diff
@@ -15,6 +16,7 @@
 @Builder
 @Getter
 @Setter
+@Jacksonized
```
Joe-Edwards-GDS: Curious what this is for - there are a lot of annotations here 😅

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
MikeCollingwood: > If it's only used in tests, then it should be covered by tests

Wouldn't the priority be production code. Or do you mean the functionality should already be covered in tests?

The `@SuppressWarnings` is a better replacement, will do that

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
MikeCollingwood: Good point - I'll make it a constant defined at the top

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Joe-Edwards-GDS: As in, if something is used by tests, it should have test coverage - by definition!

(But you're right that we don't care very much if it's not also used production)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
MikeCollingwood: This comes from casting like this:

```java
JSONObject vcClaim = (JSONObject) vc.getClaimsSet().getClaim(VC_CLAIM);
JSONArray evidenceArray = (JSONArray) vcClaim.get(VC_EVIDENCE);
```

For some reason the `signedJwt` in this method needs to be parsed for this casting to work. 

- Before, that would happen because this is used in generating a string which would have to be `SignedJWT.parse`ed later on, 
- but now I'm adding in this to ensure it gets parsed, and since it is destined to be replaced when we don't need to cast (PYIC-5171)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
MikeCollingwood: Ah ok, then I need to suppress the sonarcloud warnings

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/dto/VerifiableCredentialResponseDto.java
MikeCollingwood: Ah yep, this is from experimentation - will remove

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
MikeCollingwood: `BuildProvenUserIdentityDetailsHandlerTest` > `when(mockUserIdentityService.findIdentitytClaim(any())).thenCallRealMethod()` only calls the first and last lines of each method it need to call... Why this happens I don't know but suspect that's why there was a private method there before my changes

I accept it's a bit hacky, so this could be a follow-up piece of work to tidy up

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialGenerator.java
Joe-Edwards-GDS: Ah - it'll be because when you're building the `JWTClaimsSet` you're not using the JSON classes (we're using `TestVc`), but by serializing and reparsing, it parses it into the real JSON object.

To do it 'properly' you should make sure that all the claims you put onto the `JWTClaimsSet` are the appropriate JSON types, but doing this is a fairly easy hack to do them all at once :shrug:

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
MikeCollingwood: Ah I understand - the sonarcloud complaint was with the excessive number of parameters in the constructor (8 > 7, the threshold), as opposed to test coverage

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Oh I see... If it's easier to just get rid of them that's fine, maybe reference the methods on the ticket instead then?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: That sounds real suspicious :/

If you do need to keep this here, then can we at least tidy up the comment to explain more clearly?

MikeCollingwood: I'll remove the .thenCallRealMethod() and instead just mock the response

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
MikeCollingwood: I think I will keep it as it is for now since it's at least currently quite satisfying having verifiable credential created from the validation in one method... Unless you think there is enough to merit the change despite this

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/bavCri/ContractTest.java
```diff
@@ -134,12 +136,20 @@ void fetchVerifiableCredential_whenCalledAgainstBavCri_retrievesAValidVc(MockSer
...
-                                        credential, credentialIssuerConfig, TEST_USER);
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                "dummyCriId",
+                                                credential,
+                                                VerifiableCredentialConstants
+                                                        .IDENTITY_CHECK_CREDENTIAL_TYPE,
+                                                ECKey.parse(CRI_SIGNING_PRIVATE_KEY_JWK),
+                                                "dummyBavComponentId",
```
DanCorderIPV: This is in the `TEST_ISSUER` constant (same for some of the other test files)

DanCorderIPV: If you find a test file where the equivalent value isn't in a constant could you extract it please?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/bavCri/ContractTest.java
```diff
@@ -134,12 +136,20 @@ void fetchVerifiableCredential_whenCalledAgainstBavCri_retrievesAValidVc(MockSer
                 .forEach(
                         credential -> {
                             try {
-                                verifiableCredentialJwtValidator.validate(
-                                        credential, credentialIssuerConfig, TEST_USER);
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                "dummyCriId",
```
DanCorderIPV: Would be good to put this in a constant now it's used more than once.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
Joe-Edwards-GDS: Sure - I mean we currently have two public methods here anyway, but I don't feel very strongly about it.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -1184,16 +1251,24 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidBrc(
...
                         credential -> {
                             try {
-                                verifiableCredentialJwtValidator.validate(
-                                        credential, credentialIssuerConfig, TEST_USER);
-
-                                JsonNode vc =
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                DCMAW_CRI,
```
DanCorderIPV, line 1257: This isn't `dummyCriId` here, as it is in all the other tests, is that deliberate?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/ContractTest.java
```diff
@@ -554,24 +556,32 @@ void fetchVerifiableCredential_whenCalledAgainstExperianKbvCri_retrievesAValidVc
...
                 .getVerifiableCredentials()
                 .forEach(
                         credential -> {
                             try {
-                                verifiableCredentialJwtValidator.validate(
-                                        credential, credentialIssuerConfig, TEST_USER);
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                null,
```
DanCorderIPV: Is this deliberately `null`?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -1184,16 +1251,24 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidBrc(
...
+                                                credential,
+                                                VerifiableCredentialConstants
+                                                        .IDENTITY_CHECK_CREDENTIAL_TYPE,
+                                                ECKey.parse(CRI_RSA_ENCRYPTION_PUBLIC_JWK),
```
DanCorderIPV: This is the wrong key constant

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
```diff
@@ -146,12 +148,20 @@ void fetchVerifiableCredential_whenCalledAgainstCicCri_retrievesAValidVc(MockSer
...
-                                verifiableCredentialJwtValidator.validate(
-                                        credential, credentialIssuerConfig, TEST_USER);
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                null,
+                                                credential,
+                                                VerifiableCredentialConstants
+                                                        .IDENTITY_CHECK_CREDENTIAL_TYPE,
+                                                ECKey.parse(CRI_RSA_ENCRYPTION_PUBLIC_JWK),
```
DanCorderIPV: Wrong key constant

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/ContractTest.java
```diff
@@ -554,24 +556,32 @@ void fetchVerifiableCredential_whenCalledAgainstExperianKbvCri_retrievesAValidVc
...
+                                                credential,
+                                                VerifiableCredentialConstants
+                                                        .IDENTITY_CHECK_CREDENTIAL_TYPE,
+                                                ECKey.parse(CRI_RSA_ENCRYPTION_PUBLIC_JWK),
```
DanCorderIPV: Wrong key constant

DanCorderIPV: and in a couple more place below

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/fraudCri/CredentialTests.java
```diff
@@ -123,27 +125,35 @@ public RequestResponsePact validRequestReturnsExperianFraudCheckIssuedCredential
...
-                                verifiableCredentialJwtValidator.validate(
-                                        credential, credentialIssuerConfig, TEST_USER);
+                                var vc =
+                                        verifiableCredentialJwtValidator.validate(
+                                                TEST_USER,
+                                                null,
+                                                credential,
+                                                VerifiableCredentialConstants
+                                                        .IDENTITY_CHECK_CREDENTIAL_TYPE,
+                                                ECKey.parse(CRI_RSA_ENCRYPTION_PUBLIC_JWK),
```
DanCorderIPV: Wrong key constant, and below

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/ContractTest.java
DanCorderIPV: Ok, I've traced this through and I understand why the value doesn't matter. However I do find it odd that you pass a value to a method called `validate` that isn't used in the validation. I think I'd feel happier if this method was `parse` or similar instead, then it's clearer that it is converting the credential into an object - and as part of that it is also validating it.

(I can't comment on the relevant bit of code as it hasn't changed in this PR so commenting here instead)

DanCorderIPV (changes requested):
I've only looked at the contract tests, I believe Joe is looking at the rest.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
MikeCollingwood: No, this was unintentional, mean to be consistent. I'm updating them all to take a TEST_CRI_ID = "test-cri-id" because it's a field which gets checked

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/ContractTest.java
MikeCollingwood: I could call it `parse` but I might prefer `validateAndParse`. You're right in that it would explain the operation more clearly. I would like to reference the fact it's validating as part of the process though - what do you think?

DanCorderIPV: `and` in function names is often a bit of a red flag so I'd be careful there. Parsing requires some validation (otherwise the parsing will fail), but you're right that it doesn't stress that aspect, which is actually quite big.

DanCorderIPV: Hmm, it's tricky because some bits like the signing key aren't strictly needed to parse the JWT. That would imply that you should validate the JWT first, but then you'd need to do more validation later on anyway.
Perhaps we should split the validation in two - first check the signature, then validate the JWT contents separately. I'm torn on whether signature checking should be bundled together with the rest of this. On one hand it means we won't ever forget to do it, on the other hand it feels like a separate step.

lambdas/process-async-cri-credential/src/test/java/uk/gov/di/ipv/core/processasynccricredential/pact/f2fCri/ContractTest.java
```diff
@@ -126,9 +125,9 @@ void testF2fMessageReturnsIssuedPassportCredential(
                                 getAsyncResponseMessage(message.contentsAsString()));
 
                 verifiableCredentialValidator
-                        .validate(
+                        .parseAndValidate(
                                 TEST_USER,
-                                F2F_CRI,
+                                TEST_CRI_ID,
```
DanCorderIPV: Why has this changed? We may as well use a realistic value unless it causes problems?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
```diff
@@ -149,13 +150,13 @@ void fetchVerifiableCredential_whenCalledAgainstCicCri_retrievesAValidVc(MockSer
...
-                                        verifiableCredentialJwtValidator.validate(
+                                        verifiableCredentialJwtValidator.parseAndValidate(
                                                 TEST_USER,
-                                                null,
+                                                TEST_CRI_ID,
                                                 credential,
                                                 VerifiableCredentialConstants
                                                         .IDENTITY_CHECK_CREDENTIAL_TYPE,
-                                                ECKey.parse(CRI_RSA_ENCRYPTION_PUBLIC_JWK),
+                                                ECKey.parse(EC_PUBLIC_JWK),
```
DanCorderIPV: I think it's safer to use `CRI_SIGNING_PRIVATE_KEY_JWK` here (as we were before) so that the key we use to generate our test data will always match what we use in the tests.

DanCorderIPV (dismissed):

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/ContractTest.java
MikeCollingwood: I understand your point on the naming, though I think it wouldn't be too consequential whatever naming we choose though I think splitting the method would probably extend this PR's lifespan too much because of its affect on unit tests... what do you think?

DanCorderIPV: I'm fine not to do this in this PR, but I think it's something worth considering more in future work

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
Joe-Edwards-GDS: Don't we validate signatures using the CRI's public key, not private?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -65,22 +65,26 @@ public class BuildUserIdentityHandler
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
     private final CiMitService ciMitService;
     private final CiMitUtilityService ciMitUtilityService;
+    private final VerifiableCredentialService verifiableCredentialService;
 
+    //    @SuppressWarnings()
```
Joe-Edwards-GDS: This shouldn't be in a comment, and it should reference a specific rule (not just suppress everything)

e.g. `@SuppressWarnings("java:S107") // Methods should not have too many parameters`

MikeCollingwood: Ah yep I didn't know what number to suppress by though - so was waiting to get another sonarcloud report

Joe-Edwards-GDS: You can still see the reports if you go to sonar cloud: https://sonarcloud.io/project/pull_requests_list?id=ipv-core-back

(but also, don't delete the sonar comments..!)

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/cicCri/ContractTest.java
MikeCollingwood: You can validate with the private key as if it were the public key

DanCorderIPV: The private key blob there may be inaccurately named - it has both the private and public parts in it so can be used for both.

Joe-Edwards-GDS: It looks like Dan is right - `CRI_SIGNING_PRIVATE_KEY_JWK` contains both the public and private keys
- x and y are the public parts
- d is the private part

You can compare to the key configuration in core-common-infra, where we only have x and y.

(AFAICT we're stubbing out the crypto in the contract tests anyway, so I'm not sure how important this is!)

MikeCollingwood: Ah, I should've elaborated in my earlier comment because this is something I already understood.

I think the key does still matter because we don't mock the verifier (in the validateSignature method):
```java
ECDSAVerifier verifier = new ECDSAVerifier(signingKey);
if (!concatSignatureVerifiableCredential.verify(verifier)) {
```

Joe-Edwards-GDS: Right, so:
- All the unit tests use the test keys defined in `TestFixtures`
- All the contract tests use keys defined in each of the contract tests separately

All of these keys are actually identical, but in principle they could be different.


In terms of this change, we should use the contract test's EC signing key, not the shared fixture, as Dan suggests.

The variable name is also potentially surprising, given that you actually need the public key (not the private key) to verify the signature, but this is far out of scope of this review - it's massive enough already 😬

MikeCollingwood: I've replaced the EC_PUBLIC_JWK references with the private JWKs defined in each contract test 👍

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1696, opened 2024-02-26
BAU: Update shared deps

Joe-Edwards-GDS:
Upgrade shared dependency versions. These are not scanned by dependabot so need updating manually - something to investigate in future.

The Jackson version also needs pinning as we're currently using it in a slightly dodgy way - we're using it as a transitive dependency of lambda powertools, but this relies on some deprecated functionality, and breaks if we upgrade to the latest.

See: https://govukverify.atlassian.net/browse/PYIC-5132


build.gradle
```diff
@@ -33,25 +33,23 @@ spotless {
...
+		awsJavaSdkKms:'1.12.665',
+		awsLambdaJavaCore:'1.2.3',
+		awsLambdaJavaEvents:'3.11.4',
+		dynamodbEnhanced:'2.24.9',
+		gson:'2.10.1',
+		jacksonDataformatYaml:'2.15.3',
+		kms:'2.24.9',
+		log4j:'2.23.0',
+		lombok:'1.18.30',
 		nimbusdsOauth2OidcSdk:'9.27',
```
Joe-Edwards-GDS, line 48: This has quite a few upgrades (including some security upgrades) to pull in, but it changes the way it does JSON parsing, requiring a rewrite of lots of JSON handling code. Instead we should wait for PYIC-3908 to upgrade it.

(cc @MikeCollingwood )

libs/pact-test-helpers/build.gradle
```diff
@@ -12,7 +12,7 @@ repositories {
 dependencies {
 	implementation "com.nimbusds:oauth2-oidc-sdk:$rootProject.ext.dependencyVersions.nimbusdsOauth2OidcSdk",
 			"au.com.dius.pact.consumer:junit5:4.6.5",
-			'com.fasterxml.jackson.core:jackson-databind:2.15.2'
+			"com.fasterxml.jackson.core:jackson-databind:2.15.3"
```
Joe-Edwards-GDS, line 15: We should _really_ add this is an explicit dependency everywhere, rather than relying on the transitive version from powertools.

This is the version currently used by powertools, and upgrading to 2.16 breaks things.

lwebb-iig (approved):

PR #1697, opened 2024-02-26
PYIC-4727 To resolve issue with dynamo serialization for ipvSession. …

rpayal:
…(storing featureSet as list instead of String)

## Proposed changes
To resolve issue with dynamo serialization for ipvSession (storing featureSet as list instead of String)

### What changed
Rename methods in IpvSessionItem to getFeatureSetAsList & setFeatureSetFromList. 
- also added a check for existing session items (featureSet.equals("[ { \"S\" : \"\" } ]")) - which can be removed later on.

### Issue tracking

### Other considerations

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
@@ -101,9 +101,13 @@ public class RequestHelper {
     }
 
     public static List<String> getFeatureSet(JourneyRequest request) {
-        List<String> featureSet = request.getFeatureSet();
-        LogHelper.attachFeatureSetToLogs(featureSet);
-        return (featureSet != null) ? featureSet : Collections.emptyList();
+        String featureSet = request.getFeatureSet();
+        List<String> featureSetList =
+                (featureSet != null && !featureSet.isBlank())
+                        ? Arrays.asList(featureSet.split(","))
+                        : Collections.emptyList();
+        LogHelper.attachFeatureSetToLogs(featureSetList);
+        return featureSetList;
     }
 
     public static List<String> getFeatureSet(APIGatewayProxyRequestEvent event) {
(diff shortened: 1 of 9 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -101,9 +101,13 @@ public static String getIpvSessionId(JourneyRequest request, boolean allowNull)
...
 
     public static List<String> getFeatureSet(JourneyRequest request) {
-        List<String> featureSet = request.getFeatureSet();
-        LogHelper.attachFeatureSetToLogs(featureSet);
-        return (featureSet != null) ? featureSet : Collections.emptyList();
+        String featureSet = request.getFeatureSet();
+        List<String> featureSetList =
+                (featureSet != null && !featureSet.isBlank())
+                        ? Arrays.asList(featureSet.split(","))
+                        : Collections.emptyList();
```
Joe-Edwards-GDS, line 108: (Still feels as though there's potential to pull this out as a method - we've repeated it 4 or 5 times)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -65,13 +65,14 @@ public void addVcReceivedThisSession(String vc) {
         vcReceivedThisSession.add(vc);
     }
 
-    public List<String> getFeatureSet() {
-        return (featureSet != null)
-                ? Stream.of(featureSet.split(",")).toList()
+    public List<String> getFeatureSetAsList() {
+        return (featureSet != null && !featureSet.equals("[ { \"S\" : \"\" } ]"))
```
Joe-Edwards-GDS, line 69: (If this doesn't work, then I think we can probably be a bit more restrictive and use the same regex as the FE to discard invalid featureset configurations)

Joe-Edwards-GDS (commented):
I'll approve once you confirm the test case :)

Joe-Edwards-GDS (approved):

rpayal:
Tested for in-flight scenario - deployed main and then go through the journey till address cri (it created a session record with featureSet([ { "S" : "" } ])
THEN deployed the branch with changes......and then in browser went ahead with journey, no error (it created a new session record with featureSet null)

PR #1708, opened 2024-02-29
PYIC-4490: Added journey-map routing for NINO/BAV escape pages

RogVanGDS:
### What changed

Added journey-map routing for new dynamic NINO/BAV escape pages

### Why did it change

To correctly route users abandoning the M2B journey at the NINO/BAV stages

### Issue tracking

- [PYIC-4490](https://govukverify.atlassian.net/browse/PYIC-4490)

[PYIC-4490]: https://govukverify.atlassian.net/browse/PYIC-4490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -577,12 +581,25 @@ PYI_ESCAPE_M2B:
...
+    type: page
+    context: abandon
+    pageId: pyi-escape-m2b
+  events:
     end:
       targetState: RETURN_TO_RP
+    next:
+      targetState: IDENTITY_START_PAGE
+    bankAccount:
+      targetState: BANK_ACCOUNT_START_PAGE
```
Joe-Edwards-GDS: If you think the test is enforcing this incorrectly, then we may want to change that instead - I don't think we should be adding routes that don't exist! The assumption here is that a given page id will always need to handle a consistent set of events...

RogVanGDS: Yeah I'm thinking of changing it so we are instead testing if given a page id AND context we handle a consistent set of events

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -577,12 +581,23 @@ PYI_ESCAPE_M2B:
...
+PYI_ESCAPE_ABANDON_M2B:
+  response:
+    type: page
+    context: abandon
+    pageId: pyi-escape-m2b
+  events:
     end:
       targetState: RETURN_TO_RP
+    next:
+      targetState: IDENTITY_START_PAGE
```
blakeyp: A thought, do we need/want to go through `RESET_IDENTITY` first here 🤔

RogVanGDS: I guess it depends on what user behaviour we want. 

Do we want this to act as if you were starting the entire journey again? 

Or are we simply routing the user to the DCMAW page so they can submit passport details?

blakeyp: I think we probably do rather than having orphaned CIC/BAV VCs hanging around
but be keen to get @Joe-Edwards-GDS's thoughts on this?

Joe-Edwards-GDS: To check, this is if you abandon the M2B journey in the NINO CRI? In this case it makes sense to me to reset, since the CIC/BAV VCs won't be useful, however it's not strictly necessary.

The main impact is whether we want to restrict the user to the same name/dob as they entered in CIC/BAV the first time round...

(The obvious parallel is the existing 'cri escape' route (i.e. for kbv thin file), where actually we don't clear old VCs, but in that case there is some value to reusing address and fraud, it's not just starting from scratch).

blakeyp: Agreed, yeah this is abandoning at either BAV or NINO CRI

RogVanGDS: Ok that makes sense, I'll switch it to RESET_IDENTITY then

blakeyp (approved):

PR #1721, opened 2024-03-05
PYIC-5233 Update config and cimitUtility service to process newly updated cimit config.

rpayal:
## Proposed changes
Update config and cimitUtility service to process newly updated cimit config.

### What changed
Updated config and cimitUtility service to process newly updated cimit config.

### Issue tracking

- [PYIC-5233](https://govukverify.atlassian.net/browse/PYIC-5233)

### Other considerations

[PYIC-5233]: https://govukverify.atlassian.net/browse/PYIC-5233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Wynndow (commented):
I think that these changes will break with the existing config structure. We're going to need core to be able to handle both the old style and the new style of cimit config at the same time. Once it can, we can update the cimit config to the new style, and then remove the ability to handle the old style from core. It's a bit of a pain but it avoids anything breaking.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -38,12 +43,29 @@ public Optional<JourneyResponse> getCiMitigationJourneyStep(ContraIndicators con
         var cimitConfig = configService.getCimitConfig();
         for (var ci : contraIndicators.getContraIndicatorsMap().values()) {
             if (isCiMitigatable(ci) && !isBreachingCiThresholdIfMitigated(ci, contraIndicators)) {
-                return Optional.of(new JourneyResponse(cimitConfig.get(ci.getCode())));
+                MitigationRoute mitigationRoute =
+                        getMitigationRoute(cimitConfig.get(ci.getCode()), ci.getDocument());
+                return mitigationRoute == null
+                        ? Optional.empty()
```
Wynndow: I'm wondering if we should throw here, rather than returning an empty optional. We've already checked that the CI can be mitigated, so we _should_ be able to find a route. If this was null I think it would mean that we've somehow not matched the document ID on the CI to a document in the config, which I don't think should happen. Rather than using `orElseGet` in the stream in `getMitigationRoute` could use `orElseThrow`?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
```diff
@@ -759,4 +773,14 @@ private static List<String> getFeatureSet(String featureSet) {
                 ? Collections.singletonList(featureSet)
                 : Collections.emptyList();
     }
+
+    private static Stream<Arguments> provideConfiguredSsmCimitConfig() {
```
Wynndow: Could this get moved to just under the test that uses it? Would make it easier to understand what the test is doing.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
```diff
@@ -153,9 +154,17 @@ void getMitigationJourneyResponseShouldReturnMitigationWhenCiCanBeMitigated() th
         // arrange
         var code = "ci_code";
         var journey = "some_mitigation";
-        var ci = ContraIndicator.builder().code(code).issuanceDate("some_date").build();
+        String document = "doc_type/213123";
```
Wynndow, line 159: This tests the scenario of a CI having a document identifier. Would be good to have one for a CI that doesn't have an document identifier too.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/MitigationRoute.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.fasterxml.jackson.annotation.JsonRootName;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+@Getter
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonRootName(value = "routes")
```
Wynndow: I don't think this annotation is needed?  Might also be worth making this a Record instead of a Class?

Wynndow (commented):
This looks good! Just a few comments.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
rpayal: done

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/MitigationRoute.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
rpayal: Earlier I thought of throwing exception but then thought logging it that config is missing. If we throw exception then at handler level we need to catch it and return to /journey/error route.
or should it be to /journey/fail-with-ci route.

Wynndow: I think we want to return /journey/error though, as something wrong has happened. If we just log it we won't become aware of it?

rpayal: done

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
```diff
@@ -221,6 +257,28 @@ void getMitigationJourneyResponseShouldReturnEmptyWhenMitigationDoesNotResolveBr
         var result = ciMitUtilityService.getCiMitigationJourneyStep(cis);
 
         // assert
-        assertEquals(result, Optional.empty());
+        assertEquals(Optional.empty(), result);
+    }
+
+    @Test
+    void
+            getMitigationJourneyResponseShouldThrowExceptionWhenCiMitigationJourneyConfigNotFoundForDocType()
```
Wynndow: I think this test might be the same as the test on line 183?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
Wynndow: Sorry, I more meant a test that succesfully finds the mitigation route for a CI that isn't document related.

Wynndow: So I think something like this might be the missing test:
```
    @Test
    void getMitigationJourneyResponseShouldReturnMitigationWhenCiCanBeMitigateWithNoDoc() throws Exception {
        // arrange
        var code = "ci_code";
        var journey = "some_mitigation";
        var ci =
                ContraIndicator.builder()
                        .code(code)
                        .issuanceDate("some_date")
                        .build();
        var cis = ContraIndicators.builder().contraIndicatorsMap(Map.of(code, ci)).build();
        when(mockConfigService.getCimitConfig())
                .thenReturn(Map.of(code, List.of(new MitigationRoute(journey, null))));
        Map<String, ContraIndicatorConfig> ciConfigMap =
                Map.of(code, new ContraIndicatorConfig(code, 7, -5, "X"));
        when(mockConfigService.getContraIndicatorConfigMap()).thenReturn(ciConfigMap);
        when(mockConfigService.getSsmParameter(CI_SCORING_THRESHOLD)).thenReturn("5");

        // act
        var result = ciMitUtilityService.getCiMitigationJourneyStep(cis);

        // assert
        assertEquals(Optional.of(new JourneyResponse(journey)), result);
    }

```

rpayal: Added ...coverage was not showing that.

Wynndow (approved):
Let's give it a go!

PR #1724, opened 2024-03-06
PYIC-5070: Fix missing calls to TICF

Wynndow:
### What changed

Fix missing calls to TICF

### Why did it change

A few of the escape routes weren't calling TICF. This addresses that and moves those journeys to the INELIGIBLE sub journey.

### Issue tracking

- [PYIC-5070](https://govukverify.atlassian.net/browse/PYIC-5070)

[PYIC-5070]: https://govukverify.atlassian.net/browse/PYIC-5070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ineligible.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ineligible.yaml
@@ -13,6 +13,14 @@ INELIGIBLE_NO_TICF:
     next:
       targetState: ANOTHER_WAY_PAGE
 
+INELIGIBLE_SKIP_MESSAGE:
+  events:
+    next:
+      targetState: RETURN_TO_RP
+      checkFeatureFlag:
+        ticfCriBeta:
+          targetState: CRI_TICF_BEFORE_RETURN_TO_RP
+
 # Journey states
 
 CRI_TICF_BEFORE_ANOTHER_WAY:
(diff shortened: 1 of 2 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ineligible.yaml
```diff
@@ -31,6 +39,22 @@ CRI_TICF_BEFORE_ANOTHER_WAY:
...
 
+CRI_TICF_BEFORE_RETURN_TO_RP:
+  response:
+    type: process
+    lambda: call-ticf-cri
+  events:
+    next:
+      targetState: RETURN_TO_RP
+    enhanced-verification:
+      targetState: RETURN_TO_RP
```
Joe-Edwards-GDS: Do you think this should also be `FAILED` to match `fail-with-ci`?

The only unusual case to consider is when you're already on a mitigation journey (we don't want to re-route you if you already had the CI).

I suspect in practice it'll never happen, so not too important.

Joe-Edwards-GDS (dismissed):
Looks sensible to me, but one question about an edge case

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/ineligible.yaml
Wynndow: I think you're right. The only scenario that enhanced-verification could be triggered here is if the users is on a mitigation journey and opts to bail out from an escape route. Which means they'll have failed, and it will be more useful to them to see a page telling them that than just going back to the service. Will update.

Joe-Edwards-GDS (approved):

PR #1728, opened 2024-03-07
BAU: Fix local running

Wynndow:
### What changed

Fix local running

### Why did it change

The local running setup was failing due to how we were constucting the input to the criCallBack handler - if no feature sets were active we'd get an NPE.

This updates to invoke the handleRequest method on that handler, rather than calling one of the internal methods. This allows the existing helpers to parse the feture sets and do the right thing with it.

Also some refactoring to change a big if statment to a case statement.


local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
```diff
@@ -103,71 +105,60 @@ public LambdaHandler() throws IOException {
...
+                    if (journey == null) {
                         return gson.toJson(processJourneyEventOutput);
                     }
+
+                    switch (journey) {
+                        case "/journey/check-existing-identity":
+                            lambdaOutput =
+                                    checkExistingIdentityHandler.handleRequest(
+                                            buildJourneyRequest(request, journey), EMPTY_CONTEXT);
+                            break;
```
Joe-Edwards-GDS: This looks like a prime candidate for a switch expression, e.g.:
```
lambdaOutput = switch (journey) {
    case "/journey/check-existing-identity" ->
            checkExistingIdentityHandler.handleRequest(
                    buildJourneyRequest(request, journey), EMPTY_CONTEXT);
    case "/journey/reset-identity" ->
            resetIdentityHandler.handleRequest(
                    buildProcessRequest(request, processJourneyEventOutput), EMPTY_CONTEXT);
    ...
};
```

When we update to Java 21 we can also have guarded statements to make the default case neater as well, e.g.:
```
    case String j when j.matches("/journey/cri/build-oauth-request/.*") ->
            buildCriOauthRequestHandler.handleRequest(
                    buildJourneyRequest(request, journey),
                    EMPTY_CONTEXT);
```

local-running/src/test/java/uk/gov/di/ipv/coreback/handlers/HomeHandlerTest.java
Joe-Edwards-GDS, line 1: I don't think we should be adding this test just to satisfy Sonar - if we're adding tests then they should be doing something valuable...

If we can't write useful tests, then we should probably just exclude this entire project from coverage reporting.

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
Wynndow: ooOOoo nice. Didn't realise you could do the first thing. And the second thing will be nice!

local-running/src/test/java/uk/gov/di/ipv/coreback/handlers/HomeHandlerTest.java
Wynndow: It's not that we can't write useful tests, it's more about is it worth it for this tool that pretty much no one uses and isn't going anywhere near prod. Some would argue yes.

I'd have to look at Sonar to get it to ignore this project...

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
Joe-Edwards-GDS: Yep - introduced in Java 14 I think.

It'll also let you do fancy 'Pattern matching', where you can deconstruct record classes too:
```
case ReturnCode("A") -> /* do something */
case ReturnCode(var code) -> /* do something with code */
// etc.
```

It's one of my favourite features from Scala, which takes this to a somewhat extreme level but is totally idiomatic.

Joe-Edwards-GDS (approved):

PR #1735, opened 2024-03-08
PYIC-5272 Journey map changes for alternative doc session and separate session mitigation.

rpayal:
## Proposed changes
Journey map changes for alternate doc session and separate session mitigation.

### Issue tracking

- [PYIC-5272](https://govukverify.atlassian.net/browse/PYIC-5272)

### Other considerations

[PYIC-5272]: https://govukverify.atlassian.net/browse/PYIC-5272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -45,7 +45,13 @@ CHECK_EXISTING_IDENTITY:
...
       targetJourney: NEW_P2_IDENTITY
       targetState: ENHANCED_VERIFICATION
-
+    alternate-doc-invalid-dl:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
+      checkFeatureFlag:
+        d02MitigationEnabled:
+          targetJourney: NEW_P2_IDENTITY
+          targetState: ALTERNATE_DOC
```
Wynndow: I'm wondering if it might be better, rather than handling the feature flag in the journey map like this, to hide the emitting of this event behind the flag in the code.

With the flag disabled, this event should be treated the same as the `fail-with-ci` event. Which is complicated by the fact that we also have `fail-with-ci-and-forced-reset`.

If we ensure the code only emits the `alternate-doc-invalid-dl` event in exactly the right circumstance (when the feature flag is active), we can avoid handling it in the journey map like this, which should keep the `fail-with-ci/fail-with-ci-and-forced-reset` situation simpler.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -386,6 +394,12 @@ CRI_DRIVING_LICENCE_J3:
       checkIfDisabled:
         f2f:
           targetState: MULTIPLE_DOC_CHECK_PAGE
+    alternate-doc-invalid-dl:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
```
Wynndow: I think that this should be `targetJourney: FAILED, targetState: FAILED`?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +807,37 @@ PYI_POST_OFFICE:
...
+      targetState: CRI_UK_PASSPORT_J2
+    end:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
+
+# Mitigation journey (04) separate session - invalid-dl
+MITIGATION_04:
+  response:
+    type: page
+    pageId: pyi-driving-license-no-match
```
Wynndow: I think that `mitigationStart: invalid-dl` should be defined on this event, and not the `MITIGATION_04_IDENTITY_START_PAGE` state.

rpayal: thought this an intermediate page .......actual option to select Passport journey is in next page pyi-continue-with-passport...thats way added in state MITIGATION_04_IDENTITY_START_PAGE

Wynndow: Yep fair, looking at the content on those pages I think you're right.

Wynndow:
Just noting that we've discussed this offline and agreed to create new states for the entire mitigation journey - passport, address/fraud and KBV. This will allow us to keep the user confined to the correct path.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -15,6 +15,8 @@ ADDRESS_AND_FRAUD:
           targetState: CRI_FRAUD
         enhanced-verification:
           targetState: CRI_FRAUD
+        alternate-doc-invalid-dl:
+          targetState: CRI_FRAUD
```
Wynndow: I'm not sure these changes are needed. If the user is the mitigation jouney, they'll have already mitigated the CI by the time they reach the address CRI - they'll have done it by succesfully coming throught the passport CRI. So this event shouldn't get thrown by the address CRI.
We need to handle the `enhanced-verification` event here due the way separate session mitigation of a different CRI works - but thankfully we don't have that issuer here.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -25,3 +27,5 @@ ADDRESS_AND_FRAUD:
           exitEventToEmit: next
         enhanced-verification:
           exitEventToEmit: enhanced-verification
+        alternate-doc-invalid-dl:
+          exitEventToEmit: alternate-doc-invalid-dl
```
Wynndow: Same situation here.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -10,6 +10,11 @@ ENHANCED_VERIFICATION:
     next:
       targetState: MITIGATION_01
 
+ALTERNATE_DOC:
```
Wynndow: Might be better called something like `ALTERNATE_DOC_PASSPORT`. Just thinking ahead to when we bring in the failed passport mitigation route.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -342,6 +350,9 @@ ADDRESS_AND_FRAUD_J1:
     enhanced-verification:
       targetJourney: FAILED
       targetState: FAILED
+    alternate-doc-invalid-dl:
+      targetJourney: FAILED
+      targetState: FAILED
```
Wynndow: If you remove the event from the nestesd journey definitions we won't need this.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -371,6 +382,11 @@ ADDRESS_AND_FRAUD_J2:
       checkIfDisabled:
         hmrcKbv:
           targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE
+    alternate-doc-invalid-dl:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
```
Wynndow: Same here - I think we can lose this.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -394,6 +416,8 @@ ADDRESS_AND_FRAUD_J3:
       targetState: CHECK_FRAUD_SCORE_J3
     enhanced-verification:
       targetState: CHECK_FRAUD_SCORE_J3
+    alternate-doc-invalid-dl:
+      targetState: CHECK_FRAUD_SCORE_MITIGATION
```
Wynndow: I think this can be dropped too.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -437,6 +461,9 @@ ADDRESS_AND_FRAUD_J4:
       checkFeatureFlag:
         ticfCriBeta:
           targetState: CRI_TICF_BEFORE_F2F
+    alternate-doc-invalid-dl:
+      targetJourney: FAILED
+      targetState: FAILED
```
Wynndow: Same here - this can go

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -533,6 +560,11 @@ ADDRESS_AND_FRAUD_M2B:
       checkIfDisabled:
         hmrcKbv:
           targetState: PRE_EXPERIAN_KBV_TRANSITION_PAGE_M2B
+    alternate-doc-invalid-dl:
+      targetState: CRI_HMRC_KBV_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
```
Wynndow: This too

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +825,146 @@ PYI_POST_OFFICE:
...
+  response:
+    type: page
+    pageId: pyi-continue-with-passport
+    mitigationStart: invalid-dl
+  events:
+    next:
+      targetState: CRI_UK_PASSPORT_MITIGATION
+    end:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
```
Wynndow: I'm going to double check with UCD that INELIGBLE is more appropriate here than FAILED

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +825,146 @@ PYI_POST_OFFICE:
...
+    enhanced-verification:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
+    alternate-doc-invalid-dl:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
```
Wynndow: I think this event can be dropped

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +825,146 @@ PYI_POST_OFFICE:
...
+    next:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
+    enhanced-verification:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
```
Wynndow: This event shouldn't be possible - it would mean that the passport or DL CRIs have issued an unexpected CI. I'm wondering if we should be sending the user to an error page in that scenario.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +825,146 @@ PYI_POST_OFFICE:
...
+    mitigationStart: invalid-dl
+  events:
+    next:
+      targetState: CRI_UK_PASSPORT_MITIGATION
+    end:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
+
+# Passport journey (MITIGATION)
+CRI_UK_PASSPORT_MITIGATION:
```
Wynndow: I think we should prefix all these new states with something like`MITIGATION_PP`. 
I think we're going to have to define the states againw when we implement the driving licence mitigation route - this is because for the driving licence route we'll need to include the fraud score check, which we shouldn't in the passport route.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -793,3 +825,146 @@ PYI_POST_OFFICE:
...
+      scoreThreshold: 2
+  events:
+    met:
+      targetState: CRI_NINO_MITIGATION
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: PRE_EXPERIAN_KBV_TRANSITION_MITIGATION_PAGE
+    unmet:
+      targetJourney: FAILED
+      targetState: FAILED
```
Wynndow: Don't think we'll need this until we implement the driving licence mitigation route. Might be worth dropping it for now to keep the change set smaller.

Wynndow (commented):
Few things to look at here - do you want to pair on it to help get it over the line?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: Sure later will be needed one for DL. Done.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
rpayal: DOne

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: May not need this...will discuss.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
rpayal: Journey event alternate-doc-invalid-dl is only return if mitigation feature flag is enabled.

Left the feature flag config here......as in Journey MAP this will only be displayed when mitigation flag is turned on.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -810,3 +823,122 @@ PYI_POST_OFFICE:
...
+    type: process
+    lambda: reset-identity
+    lambdaInput:
+      isUserInitiated: false
+      deleteOnlyGPG45VCs: true
+  events:
+    next:
+      targetState: MITIGATION_04_DL_NO_MATCH
+
+MITIGATION_04_DL_NO_MATCH:
```
Wynndow:
```suggestion
MITIGATION_04_DL_NO_MATCH_PAGE:
```

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -810,3 +823,122 @@ PYI_POST_OFFICE:
...
+  nestedJourney: ADDRESS_AND_FRAUD
+  exitEvents:
+    next:
+      targetState: MITIGATION_PP_CRI_NINO
+      checkIfDisabled:
+        hmrcKbv:
+          targetState: MITIGATION_PRE_EXPERIAN_KBV_TRANSITION_PAGE
+    enhanced-verification:
+      targetJourney: FAILED
+      targetState: FAILED
```
Wynndow, line 896: I don't think we need these

rpayal: Will fail test as nested journey - ADDRESS_AND_FRAUD emit event enhanced-verification.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -810,3 +823,122 @@ PYI_POST_OFFICE:
...
+  response:
+    type: page
+    pageId: pyi-driving-licence-no-match-another-way
+    mitigationStart: invalid-dl
+  events:
+    next:
+      targetState: MITIGATION_PP_CRI_UK_PASSPORT
+    end:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
```
Wynndow: I've just heard back about the UCD, and we should be using `targetState: INELIGIBLE_SKIP_MESSAGE` here. That will take the user straight back to the service without showing them an extra "another way" page.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -810,3 +823,122 @@ PYI_POST_OFFICE:
...
+  response:
+    type: page
+    pageId: pyi-continue-with-passport
+    mitigationStart: invalid-dl
+  events:
+    next:
+      targetState: MITIGATION_PP_CRI_UK_PASSPORT
+    end:
+      targetJourney: INELIGIBLE
+      targetState: INELIGIBLE
```
Wynndow: Same here for `INELIGIBLE_SKIP_MESSAGE`.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done...return to RP instead of another way.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
rpayal: done

libs/journey-uris/src/main/java/uk/gov/di/ipv/core/library/journeyuris/JourneyUris.java
```diff
@@ -28,4 +28,7 @@ private JourneyUris() {
     public static final String JOURNEY_TEMPORARILY_UNAVAILABLE_PATH =
             "/journey/temporarily-unavailable";
     public static final String JOURNEY_UNMET_PATH = "/journey/unmet";
+    public static final String JOURNEY_ALTERNATE_DOC_PATH = "/journey/alternate-doc";
+    public static final String JOURNEY_ALTERNATE_DOC_INVALID_DL_PATH =
```
Wynndow, line 32: I can't see this being used anywhere, can it be dropped?

rpayal: used in test.
Also updated test names.

Wynndow (approved):
Nice! Thanks for this, sorry there was so much back and forth.

PR #1740, opened 2024-03-11
BAU: Add search box to journey map

Wynndow:
### What changed

This adds a search box to the journey map. It will center the state (if found), maintaining the current zoom level, and highlight it. It's case insensitive.

### Why did it change

Because looking for a particular state with just your eyeballs is annoying.

### Screenshot
Screenshot 2024-03-11 at 14 12 40

### DIFF
--- a/journey-map/public/index.html
+++ b/journey-map/public/index.html
@@ -14,6 +14,7 @@
                     </svg>
                     <span>IPV Core Journey Map</span>
                 </h1>
+                <input id="state-search" type="search" placeholder="Search for state...">
             </div>
             <div id="header-content">
                 <form id="configuration-form">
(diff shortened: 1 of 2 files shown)


journey-map/public/index.html
```diff
@@ -14,6 +14,7 @@ <h1>
                     </svg>
                     <span>IPV Core Journey Map</span>
                 </h1>
+                <input id="state-search" type="text" placeholder="Search for state...">
```
lwebb-iig: Could consider making this a `search` type (if you haven't already), get a few minor UX things for free:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#differences_between_search_and_text_types

lwebb-iig (dismissed):

journey-map/public/index.html
Wynndow: Sounds good, will update it

lwebb-iig (approved):

PR #1742, opened 2024-03-12
PYIC-4942: CIMIT D02 - Always re-mitigate in separate session

sammyiyke:
### What changed

[PYIC-4942](https://govukverify.atlassian.net/browse/PYIC-4942) CIMIT D02 - Always re-mitigate in separate session.

Adds new rules for CI Mitigations.

- At the start of a journey users still have to follow same mitigation journey for any CIs mitigated in a previous incomplete journey
- You can't follow two mitigation routes at the same time
- Users cannot proceed on a new journey if they have a mitigated CI and new unmitigated CIs

### Why did it change

When returning, the user will be forced to take a mitigation route after returning to IPV Core, even if that CI is mitigated: 

- Mitigated D02/V03 + Complete identity => Reuse

- Mitigated D02 + Incomplete identity => Mitigation route

- Mitigated V03 + Incomplete identity => Mitigation route

- Mitigated D02 + V03 => No Match

- Mitigated D02 + D02 => No Match

### Issue tracking

- [PYIC-4942](https://govukverify.atlassian.net/browse/PYIC-4942)

[PYIC-4942]: https://govukverify.atlassian.net/browse/PYIC-4942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-4942]: https://govukverify.atlassian.net/browse/PYIC-4942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -379,7 +379,8 @@ public class CheckExistingIdentityHandler
                     .getMitigatedCiJourneyStep(mitigatedCI.get())
                     .orElse(JOURNEY_RESET_GPG45_IDENTITY);
         }
-        if (!VcHelper.filterVCBasedOnProfileType(verifiableCredentials, ProfileType.GPG45).isEmpty()) {
+        if (!VcHelper.filterVCBasedOnProfileType(verifiableCredentials, ProfileType.GPG45)
+                .isEmpty()) {
             LOGGER.info(
                     LogHelper.buildLogMessage("Failed to match profile so resetting identity."));
             sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_RESET, auditEventUser);
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -376,10 +376,11 @@ public class CheckExistingIdentityHandler
             LOGGER.info(
                     LogHelper.buildLogMessage("Failed to match profile so resetting identity."));
             sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_RESET, auditEventUser);
-            if (ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators).isPresent()) {
+            var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
+            if (mitigatedCI.isPresent()) {
                 return ciMitUtilityService
-                        .getCiMitigationJourneyStep(contraIndicators)
-                        .orElse(JOURNEY_FAIL_WITH_CI);
+                        .getMitigatedCiJourneyStep(mitigatedCI.get())
+                        .orElse(JOURNEY_RESET_GPG45_IDENTITY);
             }
             return JOURNEY_RESET_GPG45_IDENTITY;
         }
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -380,13 +380,14 @@ public class CheckExistingIdentityHandler
         if (mitigatedCI.isPresent()) {
             var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
             if (mitigationRoute.isPresent()) {
-                if (mitigationRoute.get().getJourney().equals(JOURNEY_ENHANCED_VERIFICATION_PATH)) {
-                    return JOURNEY_ENHANCED_VERIFICATION_F2F_FAIL;
+                JourneyResponse journeyResponse = mitigationRoute.get();
+                if (!journeyResponse.getJourney().equals(JOURNEY_ENHANCED_VERIFICATION_PATH)) {
+                    throw new UnsupportedMitigationRouteException(
+                            String.format(
+                                    "Unsupported mitigation route: %s",
+                                    journeyResponse.getJourney()));
                 }
-                throw new UnsupportedMitigationRouteException(
-                        String.format(
-                                "Unsupported mitigation route: %s",
-                                mitigationRoute.get().getJourney()));
+                return JOURNEY_ENHANCED_VERIFICATION_F2F_FAIL;
             }
         }
         return JOURNEY_F2F_FAIL;
@@ -396,13 +397,17 @@ public class CheckExistingIdentityHandler
             List<VerifiableCredential> verifiableCredentials,
             AuditEventUser auditEventUser,
             ContraIndicators contraIndicators)
-            throws SqsException, MitigationRouteConfigNotFoundException, ConfigException {
+            throws SqsException, MitigationRouteConfigNotFoundException, ConfigException, UnsupportedMitigationRouteException {
 
         var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
         if (mitigatedCI.isPresent()) {
-            return ciMitUtilityService
-                    .getMitigatedCiJourneyStep(mitigatedCI.get())
-                    .orElse(JOURNEY_RESET_GPG45_IDENTITY);
+            var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
+            if (mitigationRoute.isEmpty()) {
+                throw new UnsupportedMitigationRouteException(
+                        String.format(
+                                "Empty mitigation route for mitigated CI: %s", mitigatedCI.get()));
+            }
+            return mitigationRoute.get();
         }
         if (!VcHelper.filterVCBasedOnProfileType(verifiableCredentials, ProfileType.GPG45)
                 .isEmpty()) {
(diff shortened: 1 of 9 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -381,11 +378,19 @@ private JourneyResponse buildF2FNoMatchResponse(
...
+            throws SqsException, MitigationRouteConfigNotFoundException, ConfigException {
         if (!VcHelper.filterVCBasedOnProfileType(vcStoreItems, ProfileType.GPG45).isEmpty()) {
             LOGGER.info(
                     LogHelper.buildLogMessage("Failed to match profile so resetting identity."));
             sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_RESET, auditEventUser);
+            if (ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators).isPresent()) {
+                return ciMitUtilityService
+                        .getCiMitigationJourneyStep(contraIndicators)
+                        .orElse(JOURNEY_FAIL_WITH_CI);
+            }
```
Wynndow: I think this might need to be outside of this `if` block. That `if` is there to ensure we reset the users VC before sending them on a full IPV journey, if we need to.

The separate session mitigation routes will all reset the users VC anyway, so we don't need to worry about that.

In this `buildNoMatchResponse` method I think we want to do three things:
1. Check if the user has mitigated CI already - if they do find the mitigation journey step from CI config for that CI and send them on it. If they don't have a mitigated CI, continue.
2. Check if they've got any GPG45 VC. If they do, send them on `JOURNEY_RESET_GPG45_IDENTITY`. If they don't, continue.
3. If we've not returned already, return `JOURNEY_IPV_GPG45_MEDIUM`.

We're already doing step 2 and 3, so you should be able to just slide in above them.

I don't think your use of `getCiMitigationJourneyStep` will give you what you want though. That method will only look at unmitigated CI. Maybe you need a new method in the `cimitUtilitySerice` - something like `getMitigatedCiJourneyStep` which takes a single CI. You could pass it the result of `hasMitigatedContraIndicator` which returns an optional of the mitigated CI.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -360,24 +365,51 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
+            List<VerifiableCredential> verifiableCredentials,
+            AuditEventUser auditEventUser,
+            ContraIndicators contraIndicators)
+            throws SqsException, MitigationRouteConfigNotFoundException, ConfigException {
+
+        var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
+        if (mitigatedCI.isPresent()) {
+            return ciMitUtilityService
+                    .getMitigatedCiJourneyStep(mitigatedCI.get())
+                    .orElse(JOURNEY_RESET_GPG45_IDENTITY);
```
Wynndow: I think here, rather than returning a `JOURNEY_RESET_GPG45_IDENTITY`, we'd probably want to throw. We've found they've got a mitigated CI, so there should never be a reason we can't get the mitigation journey step for that CI. If we can't, something's gone badly wrong.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -360,24 +365,51 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
         LOGGER.info(LogHelper.buildLogMessage("F2F return - failed to match a profile."));
         sendAuditEvent(
                 !areGpg45VcsCorrelated
                         ? AuditEventTypes.IPV_F2F_CORRELATION_FAIL
                         : AuditEventTypes.IPV_F2F_PROFILE_NOT_MET_FAIL,
                 auditEventUser);
+        var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
+        if (mitigatedCI.isPresent()) {
+            var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
+            if (mitigationRoute.isPresent()) {
```
Wynndow, line 382: I think we should also throw here - same sitation as my comment below - we should always have found a mitigation route.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -360,24 +365,51 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
+                if (mitigationRoute.get().getJourney().equals(JOURNEY_ENHANCED_VERIFICATION_PATH)) {
+                    return JOURNEY_ENHANCED_VERIFICATION_F2F_FAIL;
+                }
+                throw new UnsupportedMitigationRouteException(
+                        String.format(
+                                "Unsupported mitigation route: %s",
+                                mitigationRoute.get().getJourney()));
```
Wynndow: Looking at this I think it might be more readable if the logic was inverted so `if (!mitigationRoute.get()...) { throw new Unsupported....}`, and then returne the journey event after.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -43,6 +43,9 @@ public Optional<JourneyResponse> getCiMitigationJourneyStep(ContraIndicators con
         var cimitConfig = configService.getCimitConfig();
         for (var ci : contraIndicators.getContraIndicatorsMap().values()) {
             if (isCiMitigatable(ci) && !isBreachingCiThresholdIfMitigated(ci, contraIndicators)) {
+                if (hasMitigatedContraIndicator(contraIndicators).isPresent()) {
```
Wynndow, line 47: Maybe stick a comment in here to explain what this is for, it might be hard to work out for future us.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -68,8 +71,30 @@ private MitigationRoute getMitigationRoute(
...
+
+    public Optional<ContraIndicator> hasMitigatedContraIndicator(
+            ContraIndicators contraIndicators) {
+        for (var ci : contraIndicators.getContraIndicatorsMap().values()) {
+            if (ci.isMitigated()) {
+                return Optional.of(ci);
+            }
+        }
+        return Optional.empty();
+    }
```
Wynndow: This might be a good place to use a stream
```suggestion
        return contraIndicators.getContraIndicatorsMap().values().stream()
                .filter(ContraIndicator::isMitigated)
                .findFirst();
    }
```

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -380,13 +380,14 @@ private JourneyResponse buildF2FNoMatchResponse(
         if (mitigatedCI.isPresent()) {
             var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
             if (mitigationRoute.isPresent()) {
-                if (mitigationRoute.get().getJourney().equals(JOURNEY_ENHANCED_VERIFICATION_PATH)) {
-                    return JOURNEY_ENHANCED_VERIFICATION_F2F_FAIL;
+                JourneyResponse journeyResponse = mitigationRoute.get();
+                if (!journeyResponse.getJourney().equals(JOURNEY_ENHANCED_VERIFICATION_PATH)) {
```
Wynndow: Might be slightly safer to switch this around - if `getJourney()` was ever null we'd get a NPE. If you do it the other way around you won't. So:
`if (!JOURNEY_ENHANCED_VERIFICATION_PATH.equals(journeyResponse.getJourney()))`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -360,24 +365,57 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
+
+        var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
+        if (mitigatedCI.isPresent()) {
+            var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
+            if (mitigationRoute.isEmpty()) {
+                throw new UnsupportedMitigationRouteException(
+                        String.format(
+                                "Empty mitigation route for mitigated CI: %s", mitigatedCI.get()));
+            }
+            return mitigationRoute.get();
```
Wynndow: Because you're using optionals here, you could avoid the if statement by doing something like this:
```
return ciMitUtilityService
                    .getMitigatedCiJourneyStep(mitigatedCI.get())
                    .orElseThrow(
                            () ->
                                    new UnsupportedMitigationRouteException(
                                            String.format(
                                                    "Empty mitigation route for mitigated CI: %s",
                                                    mitigatedCI.get())));
```

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -360,24 +365,57 @@ private Optional<JourneyResponse> checkForProfileMatch(
...
-            throws SqsException {
+            List<VerifiableCredential> verifiableCredentials,
+            AuditEventUser auditEventUser,
+            ContraIndicators contraIndicators)
+            throws SqsException, MitigationRouteConfigNotFoundException, ConfigException,
+                    UnsupportedMitigationRouteException {
+
+        var mitigatedCI = ciMitUtilityService.hasMitigatedContraIndicator(contraIndicators);
+        if (mitigatedCI.isPresent()) {
+            var mitigationRoute = ciMitUtilityService.getMitigatedCiJourneyStep(mitigatedCI.get());
```
Wynndow: I think this is always going to come back as empty. The `getMitigatedCiJourneyStep` method uses the `isCiMitigatable` method. Which only returns true if the CI being mitigated hasn't already been mitigated. We need to know the mitigation route for an already mitigated CI.

Wynndow (approved):
🚀

PR #1750, opened 2024-03-13
BAU: Prevent duplicate keys in journey maps

Wynndow:
### What changed

Prevent duplicate keys in journey maps

### Why did it change

We recently had an issue where a duplicate state was added to one of them journey maps. This configures the object mapper used to parse the yaml files to not allow duplicates.


Joe-Edwards-GDS (approved):
(And I assume that duplicates in the main maps are caught by the 'should correctly deserialise' test)

Wynndow:
> (And I assume that duplicates in the main maps are caught by the 'should correctly deserialise' test)

Yep, I tested this change on the version of the joureny map that had that original problem in it, and it blew up.

PR #1755, opened 2024-03-18
PYIC-5137: Return a custom OAuth error code for issues with inhertied identity JWT

Wynndow:
For reviewing the tests, suggest commit by commit. Commit two is just a big refactor.

### What changed

Return a custom OAuth error code for issues with inhertied identity JWT.

### Why did it change

[PYIC-5137: Return specific error code for inherited identities](https://github.com/govuk-one-login/ipv-core-back/commit/02d772e6fb8d86538a952d429d3831c9a6f9655b)

Orchestrator would like to receive a specific OAuth error code for if
the JAR validation fails due to an inherited identity. Currently they
will receive "invalid_request_object" for all issues with the JAR.

This creates a new `ErrorObject` that can be returned in the event of a
problem validating the inherited identity JWT contained within the JAR.

[PYIC-5137: Tidy up initalise ipv session tests](https://github.com/govuk-one-login/ipv-core-back/commit/ccfc364a83efaa4e0c149bb290467a68063dec21)

Moves the inherited identity tests to their own nested block, which
allows for some code to be moved to a setup block.

Also moves some stuff to constants so it can be reused by each test
rather than setting it up each time.

### Issue tracking

- [PYIC-5137](https://govukverify.atlassian.net/browse/PYIC-5137)

[PYIC-5137]: https://govukverify.atlassian.net/browse/PYIC-5137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jose.JWEObject;
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.oauth2.sdk.ErrorObject;
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.util.StringUtils;
 import org.apache.http.HttpStatus;
@@ -80,6 +81,8 @@ public class InitialiseIpvSessionHandler
     private static final String REQUEST_EMAIL_ADDRESS_KEY = "email_address";
     private static final String REQUEST_VTR_KEY = "vtr";
     private static final List<Vot> HMRC_PROFILES_BY_STRENGTH = List.of(Vot.PCL250, Vot.PCL200);
+    private static final ErrorObject INVALID_INHERITED_IDENTITY_ERROR_OBJECT =
+            new ErrorObject("invalid_inherited_identity");
     private final ConfigService configService;
     private final IpvSessionService ipvSessionService;
     private final ClientOAuthSessionDetailsService clientOAuthSessionService;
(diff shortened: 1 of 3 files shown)


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -319,7 +322,7 @@ private void validateAndStoreHMRCInheritedIdentity(
                 | UnrecognisedVotException
                 | AuditExtensionException e) {
             throw new RecoverableJarValidationException(
-                    OAuth2Error.INVALID_REQUEST_OBJECT.setDescription(
+                    INVALID_INHERITED_IDENTITY_ERROR_OBJECT.setDescription(
                             "Inherited identity JWT failed to validate"),
                     claimsSet,
                     e);
```
Joe-Edwards-GDS, line 328: Do we also need to catch/rethrow other `JarValidationException`/`RecoverableJarValidationException`s?

(Would we ever want to throw another error code?)

Wynndow: Are you asking if there are other issues we would want to catch that would have a different description?
For inherited identity JWTs we're catching three scenarios - failure to validate, jwt is null, and more/less than 1 jwt recieved.

Joe-Edwards-GDS: Ah, I was thinking of the first case - but those will all be caught under `VerifiableCredentialException`. For some reason in my head that might throw a `JarValidationException`, but I don't think that's possible...

Wynndow: We're throwing the JarValidationException if the inherited jwt is null, 0, or more than 1 (and then catching it as a recoverable.)
And catching everything else from validating the jwt itself and rethrowing.

Joe-Edwards-GDS (approved):

PR #1756, opened 2024-03-18
PYIC 5429 Journey map changes to implement Passport- alternate document mitigation

rpayal:
## Proposed changes
- Journey map changes to implement Passport-alternate document  mitigation
- Changes to handle multiple CI's with  same or different unmitigated CI. 
- Alternate doc mitigation then kbv thin file should display another-way instead of no-match page

### Issue tracking

- [PYIC-5429](https://govukverify.atlassian.net/browse/PYIC-5429)

### Other considerations

[PYIC-5429]: https://govukverify.atlassian.net/browse/PYIC-5429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -27,24 +31,38 @@ public boolean isBreachingCiThreshold(ContraIndicators contraIndicators) {
...
             throws ConfigException, MitigationRouteConfigNotFoundException {
-        // Try to mitigate an unmitigated ci to resolve the threshold breach
         var cimitConfig = configService.getCimitConfig();
-        for (var ci : contraIndicators.getContraIndicatorsMap().values()) {
+        // Check for multiple unmitigated ci's
+        Collection<ContraIndicator> contraIndicatorsMapValues =
+                contraIndicators.getContraIndicatorsMap().values();
+        if (checkIfMultipleCIForMitigation(contraIndicatorsMapValues)) {
+            return Optional.of(new JourneyResponse(JOURNEY_MULTIPLE_UNMITIGATED_CIS_PATH));
+        }
```
Wynndow: Do we need this check? If a user has multiple CIs then the check for `isCiMitigatable(ci)` will usually return false (case when true described below), as it includes a check that the users score will be below the threshold if it gets mitigated. Which it won't be if there are multiple. This should allow us to just return the usual `fail-with-ci` journey response.   

Sometimes a user can pick up a CI that is lower than the threshold. But we don't let that stop their journey. If they then picked up a mitigatable CI, this check would prevent them from trying to mitigate, which we shouldn't.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -27,24 +31,38 @@ public boolean isBreachingCiThreshold(ContraIndicators contraIndicators) {
...
                         getMitigationRoute(cimitConfig.get(ci.getCode()), ci.getDocument()).event();
+                if (contraIndicatorsMapValues.stream()
+                                .filter(
+                                        contraIndicator ->
+                                                (!contraIndicator.getCode().equals(ci.getCode())
+                                                        && contraIndicator.isMitigated()))
+                                .count()
+                        > 0) {
+                    return Optional.of(new JourneyResponse(JOURNEY_MULTIPLE_UNMITIGATED_CIS_PATH));
+                }
```
Wynndow: For each CI, this is checking if there are any other mitigated CI and failing the journey if there are?
I think [Sammy's changes](https://github.com/govuk-one-login/ipv-core-back/pull/1742/files#diff-2d71258ca3bf7567d48888471013f3a02da3de7c209c2490d6042a1756ec57cfR46), merged yesterday, are already covering this scenario.

Wynndow (commented):
I think we might have both fixed the same issuer regarding mulltiple similar CIs, but in different ways. I've got a PR for it here: https://github.com/govuk-one-login/ipv-core-back/pull/1758
Interested in your thoughts on it.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
rpayal: Just for separation I'd exposed new event which also take us to no-match page (as done by fail-with-ci)  in case of multiple ci's for mitigation.
As discussed will revert changes done as part of multiple similar CI's fix and will pick master changes.

rpayal:
Seems working ok, tested in dev env.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -900,6 +914,89 @@ MITIGATION_PP_ADDRESS_AND_FRAUD:
...
+      targetJourney: FAILED
+      targetState: FAILED
+
+MITIGATION_DL_ADDRESS_AND_FRAUD:
+  nestedJourney: ADDRESS_AND_FRAUD
+  exitEvents:
+    next:
+      targetState: MITIGATION_CHECK_FRAUD_SCORE
+    enhanced-verification:
+      targetState: MITIGATION_CHECK_FRAUD_SCORE
```
Wynndow: This isn't an event we should ever see from this state, but we should probably send it to `FAILED` rather than continue mitigating

rpayal: done

Wynndow (approved):
Nice!

PR #1758, opened 2024-03-19
BAU: Allow multiple same CI codes

Wynndow:
### What changed

Allow multiple same CI codes

### Why did it change

Core was blowing up when a user had more than one CI with the same code. This is likely to happen with the new mitigation work.

Internally we were converting the list of CI received from CiMit into a map, with the CI code as the key. This doesn't work for multiple same CI codes.

This change updates the internal structure to a list, and fixes up all the methods that previosuly operated on the map to operate on the list instead.

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
@@ -6,17 +6,15 @@ import lombok.ToString;
 import uk.gov.di.ipv.core.library.domain.cimitvc.ContraIndicator;
 import uk.gov.di.ipv.core.library.exceptions.UnrecognisedCiException;
 
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 @Getter
 @Builder(toBuilder = true)
 @ToString
 public class ContraIndicators {
-    private final Map<String, ContraIndicator> contraIndicatorsMap;
+    private final List<ContraIndicator> usersContraIndicators;
 
     public Integer getContraIndicatorScore(
             final Map<String, ContraIndicatorConfig> contraIndicatorScores)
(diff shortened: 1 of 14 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
```diff
@@ -27,20 +27,32 @@ public Integer getContraIndicatorScore(
     }
 
     public Optional<ContraIndicator> getLatestContraIndicator() {
-        return contraIndicatorsMap.values().stream()
```
Wynndow, line 30: The existing test for this method was only passing by chance as far as I could tell, relying on the ordering of a map's values. Hence the update.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
```diff
@@ -27,20 +27,32 @@ public Integer getContraIndicatorScore(
...
+                .max(
+                        Comparator.comparing(ContraIndicator::getIssuanceDate)
+                                .thenComparing(
+                                        (a, b) ->
+                                                String.format("%s%s", a.getCode(), a.getDocument())
+                                                        .compareToIgnoreCase(
+                                                                String.format(
+                                                                        "%s%s",
+                                                                        b.getCode(),
+                                                                        b.getDocument()))));
```
Joe-Edwards-GDS: Is this equivalent to:
```
return usersContraIndicators.stream()
                .max(
                        Comparator.comparing(ContraIndicator::getIssuanceDate)
                                .thenComparing(ContraIndicator::getCode)
                                .thenComparing(ContraIndicator::getDocument));
```
Which I think conveys the intent more clearly?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
Joe-Edwards-GDS: Is this method actually used anywhere?

Joe-Edwards-GDS (commented):
Funny that this actually makes things simpler

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
Wynndow: haha. yeah, much better

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicators.java
Wynndow: On closer inspection, no, only by the tests. It would be useful if intellij would highlight that situation.\
I'll remove it.

Joe-Edwards-GDS (dismissed):
I think you've already observed that this'll conflict with some of @rpayal 's work, but I think this is probably the right approach.

This leaves the responsibility for deduplicating CIs to CIMIT, which is the final arbiter.

rpayal (approved):
LGTM

PR #1762, opened 2024-03-20
BAU: Fix typo in page ID

Wynndow:
### What changed

Fix typo in page ID

### Why did it change

The original page name doesn't match with the file in the frontend, which was causing a technical error page to render.


rpayal (approved):
nice spot

PR #1763, opened 2024-03-21
PYIC-3984: Show error page if dcmaw disabled in mitigation journeys

lwebb-iig:
### What changed

On mitigation journeys where dcmaw is a journey step, add checks to see if dcmaw is enabled and if not, show an error page.

### Why did it change

Disabling DCMAW still allowed users to reach the DCMAW CRI

### Issue tracking

- [PYIC-3984](https://govukverify.atlassian.net/browse/PYIC-3984)

### Other considerations

[PYIC-3984]: https://govukverify.atlassian.net/browse/PYIC-3984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -677,6 +677,10 @@ MITIGATION_01_IDENTITY_START_PAGE:
   events:
     next:
       targetState: MITIGATION_01_CRI_DCMAW
+      checkIfDisabled:
+        dcmaw:
+          targetJourney: TECHNICAL_ERROR
+          targetState: ERROR_NO_TICF
     end:
       targetState: MITIGATION_01_F2F_START_PAGE
       checkIfDisabled:
@@ -756,6 +760,10 @@ MITIGATION_02_OPTIONS:
     mitigationStart: enhanced-verification
   events:
     next:
+      checkIfDisabled:
+        dcmaw:
+          targetJourney: TECHNICAL_ERROR
+          targetState: ERROR_NO_TICF
       targetState: CRI_DCMAW_PYI_ESCAPE
     end:
       targetJourney: INELIGIBLE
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -677,6 +677,10 @@ MITIGATION_01_IDENTITY_START_PAGE:
   events:
     next:
       targetState: MITIGATION_01_CRI_DCMAW
+      checkIfDisabled:
+        dcmaw:
+          targetJourney: TECHNICAL_ERROR
+          targetState: ERROR_NO_TICF
```
Joe-Edwards-GDS: I don't think we want to skip TICF in these scenarios(?)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Joe-Edwards-GDS, line 1: I think you're missing at least a few more cases - have a look for other states that route to `CRI_DCMAW_PYI_ESCAPE`

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
lwebb-iig: @Joe-Edwards-GDS Wasn't sure on this either - F2F uses this error sate, but maybe we want to use this error state/journey instead?
```
      targetState: ERROR
```

Joe-Edwards-GDS: The idea is that we should call TICF exactly once in each journey, as close to the end as possible.

In a F2F journey, we have to call it _before_ routing to the F2F CRI, so any errors that happen afterwards shouldn't call TICF again. (Similarly, with any errors that come from the TICF CRI itself)

In this case, I think we won't have called TICF, so we can just use the 'normal' error state :)

Joe-Edwards-GDS (approved):

PR #1761, opened 2024-03-19
PYIC-5256: Added new sub-journey for 6MFC

RogVanGDS:
### What changed

Added new sub-journey for 6MFC

### Why did it change

Added journey map config to route the user through the new 6MFC journey. This can be accessed from the reuse-page with the sixMonthFraudCheckEnabled feature set turned on for testing while the activation logic is being worked on.

### Issue tracking

- [PYIC-5256](https://govukverify.atlassian.net/browse/PYIC-5256)

### Environment variables or secrets

- [ ] No environment variables or secrets were added or changed

[PYIC-5256]: https://govukverify.atlassian.net/browse/PYIC-5256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1768, opened 2024-03-25
PYIC-5358: Logic to route users to 6MFC journey

sammyiyke:
### What changed

[PYIC-5358](https://govukverify.atlassian.net/browse/PYIC-5358): Logic to route users to 6MFC journey

Introduces logic within IPV Core to manage and route users to the six-month fraud check when the `REPEAT_FRAUD_CHECK` is set

Also --->>>

- Introduces method `filterVCBasedOnEvidenceType` to filter VCs based on the evidence type
- Introduces method `checkIfExpiredVC` to check if a VC is expired using the `FRAUD_CHECK_EXPIRY_PERIOD_HOURS` parameter

### Why did it change

To ensure IPV Core remains compliant with identity verification standards and fraud prevention measures, it's essential to periodically renew fraud checks. This logic automates the determination and initiation of fraud check renewals, enhancing the security and integrity of the identity proofing process.

### Issue tracking

- [PYIC-5358](https://govukverify.atlassian.net/browse/PYIC-5358)

[PYIC-5358]: https://govukverify.atlassian.net/browse/PYIC-5358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-5358]: https://govukverify.atlassian.net/browse/PYIC-5358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -437,9 +446,32 @@ private JourneyResponse buildReuseResponse(
...
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Fraud VC found and expired, resetting identity for GPG45 evaluation."));
+            return JOURNEY_REPEAT_FRAUD_CHECK;
+        }
+
         return JOURNEY_REUSE;
     }
 
+    private boolean checkIfExpiredFraudVc(List<VerifiableCredential> vcs) {
```
Joe-Edwards-GDS: (minor, but I prefer boolean methods to have a verb that makes it clearer what true/false means - e.g. 'hasExpiredFraudVc')

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -437,9 +446,32 @@ private JourneyResponse buildReuseResponse(
...
+        var fraudVCs =
+                VcHelper.filterVCBasedOnEvidenceType(
+                        vcs, EvidenceType.IDENTITY_FRAUD, EvidenceType.FRAUD_WITH_ACTIVITY);
+        // validate that it is not empty returns one and only one fraud VC
+        var fraudVC = fraudVCs.stream().findFirst();
```
Joe-Edwards-GDS: I don't think we necessarily need to validate that there is only a single result (although we expect this always to be the case). The main thing is whether they have at least one valid non-expired fraud VC - they can have as many expired VCs as they like.

Perhaps we should invert the condition and call it `hasCurrentFraudVc` to better represent the intent?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -97,6 +105,38 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
 
+    public static List<VerifiableCredential> filterVCBasedOnEvidenceType(
+            List<VerifiableCredential> vcs, EvidenceType... evidenceType) {
+
+        return vcs.stream()
+                .filter(vc -> matchesEvidenceType(vc, Arrays.asList(evidenceType)))
+                .toList();
+    }
+
+    private static boolean matchesEvidenceType(
```
Joe-Edwards-GDS: It might be easier to just check `vc.getCriId().equals(EXPERIAN_FRAUD_CRI)` here?

(I'm somewhat keen to get rid of the 'EvidenceType' thing in future, since it's not very well-defined)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -228,4 +268,24 @@ public static Vot getVcVot(VerifiableCredential vc) throws UnrecognisedVotExcept
             throw new UnrecognisedVotException("Invalid VOT found for this VC");
         }
     }
+
+    public static boolean checkIfExpiredVC(VerifiableCredential vc) {
```
Joe-Edwards-GDS: I think this is specifically checking if the fraud VC is expired? Other VCs don't expire over the same duration.

Perhaps `isExpiredFraudVc`? (or `isCurrentFraudVc` to make it positive)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -228,4 +268,24 @@ public static Vot getVcVot(VerifiableCredential vc) throws UnrecognisedVotExcept
...
+        var jwtClaimsSet = vc.getClaimsSet();
+        var nbf = Optional.ofNullable(jwtClaimsSet.getClaim(NOT_BEFORE));
+        if (nbf.isEmpty()) {
+            LOGGER.error("VC does not have a nbf claim");
+            return false;
```
Joe-Edwards-GDS: I think we should return true in this case, or throw an error - we can't assume it's valid if we don't know how old it is!

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -228,4 +268,24 @@ public static Vot getVcVot(VerifiableCredential vc) throws UnrecognisedVotExcept
...
+        }
+
+        var expiryPeriod =
+                Integer.parseInt(configService.getSsmParameter(FRAUD_CHECK_EXPIRY_PERIOD_HOURS));
+        var nbfValue = nbf.get().toString();
+        DateTimeFormatter formatter =
+                DateTimeFormatter.ofPattern("EEE MMM dd HH:mm:ss zzz yyyy", Locale.ENGLISH);
+        ZonedDateTime zonedDateTime = ZonedDateTime.parse(nbfValue, formatter);
```
Joe-Edwards-GDS: Are you sure this is right - I think `nbf` is a unix timestamp, so we shouldn't need to mess around with time zones at all, we can just do arithmetic on `Instant`s.

```
var nbf = Instant.ofEpochSecond(nbfClaim);
var now = Instant.now();
return nbf.plus(expiryPeriod, ChronoUnit.HOURS).isBefore(now);
```
(You could also do arithmetic on the timestamp values themselves, but this is a bit neater)

Joe-Edwards-GDS (commented):
We'll also need to address https://github.com/govuk-one-login/ipv-core-back/pull/1761/files#r1531781016

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
sammyiyke: Used this instead, figured it should be a neater solution:

```
var nbfClaim = jwtClaimsSet.getNotBeforeTime(); //gets the nbf as a Date
var nbf = nbfClaim.toInstant(); // coverts Date to Instant
```

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
sammyiyke: Yea, true, the `EvidenceType` is not well defined, part of the reason for this method, as there are two different types currently that relates to fraud - hence the support for list of evidence types.  Was hoping this would be useful in filtering VCs based on evidence type.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sammyiyke: I think that in an unlikely case where they have multiple fraud VCs then we would be checking if they have an expired fraud VC hence `hasCurrentFraudVc` not a current VC

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -1069,7 +1075,10 @@ void shouldRemoveOperationalProfileIfMatchingGpg45ProfileIsPresent() throws Exce
                 .thenReturn(clientOAuthSessionItem);
         when(userIdentityService.checkRequiresAdditionalEvidence(any())).thenReturn(false);
         when(userIdentityService.areVcsCorrelated(any())).thenReturn(true);
-        when(configService.enabled(anyString())).thenReturn(false).thenReturn(true);
+        when(configService.enabled(anyString()))
+                .thenReturn(false)
+                .thenReturn(true)
+                .thenReturn(false);
```
RogVanGDS: Is it possible to write this a bit more explicitly? It's quite hard to see what's going on at a glance without just knowing what order the feature flags will be called in any given method. 

It may be that having multiple mock returns based on an input string isn't possible, but I think it would be good to have if we can

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -522,6 +523,43 @@ static VerifiableCredential vcExperianFraudScoreOne() {
...
+                        .build();
+        return generateVerifiableCredential(
+                TEST_SUBJECT,
+                EXPERIAN_KBV_CRI,
+                TestVc.builder()
+                        .evidence(FRAUD_EVIDENCE_NO_CHECK_DETAILS)
+                        .credentialSubject(credentialSubject)
+                        .build(),
+                "https://review-f.integration.account.gov.uk",
+                Instant.ofEpochSecond(1658829758));
```
RogVanGDS, line 540: ~Is it possible to pass through a clock into this builder, so we can guarantee that we always have a valid/expired VC for these tests?~

~Then we may be able to get around having to set the expiry duration to something massive as well~ 

Ignore me, I can see you've already done this below

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -54,6 +54,9 @@ CHECK_EXISTING_IDENTITY:
     enhanced-verification-f2f-fail:
       targetJourney: NEW_P2_IDENTITY
       targetState: ENHANCED_VERIFICATION_F2F_FAIL
+    repeat-fraud-check:
+      targetJourney: REPEAT_FRAUD_CHECK
+      targetState: CONFIRM_NAME_DOB
```
RogVanGDS: I think this can just be `START`, and the entry routing in the journey definition will take care of the rest

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: Ok sure, we can revisit later if need be.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: What I mean is: if you have two fraud VCs, one expired and one current, then you _don't_ need to do an additional fraud check journey (you have a valid fraud VC, the expired one is irrelevant!)

(But this should never occur in practice)

sammyiyke: Got it.. 👍

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
sammyiyke: Yea, thanks Roger. Noticed this earlier and fixed it already.

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
sammyiyke: I believed I tried that earlier and it moaned about it. I think the error it gave was related to stubbing. But I would try doing that again and see if it would work this time

RogVanGDS (approved):

PR #1776, opened 2024-03-26
PYIC-5403: Convert BuildProvenUserIdentityDetails to an API GW proxy

Joe-Edwards-GDS:
Previously this endpoint used VTL to transform the request/response, but a proxy passthrough removes the need to use VTL, and keeps all processing logic in the lambda code.

### DIFF
--- a/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
+++ b/lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
@@ -2,11 +2,14 @@ package uk.gov.di.ipv.core.buildprovenuseridentitydetails;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
 import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -20,14 +23,13 @@ import uk.gov.di.ipv.core.library.domain.Address;
 import uk.gov.di.ipv.core.library.domain.BirthDate;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.IdentityClaim;
-import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
-import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.ProfileType;
 import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
 import uk.gov.di.ipv.core.library.dto.VcStatusDto;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.NoVcStatusForIssuerException;
+import uk.gov.di.ipv.core.library.helpers.ApiGatewayResponseGenerator;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.helpers.RequestHelper;
 import uk.gov.di.ipv.core.library.persistence.item.ClientOAuthSessionItem;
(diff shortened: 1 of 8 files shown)


lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandler.java
```diff
@@ -140,11 +140,14 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
+        return buildJourneyErrorResponse(errorResponse, HttpStatus.SC_INTERNAL_SERVER_ERROR);
+    }
+
+    private APIGatewayProxyResponseEvent buildJourneyErrorResponse(
+            ErrorResponse errorResponse, int statusCode) {
         LOGGER.error(LogHelper.buildLogMessage(errorResponse.getMessage()));
-        return new JourneyErrorResponse(
-                        JOURNEY_ERROR_PATH, HttpStatus.SC_INTERNAL_SERVER_ERROR, errorResponse)
-                .toObjectMap();
+        return ApiGatewayResponseGenerator.proxyJsonResponse(statusCode, errorResponse);
```
Joe-Edwards-GDS, line 150: This is a small change, so error responses from this endpoint no longer look like journey events.

old:
```
{
  "journey": "/journey/error",
  "statusCode": 500,
  "code": 1234,
  "message": "an error happened"
}
```

new:
```
{
  "code": 1234,
  "message": "an error happened"
}
```
core-front doesn't do anything with this anyway, except include it in the logs

lambdas/build-proven-user-identity-details/src/main/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/domain/ProvenUserIdentityDetails.java
```diff
@@ -7,15 +7,14 @@
...
 import uk.gov.di.ipv.core.library.domain.Address;
-import uk.gov.di.ipv.core.library.domain.BaseResponse;
 
 import java.util.List;
 
 @EqualsAndHashCode(callSuper = false)
 @ExcludeFromGeneratedCoverageReport
 @Data
 @Builder
-public class ProvenUserIdentityDetails extends BaseResponse {
```
Joe-Edwards-GDS, line 18: (`BaseResponse` looks a bit suspect in general, but we don't need it here - we can serialize this class normally)

lambdas/build-proven-user-identity-details/src/test/java/uk/gov/di/ipv/core/buildprovenuseridentitydetails/BuildProvenUserIdentityDetailsHandlerTest.java
```diff
@@ -121,7 +119,7 @@ void shouldReceive200ResponseCodeProvenUserIdentityDetails() throws Exception {
         when(mockClientOAuthSessionDetailsService.getClientOAuthSession(any()))
                 .thenReturn(clientOAuthSessionItem);
 
-        JourneyRequest input = createRequestEvent();
+        var input = createRequestEvent();
```
Joe-Edwards-GDS, line 122: (a good advertisement for using `var`, I think)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -82,33 +86,11 @@ public enum ErrorResponse {
...
-        return String.format("{\"code\":%d,\"message\":\"%s\"}", code, message);
-    }
-
-    public String getMessage() {
-        return message;
-    }
-
     @JsonCreator
-    public static ErrorResponse forCode(int code) {
+    public static ErrorResponse forCode(@JsonProperty("code") int code) {
```
Joe-Edwards-GDS, line 93: This makes the serialisation and deserialisation for this class symmetric - previously it serialised to an object, but deserialised from an int (the annotation here turns this into a 'properties' JsonCreator, which reads the property from a JSON object).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyErrorResponse.java
Joe-Edwards-GDS, line 1: The JSON properties on this class are a bit of a mess, but it needed a small change as previously it was relying on the asymmetric serialisation of `ErrorResponse` (now it only ever serialises/deserialises using ints).

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/LambdaHandler.java
```diff
@@ -144,39 +132,28 @@ public LambdaHandler() throws IOException {
...
-        APIGatewayProxyRequestEvent apiGatewayProxyRequestEvent = new APIGatewayProxyRequestEvent();
-        apiGatewayProxyRequestEvent.setBody(request.body());
-        apiGatewayProxyRequestEvent.setHeaders(getHeadersMap(request));
-
-        APIGatewayProxyResponseEvent responseEvent =
-                handler.handleRequest(apiGatewayProxyRequestEvent, EMPTY_CONTEXT);
-
-        response.type(APPLICATION_JSON);
-        return responseEvent.getBody();
+        return (Request request, Response response) -> {
```
Wynndow, line 145: nice

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
Wynndow: I've actually had a long standing desire to remove this ErrorResponse class completely. I'm not sure it actually does anything for us.

Wynndow (approved):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
Joe-Edwards-GDS: Having a standard list of error codes is potentially useful, but I agree we're not really making any use of it!

In principle, these errors should be monitored and visible in logs etc., with a corresponding run-book entry for each (i.e. what to do if each of the errors occurs).

(If you're not doing that, you may as well just inline the error messages...)

PR #1795, opened 2024-03-28
PYIC-5675: Add SessionUserIssuedCredentialsTable

Wynndow:
### What changed

Add SessionUserIssuedCredentialsTable

### Why did it change

This adds a new DynamoDB table for storing users credentials gathered during their session, before writing them to a long term store. This change is just to create the table, and not start writing to it.

It also adds a class to represent the data that will be stored in it - I'm adding this with the table to ensure the correct format of the data, specifically the sort key. The class may change slightly when we acutally implement the writes.

The Sonar quality check is failing due to code smells. These are from the new `SessionCredentialItem` not being used yet, so should be safe to ignore.

### Issue tracking

- [PYIC-5675](https://govukverify.atlassian.net/browse/PYIC-5675)

[PYIC-5675]: https://govukverify.atlassian.net/browse/PYIC-5675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -2543,6 +2543,34 @@ Resources:
         SSEType: KMS
         KMSMasterKeyId: !GetAtt DynamoDBKmsKey.Arn
 
+  SessionCredentialsTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      # checkov:skip=CKV_AWS_28: Point in time recovery is not necessary for this table
+      TableName: !Sub "session-credentials-${Environment}"
+      BillingMode: "PAY_PER_REQUEST"
+      DeletionProtectionEnabled: !If
+        - IsProduction
+        - true
+        - false
+      AttributeDefinitions:
+        - AttributeName: "ipvSessionId"
+          AttributeType: "S"
+        - AttributeName: "sortKey"
+          AttributeType: "S"
+      KeySchema:
+        - AttributeName: "ipvSessionId"
+          KeyType: "HASH"
+        - AttributeName: "sortKey"
+          KeyType: "RANGE"
+      TimeToLiveSpecification:
+        AttributeName: "ttl"
+        Enabled: true
+      SSESpecification:
+        SSEEnabled: true
+        SSEType: KMS
+        KMSMasterKeyId: !GetAtt DynamoDBKmsKey.Arn
+
   RevokedUserCredentialsTable:
     Type: AWS::DynamoDB::Table
     Properties:
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
@@ -1,0 +1,45 @@
+package uk.gov.di.ipv.core.library.persistence.item;
+
+import com.nimbusds.jwt.SignedJWT;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSortKey;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@DynamoDbBean
+public class SessionCredentialItem implements DynamodbItem {
+
+    private static final String SORT_KEY_TEMPLATE = "%s#%s";
+    private final String ipvSessionId;
+    private final String sortKey;
+    private final String credential;
+    private final boolean receivedThisSession;
+    private long ttl;
+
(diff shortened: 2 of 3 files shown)


deploy/template.yaml
```diff
@@ -2543,6 +2543,36 @@ Resources:
         SSEType: KMS
         KMSMasterKeyId: !GetAtt DynamoDBKmsKey.Arn
 
+  SessionCredentialsTable:
```
lwebb-iig, line 2546: Should we be adding a [TTL specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-timetolivespecification.html) to the table as well?

Wynndow: Oh snap. That's what you get for copying a table without one 😓

deploy/template.yaml
```diff
@@ -2543,6 +2543,39 @@ Resources:
...
+    Type: AWS::DynamoDB::Table
+    Properties:
+      # checkov:skip=CKV_AWS_28: Point in time recovery is not necessary for this table.
```
Joe-Edwards-GDS: This seems to contradict the PITR config below? I think I agree with the comment and we should remove PITR...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
```diff
@@ -0,0 +1,45 @@
...
+    private final boolean receivedThisSession;
+    private long ttl;
+
+    public SessionCredentialItem(
+            String ipvSessionId,
+            String criId,
+            SignedJWT signedCredJwt,
+            boolean receivedThisSession) {
+        this.ipvSessionId = ipvSessionId;
+        this.sortKey = String.format(SORT_KEY_TEMPLATE, criId, signedCredJwt.getSignature());
```
Joe-Edwards-GDS, line 26: Do you think it'd be useful to include the reverse, i.e. so there is a `getCriId` method? You could also consider making the private field `criId`, and only generate the sort key on-demand in `getSortKey` - that might simplify the constructor.

(The only thing to watch out for is I'm not quite sure how `@DynamoDbBean` selects the fields to write into the record...)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
```diff
@@ -0,0 +1,45 @@
+package uk.gov.di.ipv.core.library.persistence.item;
+
+import com.nimbusds.jwt.SignedJWT;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSortKey;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@DynamoDbBean
```
Joe-Edwards-GDS, line 10: Do we need this to include `@Getter` (or `@Data`) so we can actually read the credential...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
Joe-Edwards-GDS: If this is tricky to work out now, you could also leave it to the next ticket.

deploy/template.yaml
Wynndow: We have this on the other tables - and I reasoned it in my head that in envs other than prod we don't enable it. Which is why the `checkov:skip` is neccesary, to allow the pipelines to pipeline in the lower envs. 
I guess that as this is such a short lived table if we ever actually needed it, it'd probably be pointless by the time we'd actually managed to use the functionality. Will remove.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
Wynndow: I was mostly aiming for the minimum here to express what the sort key should be - there's a good chance stuff will move around in the next ticket. Will bear this in mind.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/SessionCredentialItem.java
Wynndow: As above - intentially left off as not needed right now. Will add them as required.

deploy/template.yaml
Joe-Edwards-GDS: Yeah - e.g. it looks like the main session table doesn't have it

Joe-Edwards-GDS (approved):

PR #1798, opened 2024-04-02
PYIC-5685: Remove local running setup from datastore

Wynndow:
### What changed

Remove local running setup from datastore

### Why did it change

We had some logic to determine if core was running locally, and if so use a locally hosted version of dynamodb. This is a legacy thing that we don't use anymore, and can remove. Which will simplify a few things, including the config service.


Joe-Edwards-GDS (commented):
Looks like this was used for local DynamoDB and SSM?

Interestingly, I was thinking about whether we want to move towards mock/local AWS dependencies - i.e. so we can run core-back entirely locally without an AWS account.
- DynamoDB using https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html
- SSM using a stub that just reads from a JSON config file (this might also alleviate the need for complicated dev env overrides!)
- SQS using a stub that records audit events in memory and can be inspected by tests
- KMS using a stub that does encryption using the test keys

However, I'm not sure how much of the code you're removing will be useful for this, and having it affect all the unit tests feels like a bit of a red flag.

Once you sort the tests then this is probably fine...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -77,7 +79,15 @@ public ConfigService(SSMProvider ssmProvider, SecretsProvider secretsProvider) {
         this(ssmProvider, secretsProvider, null);
     }
 
-    public ConfigService(List<String> featureSet) {
+    @ExcludeFromGeneratedCoverageReport
+    public ConfigService() {
+        if (Boolean.parseBoolean(getEnvironmentVariable(UNMOCKABLE_CONFIG_SERVICE_TEST))) {
+            // For journey map tests that deserialize events that create a config service
```
Joe-Edwards-GDS: Is this because the `BasicState` initialiser constructs a new config service each time?

This feels a bit suspect - I wonder if we can put the `configService` into `JourneyContext` instead, to avoid the states themselves being stateful (ha). If that makes this PR too complicated then we can leave it like this...

Wynndow: ...interesting. That'll be doable, and remove the need for the tests flag...

Joe-Edwards-GDS (dismissed):
Love changes like this
image

Joe-Edwards-GDS (approved):

PR #1800, opened 2024-04-02
PYIC-5685: Write to session credentials table

Wynndow:
### What changed

Write to session credentials table

### Why did it change

[PYIC-5685: Store credentials in SessionCredentialsTable](https://github.com/govuk-one-login/ipv-core-back/commit/7e49c33d4b083e9106c01dc4a3d6ba558fd63313)

This adds logic to start storing VCs received in the new
SessionCredentialsTable. It creates a new service to do this, and wires
it into the CriStoringService.

It also updates our VerifiableCredential class to store the parsed
credential alongside the raw format. This will allow us to easily access
the signature when needed. We've already parsed it so might as well keep
it.

[PYIC-5685: Store VC in session credential table on reuse](https://github.com/govuk-one-login/ipv-core-back/commit/39471a56543b19ec468943fa9e9edfb812697030) 

If a user has already proved their identity, we want to move their VC to
the session credentials table. This is so that when we come to build
their identity, the VC required to do so are in the same place as if
they'd just completed a full identity proving journey.

[PYIC-5685: Remove creds from session credential store in reset-identity](https://github.com/govuk-one-login/ipv-core-back/commit/1b26cd3dd57197435e2894f6990454305ebfa8e5)

This updates the reset-identity handler to remove any VC from the
session credential table for the users current session, as well as from
the tactical store.

I'm a bit dubious that there will ever actually be anything to remove
here. If we're resetting the users identity they'll have failed to match
a profile and so won't have anything in the session credentials store
yet. And this happens at the start of their journey, so they shouldn't
have visited any CRI yet.

[PYIC-5685: Feature flag writes to session-credentials table](https://github.com/govuk-one-login/ipv-core-back/pull/1800/commits/aeda0ccba5e03346a67f0f0c5a5f7c3589b21104)

We need sign off from security before we can start writing to this new
table in prod. This puts the writes behind a flag which we'll be able to
control in lower envs in the usual way.

[PYIC-5685: Write VC to session credentials table for 6MFC](https://github.com/govuk-one-login/ipv-core-back/pull/1800/commits/b17c9c442ad5438e8413bd9278beb6a676551478)

When a user is sent on the 6 month fraud check journey we need to ensure
that we have all their existing VC, except for the fraud one, in the
session credentials table. This is so we can do the usual GPG45
calculations once they've collected their new fraud VC.

### Issue tracking

- [PYIC-5685](https://govukverify.atlassian.net/browse/PYIC-5685)

[PYIC-5685]: https://govukverify.atlassian.net/browse/PYIC-5685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -75,6 +76,7 @@ public CallTicfCriHandler() {
                         new AuditService(AuditService.getDefaultSqsClient(), configService),
                         null,
                         new VerifiableCredentialService(configService),
+                        new SessionCredentialsService(configService),
```
Joe-Edwards-GDS, line 79: Is it irrational to be annoyed that one is plural and one is singular 👀

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
```diff
@@ -0,0 +1,83 @@
...
+            LOGGER.error("Error persisting session credential: {}", e.getMessage(), e);
+            throw new VerifiableCredentialException(
+                    HTTPResponse.SC_SERVER_ERROR, ErrorResponse.FAILED_TO_SAVE_CREDENTIAL);
+        }
+    }
+
+    public void persistCredentials(List<VerifiableCredential> vcs, String ipvSessionId)
+            throws VerifiableCredentialException {
+        for (var vc : vcs) {
+            persistCredential(vc, ipvSessionId, false);
```
Joe-Edwards-GDS: It feels a bit surprising that this sets `receivedThisSession = false` but the singular method requires you to specify.

Can we add the flag here as well, or create distinct methods for each type (e.g. `persistNewCredential` vs. `persistExistingCredential`)?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
```diff
@@ -0,0 +1,83 @@
...
+    public void persistCredential(
+            VerifiableCredential credential, String ipvSessionId, boolean receivedThisSession)
+            throws VerifiableCredentialException {
+        try {
+            if (configService.enabled(SESSION_CREDENTIALS_TABLE_WRITES)) {
+                dataStore.create(
+                        credential.toSessionCredentialItem(ipvSessionId, receivedThisSession),
+                        ConfigurationVariable.SESSION_CREDENTIALS_TTL);
+            } else {
+                LOGGER.info(
```
Joe-Edwards-GDS: (might this be better as `debug`?)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -444,16 +456,22 @@ private JourneyResponse buildReuseResponse(
         }
 
         // check the result of 6MFC and return the appropriate journey
-        if (configService.enabled(REPEAT_FRAUD_CHECK.getName()) && !hasCurrentFraudVc(vcs)) {
+        if (configService.enabled(REPEAT_FRAUD_CHECK) && !hasCurrentFraudVc(vcs)) {
             LOGGER.info(
                     LogHelper.buildLogMessage(
                             "Fraud VC found and expired, resetting identity for GPG45 evaluation."));
+            sessionCredentialsService.persistCredentials(
+                    allVcsExceptFraud(vcs), auditEventUser.getSessionId());
```
Joe-Edwards-GDS: In this scenario, I think we'll have already persisted _all_ the credentials in `checkForProfileMatch`?

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
Wynndow: Not at all. But I'd be annoyed with the original one...

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
Wynndow: Yep, this was mostly for convenience (when would we be storing multiple VC received this session?), but you're right it's probably better to be explicit.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
Wynndow: To be honest, if it's getting moved to `debug`, i'll probably just remove it.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Ooo yeah. So we need to remove the fraud VC, rather than add everything except. Althogh that raises the question of if we should be allowing expired VC to contribute when calculating gpg45 scores?

Joe-Edwards-GDS: It's a good question... Strictly speaking we probably should exclude it (since it's no longer a valid fraud score), but it's quite convenient to allow it...
- We want to check if you meet a GPG45 profile with the expired fraud VC before allowing you to take a repeat fraud check journey
- We are testing this by getting the fraud CRI to generate an expired fraud VC as part of a journey

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -431,29 +438,43 @@ private JourneyResponse buildReuseResponse(
...
+            throws SqsException, VerifiableCredentialException {
         LOGGER.info(LogHelper.buildLogMessage("Returning reuse journey"));
         sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_COMPLETE, auditEventUser);
 
         if (attainedVot.getProfileType() == OPERATIONAL_HMRC) {
             // the only VC we should possibly have collected this session at this point is a
             // migration VC
-            return vcReceivedThisSession == null || vcReceivedThisSession.isEmpty()
+            boolean isOpProfileReuse =
+                    vcReceivedThisSession == null || vcReceivedThisSession.isEmpty();
```
Joe-Edwards-GDS, line 449: We might need to rethink this a bit once we do the second phase - I was hoping we could remove `vcReceivedThisSession` entirely...

(No need to worry about backwards compat since operational profiles aren't live)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -431,29 +438,43 @@ private JourneyResponse buildReuseResponse(
...
+            LOGGER.info(LogHelper.buildLogMessage("Expired fraud VC found"));
+            sessionCredentialsService.persistCredentials(
+                    allVcsExceptFraud(vcs), auditEventUser.getSessionId(), false);
             return JOURNEY_REPEAT_FRAUD_CHECK;
         }
 
+        sessionCredentialsService.persistCredentials(
+                VcHelper.filterVCBasedOnProfileType(vcs, attainedVot.getProfileType()),
+                auditEventUser.getSessionId(),
+                false);
```
Joe-Edwards-GDS, line 470: I wonder slightly if we could simplify by persisting at the top of the method and always filtering by a) profile type and b) removing any expired fraud VCs?

This is also fine, it just feels awkward to have the repetition.

Joe-Edwards-GDS (approved):
We might still be able to tidy up the storing a little bit, but happy to go as-is if it doesn't work.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Yeah that's fair - i'll do it in a follow up PR though as you're away today and the PR is approved.

PR #1801, opened 2024-04-02
PYIC-5403: Remove dedicated build-client-oauth-response endpoint

Joe-Edwards-GDS:
### What changed

Removed the dedicated `build-client-oauth-response` endpoint so it is handled as a special case in the journey engine.

### Why did it change

This is already effectively just bypassing the journey engine to hit one of the internal lambdas. This keeps the internals private to the journey engine and removes the confusing parallel `/journey/` endpoint.

### Issue tracking

- [PYIC-5403](https://govukverify.atlassian.net/browse/PYIC-5403)

[PYIC-5403]: https://govukverify.atlassian.net/browse/PYIC-5403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -39,6 +39,7 @@ import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.BasicState;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.JourneyChangeState;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.State;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.JourneyContext;
+import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.ProcessStepResponse;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.StepResponse;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -59,6 +60,9 @@ public class ProcessJourneyEventHandler
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String CORE_SESSION_TIMEOUT_STATE = "CORE_SESSION_TIMEOUT";
     private static final String NEXT_EVENT = "next";
+    private static final String END_SESSION_EVENT = "build-client-oauth-response";
+    private static final StepResponse END_SESSION_RESPONSE =
+            new ProcessStepResponse("build-client-oauth-response", null, null);
```
Joe-Edwards-GDS, line 65: These are intentionally different, and ideally we'd change the first one to reflect that.
- `END_SESSION_EVENT` is the journey _event_ that the frontend triggers to indicate that the session should be immediately ended and returned to the RP (so perhaps could be `end-session` or similar)
- `END_SESSION_RESPONSE` is the journey _step_ that is executed in response to that event (and refers to the lambda name)

Wynndow: Am I correct in thinking that the only place that can generate the `END_SESSION_EVENT` is from the `pyi-timeout-recoverable` page in the frontend?

Joe-Edwards-GDS: That's my understanding - but if you think it can happen elsewhere then let me know!

Wynndow: No, all good, just getting my head around it.

Wynndow (dismissed):

Wynndow (approved):

PR #1803, opened 2024-04-03
BAU Remove default content type for CRI credential request

blakeyp:
### What changed

remove default content type header for CRI credential request

### Why did it change

nimbus (possibly since the upgrade to 11) sets a default Content-Type header for a request but the call to CRI issue credential doesn't have a request body so we shouldn't need to send a Content-Type header at all

NINO CRI is returning a 415 'unsupported media type' response because the Content-Type is not expected

### DIFF
--- a/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
+++ b/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
@@ -250,6 +250,8 @@ public class CriApiService {
         var apiKey = getApiKey(criConfig, criOAuthSessionItem);
 
         var request = new HTTPRequest(HTTPRequest.Method.POST, criConfig.getCredentialUrl());
+        request.setHeader(
+                "Content-Type", ""); // remove the default, no request body so we don't need
 
         if (apiKey != null) {
             LOGGER.info(


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
```diff
@@ -250,6 +250,8 @@ public HTTPRequest buildFetchVerifiableCredentialRequest(
         var apiKey = getApiKey(criConfig, criOAuthSessionItem);
 
         var request = new HTTPRequest(HTTPRequest.Method.POST, criConfig.getCredentialUrl());
+        request.setHeader(
+                "Content-Type", null); // remove the default, no request body so we don't need
```
surajkumar: `setHeader` uses varargs for the value, this will throw a compiler warning. 

```java
warning: non-varargs call of varargs method with inexact argument type for last parameter;
        request.setHeader("Content-Type", null);
                                          ^
  cast to String for a varargs call
  cast to String[] for a non-varargs call and to suppress this warning
1 warning
```

You will need to pass an empty string.

surajkumar (changes requested):

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriApiService.java
blakeyp: Thanks, I've updated - it looks like empty string has the same effect of removing the header

surajkumar (approved):

DanCorderIPV:
This looks fine given the situation, but it seems odd that we're making a POST request but not sending any data.

DanCorderIPV (approved):

PR #1804, opened 2024-04-03
PYIC-5403: Add JSON version of journey API

Joe-Edwards-GDS:
The existing journey API uses `application/x-www-form-urlencoded`. This is largely irrelevant, since there is no request body, but it would be nice to stick with JSON for consistency.

Also refactors the VTL to (try) and make it a bit more readable...

### DIFF
--- a/openAPI/core-back-internal.yaml
+++ b/openAPI/core-back-internal.yaml
@@ -60,6 +60,21 @@ paths:
           Fn::Sub: arn:aws:apigateway:${AWS::Region}:states:action/StartSyncExecution
         passthroughBehavior: "when_no_match"
         requestTemplates:
+          application/json:
+            # This VTL template maps the request into the format required by the step function API
+            # Note that the lack of indentation is deliberate:
+            # the 'input' field must be condensed into a single line to be a valid JSON string
+            Fn::Sub: >
+              {
+              "input": "{
+              \"journey\":\"/journey/$util.escapeJavaScript($input.params('journeyStep'))\",
+              \"ipAddress\":\"$util.escapeJavaScript($input.params('ip-address'))\",
+              \"featureSet\":\"$util.escapeJavaScript($input.params('feature-set'))\",
+              \"ipvSessionId\":\"$util.escapeJavaScript($input.params('ipv-session-id'))\",
+              \"clientOAuthSessionId\":\"$util.escapeJavaScript($input.params('client-session-id'))\"
+              }",
+              "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}"
+              }
           application/x-www-form-urlencoded:
             Fn::Sub: |
               {


openAPI/core-back-internal.yaml
```diff
@@ -60,6 +60,21 @@ paths:
           Fn::Sub: arn:aws:apigateway:${AWS::Region}:states:action/StartSyncExecution
         passthroughBehavior: "when_no_match"
         requestTemplates:
+          application/json:
+            # This VTL template maps the request into the format required by the step function API
+            # Note that the lack of indentation is deliberate:
+            # the 'input' field must be condensed into a single line to be a valid JSON string
```
Joe-Edwards-GDS, line 66: I would welcome alternate approaches - the problematic constraint is that the `input` value is a JSON string containing a JSON object (not a raw JSON object), so must be a single line.

This is the best I could come up with that still allows us to write the different parts on different lines (so you can easily see all the properties being passed through) - but you can't do indentation or YAML won't strip the newlines.

openAPI/core-back-internal.yaml
```diff
@@ -60,6 +60,21 @@ paths:
...
+              {
+              "input": "{
+              \"journey\":\"/journey/$util.escapeJavaScript($input.params('journeyStep'))\",
+              \"ipAddress\":\"$util.escapeJavaScript($input.params('ip-address'))\",
+              \"featureSet\":\"$util.escapeJavaScript($input.params('feature-set'))\",
+              \"ipvSessionId\":\"$util.escapeJavaScript($input.params('ipv-session-id'))\",
+              \"clientOAuthSessionId\":\"$util.escapeJavaScript($input.params('client-session-id'))\"
+              }",
+              "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}"
+              }
```
Joe-Edwards-GDS, line 77: Two changes from the old one (below):
- The properties are also run through `$util.escapeJavaScript`, to avoid risk of injection if a user gains control of any of these parameters
- Re-ordered and removed a duplicate `featureSet` entry

openAPI/core-back-internal.yaml
Wynndow: We could use a flow scalar, which means that new lines are removed, even if there's an indent. It's a little heavy on the escapes though. I've tested this and it seems to work.
```
          application/json:
            Fn::Sub: "{
              \"input\": \"{
                \\\"journey\\\":\\\"/journey/$util.escapeJavaScript($input.params('journeyStep'))\\\",
                \\\"ipAddress\\\":\\\"$util.escapeJavaScript($input.params('ip-address'))\\\",
                \\\"featureSet\\\":\\\"$util.escapeJavaScript($input.params('feature-set'))\\\",
                \\\"ipvSessionId\\\":\\\"$util.escapeJavaScript($input.params('ipv-session-id'))\\\",
                \\\"clientOAuthSessionId\\\":\\\"$util.escapeJavaScript($input.params('client-session-id'))\\\"
              }\",
              \"stateMachineArn\": \"arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}\"
            }"
```

It's annoying that the argument to input params needs to be quoted, otherwise we could have done something like below. You can't use escapes in a single quoted scalar, and I can't figure out a way to make it work...
```
Fn::Sub: '{
  "input": "{
    \"journey\":\"/journey/$util.escapeJavaScript($input.params(journeyStep))\",
    \"ipAddress\":\"$util.escapeJavaScript($input.params(ip-address))\",
    \"featureSet\":\"$util.escapeJavaScript($input.params(feature-set))\",
    \"ipvSessionId\":\"$util.escapeJavaScript($input.params(ipv-session-id))\",
    \"clientOAuthSessionId\":\"$util.escapeJavaScript($input.params(client-session-id))\"
  }",
  "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}"
}'
```

Joe-Edwards-GDS: Hmm, I can't quite decide which I think is better!

Wynndow: I'll appove and then you can decide if you want to change it!

Wynndow (approved):

PR #1805, opened 2024-04-03
BAU: Update DCMAW VC tests to have unambiguous names and fix some tes…

DanCorderIPV:
…t case numbers

### What changed

DCMAW contract tests

### Why did it change

Some of the test names were wrong and some of the test case reference numbers were wrong


lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -1219,92 +1219,6 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidNldPassp
                         });
     }
 
-    @Test
```
DanCorderIPV, line 1222: This has just moved down to below the pact it uses. The test hasn't actually changed

Joe-Edwards-GDS (approved):

PR #1809, opened 2024-04-04
PYIC-5537: Added backend event validation

RogVanGDS:
### What changed

Added validation to the journey events submitted from front-end to back-end

### Why did it change

If a user disabled Javascript in the browser there is nothing stopping them from submitting multiple journey events to the journey engine. 

Given that most of the journey events that a user can submit are “next” or “end”,  a user can quite easily skip or jump pages, because we transition the user’s state whenever we receive an event (provided that event exists on the current user state retrieved by the backend), regardless of the page the user sees on the screen.

### Issue tracking

- [PYIC-5537](https://govukverify.atlassian.net/browse/PYIC-5537)

[PYIC-5537]: https://govukverify.atlassian.net/browse/PYIC-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
```diff
@@ -27,6 +38,28 @@ public State transition(String startState, String event, JourneyContext journeyC
                     String.format("Unknown state provided to state machine: %s", startState));
         }
 
+        if (currentPage.isPresent() && !currentPage.get().equals("oauth")) {
```
Joe-Edwards-GDS: what's the `oauth` case for here?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
```diff
@@ -27,6 +38,28 @@ public State transition(String startState, String event, JourneyContext journeyC
...
+            if (state instanceof BasicState) {
+
+                if (((BasicState) state).getResponse() instanceof PageStepResponse) {
+
+                    String pageId =
+                            ((PageStepResponse) ((BasicState) state).getResponse()).getPageId();
```
Joe-Edwards-GDS: You can use pattern matching if conditions to avoid needing any of these casts:
```
if (state instanceof BasicState basicState) {
    basicState.getResponse();
}
```

If you're feeling particularly adventurous you can even combine in the same expression, e.g.:
```
// Must be a page step with a matching id, otherwise do nothing
if (!(basicState.getResponse() instanceof PageStepResponse pageResponse
        && pageResponse.getPageId().equals(currentPage.get()))) {
    return state;
}
```

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
```diff
@@ -27,6 +38,28 @@ public State transition(String startState, String event, JourneyContext journeyC
...
+
+                    if (!currentPage.get().equals(pageId)) {
+                        return state;
```
Joe-Edwards-GDS, line 45: Are you happy that this just returns the current state and not `pyi-attempt-recovery`? I'm just a little nervous that this could lead to a surprising user experience if it's not clear that anything has gone wrong.

We could try asking Simon if we need a product decision.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
```diff
@@ -53,16 +56,33 @@ public static List<String> getFeatureSet(Map<String, String> input) {
...
+            throws HttpResponseExceptionWithErrorBody, UnsupportedEncodingException {
+        String journeyEvent =
+                Optional.ofNullable(input.get(JOURNEY))
+                        .orElseThrow(
+                                () ->
+                                        new HttpResponseExceptionWithErrorBody(
+                                                HttpStatus.SC_BAD_REQUEST,
+                                                ErrorResponse.MISSING_JOURNEY_EVENT));
+
+        String[] decodedParts = URLDecoder.decode(journeyEvent, StandardCharsets.UTF_8).split("/");
```
Joe-Edwards-GDS: Do we definitely need this decoding step? I'm a bit nervous that we're ending up with a partially encoded journey event which might lead to confusion.

The VTL for the endpoint prepends the `journeyAction` with `/journey/<journeyAction>`, so if `journeyAction` is actually URL-encoded then we might end up with `/journey/pyi-something%2Fnext` which is a bit weird!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
Joe-Edwards-GDS, line 1: We've talked about it, but I don't love how fragile we're making the journey event syntax here! Ideally we'd pass it as an extra parameter, but appreciate that's hard :(

Happy to stick with this approach for now to help get it across the line.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
```diff
@@ -53,16 +56,33 @@ public static List<String> getFeatureSet(Map<String, String> input) {
...
                                         new HttpResponseExceptionWithErrorBody(
                                                 HttpStatus.SC_BAD_REQUEST,
-                                                ErrorResponse.MISSING_JOURNEY_EVENT))
-                        .split("/");
-        return parts[parts.length - 1];
+                                                ErrorResponse.MISSING_JOURNEY_EVENT));
+
+        String[] decodedParts = URLDecoder.decode(journeyEvent, StandardCharsets.UTF_8).split("/");
+
+        return Optional.ofNullable(decodedParts.length == 4 ? decodedParts[2] : null);
```
Joe-Edwards-GDS: What if there are more parts? e.g. if I maliciously submit an event for `foo/next`, can I bypass the validation?

(This is the sort of reason why supplying the page id as a separate parameter would be nicer 😬)

Joe-Edwards-GDS (commented):
I think this is roughly what I was expecting, but a few things to think about.

(It also needs rebasing :) )

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
RogVanGDS: Yeah I'm sure it will work for page requests, we are basically saying if the current backend page is not the same as the current front-end page, ignore what the front-end is asking us to do and just take the user to where the backend thinks we are.

Then the front-end will set the current page correctly and we will be back in sync.

and fairly confident on CRI pages (although I've only been able to test it with the stubs, not the actual CRI flow so it's possibly worth disabled the transition above staging to test it thoroughly before we hit production)

With lambda function calls we will need to error the journey, there's not really much more we can do here I don't think

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
RogVanGDS: Yeah I'm really not a fan but I don't see what else we can do here.

I think this is where it might be better to change the API gateway.

Once we have the front-end submitting all requests in the new format, there might be the option to change the API gateway to something more defined (without the greedy routing).

Then we can add pageId as an extra input in the velocity templates as well and this will hopefully be more robust

RogVanGDS: Actually we would need to permanently support two separate API Gateways, one for all user-facing journey-event requests, and one for internal journey-event requests.

(Unless we changed all the internal journey-event requests to contain a pageId as well, which would involve updating the step function I believe)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
Joe-Edwards-GDS: Yeah ok - this approach is probably slightly better for the double-click case, and slightly worse for the duplicate tab case - but both should be unlikely.

For lambda calls I'm inclined to agree - in an ideal world we could ask the frontend to perform some kind of delay + retry (since lambda calls should be quick), but that's definitely not worth it now!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
Joe-Edwards-GDS: If we stick with this approach, we might just have to reject events that contain too many parts.

Otherwise, I don't think it should be _that_ hard to update the API gateway in a single shot, we might just need to get the frontend to start sending the page id first (in a query param) and _then_ update the backend to pass it into the step function + lambda.

RogVanGDS: Yeah that's true, we might need to name the gateway address something different from `/journey/*` so we can differentiate between them easily in the meantime.

I think that would be my preferred approach, if we are going to this length to refactor and make the journey more robust, we might as well do it properly

PR #1817, opened 2024-04-08
PYIC-5685: Revert reversion and fix bulk deletion

Wynndow:
### What changed

The first 8 commits are just the previously merged and then reverted commits from this PR: https://github.com/govuk-one-login/ipv-core-back/pull/1800

The last commit fixes the issue that caused the reversion:

### Why did it change

[PYIC-5685: Allow DataStore batch deletes](https://github.com/govuk-one-login/ipv-core-back/commit/400360860162c1422e87e4a5043ddd2a4b7a7fba)

The `delete(String partitionValue)` method on the data store implied that
it would delete all items that had that partition key. It actually would
have only worked on a table that only defined a partition key as part of
it's schema. Use on any table that also used a hash key would result in
an error.

This updates that method to now remove all items that use that partition
key. It fetches all items, and then adds them all to a batch delete
operation.

### Issue tracking

- [PYIC-5685](https://govukverify.atlassian.net/browse/PYIC-5685)

[PYIC-5685]: https://govukverify.atlassian.net/browse/PYIC-5685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -145,9 +148,24 @@ public T delete(String partitionValue, String sortValue) {
...
+        var writeBatchBuilder = WriteBatch.builder(typeParameterClass).mappedTableResource(table);
+        itemsToDelete.forEach(writeBatchBuilder::addDeleteItem);
+        var batchWriteResult =
+                client.batchWriteItem(b -> b.writeBatches(writeBatchBuilder.build()));
+
+        if (!batchWriteResult.unprocessedDeleteItemsForTable(table).isEmpty()) {
+            var joiner = new StringJoiner(", ");
+            batchWriteResult
+                    .unprocessedDeleteItemsForTable(table)
+                    .forEach(key -> joiner.add(key.toString()));
```
Joe-Edwards-GDS: Have you checked what this outputs? It doesn't look like `Key` has a `toString` override, so this might not be as pretty as you want...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -145,9 +148,24 @@ public T delete(String partitionValue, String sortValue) {
         return delete(key);
     }
 
-    public T delete(String partitionValue) {
-        var key = Key.builder().partitionValue(partitionValue).build();
-        return delete(key);
+    public void deleteAllByPartition(String partitionValue) throws DataStoreException {
```
Joe-Edwards-GDS, line 151: I slightly wonder if this is the right method signature - would it be better to have a `delete(List<T> items)` or similar to perform a generic batch delete?

(This way it could be used for other delete operations - e.g. as used by the `VerifiableCredentialStore` - although that will be deleted itself in due course, so not worth messing with too much)

Wynndow: I can certainly see that value in it, but I think this way might be cleaner for the use case.
We could have both signatures - but as far as I can see the VerifiableCredentialStore is the only other place we do any sort of bulk delete, so I wouldn't want to add it only to remove it again in the near future.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
Wynndow: I have not - will take a look

Joe-Edwards-GDS (approved):
Assuming previous commits match the previously approved code!

PR #1827, opened 2024-04-11
PYIC-5733: Read from session credentials table

Wynndow:
### What changed

Read from the new session credentials table, behind a feature flag

### Why did it change

[PYIC-5733: Add getCredentials to SessionCredentialsService](https://github.com/govuk-one-login/ipv-core-back/commit/4c2737c742fde3c058a68fbadf93078fb38278da)

This adds a method to the service to fetch all the credentials for a
particular sesion from the session-credentials table.

It also adds associated methods to convert the values retrieved to a
VerifiableCredential object, as that is what we use elsewhere in the
core.

[PYIC-5733: Read from session creds table in build-cri-oauth-request](https://github.com/govuk-one-login/ipv-core-back/commit/1fda0c3d206c03ddd9c069ed06d5151007bf9a90)

This adds the ability for the build-cri-oauth-request lambda to read
credentials from the session credentials table, rather than the
user-issued-credentials-table. Hidden behind a feature flag.

It also collapses some of the error handling logic.

[PYIC-5733: Read from session cred table in evaluate-gpg45-scores](https://github.com/govuk-one-login/ipv-core-back/commit/3a713f00b634599e7585b5d277488ad099f9adfe)

This adds the ability for the evaluate-gpg45-scores lambda to read from
the session-credentials table, or the user-issued-credentials-table,
depening on the state of the feature flag.

[PYIC-5733: Read from session cred table in CriCheckingService](https://github.com/govuk-one-login/ipv-core-back/commit/0e9d015cb2f144a07cdf56c37571955a4a843e0f)

This allows the CriCheckingService to fetch VC from the session
credentials table, or the user issued credentials table, depening on the
feature flag.

[PYIC-5733: Read from session cred table in TicfCriService](https://github.com/govuk-one-login/ipv-core-back/commit/5b2ec4e1241b1c8186398815b87303e6bc471949)

This allows the TicfCriService to fetch VC from the new session
credentials table, or the user issued credentials table, depending on
the flag.

It adds a new signature on the SessionCredentialsService that allows for
fetching credentials based on if they were received this session or not.
This required a modification to an unused method in the DataStore.

[PYIC-5733: Add getLatestCredentialsByIssuer to SessionCredentialsService](https://github.com/govuk-one-login/ipv-core-back/commit/fbdadebbd350ed2202392c1295b6f0cecb2e9001)

There is a scenario where a user may have more than one address VC in
the session-credentials-table for there current session. If they are on
a 6MFC jouurney we will have copied their address VC into the table. If
they then let us know they want to update it they will be sent to the
address CRI and collect another address VC.

This is only really a problem when building shared attributes for other
CRIs - we only want to include the newest address.

This adds a method to the sessionCredentialsService class which will
return the newest (by not before date) VC for each CRI for the users
current session from the session credentials table. We then use that in
the build-cri-oauth-request lambda.

[PYIC-5723: Read from session cred table in BuildProvenUserIdentity](https://github.com/govuk-one-login/ipv-core-back/commit/c9db179ede7922a3a3905431160d0d5a2e9b8634)

This updates the build-proven-user-identity lambda to read from the
session credentials table if the flag is set.

If we are reading from the new table, we no longer filter the VC by
profile type. This shouldn't be neccessary as we no longer display the
reuse page for operational profile reuse journeys - the user is just
sent straight back to the service. And when we _do_ show this page (for
a gpg45 reuse journey), any operational profile VC should have already
been removed for the user if previously present.

This also gets all the credentials from the table, rather than the
"latest by CRI", as we should have only stored and/or fetched the latest
versions in the long term table, rather than possibly having a duplicate
address VC.

[PYIC-5733: Read from session cred table in BuildUserIdentityHandler](https://github.com/govuk-one-login/ipv-core-back/commit/a01fbf798e6312eb03a057db690e7400bbc3eb51)

This allows the build-user-identity-lambda to read from the new session
credentials table, depending on the status of the flag.

It uses the `getLatestCredentialsByIssuer()` method, as it's important
that we only use the newest address VC in the case of a 6MFC journey.

This also refactors the ErroResponse class, and the
HttpResponseExceptionWithErrorBody to make it easier to return an error
response.

### Issue tracking

- [PYIC-5733](https://govukverify.atlassian.net/browse/PYIC-5733)

[PYIC-5733]: https://govukverify.atlassian.net/browse/PYIC-5733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
```diff
@@ -41,6 +43,27 @@ public SessionCredentialsService(ConfigService configService) {
...
+        try {
+            var verifiableCredentialList = new ArrayList<VerifiableCredential>();
+            for (var credential : dataStore.getItems(ipvSessionId)) {
+                verifiableCredentialList.add(
+                        VerifiableCredential.fromSessionCredentialItem(credential, userId));
+            }
+
+            return verifiableCredentialList;
+        } catch (CredentialParseException e) {
+            LOGGER.error("Error parsing session credential item: {}", e.getMessage(), e);
```
Joe-Edwards-GDS: Should we use `LogHelper.buildErrorMessage` here?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -13,7 +13,7 @@
 import java.util.Date;
 
 @Getter
-@EqualsAndHashCode
+@EqualsAndHashCode(exclude = "signedJwt")
```
Joe-Edwards-GDS, line 16: Do we actually use the equals at all?

If we do, I wonder if we should exclude `claimsSet` as well, since it's also derived from the VC.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/persistance/item/SessionCredentialItemTest.java
```diff
@@ -2,21 +2,37 @@
...
-
-        var item = new SessionCredentialItem("session-id", "cri-id", mockJwt, true);
+        assertEquals("cri-id#c2lnbmF0dXJl", sessionCredentialItem.getSortKey());
+    }
 
-        assertEquals("cri-id#c2lnbmF0dXJl", item.getSortKey());
+    @Test
+    void getCriIdShouldDoWhatYouThinkItShould() {
+        assertEquals(CRI_ID, sessionCredentialItem.getCriId());
     }
```
Joe-Edwards-GDS, line 37: 👀

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsServiceTest.java
```diff
@@ -109,6 +117,66 @@ void persistCredentialsShouldThrowVerifiableCredentialExceptionIfProblemStoring(
...
+                    VerifiableCredential.fromSessionCredentialItem(item2, USER_ID),
+                    retrievedVc.get(1));
+            assertEquals(
+                    VerifiableCredential.fromSessionCredentialItem(item3, USER_ID),
+                    retrievedVc.get(2));
+        }
+
+        @Test
+        void
+                getCredentialsShouldThrowVerifiableCredentialExceptionIfSessionStoreItemCanNotBeParsed() {
```
Joe-Edwards-GDS, line 167: Thanks spotless

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -221,62 +232,50 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
-                    .toObjectMap();
+            return buildJoureyErrorResponse(
+                    "Unable to determine type of credential",
+                    e,
+                    SC_INTERNAL_SERVER_ERROR,
+                    FAILED_TO_DETERMINE_CREDENTIAL_TYPE);
         }
     }
 
+    private Map<String, Object> buildJoureyErrorResponse(
```
Joe-Edwards-GDS: Typo: `buildJourneyErrorResponse`

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -221,62 +232,50 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
-                        .toObjectMap();
-            }
-            LOGGER.error(
-                    LogHelper.buildErrorMessage("Failed to create cri JAR", e.getErrorReason()));
-            return new JourneyErrorResponse(
-                            JOURNEY_ERROR_PATH,
-                            HttpStatus.SC_INTERNAL_SERVER_ERROR,
-                            e.getErrorResponse())
-                    .toObjectMap();
+        } catch (HttpResponseExceptionWithErrorBody | VerifiableCredentialException e) {
```
Joe-Edwards-GDS, line 232: We appear to have lost some logic around the error message here, do you think that's an issue?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -118,16 +118,14 @@ public List<T> getItems(String partitionValue) {
...
                 QueryConditional.keyEqualTo(Key.builder().partitionValue(partitionValue).build());
-        AttributeValue expressionValue = AttributeValue.builder().s(filterValue).build();
         var filterExpression =
                 Expression.builder()
-                        .expression("#a <= :b")
-                        .putExpressionName("#a", filterAttribute)
-                        .putExpressionValue(":b", expressionValue)
+                        .expression("#a = :b")
+                        .putExpressionName("#a", name)
+                        .putExpressionValue(":b", AttributeValue.builder().bool(value).build())
```
Joe-Edwards-GDS, line 122: I assume you've checked this works? I thought you could only query by sort key...

I was expecting the filtering to happen client-side (e.g. in `SessionCredentialsService`) since there won't be very many VCs anyway.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
```diff
@@ -76,6 +80,21 @@ public List<VerifiableCredential> getCredentials(
...
+            String ipvSessionId, String userId) throws VerifiableCredentialException {
+        return getCredentials(ipvSessionId, userId).stream()
+                .collect(
+                        groupingBy(
+                                VerifiableCredential::getCriId,
+                                maxBy(comparing(vc -> vc.getClaimsSet().getNotBeforeTime()))))
+                .values()
+                .stream()
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
Joe-Edwards-GDS: You can shorten this slightly with `.flatMap(Optional::stream)`

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -312,7 +312,7 @@ private JWEObject signEncryptJar(
 
         var vcs =
                 configService.enabled(SESSION_CREDENTIALS_TABLE_READS)
-                        ? sessionCredentialsService.getCredentials(
+                        ? sessionCredentialsService.getLatestCredentialsByIssuer(
```
Joe-Edwards-GDS: Presumably this is so we send the correct address in the shared claims?

I'm wondering if this is really the right approach - would it be better to move this logic to the bit that extracts the shared claims, i.e. distinguishing:
- 'We ignore any VCs that aren't the latest from a given issuer'
- 'When finding the address/name/dob, we use the latest VC with that information'

(I'm less convinced the first condition will be true forever than the latter!)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -159,7 +167,12 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             var contraIndicators = ciMitService.getContraIndicators(contraIndicatorsVc);
 
-            var vcs = verifiableCredentialService.getVcs(userId);
+            var vcs =
+                    configService.enabled(SESSION_CREDENTIALS_TABLE_READS)
+                            ? sessionCredentialsService.getLatestCredentialsByIssuer(
+                                    ipvSessionId, userId)
+                            : verifiableCredentialService.getVcs(userId);
```
Joe-Edwards-GDS, line 173: This makes me think - we should probably check with Orchestration and SPOT that sending two address VCs (in 6MFC) is desirable.

Otherwise we might need to drop the old address VC _somewhere_ in the journey...

Joe-Edwards-GDS (commented):
Beautiful commit hygiene!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
Wynndow: Yeah, we use it now in the tests a bit when comparing VerifiableCredential instances. I added this because `SignedJwt` defers to Object for its equals method and so compares by identity, which isn't useful. I suspect because of signatures.
`JwtClaimsSet` on the other hand does a comparison of the claims `Map`, which will end up just comparing maps and strings I think, and seems to work.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Wynndow: No I don't - I've checked the response codes generated by the exceptions that would fall in to the non missing ipv session id branch, and they're all 500s. The only thing we lose I think are the words "Failed to create cri Jar". But I think that would be fairly implicit if we ever actually threw here.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
Wynndow: Yep, it works. I did think about doing the filtering client side, but then figured if we've got the ability to do it in the query it might make things a bit cleaner.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
Wynndow: Sneaky

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Wynndow: If we use this implementation as written here, we won't be sending two address VC?
Is 6MFC live yet?

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Wynndow: Yeah, that's why this is here. Thought having the convenience method would be useful. It's used again in the build-user-identity-lambda for similar reasons.
I did actually consider making `getCredentials` do this and not adding the extra method, as I actually don't think it would be an issue for us to do this everywhere currently.
When do you forsee us needing to reference an older VC from an issuer from the users current session?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
Joe-Edwards-GDS: Ah - you can have both, but this one happens post-query (but before sending to us):
>[FilterExpression](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#API_Query_RequestSyntax)
A string that contains conditions that DynamoDB applies after the Query operation, but before the data is returned to you. Items that do not satisfy the FilterExpression criteria are not returned.
>
>A FilterExpression does not allow key attributes. You cannot define a filter expression based on a partition key or a sort key.

(I was thinking of the `KeyCondition`, which changes what's actually read from the table)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Joe-Edwards-GDS: 6MFC is planned to go live on Wednesday :)

Thinking about it after the weekend, I agree we shouldn't send a second address VC for now (as you've implemented). However it's worth noting, there have been definite rumblings about sending an 'address history' in some cases - which presumably _would_ include all address VCs - so we don't want to design ourselves into a corner.

As mentioned on the other PR, it feels like it'd better to just have the 'correct' set of VCs in the session store - if we no longer consider the old address as part of the user's identity.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
Joe-Edwards-GDS: Cool, I don't think there's much value in comparing the `claimsSet` because it's purely derived from the `vcString`, but it doesn't matter very much if it's only used in tests.

Joe-Edwards-GDS (approved):

PR #1829, opened 2024-04-11
PYIC-5851: Create store-indentity lambda

Wynndow:
### What changed

Create store-indentity lambda

### Why did it change

[PYIC-5851: New store-identity lambda skeleton](https://github.com/govuk-one-login/ipv-core-back/commit/0b93550e8d1c0e593c971252c800c09c42d6d753)

This adds all the boilerplate to get the new lambda up and running

[PYIC-5851: Implement store-identity lambda](https://github.com/govuk-one-login/ipv-core-back/commit/d3fa157c6646954957ef609420e0fe023765b7d3)

This adds the logic to the new store-identity lambda.

It fetches the credentials for the users current session from the
session credentials table (the most recent by CRI), and then stores
them.

To store them there is a new method on the VerifiableCredentialService
that first clears all the users VC, and then writes the current session
ones as a batch. To do the batch write, the DataStore class required a
new method too.

[PYIC-5851: Wire in the store-identity lambda](https://github.com/govuk-one-login/ipv-core-back/commit/3c12c4a7a34416690719a1147a97b470a81fa177)

This adds some new states for the store-identity lambda, and wires them
in to the journey map before the success page, and before the f2f
handoff page.

There's an interesting wrinkle due to having to take both the
sessionCredentialsTableReads and ticfCriBeta feature flags. The EVALUATE_GPG45_SCORES
state uses both. The process-journey-event lambda will use the target
state from the first enabled feature flag it finds. I believe the way
I've set this up then we should be able to use both flags independently
or combined and get the desired effect.

### Issue tracking

- [PYIC-5851](https://govukverify.atlassian.net/browse/PYIC-5851)

[PYIC-5851]: https://govukverify.atlassian.net/browse/PYIC-5851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                    ipvSessionService.getIpvSession(RequestHelper.getIpvSessionId(input));
+            ClientOAuthSessionItem clientOAuthSessionItem =
+                    clientOAuthSessionDetailsService.getClientOAuthSession(
+                            ipvSessionItem.getClientOAuthSessionId());
+            LogHelper.attachGovukSigninJourneyIdToLogs(
+                    clientOAuthSessionItem.getGovukSigninJourneyId());
+
+            verifiableCredentialService.storeIdentity(
+                    sessionCredentialsService.getLatestCredentialsByIssuer(
+                            ipvSessionItem.getIpvSessionId(), clientOAuthSessionItem.getUserId()),
```
Joe-Edwards-GDS, line 79: This is probably fine for now, but when we introduce EVCS we may want to examine this logic a little bit - specifically whether 'latest VC from each issuer' is the right thing.

I'm wondering if instead we actually need a bit of logic that clears the current address VC from the session store if you select 'change address' - not quite sure where we'd put that logic though :/

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -66,6 +67,25 @@ public void create(T item) {
...
+        var writeBatchBuilder = WriteBatch.builder(typeParameterClass).mappedTableResource(table);
+        items.forEach(writeBatchBuilder::addPutItem);
+        var batchWriteResult =
+                client.batchWriteItem(b -> b.writeBatches(writeBatchBuilder.build()));
+
+        if (!batchWriteResult.unprocessedPutItemsForTable(table).isEmpty()) {
+            throw new DataStoreException(
+                    String.format(
+                            "Unable to create %d datastore items",
+                            batchWriteResult.unprocessedPutItemsForTable(table).size()));
```
Joe-Edwards-GDS: Do we get any more information than this (i.e. _why_ it failed)? It looks like we're losing any context from the output except for the number of unprocessed items.

Joe-Edwards-GDS: (Possibly same question for the deletes - although at least that includes the IDs for the items)

Joe-Edwards-GDS (commented):
Looks plausible - I think the only sticky question is about how we handle the address VCs

lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
Wynndow: Maybe a new lambda that can get plugged into the journey map.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
Wynndow: We don't, we just get back a list of the items that didn't create. And as it's generic we don't have a good way to get more info about the items.
One option would be to add a method to the `DynamodbItem` interface to enforce a useful way to get a string representation for any item that might be seen here?

Joe-Edwards-GDS: Hmm I see. It feels a little scary not to know what has/hasn't been created but it's also not totally clear what would be useful.

The DynamoDB documentation suggests you should just retry the request (with exponential backoff) - suggesting you shouldn't get any interesting errors. For 'normal' requests I think the AWS SDK does this automatically for us, but I'm not sure how it'll work with the bulk operation.

lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
Joe-Edwards-GDS: Yeah, I can't quite decide between a new lambda vs. just a little bit of logic at the end of `ProcessCriCallback`... It probably depends how generic it is.

I can think of a few options I'd be happy with:
- Update to `BuildCriOauthRequest` that clears any existing VC for the CRI [if specified in the journey map?]
- Update to `ProcessCriCallback` that clears any existing VC for the CRI before storing the new one
- Looking ahead to https://govukverify.atlassian.net/browse/PYIC-5597, construct some more generic deletion lambda that can be instructed to specifically delete the old address (i.e. before going into the address CRI)

WDYT?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
Wynndow: hmm. might be a bit safer to just use the per item methods then, rather than having to worry about retrying things...

lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
Wynndow: Well I've implemented the ProcessCriCallback option here, but I think there's legs in the new lambda, especially if we wanted to make this more generic.
I don't think do it in the build-cri-oauth request lambda, as we need to make sure that we've cleared the dups for the build-user-identity lambda.

Joe-Edwards-GDS: Yeah, I think this is fine for now - and feels like the path of least resistance.

If we start needing a fancier deletion lambda (either for https://govukverify.atlassian.net/browse/PYIC-5597) or when handling COI journeys (that will also need to clear session storage once you start a name change), then we can update it.

Joe-Edwards-GDS (approved):
Definitely wants testing thoroughly before turning on in prod, but looks sensible to me.

We should also inform TSD about the new lambda etc. beforehand. As per slack, Olivia is kicking off conversations with live service.

PR #1835, opened 2024-04-16
BAU: Delete and recreate event mapping

Wynndow:
Previously we were enabling and disabling the event mapping that linked an SQS queue to the process async lambda. There seems to be an issue with this. Often, after renabling, the trigger no longer works. This takes a different approach and removes the event mapping completelely, and then recreates it.

### What changed

Delete and recreate event mapping

### Why did it change

Previously we were enabling and disabling the event mapping that linked an SQS queue to the process async lambda. There seems to be an issue with this. Often, after renabling, the trigger no longer works. This takes a different approach and removes the event mapping completelely, and then recreates it.

### DIFF
--- a/local-running/setConfigForLocalOrCloudRunning.py
+++ b/local-running/setConfigForLocalOrCloudRunning.py
@@ -239,7 +239,7 @@ def determine_dev_account():
         print(f"Looks like you're not authenticated against a dev account ({account_id}). Cowardly refusing to write params.")
     return DevAccount(account_id, account_num)
 
-def set_async_lambda_event_source_mapping_enabled_state(environment, account_id, dry_run, enabled):
+def delete_async_lambda_event_source_mapping(environment, account_id, dry_run):
     get_event_mapping_uuid_result = subprocess.run(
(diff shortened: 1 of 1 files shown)


local-running/setConfigForLocalOrCloudRunning.py
```diff
@@ -239,34 +239,49 @@ def determine_dev_account():
...
+            print("Could not delete event mapping. Cowardly refusing to do anything else")
             exit(1)
-        print(f"process-async-cri-credential lambda event mapping state set to: {enabled_flag}")
+        print("process-async-cri-credential lambda event mapping deleted")
     else:
-        print(f"process-async-cri-credential lambda event mapping state would have been set to: {enabled_flag}")
+        print("process-async-cri-credential lambda event mapping would have been deleted")
+
+def create_async_lambda_event_source_mapping(environment, account_id, dry_run):
+    create_event_mapping_uuid_return_code = subprocess.run(
```
Joe-Edwards-GDS: Should this be inside the `if not dry_run`? It looks like it actually does the create...

Wynndow: Whoops

Joe-Edwards-GDS (approved):
I want to look at the local running setup generally, but this seems plausible enough for now

PR #1837, opened 2024-04-17
PYIC-5635: Upgrade AWS SDK to v2

Wynndow:
### What changed

Upgrade AWS SDK to v2

This has passed all the acceptance tests in my dev env

### Why did it change

[PYIC-5635: Upgrade AWS SDK from v1 to v2](https://github.com/govuk-one-login/ipv-core-back/commit/0521bdcdc864b72b9962c1aa6464c7b3fa9c719c)

There are quite a few changes here, but the biggest part is replacing
the sevice clients with new versions, and building the model objects
(mostly request and responses for/from the services) with builders.

For the clients, we now specify the region and the credentials providers
explicitly. This is recommended, as it prevents AWS having to look these
things up and slowing things down.

This also uses the BOM for the sdk, rather than a shared version for
each service sdk. The AWS docs state that whilst the versions across all
services do currently match, that can't be guaranteed in future. Using
the BOM should protect us if that changes.

[PYIC-5635: Don't set credentialsProvider on AWS clients](https://github.com/govuk-one-login/ipv-core-back/commit/c11dd80caf7a0133d26730583cf9ff002bee9486) 

The docs for upgrading from v1 to v2 of the SDK suggest that if you're
using lambdas you should explicitly set the credentialsProvider to env
vars, which are provided by the lambda runtime. This is to avoid the
credential provider chain having to search for credentials in different
places, and makes things faster.

This doesn't work if you're using snapstart. Using snapstart causes the
runtime to use container credentials instead, so no env vars can be
found. Doing this prevents the credentials from expiring before the
function is restored.

https://docs.aws.amazon.com/lambda/latest/dg/snapstart-activate.html#snapstart-credentials

[PYIC-5635: Use UrlConnectionHttpClient for sdk clients](https://github.com/govuk-one-login/ipv-core-back/commit/0b53737db669d621d20aa25a52a69f9c5817c2c9)

The docs suggest that for lambdas you're better to use the more
lightweight UrlConnectionHttpClient, than the default. There is an async
version too, but I'm not sure we need that.

This updates the clients to use the lightweight sync http client. The
recommendation is to pass the sdk client builder a builder for the http
client. This allows the sdk to manage the http clients lifecycle.

This also adds some config to the root gradle project to exclude the
other http clients brought in by the sdk, to keep the classpath smaller
and lighter.

https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/lambda-optimize-starttime.html

[PYIC-5635: Tidy up gradle files](https://github.com/govuk-one-login/ipv-core-back/pull/1837/commits/10138ae5d7101b8a4ffd770ca4b1ea017f99e240)

This does a few things:

* Creates a new bundle for the lambda libs that all lambdas use
* Exports the sqs sdk from the audit service so any lambda using it
  doesn't have to also explicitly import it
* Drops the redundant 'Java' word from the sdk lib definitions
* Orders the imports in the gradle files

### Issue tracking

- [PYIC-5635](https://govukverify.atlassian.net/browse/PYIC-5635)

[PYIC-5635]: https://govukverify.atlassian.net/browse/PYIC-5635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
+++ b/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
@@ -6,7 +6,6 @@ import org.apache.logging.log4j.message.StringMapMessage;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
-import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbEnhancedClient;
 import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
 import software.amazon.awssdk.enhanced.dynamodb.Key;
@@ -56,12 +55,7 @@ public class DataStoreIpvSessionIT {
 
         var enhancedClient =
                 DynamoDbEnhancedClient.builder()
-                        .dynamoDbClient(
-                                DynamoDbClient.builder()
-                                        .region(US_WEST_2)
-                                        .credentialsProvider(
-                                                EnvironmentVariableCredentialsProvider.create())
-                                        .build())
+                        .dynamoDbClient(DynamoDbClient.builder().region(US_WEST_2).build())
                         .build();
         tableTestHarness =
                 enhancedClient.table(
--- a/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
+++ b/integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
@@ -1,20 +1,17 @@
 package uk.gov.di.ipv.core.integrationtest;
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient;
-import com.amazonaws.services.dynamodbv2.document.DynamoDB;
-import com.amazonaws.services.dynamodbv2.document.Item;
-import com.amazonaws.services.dynamodbv2.document.KeyAttribute;
-import com.amazonaws.services.dynamodbv2.document.Table;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.StringMapMessage;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbEnhancedClient;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
+import software.amazon.awssdk.enhanced.dynamodb.Key;
+import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import uk.gov.di.ipv.core.library.helpers.SecureTokenHelper;
 import uk.gov.di.ipv.core.library.persistence.DataStore;
 import uk.gov.di.ipv.core.library.persistence.item.IpvSessionItem;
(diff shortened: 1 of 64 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -130,11 +130,14 @@ public static String getJourneyEvent(JourneyRequest request)
 
     public static String getJourneyParameter(JourneyRequest request, String key)
             throws HttpResponseExceptionWithErrorBody {
-        List<NameValuePair> queryParams = new URIBuilder(request.getJourneyUri()).getQueryParams();
-        return queryParams.stream()
-                .filter(query -> Objects.equals(query.getName(), key))
+        return Optional.ofNullable(request.getJourneyUri().getQuery())
```
Wynndow: this new logic is due to losing an apache lib - must have been a transative dep of the v1 sdk. Implemented this rather than pulling a library back in.

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Joe-Edwards-GDS, line 1: Unrelated to this change, but do you think we should just delete all of this?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
```diff
@@ -4,12 +4,13 @@
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
-import org.apache.http.HttpStatus;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 
 import java.net.URI;
 import java.net.URISyntaxException;
 
+import static com.nimbusds.oauth2.sdk.http.HTTPResponse.SC_BAD_REQUEST;
```
Joe-Edwards-GDS, line 12: Looks like we also have `software.amazon.awssdk.http.HttpStatusCode.BAD_REQUEST`, but I don't have a strong preference. Looks like in other places we're using the AWS types for header values?

(It's a bit awkward that we're stealing the status codes from an essentially unrelated library, but it doesn't seem worth importing a new one)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -130,11 +130,14 @@ public static String getJourneyEvent(JourneyRequest request)
...
+                .map(queryString -> queryString.split("&"))
+                .map(Arrays::stream)
+                .orElseGet(Stream::empty)
```
Joe-Edwards-GDS: You can potentially simplify this chain a little bit:
```
return Stream.ofNullable(request.getJourneyUri().getQuery())
        .flatMap(queryString -> Arrays.stream(queryString.split("&")))
        ...
```

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -130,11 +130,14 @@ public static String getJourneyEvent(JourneyRequest request)
...
+                .filter(queryParam -> queryParam.startsWith(String.format("%s=", key)))
                 .findFirst()
-                .map(NameValuePair::getValue)
+                .map(queryParam -> queryParam.split("="))
```
Joe-Edwards-GDS: Perhaps `.split("=", 2)`?

It's not exactly well-defined what multiple equals in a query segment means, but I think traditionally any additional equals are considered to be part of the value.

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
```diff
@@ -1,20 +1,17 @@
 package uk.gov.di.ipv.core.integrationtest;
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
```
Wynndow, line 3: Looking at our GHAs, I don't think we actually run these integration tests anywhere. And I suspect they're a little bit redundant.

Should we just get rid of the integration-test project completely?

Joe-Edwards-GDS (dismissed):

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Wynndow: Ha - did you see my comment right above this one?

Wynndow: I think the answer is yes, but I'll do it in another PR

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyRequest.java
Wynndow: I don't have a strong preference either - we could always define these things for ourselves, but seems a bit unneccesary.
I don't think we're going to be losing nimbus or aws libs any time soon, so I'm happy to keep it as it is.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
Wynndow: I did wonder if you'd have a better way!

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Joe-Edwards-GDS: kill-it-with-fire.gif

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
Wynndow: Ah yeah, that's a good point.

integration-test/src/integration-test/java/uk/gov/di/ipv/core/integrationtest/DataStoreIpvSessionIT.java
Joe-Edwards-GDS: I promise you hadn't written that before I wrote my comment 😅

Wynndow: I reckon you'd written the comment, but not finished the review. And I wrote mine before you submitted the review.

Joe-Edwards-GDS (approved):
One last thought - do we need to worry about encoding at all in the URI parsing?

I think it's probably fine, we construct all the URIs ourselves, and won't have any special chars anyway.

Wynndow:
> One last thought - do we need to worry about encoding at all in the URI parsing?
> 
> I think it's probably fine, we construct all the URIs ourselves, and won't have any special chars anyway.

I did think about that, and came to the same conslusion, that we don't use them. If we did in future we might have to take another look

PR #1838, opened 2024-04-17
PYIC-5948: Filter VC before checking gpg45 match

Wynndow:
### What changed

Filter VC before checking gpg45 match

### Why did it change

We had a bug where if a user completed a F2F journey, but failed without a CI (no validity score), and then followed that with a hmrc migration journey with evidence, they could incorrectly match an M1A profile.

This is because we weren't filtering out the operational profile when doing gpg45 evaluation. The validity score from the migration VC was contributing and allowing the user to match M1A.

### Issue tracking

- [PYIC-5948](https://govukverify.atlassian.net/browse/PYIC-5948)

[PYIC-5948]: https://govukverify.atlassian.net/browse/PYIC-5948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
+++ b/libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
@@ -525,7 +525,7 @@ public interface VcFixtures {
                 Instant.ofEpochSecond(1658829758));
     }
 
-    static VerifiableCredential vcExperianFraudScoreOne() {
+    static VerifiableCredential vcExperianFraudScoreTwo() {
         TestVc.TestCredentialSubject credentialSubject =
                 TestVc.TestCredentialSubject.builder().address(List.of(ADDRESS_3)).build();
         return generateVerifiableCredential(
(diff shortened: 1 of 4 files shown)


libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -525,7 +525,7 @@ static VerifiableCredential vcExperianFraudFailed() {
                 Instant.ofEpochSecond(1658829758));
     }
 
-    static VerifiableCredential vcExperianFraudScoreOne() {
```
Wynndow, line 528: These changes are unrelated to the bug, and just something I noticed.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -525,7 +525,7 @@ static VerifiableCredential vcExperianFraudFailed() {
...
+    static VerifiableCredential vcExperianFraudScoreTwo() {
```
Joe-Edwards-GDS, line 528: !

Joe-Edwards-GDS (approved):

PR #1839, opened 2024-04-17
PYIC-5932: Handle not found mitigation routes

Wynndow:
### What changed

Handle not found mitigation routes

### Why did it change

We have logic that finds a mitigation route for a users CI based on the dcoument type associated with the CI. Previously we were throwing an error if a mitigation route couldn't be found. This was because we assumed that there were no circumstances where a particular CI could be issued that we wouldn't be able to mitigate.

The NINO CRI has broken that rule - we don't mitigate a particular CI from NINO. So rather than throwing, we should just not find a mitigation route and allow the usual failure event to be issued instead.

### Issue tracking

- [PYIC-5932](https://govukverify.atlassian.net/browse/PYIC-5932)

[PYIC-5932]: https://govukverify.atlassian.net/browse/PYIC-5932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -8,7 +8,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.StringMapMessage;
 import software.amazon.lambda.powertools.logging.Logging;
 import software.amazon.lambda.powertools.tracing.Tracing;
-import uk.gov.di.ipv.core.checkexistingidentity.exceptions.UnsupportedMitigationRouteException;
+import uk.gov.di.ipv.core.checkexistingidentity.exceptions.MitigationRouteException;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.auditing.AuditEvent;
 import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
@@ -27,7 +27,6 @@ import uk.gov.di.ipv.core.library.enums.Vot;
 import uk.gov.di.ipv.core.library.exceptions.ConfigException;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
-import uk.gov.di.ipv.core.library.exceptions.MitigationRouteConfigNotFoundException;
 import uk.gov.di.ipv.core.library.exceptions.SqsException;
 import uk.gov.di.ipv.core.library.exceptions.UnrecognisedCiException;
 import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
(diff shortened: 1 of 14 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CiMitUtilityService.java
```diff
@@ -47,52 +46,54 @@ public Optional<JourneyResponse> getCiMitigationJourneyStep(ContraIndicators con
...
+                var journeyEvent =
+                        Optional.ofNullable(
+                                        getMitigationRoute(
+                                                cimitConfig.get(ci.getCode()), ci.getDocument()))
+                                .map(MitigationRoute::event);
+
+                if (journeyEvent.isPresent()
+                        && journeyEvent.get().startsWith(JOURNEY_ALTERNATE_DOC_PATH)
                         && !configService.enabled(ALTERNATE_DOC_MITIGATION_ENABLED)) {
                     return Optional.empty();
```
Wynndow: This will go away very soon when we clean up the alternate doc feature flag

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -409,17 +405,21 @@ private JourneyResponse buildF2FNoMatchResponse(
...
+                            .orElseThrow(
+                                    () ->
+                                            new MitigationRouteException(
+                                                    String.format(
+                                                            "Empty mitigation route for mitigated CI: %s",
+                                                            mitigatedCI.get())));
+            if (!JOURNEY_ENHANCED_VERIFICATION_PATH.equals(mitigationJourney)) {
+                throw new MitigationRouteException(
+                        String.format("Unsupported mitigation route: %s", mitigationJourney));
             }
```
Joe-Edwards-GDS, line 421: (I hate the special-casing here, but fine to leave...)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
```diff
@@ -286,8 +287,9 @@ void getMitigationJourneyResponseShouldReturnEmptyWhenMitigationDoesNotResolveBr
     }
 
     @Test
-    void getMitigationJourneyResponseShouldThrowWhenCiMitigationJourneyConfigNotFoundForDocType()
-            throws Exception {
+    void
+            getCiMitigationJourneyStepShouldReturnEmptyOptionalWhenCiMitigationJourneyConfigNotFoundForDocType()
```
Joe-Edwards-GDS, line 291: (If you want this on one line, I'd be happy with `...ShouldReturnEmptyWhen...`)

Joe-Edwards-GDS (approved):

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: We could take this check out - it's only really here to catch a situation which should never really happen. Perhaps I was being over cautious adding it in the first place.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/CiMitUtilityServiceTest.java
Wynndow: Spotless still loves a newline with the change...

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: No I think it makes sense - we need some kind of capability to say 'run this step before emitting the same event again'. A problem for another day...

PR #1840, opened 2024-04-18
BAU: Remove integration tests

Wynndow:
### What changed

Remove integration tests

### Why did it change

These tests are unused, unloved and unneccesssary. So they're going in the bin.


README.md
```diff
@@ -95,12 +93,6 @@ These values are automatically assigned by terraform within the `aws_lambda_func
 sam build --cached --parallel
 ```
 
-As part of the build stage tests are disabled. To enable them run the following:
-```
-# Build with Unit Tests and Integration Tests
-GRADLE_SAM_EXECUTE_TEST=1 sam build -cached --parallel
```
Wynndow, line 101: This command doesn't actually seem to do anything. Running sam build doesn't currently run the tests either.
I've googled the env var, and only found two references. One from a blog post from someone who used to work on the team, and the other is this reference in the README. So I suspect it's not actually a thing.

Joe-Edwards-GDS (approved):

PR #1841, opened 2024-04-18
BAU: Clean up alternate doc flag

Wynndow:
### What changed

Clean up alternate doc flag

### Why did it change

We've been running the alternate doc mitigations in prod for a while
now, so it's probably ok to clean up the flag.

Once this is out in prod we can remove the config variables.


Joe-Edwards-GDS (dismissed):
We could double-check in the channel that they're happy for us to remove the feature flag, but I think it's fine.

It seems to be working fine, and we can always revert if really necessary.

Joe-Edwards-GDS (approved):

PR #1842, opened 2024-04-18
PYIC-5805: Invoke CIMIT lambdas by alias

RogVanGDS:
### What changed

Changes to invoke CIMIT lambdas by `live` alias instead of just ARN

### Why did it change

We want to invoke the lambdas by their alias, so we (and SPOT team) can control which version of the lambdas we are using (also allows us to rollback to previous versions in case of incidents).

### Issue tracking

- [PYIC-5805](https://govukverify.atlassian.net/browse/PYIC-5805)

[PYIC-5805]: https://govukverify.atlassian.net/browse/PYIC-5805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (dismissed):
This looks plausible. Are you planning to remove the old permissions in a follow-up release?

I'm not totally sure whether the switch would be linked to the code change, so it's probably safer to do in two steps.

RogVanGDS:
> This looks plausible. Are you planning to remove the old permissions in a follow-up release?
> 
> I'm not totally sure whether the switch would be linked to the code change, so it's probably safer to do in two steps.

Yeah I was planning on a teardown PR to remove the old permissions once we have testing and are happy

Joe-Edwards-GDS (approved):

PR #1843, opened 2024-04-18
BAU: Read from session credentials table in check-gpg45-scores

Wynndow:
### What changed

Read from session credentials table in check-gpg45-scores

### Why did it change

The check-gpg45-scores lambda reads the users current VCs. We should be reading from the new session credentials table instead, depending on the flag. This use was missed in the original PR to read from the new table.


Joe-Edwards-GDS (approved):
Oops

PR #1844, opened 2024-04-18
Pyc 5664 remove gson

sam803:
### Issue tracking

### Other considerations


libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -51,7 +51,8 @@
 
 public class CiMitService {
     private static final Logger LOGGER = LogManager.getLogger();
-    private static final Gson gson = new Gson();
+    private static final ObjectMapper objectMapper = new ObjectMapper();
```
Joe-Edwards-GDS, line 56: This should really be `OBJECT_MAPPER`, as it's static final (i.e. constant) - I know the previous one isn't!

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -75,16 +76,21 @@ public CiMitService(
...
     public void submitVC(VerifiableCredential vc, String govukSigninJourneyId, String ipAddress)
             throws CiPutException {
+        String payload;
+        try {
+            payload =
+                    objectMapper.writeValueAsString(
+                            new PutCiRequest(govukSigninJourneyId, ipAddress, vc.getVcString()));
+
+        } catch (JsonProcessingException e) {
+            throw new CiPutException(e.getMessage());
```
Joe-Edwards-GDS, line 92: Do you think this message could include any sensitive information?

If you want to keep the exception details, I think we'd be better off:
- Adding a second constructor to `CiPutException` which takes a message + cause
- Making the message a generic `Failed to serialise PutCiRequest`

That way we should get the full details in the log, but only send a generic message back to the client.

(This should never happen anyway, so not worth worrying too much about...)

Joe-Edwards-GDS: Ditto below

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/cimitvc/ContraIndicator.java
```diff
@@ -1,11 +1,15 @@
...
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.ToString;
 
 import java.util.List;
 
+@AllArgsConstructor
+@NoArgsConstructor(force = true)
```
Joe-Edwards-GDS, line 12: Why do we need the No-args constructor?

I wonder if we should just make this a `@Data`/`record` class instead

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
```diff
@@ -381,8 +382,16 @@ private TestConfiguration getTestConfiguration(String testDataSet, String featur
     void getCriPrivateApiKeyForActiveConnectionShouldReturnApiKeySecret() {
         environmentVariables.set("ENVIRONMENT", "test");
         Map<String, String> apiKeySecret = Map.of("apiKey", "api-key-value");
+
+        String json =
+                assertDoesNotThrow(
```
Joe-Edwards-GDS, line 387: (Do we need `assertDoesNotThrow` - could we just let the exception fail the test?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/cimitvc/ContraIndicator.java
sam803: @Joe-Edwards-GDS It seems the jackson library requires a no args contructor in order to deserialize the object - otherwise we get this exception
`Cannot construct instance of uk.gov.di.ipv.core.library.domain.cimitvc.ContraIndicator (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)`

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ConfigServiceTest.java
sam803: @Joe-Edwards-GDS Yes I guess, I figured it was more in line with the testing framework, but happy either way

PR #1845, opened 2024-04-18
PYIC-5953: Clean up session credential table switch

Wynndow:
**Not to be merged until https://github.com/govuk-one-login/ipv-core-common-infra/pull/906, and we're happy with the changes running in prod**

### What changed

Clean up session credential table switch

### Why did it change

This should be merged once the reads from the sessions credenentials table are happening in prod, and we're happy. It does a few things:

* Removes use of the sessionCredentialsTableReads feature flag
* Stops dual writing VCs. This stops writing to the user-issued-credentials-table during a journey. It's now only written to at the end of a journey
* Removes now unneccesary AWS perms from some lambdas
* Removes some now unncessary filtering from the "working" set of VCs. We should be operating on VCs of a know profile type when creating shared attributes and building identities

### Issue tracking

- [PYIC-5953](https://govukverify.atlassian.net/browse/PYIC-5953)

[PYIC-5954]: https://govukverify.atlassian.net/browse/PYIC-5954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PYIC-5953]: https://govukverify.atlassian.net/browse/PYIC-5953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
@@ -6,13 +6,11 @@ import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbParti
 import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSecondaryPartitionKey;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.domain.IpvJourneyTypes;
-import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
 import uk.gov.di.ipv.core.library.dto.AccessTokenMetadata;
 import uk.gov.di.ipv.core.library.dto.AuthorizationCodeMetadata;
 import uk.gov.di.ipv.core.library.dto.ContraIndicatorMitigationDetailsDto;
 import uk.gov.di.ipv.core.library.enums.Vot;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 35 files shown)


Joe-Edwards-GDS (commented):
Looks reasonable, but won't approve until we're ready to merge.

Do you think we should also remove the feature flags from `CoreFeatureFlag.java`?

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (dismissed):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -41,7 +39,7 @@ public class IpvSessionItem implements DynamodbItem {
     // Only for passing the featureSet to the external API lambdas at the end of the user journey.
     // Not for general use.
     private String featureSet;
-    private List<String> vcReceivedThisSession;
+    private boolean inheritedIdentityReceivedThisSession;
```
Joe-Edwards-GDS, line 42: I think this is fine for now - we might need to rethink this entirely once we introduce EVCS though, see notes on https://govukverify.atlassian.net/browse/PYIC-6083

Joe-Edwards-GDS (approved):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Joe-Edwards-GDS: (It's also worth checking this won't break any existing sessions, but I think it should be fine...)

Wynndow: I figured that as we're not live with inherited identities this should be fine. The reads and writes to this are with behind a flag, or only read if the user has an operation profile.

PR #1847, opened 2024-04-19
PYIC-5597: Repurpose `reset-identity` Lambda

sammyiyke:
## Proposed changes
[PYIC-5597](https://govukverify.atlassian.net/browse/PYIC-5597): Repurpose `reset-identity` Lambda

### What changed

- ~~Removes reset-identity lambda - `ResetIdentity`~~
- Creates new reset-session-identity lambda
- Updates Journey map -- removes reset-identity lambda journeys
- Updates Journey map -- adds reset-session-identity lambda journeys
- Removes reset-identity lambda and related resources from AWS deploy template
- Adds to logic for the new `ResetSessionIdentity` lambda
- Adds tests for new `ResetSessionIdentity` lambda logic
-  Replaces `RESET_IDENTITY` with new `RESET_SESSION_IDENTITY` for `PYI_ESCAPE_ABANDON_M2B`
-  Replaces `RESET_IDENTITY` with new `RESET_SESSION_IDENTITY ` for `CONFIRM_DELETE_DETAILS_PAGE`

### Why did it change

With the change to how identities are stored in ipv-core using Session VC Storage, we need to update the reset-identity lambda to fit. 

We no longer need to reset VCs at the start of a journey as these are stored/replaced at the end of the journeys and all other VCs are lost once the session ends.

Hence the `ResetIdentityHandler` is redundant and needs to be replaced.

### Issue tracking

- [PYIC-5597](https://govukverify.atlassian.net/browse/PYIC-5597)

### Other considerations

[PYIC-5597]: https://govukverify.atlassian.net/browse/PYIC-5597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-5597]: https://govukverify.atlassian.net/browse/PYIC-5597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/journeyEngineStepFunction.asl.json
```diff
@@ -25,11 +25,6 @@
           "StringMatches": "/journey/check-existing-identity",
           "Next": "CheckExistingIdentityLambda"
         },
-        {
-          "Variable": "$.journey",
-          "StringMatches": "/journey/reset-identity",
-          "Next": "ResetIdentityLambda"
-        },
```
Joe-Edwards-GDS, line 32: Have you tried this in your dev environment? I think you'll need to add corresponding blocks for the new lambda as well

deploy/template.yaml
```diff
@@ -2613,6 +2523,74 @@ Resources:
...
+                - 'kms:Decrypt'
+                - 'kms:GenerateDataKey'
+              Resource:
+                - !ImportValue AuditEventQueueEncryptionKeyArn
+        - KMSDecryptPolicy:
+            KeyId: !Ref DynamoDBKmsKey
+        - SSMParameterReadPolicy:
+            ParameterName: !Sub ${Environment}/core/*
+        - SQSSendMessagePolicy:
+            QueueName: !ImportValue AuditEventQueueName
```
Joe-Edwards-GDS, line 2645: I think you'll probably need CRUD permissions for some of the dynamodb tables here as well

settings.gradle
```diff
@@ -16,8 +16,8 @@ include "lambdas",
 		"lambdas:process-cri-callback",
 		"lambdas:process-journey-event",
 		"lambdas:replay-cimit-vcs",
-		"lambdas:reset-identity",
 		"lambdas:store-identity",
+		"lambdas:reset-session-identity",
```
Joe-Edwards-GDS, line 20: Do you think we can try and keep these alphabetical?

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
```diff
@@ -93,8 +90,6 @@ private Map<String, Object> processJourneyStep(
         return switch (journeyStep) {
             case "/journey/check-existing-identity" -> checkExistingIdentityHandler.handleRequest(
                     buildJourneyRequest(request, journeyStep), EMPTY_CONTEXT);
-            case "/journey/reset-identity" -> resetIdentityHandler.handleRequest(
-                    buildProcessRequest(request, processJourneyEventOutput), EMPTY_CONTEXT);
```
Joe-Edwards-GDS, line 97: Similarly, this will need a corresponding addition for the new lambda.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -229,15 +219,6 @@ private JourneyResponse getJourneyResponse(
...
             var ciScoringCheckResponse = checkForCIScoringFailure(contraIndicators);
 
-            Optional<Boolean> reproveIdentity =
-                    Optional.ofNullable(clientOAuthSessionItem.getReproveIdentity());
-
-            if (reproveIdentity.orElse(false)) {
-                return buildForceResetResponse(ciScoringCheckResponse.orElse(null));
-            } else if (configService.enabled(RESET_IDENTITY)) {
-                return buildForceGpg45ResetResponse(ciScoringCheckResponse.orElse(null));
-            }
```
Joe-Edwards-GDS, line 239: We can't quite just get rid of this I'm afraid - if the reprove identity flag is set, we still need to send a response here - we just don't need to reset the identity first.

I think the logic should be:
- If the reprove identity / reset identity flag is 'true', return `/journey/reprove-identity`
- In the journey map, route that to the `REPROVE_IDENTITY` entry point of `NEW_P2_IDENTITY`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -441,13 +387,6 @@ private JourneyResponse buildNoMatchResponse(
                                                     "Empty mitigation route for mitigated CI: %s",
                                                     mitigatedCI.get())));
         }
-        if (!VcHelper.filterVCBasedOnProfileType(verifiableCredentials, GPG45).isEmpty()) {
-            LOGGER.info(
-                    LogHelper.buildLogMessage("Failed to match profile so resetting identity."));
-            sendAuditEvent(AuditEventTypes.IPV_IDENTITY_REUSE_RESET, auditEventUser);
```
Joe-Edwards-GDS, line 442: If we're losing this audit event, we should also check with the D&A and TICF teams, to make sure they aren't using it for anything.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -728,18 +727,6 @@ PYI_KBV_DROPOUT_M2B:
...
 
 # Mitigation journey (01)
-MITIGATION_01:
-  response:
-    type: process
-    lambda: reset-identity
-    lambdaInput:
-      isUserInitiated: false
-      deleteOnlyGPG45VCs: true
-    mitigationStart: enhanced-verification
```
Joe-Edwards-GDS, line 790: We'll need to move these `mitigationStart` events onto the next step in the journey now - otherwise we'll lose the corresponding audit event.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -88,7 +88,8 @@ public enum ErrorResponse {
     FAILED_TO_DELETE_CREDENTIAL(1071, "Failed to delete credential"),
     FAILED_TO_GET_CREDENTIAL(1072, "Failed to get credential"),
     INVALID_JOURNEY_EVENT(1073, "Invalid journey event in input"),
-    FAILED_TO_STORE_IDENTITY(1074, "Failed to store identity");
+    FAILED_TO_STORE_IDENTITY(1074, "Failed to store identity"),
+    FAILED_TO_RESET_IPV_SESSION(1075, "Failed to reset ipv session");
```
Joe-Edwards-GDS: Looks like this is only used in tests - can we delete?

Joe-Edwards-GDS (commented):
Looks like the right idea, but I'd definitely suggest running this in your dev environment, and ideally trying to run the integration tests against it as well.

deploy/journeyEngineStepFunction.asl.json
sammyiyke: I did. Seems to work fine. I might be wrong but I think it is related to lambdas with "stops" on the journey.

Probably need to improve my understanding of the Journey system.

sammyiyke: But it should be coming back with the new changes.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -228,14 +222,14 @@ private JourneyResponse getJourneyResponse(
...
 
-            if (reproveIdentity.orElse(false)) {
-                return buildForceResetResponse(ciScoringCheckResponse.orElse(null));
-            } else if (configService.enabled(RESET_IDENTITY)) {
-                return buildForceGpg45ResetResponse(ciScoringCheckResponse.orElse(null));
+            if (reproveIdentity.orElse(false) || configService.enabled(RESET_IDENTITY)) {
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                "resetIdentity flag is enabled, reset users identity."));
+                return JOURNEY_RESET_SESSION_IDENTITY;
```
Joe-Edwards-GDS: This doesn't actually need to reset the session identity, it can link straight through to `NEW_P2_IDENTITY` / `REPROVE_IDENTITY`.

I'd probably call the event `/journey/reprove-identity` or similar

Joe-Edwards-GDS (commented):
This looks good, just one small comment, and we need to see what comes out of the audit event conversations :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: We've done this, and agreed this is fine :)

deploy/template.yaml
```diff
@@ -1673,93 +1670,6 @@ Resources:
       FilterPattern: ""
       LogGroupName: !Ref CheckExistingIdentityFunctionLogGroup
 
-
-  ResetIdentityFunction:
```
Joe-Edwards-GDS, line 1663: Do you think there's any risk of this breaking active sessions? I'm wondering if it might be safer to split this into two PRs:
- First add the new lambda and make all the journey map changes
- Second delete the old lambda

cc @Wynndow

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: I think this comment still applies :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -61,35 +55,7 @@ CHECK_EXISTING_IDENTITY:
 RESET_IDENTITY:
   response:
     type: process
-    lambda: reset-identity
-    lambdaInput:
-      isUserInitiated: false
-      deleteOnlyGPG45VCs: false
```
Joe-Edwards-GDS, line 67: Can you remove this state entirely? I'm not sure it's valid at the moment

deploy/template.yaml
Wynndow: Yeah, I think it'd be safe to drain the lambda before removing it.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -229,7 +226,7 @@ private JourneyResponse getJourneyResponse(
                 LOGGER.info(
                         LogHelper.buildLogMessage(
                                 "resetIdentity flag is enabled, reset users identity."));
-                return JOURNEY_RESET_SESSION_IDENTITY;
+                return JOURNEY_IPV_GPG45_MEDIUM;
```
Joe-Edwards-GDS: Sorry @sammyiyke  - this needs to go to a slightly different place. We still need to display the extra reprove identity screen, so it needs to go to `NEW_P2_IDENTITY`/`REPROVE_IDENTITY` (rather than `NEW_P2_IDENTITY`/`START`)

deploy/template.yaml
sammyiyke: Brought back the `resetIdentity` lambda as part of this PR. Would create a second PR for the clean up after this is merged

Joe-Edwards-GDS: Thanks!

Joe-Edwards-GDS (approved):

PR #1849, opened 2024-04-22
PYIC-5663: remove gson 

sam803:
Consolidate JSON parsing into one library by removing GSON and replacing any instances it used with jackson. 

### What changed

 - Removed GSON library
 - Replaced GSON methods with equivalent Jackson methods

### Why did it change

To reduce dependencies and simplify development

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-5663

### DIFF
--- a/libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
+++ b/libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
@@ -1,7 +1,7 @@
 package uk.gov.di.ipv.core.library.service;
 
-import com.google.gson.Gson;
-import com.google.gson.reflect.TypeToken;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jose.jwk.ECKey;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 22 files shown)


libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -226,9 +243,19 @@ private EvidenceItem parseContraIndicatorEvidence(VerifiableCredential vc)
...
-                        GSON.toJson(claimSetJsonObject), new TypeToken<CiMitJwt>() {}.getType());
+        String valueAsString;
+        try {
+            valueAsString = OBJECT_MAPPER.writeValueAsString(claimSetJsonObject);
+        } catch (JsonProcessingException e) {
+            throw new CiRetrievalException("Failed to serialize claim set");
+        }
+        CiMitJwt ciMitJwt;
+        try {
+            ciMitJwt = OBJECT_MAPPER.readValue(valueAsString, CiMitJwt.class);
```
Joe-Edwards-GDS: You should be able to do this in one shot with Jackson: `OBJECT_MAPPER.convertValue(claimSetJsonObject, CiMitJwt.class);`

(`convertValue` goes between different types, read/write is for serialising/deserialising to string)

Joe-Edwards-GDS (commented):
One small comment, but looks good. You're lacking on code coverage on some of the error cases - ideally we'd cover those, but if you think that's a false positive then we can override it.

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
sam803: thanks, just added

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1850, opened 2024-04-22
PYIC-6011: Don't set P2 for 6MFC journeys

Wynndow:
.

### What changed

Don't set P2 for 6MFC journeys

### Why did it change

We were prematurely setting the users VOT on their session as P2 for 6MFC journeys.

This change will only set the VOT if the user is going to return to the service - either with a P2, or an operational profile

### Issue tracking

- [PYIC-6011](https://govukverify.atlassian.net/browse/PYIC-6011)

[PYIC-6011]: https://govukverify.atlassian.net/browse/PYIC-6011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -376,15 +376,10 @@ public class CheckExistingIdentityHandler
 
         // vot achieved for vtr
         if (strongestAttainedVotFromVtr.isPresent()) {
-            // Profile matched
-            Vot attainedVot = strongestAttainedVotFromVtr.get();
-
-            ipvSessionItem.setVot(attainedVot);
-            ipvSessionService.updateIpvSession(ipvSessionItem);
             return Optional.of(
                     buildReuseResponse(
-                            attainedVot,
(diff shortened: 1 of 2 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -454,39 +449,48 @@ private JourneyResponse buildNoMatchResponse(
...
             throws SqsException, VerifiableCredentialException {
+        // check the result of 6MFC and return the appropriate journey
+        if (configService.enabled(REPEAT_FRAUD_CHECK)
+                && attainedVot.getProfileType() == GPG45
+                && !hasCurrentFraudVc(vcs)) {
+            LOGGER.info(LogHelper.buildLogMessage("Expired fraud VC found"));
+            sessionCredentialsService.persistCredentials(
+                    allVcsExceptFraud(vcs), auditEventUser.getSessionId(), false);
+            return JOURNEY_REPEAT_FRAUD_CHECK;
+        }
```
Joe-Edwards-GDS, line 464: (Given the name of this method, I wonder if we should just pull this out of the method entirely - but this is fine)

Joe-Edwards-GDS (approved):

PR #1851, opened 2024-04-22
PYIC-6028: Stop dual writing credentials

Wynndow:
### What changed

Stop dual writing credentials

### Why did it change

This stops core from dual writing VCs to the session-credentials table and the user-issued-credentials table.

This was previously covered in [PR#1845](https://github.com/govuk-one-login/ipv-core-back/pull/1845), however I've noticed that deploying that PR with live traffic results in errors.

It appears the permissions removal on that PR happens before the new version of the process-cri-callback lambda is active - with about a 4 minute window where the outgoing lambda is trying to write to the user-issued-credentials table but without the correct permissions, resulting in a technical error. This could affect a lot of users.

I may split that PR further to ensure all permission changes go out separately to the code changes that require them.

### Issue tracking

- [PYIC-6028](https://govukverify.atlassian.net/browse/PYIC-6028)

[PYIC-6028]: https://govukverify.atlassian.net/browse/PYIC-6028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (commented):
This looks sensible - and good to know that we need to be careful removing permissions. (cc @sammyiyke and @RogVanGDS  - I think you're both doing that)

I won't approve yet since I think we need to enable the reads in prod first?

Joe-Edwards-GDS (approved):

PR #1854, opened 2024-04-23
PYIC-5959: configuration for update address journey

sam803:
## Proposed changes

Adds configuration for the change address COI sub journey

### What changed

 Added a new journey map config and associated errors 

### Why did it change

New Continuity of Identity journey required to change adddress only

### Issue tracking

- [PYIC-5959](https://govukverify.atlassian.net/browse/PYIC-5959)

[PYIC-5959]: https://govukverify.atlassian.net/browse/PYIC-5959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-address.yaml
```diff
@@ -0,0 +1,119 @@
...
+      targetState: ERROR
+    fail-with-ci:
+      targetJourney: FAILED
+      targetState: FAILED_RFC
+    vcs-not-correlated:
+      targetJourney: FAILED
+      targetState: FAILED_RFC
+    alternate-doc-invalid-dl:
+      targetJourney: FAILED
+      targetState: FAILED_RFC
```
RogVanGDS: We don't seem to be using this parent state anywhere, I think we can probably remove it but it might be worth checking it still works without it

sam803: @RogVanGDS I initially did not put it in but it was used in the nested ADDRESS_AND_FRAUD journey so did not work, see https://github.com/govuk-one-login/ipv-core-back/blob/77375cb41affb0869a124897acd47e30b8e316a2/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml#L1

Joe-Edwards-GDS: If we do need the parent state, should the failures go to `FAILED_ADDRESS` rather than `FAILED_RFC`? (If those are actually different)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -34,6 +34,11 @@ FAILED_RFC:
     next:
       targetState: NO_MATCH_PAGE_RFC
 
+FAILED_UPDATE_ADDRESS:
+  events:
+    next:
+      targetState: NO_MATCH_PAGE_ADDRESS
```
Joe-Edwards-GDS: I think there's probably an oversight here and for the RFC as well - these states should all go via TICF as well, when the feature flag is enabled.

It's also worth checking whether we need a separate failure state for this? In the figma it's the same as should be used for repeat fraud check, so I think we can just direct to that page instead.

(We might need a separate ticket to change the design of that page, if the current one needs updating)

Joe-Edwards-GDS: Typo - this doesn't match the state below
`NO_MATCH_PAGE_ADDRESS`
vs
`NO_MATCH_PAGE_UPDATE_ADDRESS`

sam803: @Joe-Edwards-GDS thanks, yes they are the same, I'm a bit reluctant to use the FAILED_RFC **and** change the RFC to go via TICF at the same time as the RFC part would not be tested as part of this ticket. Might be worth a separate ticket?

Joe-Edwards-GDS: Yeah - although we clearly don't have a test for it. I'll raise a bug ticket for the RFC case and we can either fix as part of this or separately: https://govukverify.atlassian.net/browse/PYIC-6122

sam803: @Joe-Edwards-GDS I've added the TICF check and removed a bit of replication in the states, should be easy enough to  do that ticket when the time comes and get them both using the same error state

Joe-Edwards-GDS (approved):
I think this looks sensible for now - we can always rejig once PYIC-6125 is sorted if necessary.

(cc @RogVanGDS )

PR #1858, opened 2024-04-24
BAU: Supply region for datastore DDB client

Wynndow:
### What changed

Supply region for datastore DDB client

### Why did it change

It's recommended to supply the region when building the clients with lambda. It prevents the SDK having to do some internal logic, and makes things faster.

This should have been included in a previous PR but was missed.

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
@@ -24,6 +24,8 @@ import uk.gov.di.ipv.core.library.service.ConfigService;
 import java.time.Instant;
 import java.util.List;
 
+import static software.amazon.awssdk.regions.Region.EU_WEST_2;
+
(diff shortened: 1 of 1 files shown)


Joe-Edwards-GDS (approved):
One comment I meant to leave on the AWS SDK upgrade (but obviously forgot)...

Do you think we should make the region configurable (perhaps from an environment variable or similar)? We're hardcoding `eu-west-2` all over the place now, which will make it a lot less portable if we ever had to spin up in a different region.

Not crucial, and probably better as a separate PR.

Wynndow:
> One comment I meant to leave on the AWS SDK upgrade (but obviously forgot)...
> 
> Do you think we should make the region configurable (perhaps from an environment variable or similar)? We're hardcoding `eu-west-2` all over the place now, which will make it a lot less portable if we ever had to spin up in a different region.
> 
> Not crucial, and probably better as a separate PR.

My feeling is this is fine - if we're ever at the point of spinning up in a different region (which I really can't see ever actually happening), this seems like a relatively trivial thing to update, or make configurable at that point.

PR #1865, opened 2024-04-26
Pyic 5757 vocab types

Joe-Edwards-GDS:
Add a dependency on vocab types, and parse VCs into them, logging a warning if it fails.

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
@@ -5,6 +5,7 @@ import com.nimbusds.jwt.SignedJWT;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
+import uk.gov.di.ipv.core.library.helpers.VerifiableCredentialParser;
 import uk.gov.di.ipv.core.library.persistence.item.SessionCredentialItem;
 import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
 
@@ -31,6 +32,9 @@ public class VerifiableCredential {
             this.claimsSet = signedJwt.getJWTClaimsSet();
             this.signedJwt = signedJwt;
             this.migrated = migrated;
+
+            // Attempt to parse into a proper VC class, but do nothing with it for now
+            VerifiableCredentialParser.parseCredential(signedJwt);
         } catch (ParseException e) {
             throw new CredentialParseException(
                     "Failed to get jwt claims to construct verifiable credential", e);
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
@@ -4,11 +4,7 @@ import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
-import uk.gov.di.ipv.core.library.helpers.LogHelper;
-import uk.gov.di.ipv.core.library.helpers.VerifiableCredentialParser;
 import uk.gov.di.ipv.core.library.persistence.item.SessionCredentialItem;
 import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
 
@@ -19,7 +15,6 @@ import java.util.Date;
 @Data
 @EqualsAndHashCode(exclude = "signedJwt")
 public class VerifiableCredential {
-    private static final Logger LOGGER = LogManager.getLogger();
     private final String userId;
     private final String criId;
     private final String vcString;
@@ -39,14 +34,6 @@ public class VerifiableCredential {
         } catch (ParseException e) {
             throw new CredentialParseException(
                     "Failed to get jwt claims to construct verifiable credential", e);
-        }
-
-        try {
-            VerifiableCredentialParser.parseCredential(claimsSet);
-        } catch (CredentialParseException e) {
-            // For now, we just log a warning here that we can fix
-            // In future this should return a CredentialParseException instead
-            LOGGER.warn(LogHelper.buildErrorMessage("Failed to parse verifiable credential", e));
         }
     }
 
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
@@ -4,7 +4,11 @@ import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
+import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.helpers.VerifiableCredentialParser;
 import uk.gov.di.ipv.core.library.persistence.item.SessionCredentialItem;
 import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
@@ -16,12 +20,14 @@ import java.util.Date;
 @Data
 @EqualsAndHashCode(exclude = "signedJwt")
 public class VerifiableCredential {
+    private static final Logger LOGGER = LogManager.getLogger();
     private final String userId;
     private final String criId;
     private final String vcString;
     private final JWTClaimsSet claimsSet;
     private final SignedJWT signedJwt;
     private Instant migrated;
+    private uk.gov.di.model.VerifiableCredential vc;
 
     private VerifiableCredential(String userId, String criId, SignedJWT signedJwt, Instant migrated)
             throws CredentialParseException {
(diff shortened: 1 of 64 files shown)


libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/TestVc.java
```diff
@@ -71,7 +71,6 @@ public static class TestEvidence {
         private Integer validityScore;
         private Integer verificationScore;
         @Builder.Default private List<Object> ci = Collections.emptyList();
-        @Builder.Default private List<Object> ciReasons = Collections.emptyList();
```
Joe-Edwards-GDS, line 74: This doesn't appear in the schema - if it's present in real VCs then we may need to add it back

settings.gradle
```diff
@@ -1,3 +1,12 @@
+import org.gradle.api.initialization.resolve.RepositoriesMode
+
+if (file("secrets.gradle").exists()) {
+	apply from: "secrets.gradle"
+} else {
+	ext.githubUsername = System.getenv("GITHUB_ACTOR")
+	ext.githubToken = System.getenv("GITHUB_TOKEN")
+}
```
Wynndow, line 8: I'm wondering if rather than managing a separate `secrets.gradle` file, we just get devs to export these values to their envs in their zshrc/bashrc files? Then there's just one way to resolve these values regardless of the environment it's running in.

Joe-Edwards-GDS: Hmm, I'm not sure... Especially if it's just for this repo, I like keeping it contained. Any project setup that involves system-wide configuration feels like a bit of a red flag!

If you prefer to configure as environment variables, you can do - just don't create the `secrets.gradle` file. (I could put this in the readme explicitly)

Wynndow: Yeah, I see your point. There's just something about secrets being written to a file in a project for each dev like this that makes me a little uneasy. I guess as long as the PAT has the correct scope, read only, it should be ok.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -33,6 +38,14 @@ private VerifiableCredential(String userId, String criId, SignedJWT signedJwt)
             throw new CredentialParseException(
                     "Failed to get jwt claims to construct verifiable credential", e);
         }
+
+        try {
+            VerifiableCredentialParser.parseCredential(claimsSet);
+        } catch (CredentialParseException e) {
+            // For now, we just log a warning here that we can fix
+            // In future this should return a CredentialParseException instead
+            LOGGER.warn(LogHelper.buildErrorMessage("Failed to parse verifiable credential", e));
```
Wynndow: This doesn't seem to be generating much that leads to "why" it failed to parse. I've run the acceptance tests against this to see what it produces, and the error messages look like:
```
{
...
  "message": {
    "description": "Failed to parse verifiable credential",
    "errorDescription": "uk.gov.di.ipv.core.library.exceptions.CredentialParseException: Failed parse verifiable credential"
  },
...
}
```

Joe-Edwards-GDS: Hmm, I'd have expected the underlying exception to form part of the stack trace. Is that not being included?

Wynndow: Not in the logs I'm streaming from my deployed stack

Joe-Edwards-GDS: Ok, I'll see if we can get more detail - not much use if we lose the original message

Wynndow: This is straight from CloudWatch
```
{
    "timestamp": "2024-04-29T09:53:53.831+0000UTC",
    "instant": {
        "epochSecond": 1714384433,
        "nanoOfSecond": 831076563
    },
    "thread": "main",
    "level": "WARN",
    "loggerName": "uk.gov.di.ipv.core.library.domain.VerifiableCredential",
    "message": {
        "description": "Failed to parse verifiable credential",
        "errorDescription": "uk.gov.di.ipv.core.library.exceptions.CredentialParseException: Failed parse verifiable credential"
    },
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.log4j.spi.AbstractLogger",
    "threadId": 1,
    "threadPriority": 5,
    "clientId": "orchestrator",
    "coldStart": "false",
    "componentId": "https://dev-chrisw.01.dev.identity.account.gov.uk",
    "functionArn": "arn:aws:lambda:eu-west-2:130355686670:function:build-user-identity-dev-chrisw:live",
    "functionMemorySize": "1024",
    "functionName": "build-user-identity-dev-chrisw",
    "functionVersion": "268",
    "function_request_id": "d96b1fe8-7d03-4131-bbe7-3a5459cd0988",
    "govuk_signin_journey_id": "09a3e119-e9df-4221-8f80-28a5f630a9ff",
    "ipvSessionId": "i9s94p1uD391OLgQASwey5rEBcVCG-xOG6c8F1EKnkw",
    "samplingRate": "0.0",
    "service": "build-user-identity-dev-chrisw",
    "xray_trace_id": "1-662f6e31-57a91daa2b6136a64a41304a"
}
```

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/LogHelper.java
```diff
@@ -139,7 +147,14 @@ public static StringMapMessage buildErrorMessage(String message, String errorDes
     }
 
     public static StringMapMessage buildErrorMessage(String message, Exception e) {
-        return buildLogMessage(message).with(LOG_ERROR_DESCRIPTION.getFieldName(), e);
+        var mapMessage = buildLogMessage(message).with(LOG_ERROR_DESCRIPTION.getFieldName(), e);
+        if (LOGGER.isDebugEnabled()) {
```
Wynndow: I'm trying to work out if this is confusing or not. If we want to get the stacktrace we need to set the `LogHelper` logger to debug level. But it's not the actual logger doing the logging - it's just being used as a flag. It does give an easy way to just target logging of these stack traces from any logger, rather than having to set a bunch of different loggers (or everything) to debug. But actually setting it to debug is a bit of a pain. If this is the behaviour we want, could it be done with a feature set?

Wynndow: Config to add to log4j2.json that turned on debugging, added to Logging.loggers
```
        {
          "name": "uk.gov.di.ipv.core.library.helpers.LogHelper",
          "level": "debug",
          "additivity": false,
          "AppenderRef": {
            "ref": "JsonAppender"
          }
        }
```

Joe-Edwards-GDS: As discussed, going to revert this for now. I still think stack traces are important, but not related to this change.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -35,6 +40,14 @@ private VerifiableCredential(String userId, String criId, SignedJWT signedJwt, I
             throw new CredentialParseException(
                     "Failed to get jwt claims to construct verifiable credential", e);
         }
+
+        try {
+            VerifiableCredentialParser.parseCredential(claimsSet);
```
Joe-Edwards-GDS: We're going to have a name clash at some point, between the model `VerifiableCredential` types and this class. Either we'll just have to live with it (and specify if we need to handle both) or rename this class (which will be very noisy)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialParser.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final ObjectMapper LAX_OBJECT_MAPPER =
+            new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    private VerifiableCredentialParser() {}
+
+    public static VerifiableCredential parseCredential(JWTClaimsSet claimsSet)
+            throws CredentialParseException {
+        try {
+            var vcClaim = claimsSet.getJSONObjectClaim("vc");
```
Wynndow:
```suggestion
            var vcClaim = claimsSet.getJSONObjectClaim(VC_CLAIM);
```
We have a bunch of constants defined in `VerifiableCredentialConstants`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialParser.java
```diff
@@ -0,0 +1,71 @@
...
+
+            if (vcClaim != null && vcClaim.get("type") instanceof List<?> type) {
```
Wynndow:
```suggestion
            if (vcClaim != null && vcClaim.get(VC_TYPE) instanceof List<?> type) {
```

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialParser.java
```diff
@@ -0,0 +1,71 @@
...
+    private static <T extends VerifiableCredential> T parseCredentialOfType(
+            Object vcClaim, Class<T> vcType) {
+        try {
+            return OBJECT_MAPPER.convertValue(vcClaim, vcType);
+        } catch (IllegalArgumentException e) {
+            // Try again with more relaxed parsing rules
+            var vc = LAX_OBJECT_MAPPER.convertValue(vcClaim, vcType);
+            LOGGER.warn(
+                    LogHelper.buildErrorMessage(
+                            "Credential contained unexpected properties: " + e.getMessage(), e));
```
Wynndow, line 70: Is it worth logging before attempting to convert again? If `LAX_OBJECT_MAPPER.convertValue()` throws for some reason we'll lose the log message.

Joe-Edwards-GDS: This was deliberate, to avoid a duplicate log message. On the assumption that if both `OBJECT_MAPPER` and `LAX_OBJECT_MAPPER` throw, it'll be for the same reason...

Wynndow (approved):

PR #1882, opened 2024-05-01
BAU: Cleanup implementation of nimbusdsOauth2OidcSdk library

sammyiyke:
## Proposed changes

Cleans up how the `nimbusdsOauth2OidcSdk` library is implemented

### What changed

Changed how `nimbusdsOauth2OidcSdk` library dependency is exposed consumers of common-services.

- Moves the `nimbusdsOauth2OidcSdk` dependency to the api level configuration in `common-services`

- Removes the `nimbusdsOauth2OidcSdk` dependency from the implementation in lambdas that implement the `common-services` library

### Why did it change

By moving the dependecy to the `api` level configurations the `nimbusdsOauth2OidcSdk` will  be transitively exposed to consumers of the library. This would speed up compilation and makes it cleaner for lambdas as they do not directly implement this libary

Some benefits include:

- Reduces duplication across the dependency declaration
- Faster compilation
- Less recompilations when implementation dependencies change
- Less dependency leaks
- Cleaner publishing

Also helps when new lambdas are created - no need to implement a library that is not directky used 
😅

### DIFF
--- a/lambdas/build-client-oauth-response/build.gradle
+++ b/lambdas/build-client-oauth-response/build.gradle
@@ -12,7 +12,6 @@ repositories {
 dependencies {
 
 	implementation libs.bundles.awsLambda,
-			libs.nimbusdsOauth2OidcSdk,
 			project(":libs:common-services"),
 			project(":libs:journey-uris"),
 			project(":libs:audit-service")
--- a/lambdas/build-cri-oauth-request/build.gradle
+++ b/lambdas/build-cri-oauth-request/build.gradle
@@ -11,7 +11,6 @@ repositories {
 
 dependencies {
 	implementation libs.bundles.awsLambda,
-			libs.nimbusdsOauth2OidcSdk,
 			project(":libs:common-services"),
 			project(":libs:kms-es256-signer"),
 			project(":libs:journey-uris"),
--- a/lambdas/build-proven-user-identity-details/build.gradle
+++ b/lambdas/build-proven-user-identity-details/build.gradle
@@ -11,7 +11,6 @@ repositories {
 
 dependencies {
 	implementation libs.bundles.awsLambda,
-			libs.nimbusdsOauth2OidcSdk,
 			project(":libs:common-services"),
 			project(":libs:verifiable-credentials"),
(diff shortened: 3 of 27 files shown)


Wynndow (commented):
I like this, but I think it has the potential to go further. A whole bunch of the other libs (most i think), are using `common-services`, and could possibly also have their direct dependency on nimbus removed

Joe-Edwards-GDS (commented):
This looks fine, but I think the main benefit is reducing duplication across the dependency declaration, not speed of compilation.

The benefits you list (reduced recompilation etc.) apply when moving dependencies the other way:
- If a library is declared as an 'API' dependency, it becomes a transitive dependency and is required when compiling all other projects that depend upon it
- If a library is declared as an 'Implementation' dependency it is not transitive, and must be explicitly declared in other projects where it's needed

In this case, the old approach requires every lambda to declare whether it needs nimbus to compile, but now we automatically require it (whether it's technically needed or not).

I like reducing duplication though, and basically everything needs all these dependencies.

sammyiyke:
> I like this, but I think it has the potential to go further. A whole bunch of the other libs (most i think), are using `common-services`, and could possibly also have their direct dependency on nimbus removed

Great. I would add those to this PR too

sammyiyke:
> This looks fine, but I think the main benefit is reducing duplication across the dependency declaration, not speed of compilation.
> 
> The benefits you list (reduced recompilation etc.) apply when moving dependencies the other way:
> 
> * If a library is declared as an 'API' dependency, it becomes a transitive dependency and is required when compiling all other projects that depend upon it
> * If a library is declared as an 'Implementation' dependency it is not transitive, and must be explicitly declared in other projects where it's needed
> 
> In this case, the old approach requires every lambda to declare whether it needs nimbus to compile, but now we automatically require it (whether it's technically needed or not).
> 
> I like reducing duplication though, and basically everything needs all these dependencies.

Make sense. 

Getting a clearer idea of how the "imports" or implementation works in Java.

Joe-Edwards-GDS (approved):

PR #1895, opened 2024-05-07
Pyic 5776 : pass audit header to TxMA events

shivanshuit914:
## Proposed changes

- pass audit header to TxMA events

### What changed

- Changed audit event implementation to include device information.

### Why did it change

- To pass user device information to TxMA.

### Issue tracking

- [PYIC-5776](https://govukverify.atlassian.net/browse/PYIC-5776)

Screenshot 2024-05-10 at 15 00 44

[PYIC-5776]: https://govukverify.atlassian.net/browse/PYIC-5776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -227,14 +227,15 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-            throws ParseCriCallbackRequestException {
+            throws ParseCriCallbackRequestException, HttpResponseExceptionWithErrorBody {
         try {
             var callbackRequest = objectMapper.readValue(input.getBody(), CriCallbackRequest.class);
-            callbackRequest.setIpvSessionId(input.getHeaders().get("ipv-session-id"));
+            callbackRequest.setIpvSessionId(RequestHelper.getIpvSessionId(input));
             callbackRequest.setFeatureSet(RequestHelper.getFeatureSet(input.getHeaders()));
-            callbackRequest.setIpAddress(input.getHeaders().get("ip-address"));
-
+            callbackRequest.setIpAddress(RequestHelper.getIpAddress(input));
```
blakeyp: Looking at the implementation of RequestHelper.getIpvSessionId and RequestHelper. getIpAddress I think this changes the behaviour - will throw error if the header is missing. I don't think we want that, especially for ipvSessionId which may be null in the case of the cross-browser issue

blakeyp:
Can we default to empty string rather than null when the header is missing/header value is empty?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
shivanshuit914: Thanks, Replace it with `input.getHeaders()`

blakeyp (approved):

PR #1904, opened 2024-05-10
PYIC-5809: add page-update-name step to update name

sam803:
### What changed

The update name subjourney is missing a step that takes it to the page-update-name before going through the steps to initiate a name change. This page gives the user an opportunity to cancel/go back

The context sent to the dcmaw-success page also need changing to reflect what the front end is expecting - this was preventing the contact us details being shown for this journey

I've also updated the UPDATE_NAME_DOB_PAGE state action for cancelling  in the reuse-identity journey to go be consistent with the update-name subjourney and go back  to the reuse-identity start when selecting 'go back and check your details'

### Issue tracking
- https://govukverify.atlassian.net/browse/PYIC-5809


RogVanGDS (approved):
Nice one, thanks for sorting this 👍

PR #1915, opened 2024-05-14
PYIC-5961: update 6MFC routing

thebauSoftwire:
### What changed
Updated routing for 6MFC to route to COI journey if they choose to update their details.

***Note:*** This hasn't been manually tested yet, just waiting for the work for the page to be done before testing ([PYIC-5960](https://govukverify.atlassian.net/browse/PYIC-5960)).

### Why did it change

This updates the 6MFC process to use the COI process if the user chooses to update their details. This process is more consistent with the initial identity verification and fraud check processes.

### Issue tracking

- [PYIC-5961](https://govukverify.atlassian.net/browse/PYIC-5961)

[PYIC-5961]: https://govukverify.atlassian.net/browse/PYIC-5961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PYIC-5960]: https://govukverify.atlassian.net/browse/PYIC-5960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -35,6 +35,9 @@ states:
     events:
       next:
         targetState: REPROVE_IDENTITY_START
+        checkFeatureFlag:
+          coiEnabled:
+            targetState: REPROVE_IDENTITY_START_COI
 
   # Parent states
 
@@ -94,6 +97,15 @@ states:
     response:
       type: page
       pageId: reprove-identity-start
+    events:
+      next:
+        targetState: IDENTITY_START_PAGE
+
+  REPROVE_IDENTITY_START_COI:
+    response:
+      type: page
+      pageId: reprove-identity-start
+      context: coi
     events:
       address-only:
         targetJourney: UPDATE_ADDRESS
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -133,7 +133,7 @@ states:
       pageId: update-name-date-birth
     events:
       end:
-        targetState: REPROVE_IDENTITY_START
+        targetState: REPROVE_IDENTITY_START_COI
 
   RESET_SESSION_IDENTITY:
     response:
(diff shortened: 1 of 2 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -98,6 +101,40 @@ states:
...
+      next:
+        targetState: IDENTITY_START_PAGE
+
+  UPDATE_NAME_DOB:
+    response:
+      type: page
+      pageId: update-name-date-birth
+    events:
+      end:
+        targetState: REPROVE_IDENTITY_START
```
sam803: should this have a feature flag check for `coiEnabled` and send to `REPROVE_IDENTITY_START_COI` if so?

MikeCollingwood: The check is already done in REPROVE_IDENTITY, so should be alright without it here as UPDATE_NAME_DOB can only be reached from it?

sam803: Ok, in that case it should probably just go to `REPROVE_IDENTITY_START_COI`

Wynndow (commented):
This looks like it's making changes to the reprove identity journey, which is a different thing to the 6MFC journey. Reprove identity happens when the we're passed a flag by orchestrator asking us to make sure this user goes through a full IPV journey again.

i think I was expecting to see the new events coming from the a new state in `repeat-fraud-check.yml`. That new state will be the one that renders the new page that @sam803 is working on updating.

That new state will be routed to when a user starts a 6MFC journey instead of the existing first state (`CONFIRM_NAME_DOB`), if the `coiEnabled` flag is active.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
```diff
@@ -45,6 +48,32 @@ states:
 
   # Journey States
 
+  CONFIRM_DETAILS:
+    response:
+      type: page
+      pageId: confirm-your-details
+      context: coi
```
Wynndow: Looking at Sam's PR for this page, I don't think we need to include this context. The page will only have one form.

sam803: @Wynndow yes thats right - sorry I wrongly told @thebauSoftwire to add it when I as under the impression we would be editing an existing page

sam803 (approved):
LGTM

Wynndow (approved):

PR #1921, opened 2024-05-15
PYIC-6200: Add coi-check lambda

Wynndow:
### What changed

Add coi-check lambda

### Why did it change

This adds the coi-check lambda, which will ensure that for COI journeys the old identity matches the new identity. This is done by checking the names and DOB from both the new and old.

This change updates and reuses the correlation checking logic already available in the user identity service.

There will need to be subsequent PRs to wire the new lambda in.

### Issue tracking

- [PYIC-6200](https://govukverify.atlassian.net/browse/PYIC-6200)

[PYIC-6200]: https://govukverify.atlassian.net/browse/PYIC-6200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -0,0 +1,137 @@
...
+                            .toList();
+
+            var coiSubjourneyType = ipvSession.getCoiSubjourneyType();
+            var successfulCheck =
+                    switch (coiSubjourneyType) {
+                        case GIVEN_NAMES_ONLY, GIVEN_NAMES_AND_ADDRESS -> userIdentityService
+                                .areFamilyNameAndDobCorrelatedForCoiCheck(credentials);
+                        case FAMILY_NAME_ONLY, FAMILY_NAME_AND_ADDRESS -> userIdentityService
+                                .areGivenNamesAndDobCorrelated(credentials);
+                        case ADDRESS_ONLY -> userIdentityService.areVcsCorrelated(credentials);
```
sam803, line 107: Might be good practice to put a default case in (false) ?

Wynndow: I'm not 100% sure it's necessary - it currently covers all enum's, and if we extend the enum the switch will fail to compile unless we handle it.
It might make it more obvious that we're covering all bases though.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -361,6 +436,44 @@ private List<String> getFullNamesFromCredentials(List<IdentityClaim> identityCla
                 .toList();
     }
 
+    private List<String> getGivenNames(List<IdentityClaim> identityClaims) {
+        return identityClaims.stream()
```
sam803: Would we be able to reduce code duplication here by having a method e.g. 

```
private Stream getProperty(List<IdentityClaim> identityClaims, String propertyName) {
... 
}
```

and then use it in `getFamilyNameForCoiCheck` and `getGivenNames`

sam803 (dismissed):
LGTM

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Wynndow: Sounds good, will update

sam803 (approved):

PR #1926, opened 2024-05-16
PYIC-6397: Add back event to update-details page

Wynndow:
### What changed

 Add back event to update-details page. See the core-front PR that makes use of it here: https://github.com/govuk-one-login/ipv-core-front/pull/1447

### Why did it change

The update details page now has a back link. That link should take the
user back to the reuse landing page.

There are other ways to get to the update details page, but they are all
"back" events from pages a user would reach directly after the update
details page, so we don't want the back link to ever return the user to
those pages.

### Issue tracking

- [PYIC-6397](https://govukverify.atlassian.net/browse/PYIC-6397)

[PYIC-6397]: https://govukverify.atlassian.net/browse/PYIC-6397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reuse-existing-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reuse-existing-identity.yaml
@@ -138,6 +138,8 @@ states:
         targetState: UPDATE_NAME_DOB_PAGE
       cancel:
         targetState: RETURN_TO_RP
+      back:
+        targetState: IDENTITY_REUSE_PAGE_COI
 
   DETAILS_DELETED_PAGE:
     response:


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reuse-existing-identity.yaml
```diff
@@ -138,6 +138,8 @@ states:
         targetState: UPDATE_NAME_DOB_PAGE
       cancel:
         targetState: RETURN_TO_RP
+      back:
+        targetState: IDENTITY_REUSE_PAGE_COI
```
MikeCollingwood: I think a few states can lead to the `UPDATE_DETAILS_PAGE` state

sam803: I think the user could also have come from a some other places e.g. https://github.com/govuk-one-login/ipv-core-back/blob/706eb2d215ba45e5f238a1723d3abdd5082cd64e/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml#L67 

Is there a way to determine that we should go back to a different place?

sam803: Looks like they all are as a result of clicking a 'go back' option themselves so its probably ok for the 'back' here to go back to IDENTITY_REUSE_PAGE_COI

Wynndow: Bah. Fix incoming

sam803 (approved):

PR #1931, opened 2024-05-20
PYIC-6057: Add f2f volumes routes

MikeCollingwood:
### What changed

- Added routes from passport or driving licence cris to new pages designed to reduce f2f volumes

### Why did it change

- To reduce f2f volumes

### Issue tracking

- [PYIC-6057](https://govukverify.atlassian.net/browse/PYIC-6057)

Screenshot 2024-05-20 at 13 13 13

[PYIC-6057]: https://govukverify.atlassian.net/browse/PYIC-6057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -566,7 +566,8 @@ states:
       f2f:
         targetState: F2F_PYI_POST_OFFICE
       returnToRp:
-        targetState: RETURN_TO_RP
+        targetJourney: INELIGIBLE
+        targetState: INELIGIBLE_SKIP_MESSAGE
 
   ADDRESS_AND_FRAUD_J2:
     nestedJourney: ADDRESS_AND_FRAUD
@@ -613,7 +614,8 @@ states:
       f2f:
         targetState: F2F_PYI_POST_OFFICE
       returnToRp:
-        targetState: RETURN_TO_RP
+        targetJourney: INELIGIBLE
+        targetState: INELIGIBLE_SKIP_MESSAGE
 
   ADDRESS_AND_FRAUD_J3:
     nestedJourney: ADDRESS_AND_FRAUD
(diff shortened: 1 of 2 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -549,9 +549,25 @@ states:
...
+      type: page
+      pageId: prove-identity-another-type-photo-id
+      context: passport
+    events:
+      otherPhotoId:
+        targetState: CRI_DRIVING_LICENCE_J3
+      f2f:
+        targetState: F2F_PYI_POST_OFFICE
+      returnToRp:
+        targetState: RETURN_TO_RP
```
Wynndow: These returnToRp events will need to go via TICF before sending the user back to the service. I think you can use the `INELIGIBLE/INELIGIBLE_SKIP_MESSAGE` state for where a user doesn't have a CI

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -1253,6 +1285,20 @@ states:
...
+  MITIGATION_PP_PROVE_ANOTHER_WAY:
+    response:
+      type: page
+      pageId: prove-identity-no-other-photo-id
+      context: passport
+    events:
+      back:
+        targetState: MITIGATION_PP_CRI_UK_PASSPORT
+      returnToRp:
+        targetState: RETURN_TO_RP
```
Wynndow: The returnToRp events where a user is mitigationg will also need to go to TICF. I think you might need to create a new state similar to `INELIGIBLE_SKIP_MESSAGE`, but in the failed sub journey.

MikeCollingwood: Would INELIGIBLE/INELIGIBLE_SKIP_MESSAGE not work? I see the state already fails if there's a ci - which is what we'd want since there's a doubling-up.

Wynndow: I think it's about the difference between being ineligble and failing. In this scenario the user has a CI they can't mitigate, so they've failed.

Wynndow (approved):
Noice

PR #1932, opened 2024-05-20
PYIC-6424: Read from long term store to build P2 identity

Wynndow:
### What changed

 Read from long term store to build P2 identity

### Why did it change

When a user is updating their details on the COI journeys they can fail without a CI or abandon their journey in the DCMAW CRI. When the user started the COI journey we removed all VCs from the session credentials store except their address VC. We also have already set their VOT in therr session as P2 due to them already having an identity.

The result of the above scenario is the user is returned to the service with a P2, but only an address VC. This will be rejected by SPOT.

It would be wrong to return them with anything other than a P2, as they still have a succesful identity.

We could detect this scenario and reinstate the required VCs to the session credential store. This may be quite complex though, maybe requiring another lambda.

Alternatively we can update the build-user-identity lambda to use VCs from the long term store when building a P2 identity. This ensures we're only returning VCs that have been checked and confirmed to form an identity.

For P0, PCL200 and PCL250's we can still use the session store. This will mean we don't have to do any filtering of fetched VCs.

We may in future revisit this decision and reimplement it, but it allows for the COI work to progress.

### Issue tracking

- [PYIC-6424](https://govukverify.atlassian.net/browse/PYIC-6424)

[PYIC-6424]: https://govukverify.atlassian.net/browse/PYIC-6424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -188,7 +188,7 @@
         "filename": "deploy/template.yaml",
         "hashed_secret": "6afab4c634af2dd2b9c344a98f96667277c56df0",
         "is_verified": false,
-        "line_number": 2021
+        "line_number": 2024
       }
     ],
     "lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/pact/BuildUserIdentityHandlerTest.java": [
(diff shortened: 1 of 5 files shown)


MikeCollingwood (approved):
Looks good! Will be interesting to see where this might go (using long term storage only, e.g.)

PR #1939, opened 2024-05-22
BAU: Fix back after repeat fraud check

Wynndow:
### What changed

Fix back after repeat fraud check

### Why did it change

When a user was on a 6MFC journey they could choose to go back to check their details again, after having said they wanted to change their details.

That was routing them to the COI journey start, which is incorrect.

This introduces a couple of extra states so we can handle the back event correctly.


thebauSoftwire (approved):
thanks!

PR #1941, opened 2024-05-23
PYIC-6088: handle mfa reset request

sam803:
### What changed

- Updates to the Initialise Ipv session handler to check the scope of the received jar and initialise a mfa reset journey if the scope equals "reverification" 
- Additional validation of the journey to the client using the new config parameter `validJourneyTypes`

### Why did it change

When we receive the new JAR from MFA reset, we need to identify this and start an MFA reset journey.

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6088

### DIFF
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -63,6 +63,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static uk.gov.di.ipv.core.initialiseipvsession.domain.ScopeConstants.REVERIFICATION;
+import static uk.gov.di.ipv.core.initialiseipvsession.domain.ScopeConstants.SCOPE;
 import static uk.gov.di.ipv.core.initialiseipvsession.validation.JarValidator.CLAIMS_CLAIM;
 import static uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsIpvJourneyStart.REPROVE_IDENTITY_KEY;
 import static uk.gov.di.ipv.core.library.auditing.helpers.AuditExtensionsHelper.getExtensionsForAudit;
@@ -81,8 +83,6 @@ public class InitialiseIpvSessionHandler
     private static final String REQUEST_GOV_UK_SIGN_IN_JOURNEY_ID_KEY = "govuk_signin_journey_id";
     private static final String REQUEST_EMAIL_ADDRESS_KEY = "email_address";
     private static final String REQUEST_VTR_KEY = "vtr";
-    private static final String REQUEST_SCOPE_KEY = "scope";
-    private static final String REVERIFICATION_SCOPE = "reverification";
     private static final List<Vot> HMRC_PROFILES_BY_STRENGTH = List.of(Vot.PCL250, Vot.PCL200);
     private static final ErrorObject INVALID_INHERITED_IDENTITY_ERROR_OBJECT =
             new ErrorObject("invalid_inherited_identity");
@@ -172,8 +172,7 @@ public class InitialiseIpvSessionHandler
             List<String> vtr = claimsSet.getStringListClaim(REQUEST_VTR_KEY);
 
             var isReverification =
-                    Scope.parse(claimsSet.getStringClaim(REQUEST_SCOPE_KEY))
-                            .contains(REVERIFICATION_SCOPE);
+                    Scope.parse(claimsSet.getStringClaim(SCOPE)).contains(REVERIFICATION);
             if (!isReverification && isListEmpty(vtr)) {
                 LOGGER.error(LogHelper.buildLogMessage(ErrorResponse.MISSING_VTR.getMessage()));
                 return ApiGatewayResponseGenerator.proxyJsonResponse(
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -212,17 +212,17 @@ public class InitialiseIpvSessionHandler
                             govukSigninJourneyId,
                             ipAddress);
 
-            Optional<InheritedIdentityJwtClaim> inheritedIdentityJwtClaim;
-            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)
-                    && (inheritedIdentityJwtClaim = getInheritedIdentityClaim(claimsSet))
-                            .isPresent()) {
-                validateAndStoreHMRCInheritedIdentity(
-                        clientOAuthSessionItem.getUserId(),
-                        inheritedIdentityJwtClaim.get(),
-                        claimsSet,
-                        ipvSessionItem,
-                        auditEventUser,
-                        deviceInformation);
+            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)) {
+                var inheritedIdentityJwtClaim = getInheritedIdentityClaim(claimsSet);
+                if (inheritedIdentityJwtClaim.isPresent()) {
+                    validateAndStoreHMRCInheritedIdentity(
+                            clientOAuthSessionItem.getUserId(),
+                            inheritedIdentityJwtClaim.get(),
+                            claimsSet,
+                            ipvSessionItem,
+                            auditEventUser,
+                            deviceInformation);
+                }
             }
 
             Boolean reproveIdentity =
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -212,17 +212,17 @@ public class InitialiseIpvSessionHandler
                             govukSigninJourneyId,
                             ipAddress);
 
-            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)) {
-                var inheritedIdentityJwtClaim = getInheritedIdentityClaim(claimsSet);
-                if (inheritedIdentityJwtClaim.isPresent()) {
-                    validateAndStoreHMRCInheritedIdentity(
-                            clientOAuthSessionItem.getUserId(),
-                            inheritedIdentityJwtClaim.get(),
-                            claimsSet,
-                            ipvSessionItem,
-                            auditEventUser,
-                            deviceInformation);
-                }
+            Optional<InheritedIdentityJwtClaim> inheritedIdentityJwtClaim;
+            if (configService.enabled(CoreFeatureFlag.INHERITED_IDENTITY)
+                    && (inheritedIdentityJwtClaim = getInheritedIdentityClaim(claimsSet))
+                            .isPresent()) {
+                validateAndStoreHMRCInheritedIdentity(
+                        clientOAuthSessionItem.getUserId(),
+                        inheritedIdentityJwtClaim.get(),
+                        claimsSet,
+                        ipvSessionItem,
+                        auditEventUser,
+                        deviceInformation);
             }
 
             Boolean reproveIdentity =
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -327,7 +327,7 @@ public class InitialiseIpvSessionHandler
                 : null;
     }
 
-    private static Boolean isListEmpty(List<String> list) {
+    private static boolean isListEmpty(List<String> list) {
         return list == null || list.isEmpty() || list.stream().allMatch(String::isEmpty);
     }
 
--- a/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
+++ b/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
@@ -171,8 +171,8 @@ public class InitialiseIpvSessionHandler
 
             List<String> vtr = claimsSet.getStringListClaim(REQUEST_VTR_KEY);
 
-            String requestScope = claimsSet.getStringClaim(REQUEST_SCOPE_KEY);
-            isReverification = REVERIFICATION_SCOPE.equals(requestScope);
+            isReverification =
+                    REVERIFICATION_SCOPE.equals(claimsSet.getStringClaim(REQUEST_SCOPE_KEY));
             if (!isReverification && isListEmpty(vtr)) {
                 LOGGER.error(LogHelper.buildLogMessage(ErrorResponse.MISSING_VTR.getMessage()));
                 return ApiGatewayResponseGenerator.proxyJsonResponse(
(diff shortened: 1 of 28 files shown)


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -166,7 +172,11 @@ public APIGatewayProxyResponseEvent handleRequest(
             LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
 
             List<String> vtr = claimsSet.getStringListClaim(REQUEST_VTR_KEY);
-            if (vtr == null || vtr.isEmpty() || vtr.stream().allMatch(String::isEmpty)) {
+
+            String scope = sessionParams.get(REQUEST_SCOPE_KEY);
+            isReverification = "reverification".equals(scope);
```
MikeCollingwood: Could this be a defined constant somewhere?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -175,7 +184,19 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                    ipvSessionService.generateIpvSession(clientOAuthSessionId, null, emailAddress);
+                    ipvSessionService.generateIpvSession(
+                            clientOAuthSessionId, null, emailAddress, isReverification);
+
+            // validate that client is allowed to initiate a journey of this type
+            error = validateJourneyType(ipvSessionItem.getJourneyType().getType(), clientId);
+            if (error.isPresent()) {
+                LOGGER.error(
+                        LogHelper.buildErrorMessage(
+                                "Validation of the journey type failed.",
```
MikeCollingwood: Could this error message say the something closer to `"User not allowed to initiate this journey type: " + ipvSessionItem.getJourneyType().getType()` maybe.

Maybe the error variable could also be renamed to distinguish it from a generic error. Maybe `forbiddenJourneyError`, I'm not sure though.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -444,6 +471,14 @@ private void sendInheritedIdentityReceivedAuditEvent(
         }
     }
 
+    private Optional<ErrorResponse> validateJourneyType(String type, String clientId) {
```
MikeCollingwood: It might clarify the method to append "ForClient" to the end of this method name.

Also, if the journeyTypes is null, I think we'd still want to error

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -175,7 +184,19 @@ public APIGatewayProxyResponseEvent handleRequest(
             String clientOAuthSessionId = SecureTokenHelper.getInstance().generate();
 
             IpvSessionItem ipvSessionItem =
-                    ipvSessionService.generateIpvSession(clientOAuthSessionId, null, emailAddress);
+                    ipvSessionService.generateIpvSession(
+                            clientOAuthSessionId, null, emailAddress, isReverification);
```
MikeCollingwood, line 186: I'm not sure if it's there's a workaround to get onto a prohibited journey but could make more sense to put validation before this, and use the isReverification for the validation instead

MikeCollingwood (commented):
This makes sense, the validation of the journey type then generating the ipv session with a different journey type. I like the refactoring too. I did have some Qs, mostly message-based but some on the validation too

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -166,7 +172,10 @@ public APIGatewayProxyResponseEvent handleRequest(
             LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
 
             List<String> vtr = claimsSet.getStringListClaim(REQUEST_VTR_KEY);
-            if (vtr == null || vtr.isEmpty() || vtr.stream().allMatch(String::isEmpty)) {
+
+            String scope = sessionParams.get(REQUEST_SCOPE_KEY);
```
Wynndow: Same as above - we shoould really be using the scope from the claims set, rather than the query param

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -156,6 +160,8 @@ public APIGatewayProxyResponseEvent handleRequest(
                     jarValidator.decryptJWE(
                             JWEObject.parse(sessionParams.get(REQUEST_PARAM_KEY)),
                             configService.getSsmParameter(JAR_KMS_ENCRYPTION_KEY_ID));
+
+            String clientId = sessionParams.get(CLIENT_ID_PARAM_KEY);
```
Wynndow: We get the client_id as a query param, but it will also be coming in the claims set inside the JAR. When we have the value in the JAR we should be using it from there, as then we have the security that it's signed by the client.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: It would also be good to store the scope on the users clientOauthSessionItem. That way it will be available later in the journey checks.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: I think it might be better to validate that the client is allowed to request the specified scope. This might require some changes to the config that we've set up.
Interestingly the scope can contain more than one value - a space separated string.
The two scopes that we care about are `openid` and `reverification`. We should always receive one of those. I'd suggest we make the validation logic use those, rather than adding a new `type` field to the `IpvJourneyTypes` enum

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
```diff
@@ -10,6 +10,7 @@ public enum ConfigurationVariable {
     CI_CONFIG("self/ci-config"),
     CI_SCORING_THRESHOLD("self/ciScoringThreshold"),
     CLIENT_ISSUER("clients/%s/issuer"),
+    CLIENT_VALID_JOURNEY_TYPES("clients/%s/validJourneyTypes"),
```
Wynndow: As above - I think maybe we should change this to `validScopes`

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
sam803: hmm yes I think I need to look at a way to do this as currently the generateIpvSession method will store the session even if the journey is invalid, maybe will move the logic to the IPSessionService

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
sam803: Yes sure - I generally only use constants if its used more than once but not a hill I'd die on :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -98,7 +101,11 @@ public IpvSessionItem generateIpvSession(
         ipvSessionItem.setVot(Vot.P0);
 
         if (errorObject == null) {
-            ipvSessionItem.setJourneyType(IpvJourneyTypes.INITIAL_JOURNEY_SELECTION);
+            if (isReverification) {
+                ipvSessionItem.setJourneyType(IpvJourneyTypes.MFA_RESET);
+            } else {
+                ipvSessionItem.setJourneyType(IpvJourneyTypes.INITIAL_JOURNEY_SELECTION);
+            }
```
Wynndow:
```suggestion
            ipvSessionItem.setJourneyType(isReverification ? IpvJourneyTypes.MFA_RESET : IpvJourneyTypes.INITIAL_JOURNEY_SELECTION)
```

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
sam803: @Wynndow is it definitely going to be in the claims set ? This [PR](https://github.com/govuk-one-login/ipv-stubs/pull/809/files) only adds it to the request, so might need to add it there as well

Wynndow: Yep - the claims set will always contain it. I think @thebauSoftwire might even be adding validation that it does.
And yep, it'll need adding to the stub. I'm working on an orch stub PR at the moment so can add it.

sam803: thanks, if you haven't already done it theres this pr https://github.com/govuk-one-login/ipv-stubs/pull/814

Wynndow: No, all good, approved

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
sam803: I've refactored to validate the scope rather than the journey - this way we can validate it before we get here :)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -87,6 +89,46 @@ public JWTClaimsSet validateRequestJwt(SignedJWT signedJWT, String clientId)
...
+                                .with(
+                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                        "client not allowed to issue a request with this scope")
+                                .with(LOG_SCOPE.getFieldName(), scope)
+                                .with(LOG_CLIENT_ID.getFieldName(), clientId));
+                throw new JarValidationException(
+                        OAuth2Error.ACCESS_DENIED.setDescription(
+                                "Client not allowed to issue a request with this scope"));
+            }
+            LogHelper.attachClientIdToLogs(clientId);
```
MikeCollingwood: Very minor thoughts:
- Could be good to attach client id at the start so don't need to manually add them in
- Log messages could have commonly capitalised first letter

MikeCollingwood (dismissed):
This looks good and makes sense!

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -87,6 +89,46 @@ public JWTClaimsSet validateRequestJwt(SignedJWT signedJWT, String clientId)
         }
     }
 
+    private void validateScope(String clientId, JWTClaimsSet claimsSet)
+            throws JarValidationException {
+        try {
+            String scope = claimsSet.getStringClaim("scope");
+            var validScopes = configService.getValidClientScopes(clientId);
+            if (validScopes.isEmpty() || !validScopes.contains(scope)) {
```
Wynndow: I think there's another case here worth checking. We would never expect to receive `openid` and `reverification` in the scope - it can be a space separated (I think) list of scopes.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/mfa-reset.yaml
Joe-Edwards-GDS, line 1: Do you think this is going to be specific to MFA reset, or a general 'reverification' journey?

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -87,6 +94,62 @@ public JWTClaimsSet validateRequestJwt(SignedJWT signedJWT, String clientId)
...
+                != requestedScopes.stream()
+                        .filter(scope -> ONE_OF_REQUIRED_SCOPES.contains(scope.getValue()))
+                        .count();
+    }
+
+    private boolean scopeNotValidForClient(Scope requestedScopes, String clientId) {
+        var validClientScopes =
+                Scope.parse(configService.getSsmParameter(CLIENT_VALID_SCOPES, clientId));
+        return !requestedScopes.stream()
+                .allMatch(scope -> validClientScopes.contains(scope.getValue()));
```
Joe-Edwards-GDS, line 151: Fine to leave as-is, but as a minor style thing, I generally try to keep methods 'positive' and include a verb - e.g. `hasRequiredScope` and `hasValidScopesForClient`.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
```diff
@@ -29,19 +30,25 @@
...
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_REDIRECT_URI;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_SCOPE;
 
 public class JarValidator {
+    public static final String CLAIMS_CLAIM = "claims";
+
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String REDIRECT_URI_CLAIM = "redirect_uri";
-    public static final String CLAIMS_CLAIM = "claims";
+    private static final List<String> ONE_OF_REQUIRED_SCOPES = List.of("openid", "reverification");
```
Joe-Edwards-GDS: (Should these scopes be constants somewhere - `reverification` is also used in `InitialiseIpvSessionHandler`)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/mfa-reset.yaml
Wynndow: Better to keep it generic if we think we might reuse it - i'll change it

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
Wynndow: No, that's much better. My brain sometimes has a meltdown when dealing with double negatives so will fix it up

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/validation/JarValidator.java
Wynndow: Yeah, that's a good idea

Joe-Edwards-GDS (approved):

PR #1945, opened 2024-05-23
BAU: Fix error codes

AmritSidhu:
### What changed

Fix error codes
When rebasing a previous PR the numbers here got repeated: https://github.com/govuk-one-login/ipv-core-back/pull/1918


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -90,8 +90,8 @@ public enum ErrorResponse {
     INVALID_JOURNEY_EVENT(1073, "Invalid journey event in input"),
     FAILED_TO_STORE_IDENTITY(1074, "Failed to store identity"),
     MITIGATION_ROUTE_CONFIG_NOT_FOUND(
-            1069, "No mitigation journey route event found in cimit config"),
```
DanCorderIPV, line 93: 1069 isn't used above is the gap deliberate?

AmritSidhu: Im seeing gaps inbetween a bunch of the older error codes, not sure if its done for a reason or just errors have come and gone, it would make sense that new errors need to be unique for logging sake

DanCorderIPV (approved):

PR #1948, opened 2024-05-24
PYIC-6501 Created a new EvcsService to create, update or read user VCs from EVCS store

rpayal:
## Proposed changes
Create a new EvcsService to interact with EVCS to create/update/get user VCs
### What changed
Created a new EvcsService to create, update or read user VCs from EVCS store

### Issue tracking

- [PYIC-6501](https://govukverify.atlassian.net/browse/PYIC-6501)

### Other considerations

[PYIC-6501]: https://govukverify.atlassian.net/browse/PYIC-6501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
Joe-Edwards-GDS, line 1: Do you think this is mixing levels of abstraction?

In the old world we have a distinction between the data and service layers:
- `DataStore` is the data layer, and handles interaction with the underlying store, with basic CRUD methods like `create` and `get`
- `VerifiableCredentialService` is the service layer, and handles abstracting the data store into domain-specific methods and objects, e.g. `persistUserCredentials` and `getCredentials` which returns VC objects.

I wonder if we should follow a similar pattern here, with one class (say `EncryptedVcsClient`) which handles the low-level interaction with the EVCS API, and the orchestration logic in a separate class.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -367,7 +369,8 @@ static VerifiableCredential vcPassportMissingName() {
                 TestVc.builder()
                         .credentialSubject(credentialSubject)
                         .evidence(SUCCESSFUL_EVIDENCE)
-                        .build());
+                        .build(),
+                (String) null);
```
Joe-Edwards-GDS: I'm not sure about this
- Specifying `(String) null` looks pretty sketchy - if it's common to pass null, then maybe we should examine which overloads we support to avoid them conflicting.
- Is it actually valid to have a `null` issuer - should we set this to some test value?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -176,6 +176,10 @@ public String getCriPrivateApiKeyForActiveConnection(String criId) {
         return getApiKeyFromSecretManager(criId, getActiveConnection(criId));
     }
 
+    public String getAppApiKey(String appId) {
```
Joe-Edwards-GDS, line 179: Is this going to be reusable? Could it just be `getCoreSecretValue("evcs/api-key")`?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,335 @@
...
+    public static final String X_API_KEY_HEADER = "x-api-key";
+    public static final String AUTHORISATION_HEADER_KEY = "Authorisation";
+    public static final String VC_STATE_PARAM = "state";
+    public static final List<String> CRIS_VC_TO_UPDATE = List.of(ADDRESS_CRI, EXPERIAN_FRAUD_CRI);
+    public static final String APPLICATION_JSON_CHARSET_UTF_8 = "application/json; charset=utf-8";
+    public static final String CONTENT_TYPE = "Content-Type";
+
+    private final HttpClient httpClient;
+    private final ConfigService configService;
+    private final CriResponseService criResponseService;
```
Joe-Edwards-GDS: I don't think this service should know about CRI responses - see comment about abstraction. If we need to vary behaviour, they should probably be two different methods, e.g. `storeIdentity` and `storePendingIdentity`

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,335 @@
...
+        this.criResponseService = criResponseService;
+    }
+
+    @Tracing
+    public void storeIdentity(
+            String userId, List<VerifiableCredential> credentials, boolean is6MFCJourney)
+            throws EvcsServiceException {
+        List<EvcsVCState> vcStatesToQueryFor = new ArrayList<>();
+        vcStatesToQueryFor.add(EvcsVCState.CURRENT);
+        vcStatesToQueryFor.add(EvcsVCState.PENDING_RETURN);
```
Joe-Edwards-GDS: `List.of(CURRENT, PENDING_RETURN)`?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,335 @@
...
+                                                        : EvcsVCState.CURRENT,
+                                                null,
+                                                null))
+                        .toList();
+
+        if (CollectionUtils.isEmpty(existingEvcsUserVCs)
+                || (!CollectionUtils.isEmpty(existingEvcsUserVCs) && isF2FIncomplete)) {
+            // AC1 and AC5
+            createEvcsUserVCs(userId, userVCsForEvcs);
+        } else if (is6MFCJourney) {
```
Joe-Edwards-GDS: As discussed on the ticket, I'm not sure this should be a separate case here - we should just be comparing the VCs in the current session against the VCs in EVCS.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
```diff
@@ -16,6 +16,9 @@ public enum ConfigurationVariable {
     CORE_VTM_CLAIM("self/coreVtmClaim"),
     CREDENTIAL_ISSUERS("credentialIssuers"),
     CRI_RESPONSE_TTL("self/criResponseTtl"),
+    EVCS_APP_ID("evcs"),
+    EVCS_APPLICATION_URL("evcs/application-url"),
+    EVCS_ACCESS_TOKEN("evcs/accessToken"),
```
Wynndow: Are these right? Feels like `EVCS_APP_ID` would need a nested path?
Also, maybe worth using snakeCase for the application URL to be consistant.
And I'm not sure that EVCS_ACCESS_TOKEN is used - we'll be getting that from the JAR

libs/evcs-service/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+	id "jacoco"
+}
+
+repositories {
+	mavenCentral()
+}
+
+dependencies {
+	implementation libs.jacksonDatabind,
+			libs.notificationsJavaClient,
```
Wynndow: Does this need the notifications client?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: Separated evcs api interaction from evcsService to separate evcsClient.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
rpayal: Not needed...reverted

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: can be reused for another application to access its api's.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private URI getUri(String userId, List<EvcsVCState> vcStatesToQueryFor)
+            throws URISyntaxException {
+        if (vcStatesToQueryFor != null) {
+            return new URIBuilder(
+                            configService
+                                    .getSsmParameter(EVCS_APPLICATION_URL)
+                                    .concat("/")
+                                    .concat(userId))
```
Wynndow:
```suggestion
            return new URIBuilder(
                            configService
                                    .getSsmParameter(EVCS_APPLICATION_URL))
                            .setPath(userId)
```

Wynndow: I'm also wondering if we should be URL encoding the userId. It will contain a bunch of colons which could be confusing. I suppose it depends what the actual EVCS is expecting to receive though...

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+    public EvcsGetUserVCsDto getUserVcs(
+            String userId, String evcsAccessToken, List<EvcsVCState> vcStatesToQueryFor)
+            throws EvcsServiceException {
+        try {
+            HttpRequest.Builder httpRequestBuilder =
+                    HttpRequest.newBuilder().uri(getUri(userId, vcStatesToQueryFor)).GET();
+            httpRequestBuilder.header(
+                    X_API_KEY_HEADER, configService.getAppApiKey(EVCS_APP_ID.getPath()));
+            httpRequestBuilder.header(AUTHORISATION_HEADER_KEY, "Bearer " + evcsAccessToken);
+            httpRequestBuilder.header(CONTENT_TYPE, APPLICATION_JSON_CHARSET_UTF_8);
```
Wynndow:
```suggestion
            HttpRequest.Builder httpRequestBuilder =
                    HttpRequest.newBuilder()
                            .GET()
                            .uri(getUri(userId, vcStatesToQueryFor))
                            .header(
                                    X_API_KEY_HEADER,
                                    configService.getAppApiKey(EVCS_APP_ID.getPath()))
                            .header(AUTHORISATION_HEADER_KEY, "Bearer " + evcsAccessToken)
                            .header(CONTENT_TYPE, APPLICATION_JSON_CHARSET_UTF_8);
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Tracing
+    public void storeIdentity(
+            String userId,
+            List<VerifiableCredential> credentials,
+            String evcsAccessToken,
+            boolean isF2FIncomplete)
+            throws EvcsServiceException {
+        List<EvcsVCState> vcStatesToQueryFor = List.of(CURRENT, PENDING_RETURN);
```
Wynndow: This could probably be made a static constant, and then just reused for every call to this method.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+            throws EvcsServiceException {
+        List<EvcsCreateUserVCsDto> userVCsForEvcs =
+                credentials.stream()
+                        .map(
+                                vc ->
+                                        new EvcsCreateUserVCsDto(
+                                                vc.getVcString(),
+                                                isF2FIncomplete ? PENDING_RETURN : CURRENT,
+                                                null,
+                                                null))
```
Wynndow, line 80: I think that this is the only place we're constructing a `EvcsCreateUserVCsDto`. And we're always setting two of the parameters to null - do we need them?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+                        .toList();
+
+        if (!CollectionUtils.isEmpty(existingEvcsUserVCs)
+                && (CollectionUtils.isEmpty(existingEvcsUserVCs) || !isF2FIncomplete)) {
```
Wynndow:
```suggestion
        if (!CollectionUtils.isEmpty(existingEvcsUserVCs) && !isF2FIncomplete) {
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+            List<EvcsUpdateUserVCsDto> existingCurrentEvcsUserVcsToUpdate =
```
Wynndow: using `var` for these might make this a bit cleaner

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EVCS_APP_ID;
+
+public class EvcsClient {
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    public static final String X_API_KEY_HEADER = "x-api-key";
+    public static final String AUTHORISATION_HEADER_KEY = "Authorisation";
+    public static final String VC_STATE_PARAM = "state";
+    public static final String APPLICATION_JSON_CHARSET_UTF_8 = "application/json; charset=utf-8";
+    public static final String CONTENT_TYPE = "Content-Type";
```
Joe-Edwards-GDS: If we're using the apache lib we could consider using their constants for some of these - e.g. `org.apache.http.HttpHeaders.CONTENT_TYPE`

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
Joe-Edwards-GDS: Do you need content-type for a GET with no body?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
Joe-Edwards-GDS: If we're using a URI builder we should be able to do avoid doing the escaping and concatenation operations ourselves...

A cleaner implementation might be:
```
    private URI getUri(String userId, List<EvcsVCState> vcStatesToQueryFor)
            throws URISyntaxException {
        var builder = new URIBuilder(configService.getSsmParameter(EVCS_APPLICATION_URL))
                .setPathSegments("vcs", userId);

        if (vcStatesToQueryFor != null) {
            builder
                    .addParameter(
                            VC_STATE_PARAM,
                            vcStatesToQueryFor.stream()
                                    .map(EvcsVCState::name)
                                    .collect(Collectors.joining(",")));
        }

        return builder.build();
    }
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+            throw new EvcsServiceException(
+                    HTTPResponse.SC_SERVER_ERROR, ErrorResponse.FAILED_TO_PARSE_EVCS_RESPONSE);
+        } catch (URISyntaxException e) {
+            throw new EvcsServiceException(
+                    HTTPResponse.SC_SERVER_ERROR, ErrorResponse.FAILED_TO_CONSTRUCT_EVCS_URI);
+        }
+    }
+
+    @Tracing
+    public void createEvcsUserVCs(String userId, List<EvcsCreateUserVCsDto> userVCsForEvcs)
```
Joe-Edwards-GDS: We probably don't need 'Evcs' in these method names - it's scoped by the class (e.g. `evcsService.createUserVcs(...)` is already unambiguous).

I'd also prefer a verb like store/persist/save over 'create' - the VCs are created by the CRIs, we're just storing them here.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+
+    private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
+            throws EvcsServiceException {
+        if (200 > evcsHttpResponse.statusCode() || evcsHttpResponse.statusCode() > 299) {
+            throw new EvcsServiceException(
+                    HTTPResponse.SC_SERVER_ERROR,
+                    ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE);
+        }
```
Joe-Edwards-GDS, line 167: I know this is copied from the TICF CRI, but can we include at least the status code in the error message, otherwise this will be impossible to debug.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -0,0 +1,191 @@
...
+                            .method(
+                                    "PATCH",
+                                    HttpRequest.BodyPublishers.ofString(
+                                            OBJECT_MAPPER.writeValueAsString(evcsUserVCsToUpdate)));
+            httpRequestBuilder.header(
+                    X_API_KEY_HEADER, configService.getAppApiKey(EVCS_APP_ID.getPath()));
+            httpRequestBuilder.header(CONTENT_TYPE, APPLICATION_JSON_CHARSET_UTF_8);
+
+            var evcsHttpResponse = sendHttpRequest(httpRequestBuilder.build());
+            checkResponseStatusCode(evcsHttpResponse);
```
Joe-Edwards-GDS: Should we move `checkResponseStatusCode` inside `sendHttpRequest`?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+    public EvcsService(EvcsClient evcsClient) {
+        this.evcsClient = evcsClient;
+    }
+
+    @Tracing
+    public void storeIdentity(
+            String userId,
+            List<VerifiableCredential> credentials,
+            String evcsAccessToken,
+            boolean isF2FIncomplete)
```
Joe-Edwards-GDS: Can we rename this to `isPendingIdentity` - or even better, create two separate methods for `storeCurrentIdentity` and `storePendingIdentity`?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
Joe-Edwards-GDS: I think this is probably ok - the fields are defined in the EVCS spec, but we're not using them at the moment.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Tracing
+    private void persistEvcsUserVCs(
+            String userId,
+            List<VerifiableCredential> credentials,
+            List<EvcsGetUserVCDto> existingEvcsUserVCs,
+            boolean isF2FIncomplete)
+            throws EvcsServiceException {
+        List<EvcsCreateUserVCsDto> userVCsForEvcs =
+                credentials.stream()
```
Joe-Edwards-GDS, line 64: I think we need to filter this list so we don't re-create VCs that already exist in EVCS

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+
+        if (!CollectionUtils.isEmpty(existingEvcsUserVCs)
+                && (CollectionUtils.isEmpty(existingEvcsUserVCs) || !isF2FIncomplete)) {
+            List<EvcsUpdateUserVCsDto> existingCurrentEvcsUserVcsToUpdate =
+                    existingEvcsUserVCs.stream()
+                            .filter(vc -> vc.state().equals(CURRENT))
+                            .map(
+                                    vc ->
+                                            new EvcsUpdateUserVCsDto(
+                                                    vc.vc().split("\\.")[2],
```
Joe-Edwards-GDS: (Might be worth pulling out a method for this - or leveraging other JWT-parsing utilities?)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+                                                    EvcsVCState.ABANDONED,
+                                                    null))
+                            .toList();
+            List<EvcsUpdateUserVCsDto> evcsUserVCsToUpdate =
+                    Stream.of(
+                                    existingCurrentEvcsUserVcsToUpdate,
+                                    existingPendingReturnEvcsUserVcsToUpdate)
+                            .flatMap(Collection::stream)
```
Joe-Edwards-GDS: If you skip calling `toList()` on the intermediate collections then you can just use `Stream.concat` here

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -0,0 +1,97 @@
...
+                                                isF2FIncomplete ? PENDING_RETURN : CURRENT,
+                                                null,
+                                                null))
+                        .toList();
+
+        if (!CollectionUtils.isEmpty(existingEvcsUserVCs)
+                && (CollectionUtils.isEmpty(existingEvcsUserVCs) || !isF2FIncomplete)) {
+            List<EvcsUpdateUserVCsDto> existingCurrentEvcsUserVcsToUpdate =
+                    existingEvcsUserVCs.stream()
+                            .filter(vc -> vc.state().equals(CURRENT))
```
Joe-Edwards-GDS, line 86: We need an additional filter here - if the CURRENT VCs are in the session, then we can leave them as-is.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
rpayal: noticed....updated and cleared n last commit.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: URIBuilder take care of encoding userId - tried one scenario working fine.

libs/evcs-service/build.gradle
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: cleaned

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: updated name

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: aded filter to check in session vcs before update.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: added filter to check in evcs vcs

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: just used in this method....done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -138,31 +142,27 @@ public void updateEvcsUserVCs(String userId, List<EvcsUpdateUserVCsDto> evcsUser
...
     }
 
     private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
             throws EvcsServiceException {
         if (200 > evcsHttpResponse.statusCode() || evcsHttpResponse.statusCode() > 299) {
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE.getMessage()
+                                    + String.format(
+                                            " (%s) received", evcsHttpResponse.statusCode())));
```
Joe-Edwards-GDS: An even better way of doing this would be to make it a log property:
```
LOGGER.info(
        LogHelper.buildLogMessage(
                        ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE
                                .getMessage())
                .with(LOG_STATUS_CODE.getFieldName(), evcsHttpResponse.statusCode()));
```
This would make it queryable, so we could (e.g.) easily get metrics on which status codes were being received.

Would you be able to add this to the corresponding place in the TICF service as well, for completeness?

Joe-Edwards-GDS (commented):
Looks like there are still a couple of unresolved comments (you might need to expand out the list), but we're basically there I think

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: will create two separate as -  storeCompletedIdentity() and storePendingIdentity()

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: also add log statement for TICF service

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: NA any more

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: done ...updated ticf service also for log statement.

Joe-Edwards-GDS: Thanks, that's really helpful!

Joe-Edwards-GDS (approved):
Thanks for the updates, I think this gives a good starting point for the rest of the EVCS work :)

PR #1953, opened 2024-05-28
PYIC-6486: Add reverification journey

MikeCollingwood:
### What changed

- Added mfa reset journey

### Why did it change

- To allow people to make mfa reset journeys

### Issue tracking

- [PYIC-6486](https://govukverify.atlassian.net/browse/PYIC-6486)

### Notes

- Does the entry point work with the dcmaw check?

### Verbose journey

1. Can only start if DCMAW and TICF enabled
2. page-ipv-identity-document-start (no issues here because triages if DCMAW possible)
3. Must do DCMAW (one day, STRATEGIC_APP_TRIAGE - dead end atm) otherwise INELIGIBLE > INELIGIBLE

This gives INELIGIBLE > CRI_TICF_BEFORE_ANOTHER_WAY (calls ticf) then pyi-another-way OR pyi-no-match if failed ticf

4. POST_DCMAW_SUCCESS_PAGE (or ineligible again, or failed if ci)
5. ADDRESS_AND_FRAUD (or failed: ticf, no-match)
6. EVALUATE_GPG45_SCORES (or failed: ticf, no-match)
7. CHECK_COI (or failed: ticf, no-match)
8. CRI_TICF (or FAILED_CONFIRM_DETAILS)
9. RETURN_TO_RP (or FAILED_NO_TICF)

### Notes

- I think I may need to use context for the check coi lambda because it would typically need to know the coi subjourney type, but we're not in one and we just need it to check both names and DOB

In testing:
```
https://dev-mikec.02.dev.identity.account.gov.uk/ipv/usefeatureset?featureSet=ticfCriBeta,mfaReset
```

```
https://orch-dev-mikec.02.core.dev.stubs.account.gov.uk/
```

[PYIC-6486]: https://govukverify.atlassian.net/browse/PYIC-6486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/mfa-reset.yaml
```diff
@@ -0,0 +1,189 @@
...
+        targetJourney: FAILED
+        targetState: FAILED
+      vcs-not-correlated:
+        targetJourney: FAILED
+        targetState: FAILED
+
+  CHECK_COI:
+    response:
+      type: process
+      lambda: check-coi
```
MikeCollingwood, line 166: Need to add context so it nows to try to correlate all VCs

MikeCollingwood:
So on this ticket getting merged, I'll:
- Need to rename to reverification.yaml
- Need to set the coiSubjourneyType when the state/ journey map are set

After this:
- We should look into removing coiSubjourneyType as a latent variable
- This will require us to expand out the UPDATE_NAME journey map
- We should also remove the coi check from UPDATE_ADDRESS because not needed
- The UPDATE_NAME journey map will need to separate flows for which name part is getting updated, then pass context to CheckCoi states.
- This could benefit from using a nested state for the two types of journey in UPDATE_NAME
- (Nested journeys should be fixed in [the deployed map diagram](https://govuk-one-login.github.io/ipv-core-back))
- This could additionally benefit from an enhancement where we use the session for "back". This is one of the few cases where sessions might be appropriate for handling routing. This might look like a stack of states which can be gone back through up to a CRI state.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
```diff
@@ -1,23 +1,189 @@
...
-
   START:
     events:
       next:
-        # should be filled out in https://govukverify.atlassian.net/browse/PYIC-6486
-        targetState: PLACEHOLDER
+        targetJourney: TECHNICAL_ERROR
+        targetState: ERROR_NO_TICF
+        checkFeatureFlag:
+          ticfCriBeta:
```
Joe-Edwards-GDS: That doesn't look like the right feature flag!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
```diff
@@ -1,23 +1,189 @@
...
+            targetState: IDENTITY_START_PAGE
+        checkIfDisabled:
```
Joe-Edwards-GDS: I think this should strictly be inside the `checkFeatureFlag`, although it's very unlikely to matter!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
```diff
@@ -1,23 +1,189 @@
...
+        targetJourney: INELIGIBLE
+        targetState: INELIGIBLE
+      access-denied:
+        targetJourney: INELIGIBLE
+        targetState: INELIGIBLE
+      temporarily-unavailable:
+        targetJourney: INELIGIBLE
+        targetState: INELIGIBLE
+
+  STRATEGIC_APP_TRIAGE:
```
Joe-Edwards-GDS: FYI @blakeyp - this will need keeping in line with any changes in the 'main' journey flow.

I'm wondering if we should just remove all the strategic app stuff here for now, and have a ticket in the strategic app epic for rolling out across other journey types when it's closer to completion?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
```diff
@@ -1,23 +1,189 @@
...
+        targetJourney: FAILED
+        targetState: FAILED
+
+  CHECK_COI:
+    response:
+      type: process
+      lambda: check-coi
+    events:
+      coi-check-passed:
+        targetState: CRI_TICF
```
Joe-Edwards-GDS: We probably still need to handle TICF being disabled, unless you're sure TICF is going live first?

Joe-Edwards-GDS (commented):
This looks ok (apart from the feature flag)

It will almost certainly need updating once the UCD and new screens are confirmed, but this will be a useful starting point.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
MikeCollingwood: This was intentional as the flow requires ticf & I believe you can only get onto this journey if the mfaReset feature flag is set - in IpvSessionService

Joe-Edwards-GDS: What do you mean by 'requires TICF', do you mean we shouldn't support this journey if TICF is unavailable?

MikeCollingwood: It looked like it was important in the MURAL design. Since I don't know when each is getting rolled out I thought that implied it was a dependency but I don't know the ordering of roll-outs so happy to make this flow inclusive

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
MikeCollingwood: Will remove and add the ticket to the epic

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reverification.yaml
Joe-Edwards-GDS: Yeah, I would treat it like the TICF CRI in all the other places.

I think it's meant to be enabled soon(TM) which will make this easier to manage!

Joe-Edwards-GDS (approved):
This is confirmed changing now (as mentioned on the ticket), but this will be a useful starting point and allow us to run through the journey now.

PR #1956, opened 2024-05-29
PYIC-6088: Fix scope validation

Wynndow:
### What changed

Fixes the scope validation issues that previously caused a P1 incident, and moves reverification logic behind a feature flag.

### Why did it change

[Revert "BAU: Remove scope checks"](https://github.com/govuk-one-login/ipv-core-back/commit/78b6b871de9dfefcbe418b6052a838fa0a6f147d)

This reverts commit https://github.com/govuk-one-login/ipv-core-back/commit/52abdbef5d3eb80d012deadb768e370c5e909b7a.
This was the fix to end the P1 incident. Bringing the changes back in so we can modify them.

[PYIC-6088: Fix scope validation](https://github.com/govuk-one-login/ipv-core-back/commit/85cc7ae9c0d6a2f26dd7a6948a4325292dbfa65e)

We were validating that all scopes provided were defined in config for a
given client. This was too strict. Orch gives us more than just
`openid`, which is the only scope we care about for them.

This updates the logic to ensure that only clients configured for openid
can use it, and only clients configured for reverification can use it.
If the client doesn't give us either of these we log it and allow an IPV
journey. This is due to the values in the scope being passed on to us
from them by orch, and we don't want to block unneccessarily. If we see
logs showing neither provided we can address this.

[PYIC-6088: Hide scope validation behind flag](https://github.com/govuk-one-login/ipv-core-back/commit/b118487d0b987b9828c8e3834a7df9c404b9cbf9)

This moves all the scope checking and validation behind the mfa reset
feature flag, so that we're only running the new code paths when we
want to.

### Issue tracking

- [PYIC-6088](https://govukverify.atlassian.net/browse/PYIC-6088)

[PYIC-6088]: https://govukverify.atlassian.net/browse/PYIC-6088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -76,6 +76,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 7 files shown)


Joe-Edwards-GDS (approved):
Thanks - maybe release tomorrow just in case, but this looks much safer :)
- We only fail if the client explicitly tries to do something they aren't allowed to do
- It's behind a feature flag

PR #1958, opened 2024-05-30
BAU: Move setting of COI journey type in session

Wynndow:
### What changed

Move setting of COI journey type in session

### Why did it change

@MikeCollingwood highlighted this as a potential issue.

The COI journey type was being set on the users session before we'd validated that the event was legitimate for the users current state.

The implication of this was that a user could potentially update there COI journey type by sending a new request in to the journey engine via the frontend.

The only thing preventing this from happening was that we were not saving the users session before throwing an error for an invalid event.

This feels a little risky - future updates could move the point at which we store the users sessiona and introduce a issue with the COI journey type.


MikeCollingwood (approved):
Looks good!

PR #1959, opened 2024-05-30
PYIC-6089: add external MFA reset endpoint

thebauSoftwire:
### What changed
- Adds new external `/reverification` endpoint to return success/fail response not containing any authentication details based on MFA reset journey.
- Refactors and adds unit tests.
- Adds endpoint to enable local-running
- Updates openAPI docs to include new endpoint
- Updates sam template.yaml to call `BuildUserIdentity` lambda with new endpoint

### Why did it change
New endpoint needed for new MFA reset journey to return MFA reset result without authentication details.

### Issue tracking

- [PYIC-6089](https://govukverify.atlassian.net/browse/PYIC-6089)

[PYIC-6089]: https://govukverify.atlassian.net/browse/PYIC-6089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

| Failed response | Success response |
| --------------- | ------------|
| Screenshot 2024-05-31 at 09 20 01 | Screenshot 2024-05-31 at 09 20 40 |


lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -61,6 +60,11 @@ public class BuildUserIdentityHandler
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String AUTHORIZATION_HEADER_KEY = "Authorization";
 
+    private static final String REVERIFICATION_SCOPE = "reverification";
```
Joe-Edwards-GDS: This will be new since you wrote this, but there are now constants for these: `uk.gov.di.ipv.core.initialiseipvsession.domain.ScopeConstants`

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -159,35 +163,17 @@ public APIGatewayProxyResponseEvent handleRequest(
                     ciMitService.getContraIndicatorsVc(
                             userId, clientOAuthSessionItem.getGovukSigninJourneyId(), null);
```
Joe-Edwards-GDS: I think this call to the CIMIT endpoint should happen inside the user-identity call - I don't think it's used for reverification?

(It can also be pretty slow!)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -205,6 +191,85 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            IpvSessionItem ipvSessionItem,
+            String userId,
+            AuditEventUser auditEventUser,
+            VerifiableCredential contraIndicatorsVc)
+            throws CiRetrievalException, VerifiableCredentialException, SqsException,
+                    HttpResponseExceptionWithErrorBody, CredentialParseException,
+                    UnrecognisedCiException {
+        if (!Arrays.asList(scopeClaims).contains(OPENID_SCOPE)) {
+            return getAccessDeniedApiGatewayProxyResponseEvent();
+        }
```
Joe-Edwards-GDS: I'd probably move this check out of this method - so the main handler method determines if the call is authorized (i.e. alongside the access token checks) and these methods handle the action.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -205,6 +191,85 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+
+        String ipvSessionId = ipvSessionItem.getIpvSessionId();
```
Joe-Edwards-GDS: We seem to be slightly inconsistent about var vs. specifying the type? I'd prefer using `var` except where it might be ambiguous

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
Joe-Edwards-GDS, line 1: I'm not totally sold on this being a single lambda to handle both endpoints - the meat of what they're doing seems to be quite different, and it's really just the access token validation that's shared?

Perhaps we should create a new class to handle the access token validation logic, which can be reused across two separate handlers?

I think you discussed this with @Wynndow , and don't want to waste a load of time changing it again, but curious about the rationale here. 🙂

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -22,13 +22,11 @@
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionsUserIdentity;
 import uk.gov.di.ipv.core.library.cimit.exception.CiRetrievalException;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
-import uk.gov.di.ipv.core.library.domain.AuditEventReturnCode;
-import uk.gov.di.ipv.core.library.domain.ContraIndicatorConfig;
-import uk.gov.di.ipv.core.library.domain.ContraIndicators;
-import uk.gov.di.ipv.core.library.domain.ErrorResponse;
-import uk.gov.di.ipv.core.library.domain.ReturnCode;
-import uk.gov.di.ipv.core.library.domain.UserIdentity;
+import uk.gov.di.ipv.core.library.domain.*;
```
Joe-Edwards-GDS: We tend not to use wildcard imports in our project - if this was IntelliJ, you can configure it to avoid doing this automatically! (If you can't find the setting let me know and I can try to dig it out)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/reverification/ReverificationSuccessResponse.java
Joe-Edwards-GDS, line 1: A more streamlined approach might be to use a single record class and factory methods for the variants:
```
@JsonInclude(NON_NULL)
public record ReverificationResponse(
        boolean success,
        String sub,
        @JsonProperty("error_code") String errorCode,
        @JsonProperty("error_description") String errorDescription
) {
    public static ReverificationResponse success(String sub) {
        return new ReverificationResponse(true, sub, null, null);
    }

    public static ReverificationResponse failure(String sub, String errorCode, String errorDescription) {
        return new ReverificationResponse(false, sub, errorCode, errorDescription);
    }
}
```
This doesn't have as strong a distinction between the different structures, so what you've got is also fine, up to you :)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
thebauSoftwire: out of curiosity, why do you have a preference for using `var` over an explicit type?

Joe-Edwards-GDS: That's a good question, and it's definitely a bit subjective, but for me the main reasons are:
- It clearly indicates a variable declaration and is more consistent with other languages
- It makes it easier to change/rename types later - e.g. recently we changed a bunch of code to pass `VerifiableCredential` rather than `SignedJWT`, and we had to change a bunch of intermediate variables that just added noise
- It's usually pretty obvious what the type is, either from context or by hovering over it in your IDE

Consider code like:
```
List<Credential> someCredentials = getSomeCredentials();
Credential specificCredential = findSpecificCredential(someCredentials);
doSomethingWith(specificCredential);
```
If you modify the name of the `Credential` type (or switch to a different class), you have to update these variable declarations, even though you aren't actually changing any logic in the method.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -205,6 +197,86 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+        if (ipvSessionItem.getVot().equals(Vot.P2)) {
+            ipvSessionService.revokeAccessToken(ipvSessionItem);
+            deleteSessionCredentials(ipvSessionId);
+
+            ReverificationResponse successResponse = ReverificationResponse.successResponse(userId);
+
+            return ApiGatewayResponseGenerator.proxyJsonResponse(
+                    HTTPResponse.SC_OK, successResponse);
+        } else {
+            ipvSessionService.revokeAccessToken(ipvSessionItem);
```
Joe-Edwards-GDS: The first two lines of both branches are identical, so we could pull them out.

However, I think it would be even better to move this to the end of the handler method instead (since it also happens in the user-identity endpoint) - we should invalidate the access token only once we're confident we'll return a proper response (i.e. 200).

I don't think the rest of this method can actually fail, but it would convey intent more clearly :)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -61,6 +66,11 @@ public class BuildUserIdentityHandler
     private static final Logger LOGGER = LogManager.getLogger();
     private static final String AUTHORIZATION_HEADER_KEY = "Authorization";
 
+    //    private static final String REVERIFICATION_SCOPE = "reverification";
```
Joe-Edwards-GDS: Remove this commented code?

deploy/template.yaml
```diff
@@ -1161,6 +1161,79 @@ Resources:
...
+                  "lambda:VpcIds":
+                    - Fn::ImportValue: !Sub ${VpcStackName}-VpcId
+        - KMSDecryptPolicy:
+            KeyId: !Ref DynamoDBKmsKey
+        - DynamoDBCrudPolicy:
+            TableName: !Ref SessionCredentialsTable
+        - DynamoDBReadPolicy:
+            TableName: !Ref SessionsTable
+        - DynamoDBWritePolicy:
+            TableName: !Ref SessionsTable
```
Joe-Edwards-GDS, line 1229: (Not important, but should these two be combined into one CrudPolicy?)

libs/pact-test-helpers/bin/test/dvlaVc/body.json
Joe-Edwards-GDS, line 1: Where has this come from?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -148,6 +139,13 @@ public APIGatewayProxyResponseEvent handleRequest(
                     clientOAuthSessionItem.getGovukSigninJourneyId());
 
             String userId = clientOAuthSessionItem.getUserId();
+
+            var scopeClaims = clientOAuthSessionItem.getScope().split(" ");
+            if (!input.getPath().contains(USER_IDENTITY_ENDPOINT)
+                    || !Arrays.asList(scopeClaims).contains(OPENID)) {
+                return ApiGatewayResponseGenerator.getAccessDeniedApiGatewayProxyResponseEvent();
+            }
```
Joe-Edwards-GDS: This will add some extra validation on the scope in prod, so out of an abundance of paranoia can we put this behind the MFA_RESET feature flag?

(You were away, but we had a P1 incident last week from the validation on `scope` we added at the start of the journey 😬 )

local-running/src/main/java/uk/gov/di/ipv/coreback/CoreBack.java
```diff
@@ -28,6 +28,8 @@ public CoreBack() throws IOException {
         Spark.post("/token", lambdaHandler.getToken());
         Spark.get("/user-identity", lambdaHandler.getUserIdentity());
 
+        Spark.get("/reverification", lambdaHandler.getUserIdentity());
```
Joe-Edwards-GDS: This probably needs updating now

openAPI/core-back-external.yaml
```diff
@@ -33,6 +33,29 @@ paths:
...
+          description: "403 Response"
+          content: {}
+      x-amazon-apigateway-integration:
+        type: "aws_proxy"
+        httpMethod: "POST"
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${BuildUserIdentityFunction.Arn}:live/invocations
+        passthroughBehavior: "when_no_match"
+
+  /reverification:
```
Joe-Edwards-GDS, line 46: I think the api gw integration needs updating here as well

Joe-Edwards-GDS (commented):
A few comments, but I think the separate lambda approach is good :)

libs/pact-test-helpers/bin/test/dvlaVc/body.json
sam803: hmm not sure must have commited by accident - will add to .gitignore

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/UserIdentityRequestHandler.java
```diff
@@ -0,0 +1,194 @@
...
+            SessionCredentialsService sessionCredentialsService) {
+        this.allowedPath = allowedPath;
+        this.allowedScope = allowedScope;
+        this.ipvSessionService = ipvSessionService;
+        this.configService = configService;
+        this.clientOAuthSessionDetailsService = clientOAuthSessionDetailsService;
+        this.sessionCredentialsService = sessionCredentialsService;
+    }
+
+    protected IpvSessionItem initialiseIpvSession(APIGatewayProxyRequestEvent input)
```
Joe-Edwards-GDS: I think this should be something like `validateAccessTokenAndGetIpvSession` (or similar) - the most important part of this is that it checks the access token, not that it retrieves the session.

(If you wanted to modularise this, you could put all these methods in an `AccessTokenValidationService` instead, but don't want to spend too much more time on this)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/UserIdentityRequestHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        String ipvSessionId = ipvSessionItem.getIpvSessionId();
+        LogHelper.attachIpvSessionIdToLogs(ipvSessionId);
+
+        return ipvSessionItem;
+    }
+
+    protected ClientOAuthSessionItem getClientOAuthSessionItem(
+            String urlPath, IpvSessionItem ipvSessionItem) throws InvalidScopeException {
```
Joe-Edwards-GDS: Now that we've got separate handlers, I don't think we need to check the `urlPath` - we can just store and validated the `allowedScope` directly.

(If we change the path of the endpoint in API GW, it shouldn't break this)

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/UserIdentityRequestHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        var scopeClaims = clientOAuthSessionItem.getScope().split(" ");
+        if (configService.enabled(MFA_RESET)
+                && (!urlPath.contains(this.allowedPath)
+                        || !Arrays.asList(scopeClaims).contains(this.allowedScope))) {
+            throw new InvalidScopeException();
+        }
+        return clientOAuthSessionItem;
+    }
+
+    protected void deleteSessionCredentials(String ipvSessionId) {
```
Joe-Edwards-GDS: I wonder if this should be combined with revoking the access token, and phrased more like 'close session' or similar 🤔

Not necessary

Joe-Edwards-GDS (commented):
A handful of comments but I think this is basically fine, so let me know and I will approve to get this merged and we can continue tweaking later if we want.

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/UserIdentityRequestHandler.java
sam803: I like it, will add

Joe-Edwards-GDS (dismissed):
Thanks for all the refactoring :)

Joe-Edwards-GDS (approved):

PR #1961, opened 2024-05-30
PYIC-6579: Remove COI check from address only journeys

Wynndow:
### What changed

Remove COI check from address only journeys

### Why did it change

If a user is updating just their address then there is no need to run a COI check. Their name and DOB won't have changed.

This also removes the actual check from the lambda and replaces it with some logging. We should never see that log.

This leaves the ADDRESS_ONLY enum in place - it's used by the reset session lambda to remove the correct VCs. This will probably get addressed in an upcoming ticket to use contexts instead of a value in session.

### Issue tracking

- [PYIC-6579](https://govukverify.atlassian.net/browse/PYIC-6579)

[PYIC-6579]: https://govukverify.atlassian.net/browse/PYIC-6579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
+++ b/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
@@ -104,7 +104,11 @@ public class CheckCoiHandler implements RequestHandler<ProcessRequest, Map<Strin
                                 .areFamilyNameAndDobCorrelatedForCoiCheck(credentials);
                         case FAMILY_NAME_ONLY, FAMILY_NAME_AND_ADDRESS -> userIdentityService
                                 .areGivenNamesAndDobCorrelated(credentials);
-                        case ADDRESS_ONLY -> userIdentityService.areVcsCorrelated(credentials);
+                        case ADDRESS_ONLY -> {
+                            LOGGER.warn(
+                                    LogHelper.buildLogMessage("Address only COI check requested"));
+                            yield true;
+                        }
                     };
 
             if (!successfulCheck) {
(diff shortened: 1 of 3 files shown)


lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -104,7 +104,11 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
                                 .areFamilyNameAndDobCorrelatedForCoiCheck(credentials);
                         case FAMILY_NAME_ONLY, FAMILY_NAME_AND_ADDRESS -> userIdentityService
                                 .areGivenNamesAndDobCorrelated(credentials);
-                        case ADDRESS_ONLY -> userIdentityService.areVcsCorrelated(credentials);
+                        case ADDRESS_ONLY -> {
+                            LOGGER.warn(
+                                    LogHelper.buildLogMessage("Address only COI check requested"));
+                            yield true;
```
Joe-Edwards-GDS, line 110: Look at you and your fancy new keywords ;)

Joe-Edwards-GDS (approved):

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
Wynndow: Second use in the repo, so not completely new. Old hat, in fact.

PR #1966, opened 2024-06-03
PYIC-5872: Expose core's internal API in dev

Wynndow:
### What changed

Expose core's internal API to requets with an API key.

### Why did it change

[PYIC-5872: Expose internal API in dev](https://github.com/govuk-one-login/ipv-core-back/pull/1966/commits/796c7896128268d063e020c4b9cc0bd1ba7ec859)

This change allows core's internal API to be accessed with an API key
when deployed to either of the dev accounts.

It adds a usage plan with API key to the internal API, and then makes
use of an API key required only for the dev accounts. It also adds DNS
to the API gateway to make use of the internal API easier.

Also, to allow programatic access to the API key, a new resource policy
is created to allow the OIDC identity provider enough permissions to
fetch the API key.

It removes the `CoreFrontLocal` parameter in favour of defaulting to
this behavour for dev accounts. This is because changing back to a
private API from the new regional version is a two step process - so not
as easy as just toggling the parameter.

It's a two step process as you can't change from regional to private if
it has custom domains associated with it. In CloudFormation there isn't
a way to specify that something should be removed before making changes,
resulting in having to remove the custom domain before changing the API
to private.

Ideally I would have liked to have conditionally created the UsagePlan -
only for the dev accounts. However there appears to be a bug in cfn-lint
that means we can't use an If function with the CreateUsagePlan
property.

The GHA workflow included in this PR is just to demonstrate that the
gateway is now available with an API key, and also how to
programatically access the API key. This would not be merged with the
PR.

[PYIC-5872: Don't use SAM UsagePlan](https://github.com/govuk-one-login/ipv-core-back/pull/1966/commits/e879e14f808012d0d299768a5a1d826f48ff7e13)

SAM allows an easy way to create a usage plan and API key, by defining
the `UsagePlan` property in an `Api` resource. Unfortunately you can't
use an `!If` function with it do conditionally create it.

This removes the use of it and defines the resources requied using
plain old CloudFormation. This allows us to only create them when we
want.

Using the previous method would have caused a usage plan and api key to
be created in all envs including prod. This felt like murky ground - the
non dev api's would still be private so it wouldn't have been an issue,
but feels safer to only create them when actually needed.

[PYIC-5872: Create separate API for internal testing](https://github.com/govuk-one-login/ipv-core-back/pull/1966/commits/5c8765d81953dff79d39bb5e8f680e18054b0976)

This creates a new API gateway to run in parallel with the original.
This means that the VPC protected internal API can still be used by
core-front without any changes, and the new regional, API token
protected API gateway can be used for testing.

cfn-lint has an issue with having an event defined in a lambda that
includes a reference to a conditionally created resource, even if that
resource is behind the same condition. This has lead to the ignore
written into the metadata of a number of resources.

### Issue tracking

- [PYIC-5872](https://govukverify.atlassian.net/browse/PYIC-5872)

[PYIC-5872]: https://govukverify.atlassian.net/browse/PYIC-5872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/dev-env-api-test.yaml
```diff
@@ -0,0 +1,36 @@
+name: Dev env internal API test
```
Wynndow: This is just hear to demonstrate the PR works, and how to access the API key programatically. Will be removed before merging the PR.

deploy/template.yaml
```diff
@@ -325,18 +310,20 @@ Resources:
...
         VPCEndpointIds:
-          - !If [UseLocalCoreFront, !Ref AWS::NoValue, Fn::ImportValue: !Sub "${VpcStackName}-ExecuteApiGatewayEndpointId"]
+          - !If [IsDevelopment, !Ref AWS::NoValue, Fn::ImportValue: !Sub "${VpcStackName}-ExecuteApiGatewayEndpointId"]
       DefinitionBody:
         openapi: "3.0.3" # workaround to get `sam validate` to work
         paths: # workaround to get `sam validate` to work
           /foo:
-            bar: baz # workaround to get `sam validate` to work
+            bar:
+              baz: thing # workaround to get `sam validate` to work
```
Joe-Edwards-GDS: Is this just to make it detect a change?

deploy/template.yaml
```diff
@@ -583,6 +608,11 @@ Resources:
 
   IssueClientAccessTokenFunction:
     Type: AWS::Serverless::Function
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks:
+            - W1001 # Ref to resource "IPVCoreInternalTestingApi" that may not be available when condition "IsDevelopment" is False
```
Joe-Edwards-GDS, line 614: This is a bit sad 😞 - I guess it's not clever enough to realise that the condition is the same

(I assume you've checked that it _isn't_ a problem - not actually sure which order it'll resolve everything!)

Joe-Edwards-GDS (commented):
This seems better to me, what do you think?

It makes the prod configuration cleaner, and isn't much more code than before. The main thing that's annoying is configuring the extra events and overriding that cfn-lint rule :/

deploy/template.yaml
Wynndow: I've not deployed to a non-dev environment to test it. But the validation failure breaks our existing workflows which is annoying. I think it's to do with how SAM gets transformed and generates new resources.

deploy/template.yaml
Wynndow: Good question. sam validate was moaning at me that it wasn't a map. I'm going to try put it back.

Wynndow:
> This seems better to me, what do you think?
> 
> It makes the prod configuration cleaner, and isn't much more code than before. The main thing that's annoying is configuring the extra events and overriding that cfn-lint rule :/

I think this is ok - at some point we're going to create a new lambda and forget to add the extra testing event on. We _could_ write a macro to do that for us and not have to define the event at all....

deploy/template.yaml
Wynndow: Seems happy now ¯\_(ツ)_/¯

Joe-Edwards-GDS (commented):
I'm happy to approve this, but I think we need to:
- Remove the test pipeline code
- Monitor this in `build` to make sure it doesn't cause any changes there

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1968, opened 2024-06-03
PYIC-6057: Remove references to f2fVolumes feature flag with assumption it was always true

MikeCollingwood:
### What changed

- Remove references to a feature flag to clean it up

### Issue tracking

- [PYIC-6057](https://govukverify.atlassian.net/browse/PYIC-6057)

### Merging order

This must follow this change to turn the feature flag on everywhere:
https://github.com/govuk-one-login/ipv-core-common-infra/pull/952

[PYIC-6057]: https://govukverify.atlassian.net/browse/PYIC-6057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -545,13 +545,7 @@ states:
       next:
         targetState: ADDRESS_AND_FRAUD_J2
       access-denied:
-        targetState: MULTIPLE_DOC_F2F_CHECK_PAGE
-        checkIfDisabled:
-          f2f:
-            targetState: MULTIPLE_DOC_CHECK_PAGE
-        checkFeatureFlag:
-          f2fVolumesJourneyEnabled:
-            targetState: PROVE_ANOTHER_WAY_J2
+        targetState: PROVE_ANOTHER_WAY_J2
       alternate-doc-invalid-passport:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -545,13 +545,7 @@ states:
       next:
         targetState: ADDRESS_AND_FRAUD_J2
       access-denied:
-        targetState: MULTIPLE_DOC_F2F_CHECK_PAGE
-        checkIfDisabled:
-          f2f:
-            targetState: MULTIPLE_DOC_CHECK_PAGE
```
Joe-Edwards-GDS, line 551: Are you sure we want to remove this check? I think we probably need to keep this, as we can't start routing users to F2F if it's disabled.

Joe-Edwards-GDS (commented):
~we should turn the flag on in prod before merging this~

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
MikeCollingwood: Ah you're right, PROVE_ANOTHER_WAY_J2 has a f2f route

Joe-Edwards-GDS (approved):

PR #1972, opened 2024-06-04
PYIC-6052: Store async VCs in EVCS

Wynndow:
### What changed

Store async VCs in EVCS

### Why did it change

Stores f2f VCs received by the process-async-cri-credential lambda in evcs, behind a feature flag.

This also adds a `migrated` timestamp to VCs stored in the tactical store, as well as fields on the SessionCredential and VerifiableCredential classes. This is neccessary to ensure that the `migrated` timestamp isn't lost.

### Issue tracking

- [PYIC-6052](https://govukverify.atlassian.net/browse/PYIC-6052)

[PYIC-6052]: https://govukverify.atlassian.net/browse/PYIC-6052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
@@ -2,8 +2,8 @@ package uk.gov.di.ipv.core.library.domain;
 
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
+import lombok.Data;
 import lombok.EqualsAndHashCode;
-import lombok.Getter;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
 import uk.gov.di.ipv.core.library.persistence.item.SessionCredentialItem;
 import uk.gov.di.ipv.core.library.persistence.item.VcStoreItem;
@@ -12,7 +12,7 @@ import java.text.ParseException;
 import java.time.Instant;
 import java.util.Date;
 
-@Getter
+@Data
 @EqualsAndHashCode(exclude = "signedJwt")
 public class VerifiableCredential {
     private final String userId;
@@ -20,8 +20,9 @@ public class VerifiableCredential {
     private final String vcString;
     private final JWTClaimsSet claimsSet;
     private final SignedJWT signedJwt;
+    private Instant migrated;
 
-    private VerifiableCredential(String userId, String criId, SignedJWT signedJwt)
+    private VerifiableCredential(String userId, String criId, SignedJWT signedJwt, Instant migrated)
             throws CredentialParseException {
         try {
             this.userId = userId;
@@ -29,6 +30,7 @@ public class VerifiableCredential {
             this.vcString = signedJwt.serialize();
             this.claimsSet = signedJwt.getJWTClaimsSet();
             this.signedJwt = signedJwt;
+            this.migrated = migrated;
         } catch (ParseException e) {
             throw new CredentialParseException(
                     "Failed to get jwt claims to construct verifiable credential", e);
@@ -37,7 +39,7 @@ public class VerifiableCredential {
 
     public static VerifiableCredential fromValidJwt(String userId, String criId, SignedJWT jwt)
             throws CredentialParseException {
-        return new VerifiableCredential(userId, criId, jwt);
+        return new VerifiableCredential(userId, criId, jwt, null);
     }
 
     public static VerifiableCredential fromVcStoreItem(VcStoreItem vcStoreItem)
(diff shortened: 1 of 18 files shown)


libs/common-services/src/test/java/uk/gov/di/ipv/core/library/persistance/item/SessionCredentialItemTest.java
Joe-Edwards-GDS, line 1: (I've just spotted there's a typo in the package name here - it should be `persistence`)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -53,6 +53,15 @@ public void storePendingIdentity(
         persistEvcsUserVCs(userId, credentials, existingEvcsUserVCs, true);
     }
 
+    @Tracing
+    public void storeAsyncVc(VerifiableCredential credential) {
```
Joe-Edwards-GDS: Do you think this should be `storePendingVc` (e.g. to match `storePendingIdentity` above)?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
Joe-Edwards-GDS, line 1: Can you explain the scenarios where we need the migrated flag here (and on the session item) - are there cases where we read a VC from the store and write it back (overwriting the existing value)?

This is fine, we'll just need to be a bit careful to understand what exactly it represents - particularly once we start reading these from EVCS instead: we'll probably need to treat it as migrated = true when reading from EVCS, but that could be misleading if we've never actually written it to the old store.

(The alternative would be passing `migrated` as an extra flag to `VerifiableCredentialService#persistUserCredentials` and `VerifiableCredentialService#storeIdentity`)

Joe-Edwards-GDS: I guess it's likely to be useful for @rpayal 's half of the problem - where we need to update existing records, but could save ourselves a round-trip by remembering the old value.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/VcStoreItem.java
```diff
@@ -24,6 +24,7 @@ public class VcStoreItem implements DynamodbItem {
     private String credential;
     private Instant dateCreated;
     private Instant expirationTime;
+    private Instant migrated;
```
Joe-Edwards-GDS, line 27: (I assume this will correctly parse as `null` for existing records with no value)

Joe-Edwards-GDS (dismissed):
Couple of comments but this looks sensible to me

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
Wynndow: Yep - a user achieves a P2 identity whilst the `evcsWrites` flag is on, and so any VC stored in the tactical store should have the `migrated` timestamp. 
They then run a reuse journey and either enter into a 6MFC journey (hopefully it won't take 6 months before we switch to evcs but you never know....), or a change of details journey.
We remove some VC from the session creds store and they run a new IPV journey. If succesful we write the VCs back to the tactical store.
If we didn't have the flag on the VerifiableCredential item, or the SessionCredential item, we'd lose the existing migrated timestamp.

...

And now that I've writted it out I've realised that we would (hopefully) be setting the `migrated` flag when writing back on the second IPV journey. But it does leave the question of if we are happy to lose the original timestamp in favour of the new one. If we don't care, we can probably not worry about having the `migrated` timestamp anywhere other than the `VcStoreItem`.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/VcStoreItem.java
Wynndow: Yep yep

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
Joe-Edwards-GDS: Ah I see, because when we store VCs in the tactical store we just blow away everything and write them again.

Thinking about what this timestamp might be used for, it makes sense if it's updated each time. It's meant to keep a record of which tactical store VCs are in EVCS, and it's a timestamp so if EVCS failed/needed to be reset, we could disregard any that were migrated before that point. So we want the most recent time it would have been written.

In that case, I think we could take either approach :)

Wynndow: In which case, I don't think we need to pull the `migrated` timestamp out of the tactical store and handle it - we can just let it get reset when writing back? I'll remove it from all but the `VcStoreItem`.

Wynndow: Actually, I think we just leave it as is here - we'll need `migrated` on the `VerifiableCredential` class when setting in in the process-aysnc lambda, and we'll need it on the `SessionCredential` class when setting it when storing an identity.

We just need to overwrite the timestamp if it already exists when storing.

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #1973, opened 2024-06-04
BAU: Remove redundant logger config

Wynndow:
### What changed

Remove redundant logger config

### Why did it change

We were defining an extra logger, but it had exactly the same config as the root logger.

I've deployed this to my dev env and I'm still getting JSON logs into cloudwatch.


Joe-Edwards-GDS (approved):
Obviously worth checking in other environments too, but I'm sure you will :)

/cc @sam803 - you'll want to make this change on the new reverification lambda too

PR #1977, opened 2024-06-05
PYIC-6582: Use context for coi-check and reset-session-identity lambdas

Wynndow:
### What changed

Use context for coi-check and reset-session-identity lambdas

### Why did it change

This moves from using a `coiSubjourneyType` value in the users session to injecting the required reset or check type in the journey map.

This means we're not relying on state that gets set by magic in a different lambda, which could be error prone.

This removes the existing enum as it wasn't really appropriate for both contexts, and replaces it with more specific versions. The reset-session-identity lambda is used in more than just COI journeys for example.

### Issue tracking

- [PYIC-6582](https://govukverify.atlassian.net/browse/PYIC-6582)

[PYIC-6582]: https://govukverify.atlassian.net/browse/PYIC-6582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -97,31 +99,26 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
                                             .stream())
                             .toList();
 
-            var coiSubjourneyType = ipvSession.getCoiSubjourneyType();
+            var checkType = RequestHelper.getCoiCheckType(request);
```
Joe-Edwards-GDS: (@MikeCollingwood this will conflict with your PR, but hopefully it'll be obvious how to resolve - whoever merges second!)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
Joe-Edwards-GDS, line 1: Interested to get your take on how painful this was to understand/implement - basically do you think this is still easier to understand than the session context approach. (Excluding considerations for the audit events)

Joe-Edwards-GDS, line 1: Do you think there'd be value in using nested journeys here?

We could have two:
- UPDATE_NAME (reset -> DCMAW -> Success -> Fraud)
- UPDATE_NAME_AND_ADDRESS (reset -> DCMAW -> Success -> Address -> Fraud)

I'm not sure - it doesn't meet the rule of three, so I'm reasonably ok with the duplication. (It's already split into a separate update-name subjourney, which helps a lot!)

Joe-Edwards-GDS (commented):
I won't approve this yet, before we're confident the context thing won't be needed for the audit events, but it looks like exactly what I was expecting :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
Wynndow: I think this is very easy to understand - much more so that the session based approach.

Wynndow: I don't think it's neede right now - we'd also need to work out a way to pass the lamdaInput to the reset lambda when inside a nested journey definition - that's not something we've done before.
Also, the `_AND_ADDRESS` branches are already using a nested thing, so there's less saving there.

Joe-Edwards-GDS (approved):

PR #1982, opened 2024-06-06
PYIC-5872: Open dev internal api gateway again

Wynndow:
### What changed

Open dev internal api gateway again.

I've tested that this will actually deploy to a non-dev environment.

### Why did it change

[Revert "Revert "PYIC-5872: Open API gateway""](https://github.com/govuk-one-login/ipv-core-back/commit/e55a316385596aabb4f54e5b2e4fdaa6945b5609)

This reverts commit https://github.com/govuk-one-login/ipv-core-back/commit/1143d1c4f98a647ce5b47521a240f4eb237891d7.

This reversion commit was for everthing in https://github.com/govuk-one-login/ipv-core-back/pull/1966.
Reverting the reversion to bring the changes back in, and then roll forward in the next commit.

[PYIC-5872: Use lambda perm over conditional event](https://github.com/govuk-one-login/ipv-core-back/commit/cf1ef547a3c6b84dfafd5fd8cd7ef524f287f353)

Having conditional events on functions wasn't working in non dev
environments. It appears CloudFormation wanted to resolve the nested
reference to the conditionally created API gateway before noticing that
the event itself didn't need creating. This led to errors due to an
unresolveable reference.

Having inspected the CloudFormation stacks created with and without the
event, the only extra resource being created was the permission on the
lambda to allow the testing API gateway to invoke it. All the other API
gateway resources that it might create were already handled by using the
OpenApi spec when creating the gateway itself.

We can create that permission separately and use a conditional for it.

### Issue tracking

- [PYIC-5872](https://govukverify.atlassian.net/browse/PYIC-5872)

[PYIC-5872]: https://govukverify.atlassian.net/browse/PYIC-5872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -1278,8 +1419,14 @@ Resources:
         IPVCorePrivateAPI:
           Type: Api
           Properties:
-            RestApiId:
-              Ref: IPVCorePrivateAPI
+            RestApiId: !Ref IPVCorePrivateAPI
+            Path: /journey/{journeyStep+}
+            Method: POST
+        IPVCoreInternalTestingApi:
```
Joe-Edwards-GDS, line 1425: Should this one still be here?

Joe-Edwards-GDS (commented):
Nice that we don't have a load of extra cfn-lint issues :)

This approach does seem harder to keep up-to-date, but hopefully:
- In future we'll have a bunch of automated tests using this
- We don't add new internal APIs very often!

deploy/template.yaml
Wynndow: Yep - this approach works when attaching to a step function, rather than a lambda function. Thanks CloudFormation.

Wynndow:
> Nice that we don't have a load of extra cfn-lint issues :)
> 
> This approach does seem harder to keep up-to-date, but hopefully:
> 
>     * In future we'll have a bunch of automated tests using this
> 
>     * We don't add new internal APIs very often!

Yeah - i've intentionally put the new permissions with the functions rather than all grouped with the new gateway. I'm guessing that the next time we add an internal function we'll look at prior art and it'll help future us remember.

deploy/template.yaml
Joe-Edwards-GDS: of course

Joe-Edwards-GDS (approved):

PR #1985, opened 2024-06-06
PYIC-6652: remove address/fraud + gpg45 evaluation from reverification journey

thebauSoftwire:
### What changed
- removed address/fraud + gpg45 evaluation from reverification journey
- in `CriStoringService.storeVcs`, we only VCs against a user if we are not on a reverification journey as during this journey, they are not authenticated
- in `CriCheckingService.checkVcResponse`, we only want to check CIs when not on a reverification journey
- adds reverificationStatus to ipvSession which is set in `CheckCoiHandler`
- updates `UserReverificationHandler ` to check for reverificationStatus instead of Vot
- updates `gpg45DcmawValidator` to also check verification and strength scores
- adds/updates unit tests

### Why did it change
We are removing the address and fraud checks from the MFA reset journey, and relying on the verification check in DCMAW and continuity of identity check only.

### Issue tracking

- [PYIC-6652](https://govukverify.atlassian.net/browse/PYIC-6652)

[PYIC-6652]: https://govukverify.atlassian.net/browse/PYIC-6652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -2887,6 +2887,8 @@ Resources:
             QueueName: !ImportValue AuditEventQueueName
         - DynamoDBReadPolicy:
             TableName: !Ref SessionsTable
+        - DynamoDBWritePolicy:
+            TableName: !Ref SessionsTable
```
Joe-Edwards-GDS, line 2891: (We could potentially merge this and the one above into a CRUD policy - but this is fine)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
```diff
@@ -34,4 +34,8 @@ public class ClientOAuthSessionItem implements DynamodbItem {
     public String getClientOAuthSessionId() {
         return clientOAuthSessionId;
     }
+
+    public String[] getScopeClaims() {
+        return this.scope.split(" ");
```
Joe-Edwards-GDS: Do you think it'd make sense to have this return a `List`? (Is an array ever more useful?)

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -158,13 +166,20 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
...
                 LOGGER.info(
                         LogHelper.buildLogMessage("Failed COI check")
                                 .with(LOG_COI_CHECK_TYPE.getFieldName(), coiCheckType));
 
                 return JOURNEY_COI_CHECK_FAILED.toObjectMap();
             }
 
+            if (isReverification) {
+                setIpvSessionReverificationStatus(ipvSession, ReverificationStatus.SUCCESS);
+            }
```
Joe-Edwards-GDS: It might be slightly cleaner to make this a single call above the `if (!coiCheckSuccess)` - but doesn't matter too much.

(Just so if other branches are added, it's not missed)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ReverificationStatus.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@Getter
+@ExcludeFromGeneratedCoverageReport
+public enum ReverificationStatus {
+    SUCCESS("success"),
+    FAILED("failed");
+
+    private final String status;
```
sam803: Is this status string used anywhere? If not simpler to remove and just have:
```
public enum ReverificationStatus {
    SUCCESS,  FAILED;
}
```

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingServiceTest.java
```diff
@@ -484,6 +486,25 @@ void checkVcResponseShouldReturnFailWithCiWhenUserBreachesCiThreshold() throws E
         assertEquals(new JourneyResponse(JOURNEY_FAIL_WITH_CI_PATH), result);
     }
 
+    @Test
+    void checkVcResponseDoesNotCheckForCIsWhenOnReverificationJourney() throws Exception {
+        // Arrange for VCs not correlated
```
Joe-Edwards-GDS: (I think this comment is wrong?)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45DcmawValidator.java
```diff
@@ -13,6 +16,22 @@ public static boolean isSuccessful(CredentialEvidenceItem item) {
...
+        var checkMethodWithVerificationScore =
+                checkMethods.stream()
+                        .filter(
+                                checkMethod ->
+                                        checkMethod.getBiometricVerificationProcessLevel() != null)
+                        .findFirst();
+
+        return checkMethodWithVerificationScore.isPresent()
+                ? checkMethodWithVerificationScore.get().getBiometricVerificationProcessLevel()
+                : 0;
```
Joe-Edwards-GDS: Do you think there's anywhere to commonise this? Perhaps it should be a method on `CredentialEvidenceItem` itself.

I might just override `getVerificationScore` itself so it follows this logic for DCMAW.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ReverificationStatus.java
Joe-Edwards-GDS: TBH it could also just be a boolean (although an enum does give us more flexibility to extend in future)

deploy/template.yaml
thebauSoftwire: yeah that was the other option but I was leaning towards this in keeping w least privilege

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45DcmawValidator.java
thebauSoftwire: Not sure it should be in `CredentialEvidenceItem` as I don't think it's a method that should be available if it wasn't from dcmaw. But I agree with replacing `getVerificationScore` instead and letting the validator be the source of truth for this logic. Let me know if you disagree though!

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -165,6 +163,10 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
                     govukSigninJourneyId,
                     ipAddress);
 
+            if (isReverification) {
+                setIpvSessionReverificationStatus(ipvSession, ReverificationStatus.SUCCESS);
+            }
```
Joe-Edwards-GDS, line 158: I really meant make this cover all cases for the check, i.e.:
```
if (isReverification) {
    setIpvSessionReverificationStatus(ipvSession, coiCheckSuccess
        ? ReverificationStatus.SUCCESS
        : ReverificationStatus.FAILED);
}
```
(We definitely don't want to do two writes to dynamodb if we don't need to)

lambdas/check-coi/src/test/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandlerTest.java
```diff
@@ -108,7 +109,8 @@ void shouldReturnPassedForSuccessfulGivenNamesOnlyCheck(
                                 List.of(M1A_ADDRESS_VC, M1A_EXPERIAN_FRAUD_VC)))
                         .thenReturn(true);
                 when(mockIpvSessionItem.getCoiSubjourneyType()).thenReturn(coiSubjourneyType);
-                when(mockClientSessionItem.getScopeClaims()).thenReturn(new String[] {"openid"});
+                when(mockClientSessionItem.getScopeClaims())
+                        .thenReturn(Arrays.asList(ScopeConstants.OPENID));
```
Joe-Edwards-GDS: (Personally I have a slight preference for `List.of()` - but it's fairly equivalent)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45DcmawValidator.java
Joe-Edwards-GDS: You could still include the DCMAW check if it was part of `CredentialEvidenceItem`, e.g.:
```
public Integer getVerificationScore() {
    if (isDcmaw()) {
        // get biometric verification score
    }
    return verificationScore;
}
```
but either way is fine

Joe-Edwards-GDS (commented):
Ah, looks like you've got conflicts from Chris' merge 😬

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45DcmawValidator.java
thebauSoftwire: good point!

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/domain/CredentialEvidenceItem.java
```diff
@@ -192,4 +192,26 @@ public enum EvidenceType {
...
+                    checkDetails.stream()
+                            .filter(
+                                    checkMethod ->
+                                            checkMethod.getBiometricVerificationProcessLevel()
+                                                    != null)
+                            .findFirst();
+
+            return checkMethodWithVerificationScore.isPresent()
+                    ? checkMethodWithVerificationScore.get().getBiometricVerificationProcessLevel()
+                    : 0;
```
Joe-Edwards-GDS: This is fine, but if you want a couple of neater ways of writing this:

Streams:
```
return checkDetails.stream()
        .filter(checkDetail -> checkDetail.getBiometricVerificationProcessLevel() != null)
        .findFirst()
        .map(CheckDetail::getBiometricVerificationProcessLevel)
        .orElse(0);
```

Or old-school:
```
for (var checkDetail : checkDetails) {
    if (checkDetail.getBiometricVerificationProcessLevel() != null) {
        return checkDetail.getBiometricVerificationProcessLevel();
    }
}
return 0;
```

Joe-Edwards-GDS (dismissed):
One optional comment, but looking good I think.

We should keep an eye out for the fact we've changed the validation rules for DCMAW, but it should be fine. (Maybe just wait until tomorrow morning to merge)

Joe-Edwards-GDS (approved):

PR #1986, opened 2024-06-06
PYIC-6660: Update TxMA audit event when storing identity

sam803:
## Proposed changes

Update TxMA audit event when storing identity to include the identity type (new, update, pending)

### What changed

- changed `isCompletedIdentity` boolean lambda input to `identityType` string
- added enum for the different identity types
- added identity type to `AuditExtensionVot` and renamed to `AuditExtensionIdentityType`
- changed `RequestHelper` to get the identity type
- updated associated tests

### Why did it change

When an identity is stored we need to add an extra audit field to distinguish between the different event types

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6660

### DIFF
--- a/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionIdentityType.java
+++ b/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionIdentityType.java
@@ -1,0 +1,12 @@
+package uk.gov.di.ipv.core.library.auditing.extension;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+import uk.gov.di.ipv.core.library.enums.IdentityType;
+import uk.gov.di.ipv.core.library.enums.Vot;
+
+@ExcludeFromGeneratedCoverageReport
+public record AuditExtensionIdentityType(
+        @JsonProperty(value = "identity_type", required = true) IdentityType identityType,
+        @JsonProperty(required = false) Vot vot)
+        implements AuditExtensions {}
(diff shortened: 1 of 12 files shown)


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionIdentityType.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+import uk.gov.di.ipv.core.library.enums.IdentityType;
+import uk.gov.di.ipv.core.library.enums.Vot;
+
+@ExcludeFromGeneratedCoverageReport
+public record AuditExtensionIdentityType(
+        @JsonProperty(value = "identity_type", required = true) IdentityType identityType,
+        @JsonProperty(required = false) Vot vot)
```
Joe-Edwards-GDS, line 11: I don't think the `required` will do anything here (but it doesn't do any harm 🤷)

Joe-Edwards-GDS (approved):
This looks sensible to me. We'll need to update the TxMA spreadsheet etc. when we add this attribute.

There's a chance we'll need to add further detail as well (about the new/old identity) but I'm expect that to be a separate ticket.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionIdentityType.java
sam803: Yeh I just thought it would make it more obvious that one is required and one not :)

PR #1989, opened 2024-06-07
PYIC-6672 Set migrated TS to user credentials when writing to EVCS.

rpayal:
## Proposed changes
Set migrated TS to user credentials when writing to EVCS.

### What changed
Change made to set migrated TS to all newly transferred user credentials when writing to EVCS.

### Issue tracking

- [PYIC-6672](https://govukverify.atlassian.net/browse/PYIC-6672)

### Other considerations

[PYIC-6672]: https://govukverify.atlassian.net/browse/PYIC-6672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
+++ b/lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
@@ -33,6 +33,7 @@ import uk.gov.di.ipv.core.library.service.IpvSessionService;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredentialsService;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.VerifiableCredentialService;
 
+import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
@@ -107,8 +108,6 @@ public class StoreIdentityHandler implements RequestHandler<ProcessRequest, Map<
 
             var identityType = RequestHelper.getIdentityType(input);
 
-            verifiableCredentialService.storeIdentity(credentials, userId);
-
             if (configService.enabled(EVCS_WRITE_ENABLED)) {
                 if (identityType != IdentityType.PENDING) {
                     evcsService.storeCompletedIdentity(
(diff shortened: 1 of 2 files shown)


lambdas/store-identity/src/main/java/uk/gov/di/ipv/core/storeidentity/StoreIdentityHandler.java
```diff
@@ -117,7 +116,15 @@ public Map<String, Object> handleRequest(ProcessRequest input, Context context)
                     evcsService.storePendingIdentity(
                             userId, credentials, clientOAuthSessionItem.getEvcsAccessToken());
                 }
+                credentials.forEach(
+                        credential -> {
+                            if (credential.getMigrated() == null)
+                                credential.setMigrated(Instant.now());
```
Joe-Edwards-GDS: I think we actually want to do this even if there is already a migrated timestamp - it'll be more useful to have the date as the most recent time we know it was written to the store (rather than just the first time).

rpayal: I thought we want to keep when this was first migrated.

Joe-Edwards-GDS: I think the situation where a timestamp is useful is something like:
1. We turn on EVCS migrations and records start getting written into EVCS
2. Something goes wrong, and we end up turning it off and wiping EVCS
3. We turn it back on again and start writing records again

In this case all records with a timestamp older than step 2 are actually not migrated, but if we write them again then we want to update the timestamp to indicate that they have been again.

Does that make sense? It's all a bit hypothetical - if all goes well we won't need this!

rpayal: done.

deploy/journeyEngineStepFunction.asl.json
```diff
@@ -142,7 +142,6 @@
         "ipAddress.$": "$$.Execution.Input.ipAddress",
         "clientOAuthSessionId.$": "$$.Execution.Input.clientOAuthSessionId",
         "featureSet.$": "$$.Execution.Input.featureSet",
-        "lambdaInput.$": "$.lambdaInput",
```
Joe-Edwards-GDS: I think we still need this?

rpayal: yeah sure we needed that.....not sure what I was thinking.

Joe-Edwards-GDS (approved):

PR #1969, opened 2024-06-03
PYIC-6588: send audit event when no photo id journey starts

shivanshuit914:
## Proposed changes

-  send audit event when no photo id journey starts

### What changed

- Sending new audit event for the photo id journey.

### Why did it change

- To allow analytics team to track no photo id journey.

### Issue tracking

- [PYIC-6588](https://govukverify.atlassian.net/browse/PYIC-6588)

Screenshot 2024-06-04 at 13 28 20

[PYIC-6588]: https://govukverify.atlassian.net/browse/PYIC-6588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1994, opened 2024-06-07
Pyic 6711 audit event journey event

Joe-Edwards-GDS:
### What changed

Added a new mechanism to trigger audit events upon specific journey events, e.g. in the journey event definition:
```
  SOME_JOURNEY_STATE:
    response:
      type: page
      pageId: some-page-id
    events:
      next:
        targetState: SOME_OTHER_STATE
        auditEvents:
          - IPV_SOME_AUDIT_EVENT
        auditContext:
          foo: bar
```

### Why did it change

We are getting an increasing number of audit event requirements, which are tied to specific journeys. Having a mechanism like this should make it easier to define 

### Issue tracking

- [PYIC-6711](https://govukverify.atlassian.net/browse/PYIC-6711)

[PYIC-6711]: https://govukverify.atlassian.net/browse/PYIC-6711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -15,6 +15,7 @@ import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
 import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionMitigationType;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionSubjourneyType;
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensions;
 import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestrictedDeviceInformation;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
 import uk.gov.di.ipv.core.library.domain.CoiSubjourneyType;
@@ -48,11 +49,11 @@ import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.StepRes
 
 import java.io.IOException;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 
+import static com.amazonaws.util.CollectionUtils.isNullOrEmpty;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.BACKEND_SESSION_TIMEOUT;
 import static uk.gov.di.ipv.core.library.domain.CoiSubjourneyType.isCoiSubjourneyEvent;
 import static uk.gov.di.ipv.core.library.domain.IpvJourneyTypes.SESSION_TIMEOUT;
@@ -168,22 +169,6 @@ public class ProcessJourneyEventHandler
                         auditEventUser, stepResponse.getMitigationStart(), deviceInformation);
             }
 
-            String journeyAuditEvent = stepResponse.getAuditEvent();
-            if (journeyAuditEvent != null) {
-                boolean eventTypeExists =
-                        Arrays.stream(AuditEventTypes.values())
-                                .anyMatch(eventType -> eventType.name().equals(journeyAuditEvent));
-                if (!eventTypeExists) {
-                    LOGGER.error(
-                            LogHelper.buildLogMessage("Invalid audit event type provided")
-                                    .with(LOG_JOURNEY_EVENT.getFieldName(), journeyAuditEvent));
-                    throw new JourneyEngineException(
-                            "Invalid audit event type provided, failed to execute journey engine step.");
-                }
-                AuditEventTypes auditEventType = AuditEventTypes.valueOf(journeyAuditEvent);
-                sendJourneyAuditEvent(auditEventType, auditEventUser, deviceInformation);
-            }
-
             return stepResponse.value();
         } catch (HttpResponseExceptionWithErrorBody e) {
             return StepFunctionHelpers.generateErrorOutputMap(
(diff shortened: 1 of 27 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -359,31 +366,30 @@ private Map<IpvJourneyTypes, StateMachine> loadStateMachines(
...
                         new AuditRestrictedDeviceInformation(deviceInformation)));
     }
 
+    private AuditExtensions getAuditExtensions(
+            AuditEventTypes auditEventType, Map<String, String> auditContext) {
+        return switch (auditEventType) {
+            case IPV_MITIGATION_START -> new AuditExtensionMitigationType(
+                    auditContext.get("mitigationType"));
+            default -> null;
+        };
```
Joe-Edwards-GDS, line 390: Intention here is that we have a relatively flexible system for adding metadata to audit events - we already need it for the mitigation type, but will also want it for the COI change type, etc.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/TransitionResult.java
Joe-Edwards-GDS, line 1: This has caused a bit of noise - the alternative would be for the state machine itself to record the audit events, but that has its own problems (e.g. injecting the `auditEventService` and associated user info).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -515,9 +521,17 @@ states:
...
+          - IPV_MITIGATION_START
+        auditContext:
+          mitigationType: enhanced-verification
         checkIfDisabled:
           f2f:
             targetState: MITIGATION_02_OPTIONS
+            auditEvents:
+              - IPV_MITIGATION_START
+            auditContext:
+              mitigationType: enhanced-verification
```
Joe-Edwards-GDS, line 534: Interested in opinions on this syntax, I've tried to make this as flexible as possible while staying concise, but we could streamline in a few ways:
- `auditEvents` could be singular - but that means you can't record multiple events at the same time, which we might need to do for COI
- `auditContext` could be a simple string - but an object allows much more specificity
- we could 'inherit' audit events for these disabled/feature checks to reduce duplication - at the moment that's desirable, but it wouldn't have been while these mitigations were behind a feature flag (for example)

blakeyp: Plural `auditEvents` and an object for `auditContext` sounds sensible but yeah inheriting seems a bit risky especially if it's not obvious how it works

Wynndow: Part of me is wondering if we want the same audit context tied to all the audit events? Or if there is any benefit in each event having its own context? Something like:
```
enhanced-verification:
  targetState: MITIGATION_02_OPTIONS_WITH_F2F
  auditEvents:
    - event: IPV_MITIGATION_START
      auditContext:
        mitigationType: enhanced-verification
    - event: SOMETHING ELSE
      auditContext:
        other: stuff
```

I supposed it doesn't really matter which event consumes which part of the context - it just feels a bit "loose" like that.

Joe-Edwards-GDS: Yeah, I think that's a good question - that would be more precise, but a little more verbose. Part of my motivation was to try and make the 'simple' cases as easy as possible.

I wonder if there's a compromise at:
```
enhanced-verification:
  targetState: MITIGATION_02_OPTIONS_WITH_F2F
  auditEvents:
    - event: IPV_MITIGATION_START
      mitigationType: enhanced-verification
    - event: SOMETHING_ELSE
```

It'd need a slightly more interesting type (to allow arbitrary additional parameters), but should be doable.

Joe-Edwards-GDS: In practice I think we're unlikely to often need multiple events in the same place, and even less often to need context on both.

Wynndow: Maybe we do that refactor if we need it. I think the bag of stuff approach is proabbly fine, especially if we're actually going to be limited in where we have multiple events.

Wynndow: I missed the bit about inheriting - but I also agree. Something would get missed and we'd send the wrong thing.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -704,16 +734,23 @@ states:
...
 
   # No photo id journey (M2B)
   CRI_CLAIMED_IDENTITY_M2B:
     response:
       type: cri
       criId: claimedIdentity
-      auditEvent: IPV_NO_PHOTO_ID_JOURNEY_START
```
Wynndow, line 716: Did this work before?!

Wynndow (approved):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Joe-Edwards-GDS: Yeah, this was added last week by Shiv (https://github.com/govuk-one-login/ipv-core-back/pull/1969) but superseded by this

PR #1998, opened 2024-06-10
PYIC-6581: Start storing users states in a stack

Wynndow:
### What changed

Start storing users states in a stack

### Why did it change

[BAU: Remove ContraIndicatorMitigationDetailsDto](https://github.com/govuk-one-login/ipv-core-back/commit/eb2534a7ff618f8975389f50a6d5c8010698de75)

This class and its session item are a hangover from an early iteration
on mitigating CIs. It's not used anywhere in the non-test code and can
be removed.

[PYIC-6581: Start storing users states in a stack](https://github.com/govuk-one-login/ipv-core-back/commit/09769274872dbcb6f00451e69308e6e52320f94c)

This change starts recording the users transition through the journey
states in a stack. This is to allow us to support a native "back"
functionality, rather than having to have specific events for it.

This includes a new convertor for the stack - dynamo can't save a
complex type without you telling it how to do it.

We need to merge this before any changes that might start consuming the
stack to ensure that all active sessions have a stack to work with.

### Issue tracking

- [PYIC-6581](https://govukverify.atlassian.net/browse/PYIC-6581)

[PYIC-6581]: https://govukverify.atlassian.net/browse/PYIC-6581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -213,6 +213,9 @@ private StepResponse executeJourneyEvent(
                                         journeyChangeState.getInitialState()));
                 ipvSessionItem.setJourneyType(journeyChangeState.getJourneyType());
                 ipvSessionItem.setUserState(journeyChangeState.getInitialState());
+                ipvSessionItem.pushState(
+                        journeyChangeState.getJourneyType(), journeyChangeState.getInitialState());
```
Joe-Edwards-GDS, line 209: I wonder about this - should we include this intermediate state in the stack? It's a more complete history if we do, but we'll need to handle it specially when calculating the 'back' logic.

(We do have to handle this case - currently you can go 'back' from the first screen in `UPDATE_NAME` to `REUSE_EXISTING_IDENTITY`)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -67,4 +79,16 @@ public void setFeatureSetFromList(List<String> featureSet) {
...
     }
+
+    public void pushState(IpvJourneyTypes journeyType, String state) {
+        this.stateStack.push(new JourneyState(journeyType, state));
+    }
+
+    public void pushState(String state) throws NoCurrentStateException {
+        var currentState = this.stateStack.peek();
+        if (currentState == null) {
+            throw new NoCurrentStateException();
```
Joe-Edwards-GDS, line 80: If you'd prefer to use an unchecked exception here, you could use `IllegalStateException`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/convertors/DequeJourneyStateConverter.java
Joe-Edwards-GDS, line 1: This is fine - and kinda cool.

A couple of other options occur to me:

- If we're rolling our own serialization we could consider using a less verbose (and more readable) format, e.g.: `JOURNEY_ONE/STATE_ONE,JOURNEY_TWO/STATE_TWO,...` This does make assumptions about which characters may be present in journey/state names, so would need a tiny bit of care. (But I think we might already make similar assumptions elsewhere!)

- You could drop `Deque` entirely and just use a list with built-in serialisation - the performance isn't really a concern.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -67,4 +79,16 @@ public void setFeatureSetFromList(List<String> featureSet) {
         this.featureSet =
                 (featureSet != null && !featureSet.isEmpty()) ? String.join(",", featureSet) : null;
     }
+
+    public void pushState(IpvJourneyTypes journeyType, String state) {
```
Joe-Edwards-GDS, line 74: If we plan to keep this alongside `setState` for a while, we should probably move the `setState` call in here and make it private - so you can't accidentally take them out of sync - but if it's just temporary then it doesn't matter so much.

Joe-Edwards-GDS (commented):
Looks good - I think it's worth nailing down the DynamoDB format, but what you've got should work

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: I think we should for completeness. It's also useful for keeping track of the users journey type. When we remove journey type from the session I think we'd lose it otherwise.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Wynndow: I did think about this. But wondered about handling it gracefully were it to ever happen, rather than just propergating it to the runtime. To avoid that we'd need to catch it. And at that point we might as well have a checked...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Wynndow: I think I follow - but yeah this is temporary. I'm planning on removing `userState` from the session in a follow up PR. Just got to stage it out

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/convertors/DequeJourneyStateConverter.java
Wynndow: We's still need the custom converter if we used a list with a complex type instead of a deque. But if we used the format you've suggested that'd probably be fine. And you're right, we already use the slash when tracking a users state when nested, so could reuse that.

I think I just liked using `pop()` though.

It'l be less code and less complex though if I pair it back a bit, will change it.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Sure - when we go back we'll just need to know that it's ok to skip over it.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Joe-Edwards-GDS: Fair enough, I think it basically depends whether you think it's worth handling gracefully (given that either way it'll produce an error for the user).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/convertors/DequeJourneyStateConverter.java
Joe-Edwards-GDS: You could still be able to make the interface exposed in `IpvSessionItem` a `Deque`, but use a different serialisation format internally. But using a string list seems easier anyway, since we can avoid the custom converter code entirely.

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2001, opened 2024-06-11
PYIC-5099: Enable RSA signature validation

MikeCollingwood:
### What changed

- Derive signing algorithm from signing key: link
- Use signing algorithm to verify with correct algorithm: [link](https://github.com/govuk-one-login/ipv-core-back/pull/2001/files#diff-92492052064a6d727bcaacfa28906cd42bb1d917aed6e34277c4bdbb3bb5b055)
- Added test case for verifying with RSA: [link](https://github.com/govuk-one-login/ipv-core-back/pull/2001/files#diff-5b67ba034eb9b79b4cd2f049e858d1436eab521ee171f2870d1847eb4d66b980)
- Updated the local running cri config: [link](https://github.com/govuk-one-login/ipv-core-back/pull/2001/files#diff-385c34f779b8d44e9f334f576d2fd0c75c30ae1a49fd7e3b4dce99a686384333)

### Why did it change

- So we can validate RSA-signed VCs we expect from DWP

### Issue tracking

- [PYIC-5099](https://govukverify.atlassian.net/browse/PYIC-5099)

### Remaining to do:

- Change the config manually for the DWP KBV stub to be an RSA key

[PYIC-5099]: https://govukverify.atlassian.net/browse/PYIC-5099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -108,30 +123,29 @@ public VerifiableCredential parseAndValidate(
...
-            if (!concatSignatureVerifiableCredential.verify(verifier)) {
+            var verifier = getVerifier(signingAlgorithm, signingKey);
+            if (!formattedVc.verify(verifier)) {
                 LOGGER.error(
                         LogHelper.buildLogMessage("Verifiable credential signature not valid"));
                 throw new VerifiableCredentialException(
                         HTTPResponse.SC_SERVER_ERROR,
                         ErrorResponse.FAILED_TO_VALIDATE_VERIFIABLE_CREDENTIAL);
             }
+        } catch (ParseException e) {
```
MikeCollingwood, line 152: We handle the parse exception for trying to parse a signing key here!

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -69,29 +73,40 @@ public List<VerifiableCredential> parseAndValidate(
...
         var validVcs = new ArrayList<VerifiableCredential>();
         for (String vc : vcs) {
             validVcs.add(
-                    parseAndValidate(userId, criId, vc, vcType, signingKey, componentId, false));
+                    parseAndValidate(
+                            userId,
+                            criId,
+                            vc,
+                            vcType,
+                            signingAlgorithm,
```
DanCorderIPV: Instead of passing in `signingAlgorithm`, `signingKey` and `componentId` separately could we pass in a CriConfig object instead?
@Joe-Edwards-GDS @blakeyp you may have opinions on this

local-running/setConfigForLocalOrCloudRunning.py
```diff
@@ -194,7 +194,8 @@ def get_local_running_params(environment, client_id_suffix):
             "tokenUrl":"http://host.docker.internal:3015/token",
             "credentialUrl":"http://host.docker.internal:3015/credentials/issue",
             "clientId":"ipv-core-dev{client_id_suffix}",
-            "signingKey":"{{\\"kty\\":\\"EC\\",\\"crv\\":\\"P-256\\",\\"x\\":\\"RBXnILIdExUEWUJMlYeD6agE8u9gGgA3InKrd5TKhhY\\",\\"y\\":\\"kKtt9v_xq9oqvv5_E8AHcV77IYQfyNwaTQyTYxdO_UM\\"}}",
+            "signingAlgorithm": "RSA",
```
DanCorderIPV: We should probably add `"EC"` values to the other CRIs here - assuming that the plan is to update the real configs with `signingAlgorithm` values too.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/EncryptionAlgorithm.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.di.ipv.core.library.exceptions;
```
DanCorderIPV: Not sure why this is in the `exceptions` package?

libs/cimit-service/src/test/java/uk/gov/di/ipv/core/library/service/CiMitServiceTest.java
```diff
@@ -449,6 +452,19 @@ void getContraIndicatorsVCJwtThrowsErrorWhenInvalidCimitKey() throws JsonProcess
...
+                        eq(SIGNED_CONTRA_INDICATOR_VC),
+                        eq(VerifiableCredentialConstants.SECURITY_CHECK_CREDENTIAL_TYPE),
+                        any(),
+                        eq("INVALID_CIMIT_KEY"),
+                        eq(CIMIT_COMPONENT_ID),
+                        eq(false)))
+                .thenThrow(
+                        new VerifiableCredentialException(
+                                HTTPResponse.SC_SERVER_ERROR,
+                                ErrorResponse.FAILED_TO_VALIDATE_VERIFIABLE_CREDENTIAL));
```
DanCorderIPV, line 464: It seems odd that you had to add this, but perhaps it should always have been here?
Not something you caused, but could these tests work with a real verifiableCredentialValidator instead of setting up a mock each time?

DanCorderIPV (changes requested):

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
Joe-Edwards-GDS: That seems fine to me.

One question I have here - isn't the algorithm captured in the signing key itself (the `alg` property)? Do we actually need a separate algorithm field at all?

blakeyp: The JWKs have `kty` which specifies the type of key

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -140,6 +154,28 @@ private void validateSignature(SignedJWT verifiableCredential, ECKey signingKey)
         }
     }
 
+    private JWSVerifier getVerifier(EncryptionAlgorithm signingAlgorithm, String signingKey)
+            throws ParseException, JOSEException {
+        return switch (signingAlgorithm) {
+            case EC -> new ECDSAVerifier(ECKey.parse(signingKey));
+            case RSA -> new RSASSAVerifier(RSAKey.parse(signingKey));
```
Joe-Edwards-GDS: Digging a bit further into our libraries, it looks like you might be able to do this completely automatically.

Have a look at `DefaultJWSVerifierFactory.createJWSVerifier()` and `JWK.parse()`

MikeCollingwood: I saw JWK.parse, and am now updating to use the key to determine the algorithm. Will also look at createJWSVerifier

local-running/setConfigForLocalOrCloudRunning.py
MikeCollingwood: It would've been ok since there would've been a default to EC. I'm now no longer storing the algorithm in the cir config anyway, since it can be determined from the signing key header

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
MikeCollingwood: Trying to play around with createJWSVerifier but it feels like it may not be as much of a simplification as wanted compared to the new version of the getVerifier method.

```
            var signingKeyJws = JWSObject.parse(signingKey);
            var keyFactory = KeyFactory.getInstance(signingKeyJws.getHeader().getAlgorithm().toString());
            keyFactory.generatePublic(new PKCS8EncodedKeySpec(signingKey.getBytes())
            );

            new DefaultJWSVerifierFactory().createJWSVerifier(signingKeyJws.getHeader(), Key);

            var formattedVc = EC.equals(signingKeyJws.getHeader().getAlgorithm()) ? transcodeSignatureIfDerFormat(vc) : vc;

```

libs/cimit-service/src/test/java/uk/gov/di/ipv/core/library/service/CiMitServiceTest.java
shivanshuit914: hm yes it can probably work without mock, I think its mocked because we are testing cimit service in its own isolation as a unit tests. 

may be we can add contract tests where we can test with real verifiableCredentialValidator like we did for other cri.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/TestFixtures.java
```diff
@@ -46,6 +46,9 @@ public interface TestFixtures {
...
 
+    String RSA_SIGNING_PRIVATE_JWK = RSA_ENCRYPTION_PRIVATE_KEY;
```
blakeyp: `RSA_ENCRYPTION_PRIVATE_KEY` isn't a JWK so best call this just `RSA_SIGNING_PRIVATE_KEY` ?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -127,30 +135,26 @@ private void tryParseVc(VerifiableCredential vc) {
...
-                            ? transcodeSignature(verifiableCredential)
-                            : verifiableCredential;
-        } catch (JOSEException | ParseException e) {
-            LOGGER.error(LogHelper.buildErrorMessage("Error transcoding signature", e));
-            throw new VerifiableCredentialException(
-                    HTTPResponse.SC_SERVER_ERROR,
-                    ErrorResponse.FAILED_TO_VALIDATE_VERIFIABLE_CREDENTIAL);
-        }
+            var signingKeyType = JWK.parse(signingKey).getKeyType();
+            var formattedVc = EC.equals(signingKeyType) ? transcodeSignatureIfDerFormat(vc) : vc;
```
blakeyp, line 142: aren't we skipping a check here `signatureIsDerFormat(verifiableCredential)` ?

blakeyp: ohh it's been extracted into a method, ignore

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/TestFixtures.java
shivanshuit914: Makes sense, updated it thanks

blakeyp (approved):

DanCorderIPV (approved):

PR #2004, opened 2024-06-12
PYIC-6058: evcs reads

sam803:
## Proposed changes

Adds EVCS reads (if enabled) to the CheckCoi and CheckExistingIdentity lambda handlers

### What changed

We need to read eventually from the EVCS store in these lambda handlers, if the EVCS store does not have any vcs then call the tactical store as normal

Since the EVCS store does not store the CRI-ID, a function to generate a map of issuers to CRI-IDs has been added

### Why did it change

To support the EVCS store

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6058


lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -223,4 +228,22 @@ private void sendAuditEvent(
...
+            credentials = evcsService.getVerifiableCredentials(userId, evcsAccessToken, CURRENT);
+        }
+        if (credentials == null || credentials.isEmpty()) {
+            credentials = verifiableCredentialService.getVcs(userId);
+        }
+        return Stream.concat(
+                        credentials.stream(),
+                        sessionCredentialsService.getCredentials(ipvSessionId, userId).stream())
+                .toList();
+    }
```
Joe-Edwards-GDS, line 278: I think it might be clearer if this was just `getOldIdentity` or similar (and you could have an equivalent `getNewIdentity` if you like). Would probably make the intent clearer...

(FYI this will probably conflict with @thebauSoftwire 's ticket)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -341,11 +343,18 @@ private String getSecretValue(String criId, String connection, String secretId)
...
+
+    private <T> T getCriConfigForType(
+            String connection, String criId, Class<T> configType, boolean failOnUnknown) {
         final String pathTemplate =
                 ConfigurationVariable.CREDENTIAL_ISSUERS.getPath() + "/%s/connections/%s";
         try {
             String parameter = ssmProvider.get(resolvePath(pathTemplate, criId, connection));
-            return OBJECT_MAPPER.readValue(parameter, configType);
+            return OBJECT_MAPPER
+                    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, failOnUnknown)
```
Joe-Edwards-GDS: Why is this optional? I think we should probably be consistent about failing on unknown properties, otherwise it's liable to cause some surprising behaviour...

Does this mean some of our CRI config is actually invalid?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -358,4 +367,15 @@ private <T> T getCriConfigForType(String connection, String criId, Class<T> conf
                             pathTemplate, e));
         }
     }
+
+    public Map<String, String> getCriConfigs() {
```
Joe-Edwards-GDS: Should this actually be `String getCriByIssuer(String issuer)`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -358,4 +367,15 @@ private <T> T getCriConfigForType(String connection, String criId, Class<T> conf
...
+        Map<String, String> configs = new HashMap<>();
+        for (var cri : CriIdentifer.values()) {
+            var criId = cri.getId();
+            var criConfig =
+                    getCriConfigForType(getActiveConnection(criId), criId, CriConfig.class, false);
```
Joe-Edwards-GDS: I'm a bit concerned that this only covers 'active' connections - what if the VC was generated by a (now inactive) connection.

You might need to use the SSM `getMultiple` to get all the connections for each CRI?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsCriConfig.java
Joe-Edwards-GDS, line 1: I don't think we can have this as a static singleton, because it might have weird caching characteristics with lambda.

Instead, we should be able to rely on the ConfigService caching, which will automatically cache calls to SSM for 3 minutes.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -341,11 +343,18 @@ private String getSecretValue(String criId, String connection, String secretId)
     }
 
     private <T> T getCriConfigForType(String connection, String criId, Class<T> configType) {
+        return getCriConfigForType(connection, criId, configType, true);
```
Joe-Edwards-GDS: Should we update this (and any other methods that take `credentialIssuerId` as an argument) to use the enum type instead?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriIdentifer.java
Joe-Edwards-GDS, line 1: I think this is a reasonable idea - there's probably more work to start using this enum wherever we want to use a CRI ID, but that's going to add even more noise to an already noisy review!

Joe-Edwards-GDS, line 1: Do you think `CriIdentifier` is right here? The thing that looks suspicious to me is calling `CriIdentifier.getId()`

Perhaps `CriType`, or even just `Cri`?

Joe-Edwards-GDS (commented):
This looks sensible to me, but I think we need to examine the way we map issuer -> CRI (perhaps as expected..!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriIdentifer.java
sam803: Yes, I figured the same thing (I tried putting it in VerifiableCredental but it would've required tonnes more changes)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
sam803: @Joe-Edwards-GDS We are only interested in the componentId so deserialized as a `CriConfig` (componentId, signingKey) but if we deserialized as `OauthCriConfig` we wouldn't need this option

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsCriConfig.java
sam803: ack, will remove

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: Ah I see - that makes some sense. I wonder if we can be cleverer about this... I'm thinking:
- Can we make the deserialisation automatically select the appropriate subclass? e.g. using `@JsonTypeInfo` and `@JsonSubTypes`
- Or can we mark the superclass as `@JsonIgnoreProperties(ignoreUnknown = true)`, but have the subclasses be more strict?

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
sam803: @Joe-Edwards-GDS Do you mean a getOldIdentity for the call to evcs/tactical and getNewIdentity for the call to sessionCredentialsService ?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
sam803: That would mean parsing the whole config multiple times if there are multiple cris in the vcs. Mayve  `String getCrisByIssuers(String ...issuers)` ?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
sam803: I started doing this and its resulted in lots more changes required! Might be worth tackling in a separate PR

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
Joe-Edwards-GDS: Yeah exactly

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: Ah I see. It's probably pretty fast, but it does seem slightly wasteful!

In that case, leave this as Map<String, String>, and then the consumer (i.e. the read VCs method) can call it once per batch of VCs.

Maybe call it `getCriIdsByIssuer` or something, given that it doesn't actually return any config?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: Yeah definitely - can you either raise a subtask or a separate tech debt task?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
sam803: I've used your original suggestion as was a bit messy in the calling code to get parse the JWTs extract the issuers then map to Cris

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
sam803: https://govukverify.atlassian.net/browse/PYIC-6816

.secrets.baseline
```diff
@@ -202,21 +202,21 @@
         "filename": "deploy/template.yaml",
         "hashed_secret": "38450ffe4ff65a68053ea5083d47521010709df2",
         "is_verified": false,
-        "line_number": 1994
+        "line_number": 1888
```
Joe-Edwards-GDS: Hmmm, looks like the line numbers have somehow made their way back in :(

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: Thanks!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
Joe-Edwards-GDS: Are you using this any more - can we just remove this again?

Joe-Edwards-GDS (commented):
One possible bit of cleanup and more secrets conflicts :( but looking reasonable, thanks!

sam803:
> One possible bit of cleanup and more secrets conflicts :( but looking reasonable, thanks!

Thanks, have done and regenerated the secrets baseline without line numbers

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -302,6 +309,19 @@ private JourneyResponse getJourneyResponse(
         }
     }
 
+    @Tracing
+    private List<VerifiableCredential> getVerifiableCredentials(
+            String userId, String evcsAccessToken)
+            throws CredentialParseException, EvcsServiceException {
+        if (configService.enabled(EVCS_READ_ENABLED)) {
+            var vcs = evcsService.getVerifiableCredentials(userId, evcsAccessToken, CURRENT);
```
Joe-Edwards-GDS, line 317: This is ok for now, but one thing we'll want to do (as mentioned on @rpayal's PR) is to avoid reading twice from EVCS (here and in the migration process) - we don't need to do a migration at all if the identity is already in EVCS.

It'll probably be easier to do this as part of the F2F journey work, since then we'll be reading CURRENT + PENDING, and need to handle the pending case separately.

Joe-Edwards-GDS (approved):

PR #2007, opened 2024-06-12
PYIC-6053 EVCS On-demand migration in case of reuse and repeat fraud journey

rpayal:
## Proposed changes
Enable EVCS On-demand migration in case of reuse and repeat fraud journey

### What changed
- When EVCS_WRITE flag is enabled then enabled EVCS migration in case of reuse and repeat fraud journey
- Also update migrated TS for VC's getting migrated.

### Issue tracking

- [PYIC-6053](https://govukverify.atlassian.net/browse/PYIC-6053)

### Other considerations

[PYIC-6053]: https://govukverify.atlassian.net/browse/PYIC-6053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -449,19 +458,28 @@ private JourneyResponse buildReuseResponse(
...
+            verifiableCredentialService.updateIdentity(credentials);
+        }
+    }
+
     private List<VerifiableCredential> allVcsExceptFraud(List<VerifiableCredential> vcs) {
         return vcs.stream().filter(vc -> !EXPERIAN_FRAUD_CRI.equals(vc.getCriId())).toList();
     }
 
-    private boolean hasCurrentFraudVc(List<VerifiableCredential> vcs) {
+    private boolean hasCurrentExpiredFraudVc(List<VerifiableCredential> vcs) {
```
Joe-Edwards-GDS: I don't agree with this rename - the method is checking whether the user has at least one valid (non-expired) fraud VC?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -422,10 +429,12 @@ private JourneyResponse buildReuseResponse(
...
         if (configService.enabled(REPEAT_FRAUD_CHECK)
                 && attainedVot.getProfileType() == GPG45
-                && !hasCurrentFraudVc(vcs)) {
+                && !hasCurrentExpiredFraudVc(vcs)) {
             LOGGER.info(LogHelper.buildLogMessage("Expired fraud VC found"));
+            var credentials = allVcsExceptFraud(vcs);
             sessionCredentialsService.persistCredentials(
-                    allVcsExceptFraud(vcs), auditEventUser.getSessionId(), false);
+                    credentials, auditEventUser.getSessionId(), false);
+            migrateCredentialsToEVCS(auditEventUser.getUserId(), credentials);
```
Joe-Edwards-GDS: As mentioned on the ticket, I think this _should_ include the expired fraud VC - it's still part of their 'current' identity.

Otherwise, consider the following scenario:
- User has DCMAW + Address + expired Fraud VC
- DCMAW + Address VCs are migrated
- User abandons the 6MFC
- User returns to the service
- Now we read from EVCS but only see DCMAW + Address
- We interpret this as an incomplete profile and route to a new identity journey (rather than 6MFC)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsMigrationService.java
```diff
@@ -0,0 +1,37 @@
...
+    public EvcsMigrationService(ConfigService configService) {
+        this.evcsClient = new EvcsClient(configService);
+    }
+
+    @Tracing
+    public void migrateExistingIdentity(String userId, List<VerifiableCredential> credentials)
+            throws EvcsServiceException {
+        List<EvcsCreateUserVCsDto> userVCsForEvcs =
+                credentials.stream()
+                        .filter(credential -> credential.getMigrated() != null)
```
Joe-Edwards-GDS: Do we need this filter?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -449,19 +458,28 @@ private JourneyResponse buildReuseResponse(
...
 
         return JOURNEY_REUSE;
     }
 
+    private void migrateCredentialsToEVCS(String userId, List<VerifiableCredential> credentials)
+            throws VerifiableCredentialException {
+        if (configService.enabled(EVCS_WRITE_ENABLED)) {
+            evcsMigrationService.migrateExistingIdentity(userId, credentials);
+            credentials.forEach(credential -> credential.setMigrated(Instant.now()));
+            verifiableCredentialService.updateIdentity(credentials);
```
Joe-Edwards-GDS: I wonder if this logic should also live inside `EvcsMigrationService`, so you can't migrate without doing it? Not super important

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -449,19 +458,28 @@ private JourneyResponse buildReuseResponse(
...
-                VcHelper.filterVCBasedOnProfileType(vcs, attainedVot.getProfileType()),
-                auditEventUser.getSessionId(),
-                false);
+                credentials, auditEventUser.getSessionId(), false);
+        migrateCredentialsToEVCS(auditEventUser.getUserId(), credentials);
 
         return JOURNEY_REUSE;
     }
 
+    private void migrateCredentialsToEVCS(String userId, List<VerifiableCredential> credentials)
```
Joe-Edwards-GDS: You might need to coordinate with @sam803 on this, but we need an additional check here, to make sure the user is not _already migrated_.

That is, if there are any `CURRENT` VCs in EVCS then we don't need to migrate.

Joe-Edwards-GDS: Oh I see - that's what the `filter` is for 🤔

I think we probably still need to think carefully about how this is going to work when we add EVCS reads.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: reverted

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: Done ...also done for reuse journey. 

Also observed ..we may need to do migration for '/journey/in-migration-reuse' or '/journey/operational-profile-reuse'.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsMigrationService.java
rpayal: refactored to move migrated TS update to evcsMigrationService.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: done - refactored to move migrated TS update to evcsMigrationService.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -123,9 +123,10 @@ private void persistEvcsUserVCs(
                                     existingPendingReturnEvcsUserVcsToUpdate.stream())
                             .toList();
 
-            evcsClient.updateUserVCs(userId, evcsUserVCsToUpdate);
+            if (evcsUserVCsToUpdate.size() > 0)
+                evcsClient.updateUserVCs(userId, evcsUserVCsToUpdate);
         }
-        evcsClient.storeUserVCs(userId, userVCsForEvcs);
+        if (userVCsForEvcs.size() > 0) evcsClient.storeUserVCs(userId, userVCsForEvcs);
```
Joe-Edwards-GDS: `if (!userVCsForEvcs.isEmpty())`?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Yep, I'll add a note to https://govukverify.atlassian.net/browse/PYIC-6083 - but it may never be necessary!

If we only turn on inherited ids _after_ EVCS migration, then there'll never be an operational profile to migrate :)

Joe-Edwards-GDS (approved):
We will still want to do something to avoid double-fetching the VCs from EVCS once reads are enabled (cc @sam803 ), but this should work for now.

PR #2008, opened 2024-06-12
PYIC-6767 Update evcs client.

rpayal:
## Proposed changes
- Update EVCS client to add error message from evcs call to log. 
- Also update urlBuilder to resolve url path correctly. 
- And throw no exception when 404 response received from EVCS.

### Issue tracking

- [PYIC-6767](https://govukverify.atlassian.net/browse/PYIC-6767)

### Other considerations

[PYIC-6767]: https://govukverify.atlassian.net/browse/PYIC-6767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
+++ b/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
@@ -27,12 +27,15 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static org.apache.http.HttpHeaders.AUTHORIZATION;
 import static org.apache.http.HttpHeaders.CONTENT_TYPE;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EVCS_APPLICATION_URL;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.EVCS_APP_ID;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_STATUS_CODE;
 
 public class EvcsClient {
@@ -50,6 +53,7 @@ public class EvcsClient {
         this.httpClient = HttpClient.newHttpClient();
     }
 
+    @ExcludeFromGeneratedCoverageReport
     public EvcsClient(ConfigService configService, HttpClient httpClient) {
         this.configService = configService;
         this.httpClient = httpClient;
@@ -156,11 +160,22 @@ public class EvcsClient {
     private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
             throws EvcsServiceException {
         if (200 > evcsHttpResponse.statusCode() || evcsHttpResponse.statusCode() > 299) {
+            String responseMessage;
+            try {
+                Map<String, String> responseBody =
+                        OBJECT_MAPPER.readValue(evcsHttpResponse.body(), new TypeReference<>() {});
+                responseMessage =
+                        Optional.ofNullable(responseBody.get("message"))
+                                .orElse("Received no evcs response body.");
+            } catch (JsonProcessingException e) {
+                responseMessage = "Failed to parse evcs response body.";
+            }
             LOGGER.info(
                     LogHelper.buildLogMessage(
                                     ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE
                                             .getMessage())
-                            .with(LOG_STATUS_CODE.getFieldName(), evcsHttpResponse.statusCode()));
+                            .with(LOG_STATUS_CODE.getFieldName(), evcsHttpResponse.statusCode())
+                            .with(LOG_MESSAGE_DESCRIPTION.getFieldName(), responseMessage));
             throw new EvcsServiceException(
                     HTTPResponse.SC_SERVER_ERROR,
                     ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE);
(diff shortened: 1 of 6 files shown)


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -72,7 +77,10 @@ public EvcsGetUserVCsDto getUserVcs(
             var evcsHttpResponse = sendHttpRequest(httpRequestBuilder.build());
 
             EvcsGetUserVCsDto evcsGetUserVCs =
-                    OBJECT_MAPPER.readValue(evcsHttpResponse.body(), new TypeReference<>() {});
+                    evcsHttpResponse.statusCode() != 404
```
Joe-Edwards-GDS, line 82: Ah, so it returns a 404 with a different response structure if the user has no VCs in the store? That's a bit annoying 😞

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -156,14 +167,27 @@ private URI getUri(String userId, List<EvcsVCState> vcStatesToQueryFor)
     private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
             throws EvcsServiceException {
         if (200 > evcsHttpResponse.statusCode() || evcsHttpResponse.statusCode() > 299) {
+            String responseMessage;
+            try {
+                Map<String, String> responseBody =
+                        OBJECT_MAPPER.readValue(evcsHttpResponse.body(), new TypeReference<>() {});
```
Joe-Edwards-GDS, line 181: If we do anything more complex than this, we should create a proper `EvcsErrorResponse` class, but this is fine for now.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -156,14 +167,27 @@ private URI getUri(String userId, List<EvcsVCState> vcStatesToQueryFor)
...
-                    ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE);
+                            .with(LOG_STATUS_CODE.getFieldName(), evcsHttpResponse.statusCode())
+                            .with(LOG_MESSAGE_DESCRIPTION.getFieldName(), responseMessage));
+            if (evcsHttpResponse.statusCode() != 404) {
+                throw new EvcsServiceException(
+                        HTTPResponse.SC_SERVER_ERROR,
+                        ErrorResponse.RECEIVED_NON_200_RESPONSE_STATUS_CODE);
+            }
         }
         LOGGER.info(LogHelper.buildLogMessage("Successful HTTP response from EVCS Api"));
```
Joe-Edwards-GDS, line 200: Do we still want this log message if it was a 404?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -140,9 +148,12 @@ public void updateUserVCs(String userId, List<EvcsUpdateUserVCsDto> evcsUserVCsT
 
     private URI getUri(String userId, List<EvcsVCState> vcStatesToQueryFor)
             throws URISyntaxException {
-        var uriBuilder =
-                new URIBuilder(configService.getSsmParameter(EVCS_APPLICATION_URL))
-                        .setPathSegments("vcs", userId);
+        URI evcsUri = new URI(configService.getSsmParameter(EVCS_APPLICATION_URL));
+        List<String> pathSegments =
+                evcsUri.getPath() != null
+                        ? List.of(evcsUri.getPath().replaceFirst("/", ""), "vcs", userId)
```
Joe-Edwards-GDS: Will this work if you have more than one path segment?

I know I asked you not to do this in the first place, but if we're having to break this down anyway, then it's probably easier just to construct it properly ourselves, i.e.:
```
var baseUri = "%s/vcs/%s".formatted(
        configService.getSsmParameter(EVCS_APPLICATION_URL),
        URLEncoder.encode(userId, StandardCharsets.UTF_8));

var uriBuilder = new URIBuilder(baseUri);

...
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: 404 is returned as error. (seems like for scenario when user does not exist).
Checked with Taher also for user with VC's but not with requested state ..this will also give 404 response.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: Good to print log for any response code other then 2xx

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
rpayal: Done and added test.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
Joe-Edwards-GDS: What I mean is that I think (for a 404) this will log:
> Non 200 HTTP response status code (status 404)
> Successful HTTP response from EVCS Api

Not a big deal, but slightly inaccurate.

You could fix it by
- putting the 404 check in the outer if, and treating it as a success
- putting the final log message in an `else`, and treating 404 as an error
- making 404 a special case with its own log message

Joe-Edwards-GDS (approved):
One suggestion for tweaking the 404 logs, but not required

PR #2015, opened 2024-06-13
PYIC-6724: Implement inbuilt back navigation

Wynndow:
### What changed

Implement inbuilt back navigation

### Why did it change

[PYIC-6724: Start reading user state from stack](https://github.com/govuk-one-login/ipv-core-back/commit/200ee830e3446da80e3d49e7d63677d89e0eee50)

This switches core-back to start using the users state and subjourney
type from the stack, rather than the individual fields in their session.
It removes those fields entirely.
This also updates the exception thrown if the stack is empty to a
runtime exception. Keeping it as checked would have made the
process-journey-event lambda excessivley verbose. And this is an
exception that shouldn't ever happen in reality.

[PYIC-6724: Implement back functionality in journey-engine](https://github.com/govuk-one-login/ipv-core-back/commit/39b7631d772da8030b491fa3d98d1ddea11ecd28)

This adds functionality to the journey engine to detect a back event,
and find a legitimate previous state for them from the stack.

Initially we will only support back where the current state and the
previous state are page states.

This change needs to take account of jumps between subjourneys, as well
as nested journeys.

This change removes the addition of JourneyChangeStates to the stack -
they make the logic a lot more complicated and add little of value.

This change also adds a new field to the states object - journeyType.
This in populated when the state machine is initialized, and makes
writing states to the stack a lot simpler, particularly when moving
between subjourneys.

When a user does go "back", this change actually adds the previous state
on top of the stack, rather than just popping off the top one. This will
give us a fuller picture of the users journey should we need it.

[PYIC-6724: Default to back event on state](https://github.com/govuk-one-login/ipv-core-back/commit/b4b00bf368f70b1068f39b94ba432b4d984e5745)

If a state has a back event explicitly defined we should use it rather
than the inbuilt mechanism. The motivation for this is that we currently
have back events defined that emit audit events defined in the journey
map. If we just use the inbuilt functionality we'd lose the audit
events.

This will also allow us to configure a back event that does something
other than go back to the user previous state, or to a non page state.
Such as in the passport mitigation journey.

[PYIC-6724: Remove back events from journey map](https://github.com/govuk-one-login/ipv-core-back/commit/e8d4a599b1fb79699abf9b32cbb30c4f3d92be2b)

This removes all the back events that the new functionality can replace,
leaving only those with audit events and those that go back to a
non-page state (CRI states).

### Issue tracking

- [PYIC-6724](https://govukverify.atlassian.net/browse/PYIC-6724)

[PYIC-6724]: https://govukverify.atlassian.net/browse/PYIC-6724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -44,13 +44,16 @@ import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownEve
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownStateException;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.BasicState;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.JourneyChangeState;
+import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.NestedJourneyInvokeState;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.State;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.JourneyContext;
+import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.PageStepResponse;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.ProcessStepResponse;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.StepResponse;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.List;
@@ -72,6 +75,7 @@ public class ProcessJourneyEventHandler
     private static final String END_SESSION_EVENT = "build-client-oauth-response";
     private static final StepResponse END_SESSION_RESPONSE =
             new ProcessStepResponse(END_SESSION_EVENT, null);
+    public static final String BACK_EVENT = "back";
     private final IpvSessionService ipvSessionService;
     private final AuditService auditService;
     private final ConfigService configService;
@@ -185,12 +189,13 @@ public class ProcessJourneyEventHandler
             journeyEvent = NEXT_EVENT;
         }
 
-        var initialJourneyState = ipvSessionItem.getState();
+        var currentJourneyState = ipvSessionItem.getState();
 
         try {
             var newState =
                     executeStateTransition(
-                            initialJourneyState,
+                            currentJourneyState,
+                            ipvSessionItem,
                             journeyEvent,
                             currentPage,
                             auditEventUser,
@@ -205,14 +210,13 @@ public class ProcessJourneyEventHandler
                                 .with(
                                         LOG_USER_STATE.getFieldName(),
                                         journeyChangeState.getInitialState()));
-                ipvSessionItem.pushState(
-                        journeyChangeState.getJourneyType(), journeyChangeState.getInitialState());
 
                 sendSubJourneyStartAuditEvent(
                         auditEventUser, journeyChangeState.getJourneyType(), deviceInformation);
                 newState =
                         executeStateTransition(
-                                ipvSessionItem.getState(),
+                                journeyStateFrom(journeyChangeState),
+                                ipvSessionItem,
                                 NEXT_EVENT,
                                 null,
                                 auditEventUser,
@@ -220,9 +224,9 @@ public class ProcessJourneyEventHandler
             }
 
             var basicState = (BasicState) newState;
-            ipvSessionItem.pushState(basicState.getName());
+            ipvSessionItem.pushState(basicState.getJourneyType(), basicState.getName());
 
-            logStateChange(initialJourneyState, journeyEvent, ipvSessionItem);
+            logStateChange(currentJourneyState, journeyEvent, ipvSessionItem);
 
             clearOauthSessionIfExists(ipvSessionItem);
 
@@ -259,12 +263,14 @@ public class ProcessJourneyEventHandler
     @Tracing
     private State executeStateTransition(
             JourneyState initialJourneyState,
+            IpvSessionItem ipvSessionItem,
             String journeyEvent,
             String currentPage,
             AuditEventUser auditEventUser,
             String deviceInformation)
             throws StateMachineNotFoundException, SqsException, UnknownEventException,
                     UnknownStateException {
+
         StateMachine stateMachine = stateMachines.get(initialJourneyState.subJourney());
         if (stateMachine == null) {
             throw new StateMachineNotFoundException(
@@ -277,6 +283,19 @@ public class ProcessJourneyEventHandler
                         String.format(
                                 "Found state machine for journey type: %s",
                                 initialJourneyState.subJourney())));
+
+        if (BACK_EVENT.equals(journeyEvent)) {
+            var previousJourneyState = ipvSessionItem.getPreviousState();
+
+            if (isPageState(initialJourneyState) && isPageState(previousJourneyState)) {
+                ipvSessionItem.pushState(previousJourneyState);
+                return journeyStateToBasicState(previousJourneyState);
+            }
+
+            throw new UnknownEventException(
+                    String.format(
+                            "Back event provided to state: '%s'", initialJourneyState.state()));
+        }
 
         var result =
                 stateMachine.transition(
@@ -400,5 +419,53 @@ public class ProcessJourneyEventHandler
                         auditEventUser,
                         new AuditExtensionSubjourneyType(journeyType),
                         new AuditRestrictedDeviceInformation(deviceInformation)));
+    }
+
+    private boolean isPageState(JourneyState journeyState)
+            throws StateMachineNotFoundException, UnknownStateException {
+        StateMachine stateMachine = stateMachines.get(journeyState.subJourney());
+        if (stateMachine == null) {
+            throw new StateMachineNotFoundException(
+                    String.format(
+                            "State machine not found for journey type: '%s'",
+                            journeyState.subJourney()));
+        }
+
+        var state =
+                getStateFromStatesMap(
+                        stateMachine.getStates(),
+                        new ArrayList<>(Arrays.asList(journeyState.state().split("/"))));
+        if (state == null) {
+            throw new UnknownStateException(
+                    String.format(
+                            "Unknown state provided. State machine: '%s', state: '%s'",
+                            journeyState.subJourney(), journeyState.state()));
+        }
+
+        return state instanceof BasicState basicState
+                && basicState.getResponse() instanceof PageStepResponse;
+    }
+
+    private State getStateFromStatesMap(Map<String, State> statesMap, List<String> stateParts) {
+        if (stateParts.size() == 1) {
+            return statesMap.get(stateParts.get(0));
+        } else {
+            // Recurse into nested states to find the actual state we care about
+            return getStateFromStatesMap(
+                    ((NestedJourneyInvokeState) statesMap.get(stateParts.remove(0)))
+                            .getNestedJourneyDefinition()
+                            .getNestedJourneyStates(),
+                    stateParts);
+        }
+    }
+
+    private BasicState journeyStateToBasicState(JourneyState journeyState) {
+        return (BasicState)
+                stateMachines.get(journeyState.subJourney()).getState(journeyState.state());
+    }
+
+    private JourneyState journeyStateFrom(JourneyChangeState journeyChangeState) {
+        return new JourneyState(
+                journeyChangeState.getJourneyType(), journeyChangeState.getInitialState());
     }
 }
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -44,16 +44,13 @@ import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownEve
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.exceptions.UnknownStateException;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.BasicState;
 import uk.gov.di.ipv.core.processjourneyevent.statemachine.states.JourneyChangeState;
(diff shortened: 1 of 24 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyState.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.ipv.core.library.domain;
+
+public record JourneyState(IpvJourneyTypes subJourney, String state) {
+    public static final String JOURNEY_STATE_DELIMITER = "/";
```
Joe-Edwards-GDS, line 4: If you're putting this here, maybe the serialisation/deserialisation should be too?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -253,41 +252,51 @@ private StepResponse executeJourneyEvent(
...
                         String.format(
                                 "Found state machine for journey type: %s",
-                                ipvSessionItem.getJourneyType().name())));
+                                initialJourneyState.subJourney())));
+
+        if (BACK_EVENT.equals(journeyEvent) && !isBackEventDefinedOnState(initialJourneyState)) {
+            var previousJourneyState = ipvSessionItem.getPreviousState();
+
+            if (isPageState(initialJourneyState) && isPageState(previousJourneyState)) {
+                ipvSessionItem.pushState(previousJourneyState);
```
Joe-Edwards-GDS: Do you think this is right - will this mean if you keep pressing back you'll toggle back-and-forth, rather than continuing to go back?

(Should it be `popState()` instead?)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -253,41 +252,51 @@ private StepResponse executeJourneyEvent(
...
+                            initialJourneyState.subJourney()));
         }
         LOGGER.debug(
                 LogHelper.buildLogMessage(
                         String.format(
                                 "Found state machine for journey type: %s",
-                                ipvSessionItem.getJourneyType().name())));
+                                initialJourneyState.subJourney())));
+
+        if (BACK_EVENT.equals(journeyEvent) && !isBackEventDefinedOnState(initialJourneyState)) {
```
Joe-Edwards-GDS, line 281: Do we have cases where `back` goes to a page that is not just the previous state? That's a bit annoying - but I guess a handy feature to allow.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -413,4 +417,33 @@ private void sendSubJourneyStartAuditEvent(
...
+        return new JourneyState(
+                journeyChangeState.getJourneyType(), journeyChangeState.getInitialState());
+    }
+
+    private boolean isBackEventDefinedOnState(JourneyState journeyState) {
+        return ((BasicState)
+                        stateMachines.get(journeyState.subJourney()).getState(journeyState.state()))
+                .getEvents()
+                .containsKey(BACK_EVENT);
+    }
```
Joe-Edwards-GDS, line 448: It doesn't quite feel right to have all these at the handler level, but I guess the trouble is you can't easily move it inside `StateMachine` because it might move between them :/

There's probably a refactoring begging to be let out, but perhaps not the time. (I also remember we had that chat about trying to rewrite this all in TS - perhaps that's the real answer :D )

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -413,4 +417,33 @@ private void sendSubJourneyStartAuditEvent(
...
+        if (stateMachine == null) {
+            throw new StateMachineNotFoundException(
+                    String.format(
+                            "State machine not found for journey type: '%s'",
+                            journeyState.subJourney()));
+        }
+        return stateMachine.isPageState(journeyState);
+    }
+
+    private BasicState journeyStateToBasicState(JourneyState journeyState) {
```
Joe-Edwards-GDS: Does this need to return a basic state?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -413,4 +417,33 @@ private void sendSubJourneyStartAuditEvent(
...
+                stateMachines.get(journeyState.subJourney()).getState(journeyState.state());
+    }
+
+    private JourneyState journeyStateFrom(JourneyChangeState journeyChangeState) {
+        return new JourneyState(
+                journeyChangeState.getJourneyType(), journeyChangeState.getInitialState());
+    }
+
+    private boolean isBackEventDefinedOnState(JourneyState journeyState) {
+        return ((BasicState)
```
Joe-Edwards-GDS: Can we guarantee this is a basic state?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
```diff
@@ -54,6 +59,36 @@ public TransitionResult transition(
...
+        if (stateParts.size() == 1) {
+            return statesMap.get(stateParts.get(0));
+        } else {
+            // Recurse into nested states to find the actual state we care about
+            return recurseToState(
+                    ((NestedJourneyInvokeState) statesMap.get(stateParts.remove(0)))
+                            .getNestedJourneyDefinition()
+                            .getNestedJourneyStates(),
+                    stateParts);
+        }
```
Joe-Edwards-GDS, line 89: A bit sad, but seems hard to avoid without a major refactor :/

The JJJ does it by exploding out the nested states during initialisation instead (so once the map is initialised there are no 'nested' states to deal with), which might be something to consider if we end up rewriting this.

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -206,6 +206,9 @@ private void findPageSpecificStatesAndEvents(
                     }
 
                     var pageEvents = new HashSet<>(basicState.getEvents().keySet());
+                    pageEvents.remove(
+                            "back"); // the back event is a special case that may or may not be
```
Joe-Edwards-GDS, line 211: Yeah - I also considered the option of still requiring it to be explicitly defined (but with no need to explicitly specify a target).

That would also let you attach audit events - which might be helpful for the user details aborted case?

```
events:
  next:
    targetState: something
  back:
  other:
    targetState: something


events:
  back:
    auditEvents:
      - IPV_GONE_BACK
```

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: I guess I was thinking that the user is only going to go 'back' once, and this would give us a fuller picture of there journey in the stack. But you're right, any more than once and it'll break. I'll switch it, it's probably not that important to have that information in the stack anyway.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: I don't think so? If you're referring to the second condition here, that's to allow us to override the default back behaviour. Useful for where we now have back events with an audit event attached in the journey map, and we also have a couple of cases of a back event leading to a CRI state.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: Yeah I agree - I wanted to push as much of this into the state machine, but it's hard hecause of the way our collection of state machines is structured. It would perhaps be better to just have on state machine that knows about all the journey maps.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: I think it returns a basic staate because it always will, given the context it's used in - it's only used on a journeyState we've already checked is a BasicState.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: No, possibly not. I'll update it to check before trying to cast it.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
Wynndow: Yeah, that might be a good way to go. The nested stuff is too complex.

Wynndow: Alternatively we just bin nested states. They're not as useful as we hoped they would be

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
Wynndow: Yeah - feels like we're edging closer to what we already have though - meaning we'd always need to go consult the journey map anyway.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/JourneyState.java
Wynndow: Which (de)serialisation? That's being handled for us now.
Do you mean the convenience stuff in the ipvSessionItem? If so, I'll see what I can squeeze in here.

Joe-Edwards-GDS: Yeah, I don't really mind - I think it's reasonable to leave serialisation as an implementation detail of the session, but then it's odd to keep the delimiter here!

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Yeah, the second condition. Going 'back' to a CRI state sounds odd - I wonder if they are really 'back' steps.

Still useful capability though :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Yep, agreed - the distinction is really for ease of definition. Optimising the execution isn't really that important (not until we get _really_ complicated!)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: Sort of is - it's for when a user is at a CRI and then abandon. We now show them a screen and let them go back to it if they want to try again - part of the F2F volumes stuff.

Although it is done from a radio button rather than a back link, so perhaps it should use an event other than back...

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachine.java
Joe-Edwards-GDS: Yeah, they might need overhauling slightly to get maximum value. They've been kinda useful for Address+Fraud, but it wouldn't have been that hard to just define them each time either.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: Ah ok - I might consider it more a 'try again' than a 'back', but don't worry too much for this PR.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: I think I meant - does the caller actually care that this is a Basic State?

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
Joe-Edwards-GDS: I guess we can add this if we need it

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (dismissed):

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: I think not - will get it to return the interface

Joe-Edwards-GDS (approved):

PR #2020, opened 2024-06-14
PYIC-6584: POC for async SQS messaging

Joe-Edwards-GDS:
Try creating an async SQS client, to potentially improve performance by parallelising AWS API calls.

The X-Ray traces show SQS requests running in parallel to others, so this appears to work. It'll be hard to tell without a proper performance test whether it makes an appreciable difference.

Other clients could also be made async, but the value will be reduced - e.g. most DynamoDB and SSM calls need to be synchronous as we have to wait for the result.

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -197,6 +197,8 @@ public class CheckExistingIdentityHandler
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
+        } finally {
+            auditService.awaitAuditEvents();
         }
     }
 
--- a/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
+++ b/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
@@ -2,48 +2,76 @@ package uk.gov.di.ipv.core.library.service;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
-import software.amazon.awssdk.services.sqs.SqsClient;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient;
+import software.amazon.awssdk.services.sqs.SqsAsyncClient;
 import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+import software.amazon.awssdk.services.sqs.model.SendMessageResponse;
 import uk.gov.di.ipv.core.library.auditing.AuditEvent;
 import uk.gov.di.ipv.core.library.exceptions.SqsException;
+import uk.gov.di.ipv.core.library.helpers.LogHelper;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 import static software.amazon.awssdk.regions.Region.EU_WEST_2;
 import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.SQS_AUDIT_EVENT_QUEUE_URL;
 
 public class AuditService {
-    private final SqsClient sqs;
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    private final SqsAsyncClient sqs;
     private final String queueUrl;
     private final ObjectMapper objectMapper;
 
-    public AuditService(SqsClient sqs, ConfigService configService) {
+    private List<CompletableFuture<SendMessageResponse>> events = new ArrayList<>();
+
+    public AuditService(SqsAsyncClient sqs, ConfigService configService) {
         this.sqs = sqs;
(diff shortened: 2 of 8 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -197,6 +197,8 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
+        } finally {
+            auditService.awaitAuditEvents();
```
Joe-Edwards-GDS, line 201: The scary thing about this approach is that if we _don't_ call this, then the lambda runtime will suspend and (potentially) interrupt an in-progress request.

I can't think of a great solution to this beyond 'make sure it's included in all the handlers' and 'never construct a new audit service outside of a handler'.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -2,48 +2,76 @@
...
         this.objectMapper = objectMapper;
     }
 
-    public static SqsClient getSqsClient() {
-        return SqsClient.builder()
+    public static SqsAsyncClient getSqsClient() {
+        return SqsAsyncClient.builder()
                 .region(EU_WEST_2)
-                .httpClientBuilder(UrlConnectionHttpClient.builder())
+                .httpClientBuilder(AwsCrtAsyncHttpClient.builder())
```
Joe-Edwards-GDS, line 48: The AWS docs seem to imply that we should use the AWS CRT client for the sync case as well (https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration.html#http-clients-recommend) - unless it's a snapstart thing to prefer the `UrlConnectionHttpClient`?

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -2,48 +2,76 @@
...
+
+    // Await audit events MUST be called before the end of each handler
+    public void awaitAuditEvents() {
+        try {
+            var eventsToAwait = events;
+            events = new ArrayList<>();
+            CompletableFuture.allOf(eventsToAwait.toArray(new CompletableFuture[0])).get();
+        } catch (InterruptedException | ExecutionException e) {
+            LOGGER.error(LogHelper.buildErrorMessage("Failed to send audit event(s)", e));
+            throw new AuditException("Failed to send audit event(s)", e);
```
Joe-Edwards-GDS, line 74: This is a RuntimeException to ease its use - this means an audit failure will fail the lambda, but this is probably a good thing!

I'm in discussion about whether any of the audit events can be marked as non-critical (i.e. we can ignore if they fail), but for now let's assume that they all are, because TICF may need them.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -2,48 +2,76 @@
...
         }
     }
+
+    // Await audit events MUST be called before the end of each handler
+    public void awaitAuditEvents() {
+        try {
+            var eventsToAwait = events;
+            events = new ArrayList<>();
+            CompletableFuture.allOf(eventsToAwait.toArray(new CompletableFuture[0])).get();
+        } catch (InterruptedException | ExecutionException e) {
```
Wynndow, line 72: If we get an InterruptedException (which I can't imagine we will), do we need to call `Thread.interrupted()`?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Do you think we'd catch this in the pipeline before any mistakes hit prod?

Joe-Edwards-GDS: Probably not, which is the scary part :/

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
Wynndow, line 48:
```suggestion
                .httpClient(AwsCrtAsyncHttpClient.create())
```

Joe-Edwards-GDS: What's the reason for preferring this? In any case, I probably won't change for this PR, but whoever picks up https://govukverify.atlassian.net/browse/PYIC-6883 can update if they wish

Wynndow: I think functionaly there's no difference, but if I was looking at it for the first time I'd wonder why we have to pass in a builder? Is something weird happening under the hood (it's not). But the way i've suggested seems more...expected?

Joe-Edwards-GDS: The docs show using builder unless you want to provide your own shared HTTP Client, but I suspect in our case it'll make no difference. (https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration-crt.html)

I wonder if it changes the laziness at all (e.g. will it defer creating the client until it needs it?)

Joe-Edwards-GDS:
Superseded by https://github.com/govuk-one-login/ipv-core-back/pull/2123

PR #2025, opened 2024-06-17
PYIC-none: refactor AuditEvent to have factory methods

thebauSoftwire:
### What changed
- Adds AuditEvent factory methods for creating events with/without device information and apply change to existing audit events
- Fix small bug for `AuditRestrictedInheritedIdentity` device info

### Why did it change
Where possible, all audit events should include the device information in the restricted section. The device information is only available in requests that come from core-front so aren't available in lambdas like `build-user-identity` or `process-async-cri-callback`. When it is available, it should be enforced that device information is required.


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEventFactory.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.di.ipv.core.library.auditing;
+
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensions;
+import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestricted;
+import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestrictedDeviceInformation;
+
+public class AuditEventFactory {
+    public AuditEvent createAuditEventWithDeviceInformation(
```
Joe-Edwards-GDS: I think these could be static, and probably just live on `AuditEvent` (this would let you restrict the existing constructors to force you to use one of these)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEventWithDeviceInformation.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.di.ipv.core.library.auditing;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensions;
+import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestrictedDeviceInformation;
+
+public class AuditEventWithDeviceInformation extends AuditEvent {
```
Joe-Edwards-GDS: I don't think you need this class - it's identical to `AuditEvent` except it has a different constructor (which I think you're implementing with a factory method instead)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/restricted/AuditRestrictedCheckCoi.java
```diff
@@ -9,11 +9,27 @@
...
-public record AuditRestrictedCheckCoi(
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<Name> oldName,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<Name> newName,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<BirthDate> oldBirthDate,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<BirthDate> newBirthDate,
-        @JsonProperty(value = "device_information", required = true)
-                DeviceInformation deviceInformation)
-        implements AuditRestricted {}
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class AuditRestrictedCheckCoi extends AuditRestrictedDeviceInformation {
```
Joe-Edwards-GDS: This it could be worth considering whether an interface would actually reduce code - consider:
```
interface AuditRestrictedWithDeviceInformation extends AuditRestricted {
    DeviceInformation deviceInformation();
}

public record AuditRestrictedCheckCoi(
    List<Name> oldName,
    List<Name> newName,
    List<BirthDate> oldBirthDate,
    List<BirthDate> newBirthDate,
    @JsonProperty("device_information")
    DeviceInformation deviceInformation)
  implements AuditRestrictedWithDeviceInformation {}

public record AuditRestrictedDeviceInformation(
    @JsonProperty("device_information")
    DeviceInformation deviceInformation
) implements AuditRestrictedWithDeviceInformation {
    public AuditRestrictedDeviceInformation(String deviceInformation) {
        this(new DeviceInformation(deviceInformation));
    }
}
```
But tbh I think it's always going to be a bit boilerplatey.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEvent.java
```diff
@@ -70,4 +71,22 @@ public AuditEvent(
             AuditRestricted restricted) {
         this(eventName, componentId, user, null, restricted);
     }
+
+    public static AuditEvent createAuditEventWithDeviceInformation(
```
Joe-Edwards-GDS: If they're in this class, you could probably drop duplicate `AuditEvent` - e.g. `AuditEvent.createWithDeviceInformation`, but whatever you think is clearest :)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/AuditEvent.java
```diff
@@ -35,7 +36,7 @@ public class AuditEvent {
     private final AuditRestricted restricted;
 
     @JsonCreator
-    public AuditEvent(
+    private AuditEvent(
```
Joe-Edwards-GDS, line 39: exactly - and you could remove/privatise the remaining constructors as well (but that'll require refactoring all the existing audit events)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/restricted/AuditRestrictedCheckCoi.java
```diff
@@ -10,10 +10,10 @@
...
 @ExcludeFromGeneratedCoverageReport
 public record AuditRestrictedCheckCoi(
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<Name> oldName,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<Name> newName,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<BirthDate> oldBirthDate,
-        @JsonInclude(JsonInclude.Include.NON_NULL) List<BirthDate> newBirthDate,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<Name> oldName,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<Name> newName,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<BirthDate> oldBirthDate,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @JsonProperty List<BirthDate> newBirthDate,
```
Joe-Edwards-GDS: I don't think you should need `@JsonProperty` here, but worth checking.

libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/helpers/AuditExtensionsHelperTest.java
```diff
@@ -109,7 +109,7 @@ void getAuditRestrictedInheritedIdentityShouldReturnTheRightStuff() throws Excep
                 "[{\"personalNumber\":\"AB123456C\"}]",
                 OBJECT_MAPPER.writeValueAsString(restricted.socialSecurityRecord()));
         assertEquals(
-                "{\"device_information\":{\"encoded\":\"test_device_data\"}}",
+                "{\"encoded\":\"test_device_data\"}",
                 OBJECT_MAPPER.writeValueAsString(restricted.deviceInformation()));
```
Joe-Edwards-GDS, line 113: A bit odd that this has changed? Can we double-check that we haven't accidentally changed the format of the device information (by checking logs in dev, for example)

Joe-Edwards-GDS (dismissed):
This looks good to me, but would be good to double check that we haven't inadvertently changed the format before merging (if you haven't already)

libs/audit-service/src/test/java/uk/gov/di/ipv/core/library/auditing/helpers/AuditExtensionsHelperTest.java
thebauSoftwire: Ahh yep, originally the `IPV_INHERITED_IDENTITY_VC_RECEIVED` restricted section looked like:
```
restricted: {
  deviceInformation: {
    device_information: {}
  }
}
```
but it should look like this instead:
```
restricted: {
  device_information: {}
}
```
Confirmed with Patrick that this was a bug and is fine to change :) (but I did forget to re-name the json property from deviceInformation -> device_information)

Joe-Edwards-GDS: Nice spot!

Joe-Edwards-GDS (approved):

PR #2030, opened 2024-06-18
PYIC-6053 Change to handle f2f vcs in pending_return state in evcs.

rpayal:
## Proposed changes
Correctly handle vcs - during migration - to handle f2f vcs in pending_return state in evcs.

### What changed
Update logic to evaluate vcs to store in case of storeCompleteIdentity, to handle f2f vcs in pending_return state in evcs.

### Issue tracking

- [PYIC-6053](https://govukverify.atlassian.net/browse/PYIC-6053)

### Other considerations

[PYIC-6053]: https://govukverify.atlassian.net/browse/PYIC-6053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
+++ b/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
@@ -102,7 +102,7 @@ public class EvcsService {
             List<EvcsGetUserVCDto> existingEvcsUserVCs,
             boolean isPendingIdentity)
             throws EvcsServiceException {
-        List<EvcsCreateUserVCsDto> userVCsForEvcs =
+        List<EvcsCreateUserVCsDto> userVCsToStore =
                 credentials.stream()
                         .filter(
                                 credential ->
@@ -122,8 +122,17 @@ public class EvcsService {
                                                 null))
                         .toList();
 
-        if (!CollectionUtils.isEmpty(existingEvcsUserVCs) && !isPendingIdentity) {
-            var existingCurrentEvcsUserVcsToUpdate =
+        updateExistingUserVCs(userId, credentials, existingEvcsUserVCs);
+        if (!userVCsToStore.isEmpty()) evcsClient.storeUserVCs(userId, userVCsToStore);
+    }
+
+    private void updateExistingUserVCs(
+            String userId,
+            List<VerifiableCredential> credentials,
+            List<EvcsGetUserVCDto> existingEvcsUserVCs)
+            throws EvcsServiceException {
+        if (!CollectionUtils.isEmpty(existingEvcsUserVCs)) {
+            var existingCurrentUserVcsNotInSessionToUpdate =
                     existingEvcsUserVCs.stream()
                             .filter(vc -> vc.state().equals(CURRENT))
                             .filter(
@@ -139,9 +148,17 @@ public class EvcsService {
                                             new EvcsUpdateUserVCsDto(
                                                     getVcSignature(vc), EvcsVCState.HISTORIC, null))
                             .toList();
-            var existingPendingReturnEvcsUserVcsToUpdate =
+            var existingPendingReturnUserVcsNotInSessionToUpdate =
                     existingEvcsUserVCs.stream()
                             .filter(vc -> vc.state().equals(PENDING_RETURN))
+                            .filter(
+                                    vc ->
+                                            credentials.stream()
+                                                    .noneMatch(
+                                                            credential ->
+                                                                    credential
+                                                                            .getVcString()
+                                                                            .equals(vc.vc())))
                             .map(
                                     vc ->
                                             new EvcsUpdateUserVCsDto(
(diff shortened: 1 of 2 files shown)


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -109,10 +109,12 @@ private void persistEvcsUserVCs(
...
-                                                                evcsVC.vc()
-                                                                        .equals(
-                                                                                credential
-                                                                                        .getVcString())))
+                                                                (evcsVC.vc()
+                                                                                .equals(
+                                                                                        credential
+                                                                                                .getVcString())
+                                                                        && evcsVC.state()
+                                                                                .equals(CURRENT))))
```
Joe-Edwards-GDS: I think this is not _quite_ right - this will recreate the VC, but instead you want to update it to `CURRENT`.

Instead, there might be a new category in the updates:
|  | CURRENT | PENDING |
| ------------- | ------------- | ------------- |
| In session  | No change  | Update to CURRENT |
| Not in session  | Update to HISTORIC  | Update to ABANDONED |

At the moment I think we're doing the bottom row, but not the top-right.

rpayal: yes we are doing the bottom one for update.

Joe-Edwards-GDS: I think what you've implemented is slightly different: you're updating the existing record to ABANDONED, and then creating a new CURRENT record. Instead I think we should just update it directly to CURRENT?

To expand the table out in all cases:

When storing a current identity:
| EVCS | CURRENT | PENDING | not present |
| ------------- | ------------- | ------------- | ------------- |
| In session  | No change  | Update to CURRENT | Create new CURRENT |
| Not in session  | Update to HISTORIC  | Update to ABANDONED | - |

When storing a pending identity:
| EVCS | CURRENT | PENDING | not present |
| ------------- | ------------- | ------------- | ------------- |
| In session  | -  | - | Create new PENDING |
| Not in session  | -  | Update to ABANDONED | - |

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -139,26 +148,51 @@ private void persistEvcsUserVCs(
...
+                                            credentials.stream()
+                                                    .anyMatch(
+                                                            credential ->
+                                                                    credential
+                                                                            .getVcString()
+                                                                            .equals(vc.vc())))
+                            .map(
+                                    vc ->
+                                            new EvcsUpdateUserVCsDto(
+                                                    getVcSignature(vc), EvcsVCState.CURRENT, null))
```
Joe-Edwards-GDS: N.B. the CURRENT and HISTORIC updates should only happen when storing current identities (not pending ones). I don't think this can actually happen in practice, but might be worth adding a guard?

(If we're storing a pending identity, it's just the pending -> abandoned that we care about)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -139,26 +148,51 @@ private void persistEvcsUserVCs(
...
+                            .toList();
+            var userVCsToUpdate =
                     Stream.concat(
-                                    existingCurrentEvcsUserVcsToUpdate.stream(),
-                                    existingPendingReturnEvcsUserVcsToUpdate.stream())
+                                    Stream.concat(
+                                            existingCurrentUserVcsNotInSessionToUpdate.stream(),
+                                            existingPendingReturnUserVcsNotInSessionToUpdate
+                                                    .stream()),
+                                    existingPendingReturnUserVcsInSessionToUpdate.stream())
```
Joe-Edwards-GDS: (If you drop the `.toList()`s on the variables, then you don't need the `.stream()`s here)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -131,68 +133,67 @@ private void updateExistingUserVCs(
...
+                                                                                .getVcString()
+                                                                                .equals(vc.vc())))
+                                .map(
+                                        vc ->
+                                                new EvcsUpdateUserVCsDto(
+                                                        getVcSignature(vc),
+                                                        EvcsVCState.CURRENT,
+                                                        null));
+        var existingCurrentUserVcsNotInSessionToUpdate =
+                existingPendingReturnUserVcsInSessionToUpdate.get().findAny().isPresent()
```
Joe-Edwards-GDS: I think this should be the `isPendingIdentity` parameter?

If it's a pending identity, then we only care about `existingPendingReturnUserVcsNotInSessionToUpdate`

If it's a current identity, then we _also_ care about `existingPendingReturnUserVcsInSessionToUpdate` and `existingCurrentUserVcsNotInSessionToUpdate`

You might find it easier to refactor like:

```
var vcUpdates = new ArrayList<EvcsUpdateUserVCsDto>();

// Pending VCs that are not in the session identity become abandoned
var existingPendingReturnUserVcsNotInSessionToUpdate = [...];
vcUpdates.addAll(existingPendingReturnUserVcsNotInSessionToUpdate)

if (!isPendingIdentity) {
  // Pending VCs that are in the session identity become current
  var existingPendingReturnUserVcsInSessionToUpdate = [...];
  vcUpdates.addAll(existingPendingReturnUserVcsInSessionToUpdate);

  // Current VCs that are not in the session identity become historic
  var existingCurrentUserVcsNotInSessionToUpdate = [...];
  vcUpdates.addAll(existingCurrentUserVcsNotInSessionToUpdate);
}

if (!vcUpdates.isEmpty()) {
  evcsClient.updateUserVCs(userId, vcUpdates);
}
```

rpayal: Refactored code to use available isPendingIdentity flag to update pending_return to current for completed identity only.

Joe-Edwards-GDS (approved):
Looks good, thanks :)

I think it's important to get this right, otherwise we'll end up with weird bugs later!

PR #2033, opened 2024-06-19
PYIC-6066: use f2f vcs with pending return from evcs 

sam803:
## Proposed changes

Read pending return VCs from evcs store 

### What changed

- Use pending return vcs to evaluate identity if they are available and EVCS read is enabled
- Face-to-face considered complete if we have at least one f2f pending return vc 
- Route to a new sub journey if we need to update the pending return vcs before continuing on the reuse journey

### Why did it change

To make use of the pending return state in EVCS

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6066


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -74,7 +74,9 @@ public List<VerifiableCredential> getVerifiableCredentials(
             try {
                 var jwt = SignedJWT.parse(vc.vc());
                 var cri = configService.getCriByIssuer(jwt.getJWTClaimsSet().getIssuer());
-                credentials.add(VerifiableCredential.fromValidJwt(userId, cri.getId(), jwt));
+                var credential = VerifiableCredential.fromValidJwt(userId, cri.getId(), jwt);
+                credential.setPendingReturn(PENDING_RETURN.equals(vc.state()));
```
Joe-Edwards-GDS: If we're going to do this, I think we should make it part of the constructor, and just store the `State` directly.

The challenge we have is that we cannot distinguish these states in VCs loaded from our tactical/session stores, so it might be preferable to keep that separate (see comment in CheckExistingIdentityHandler).

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -229,12 +233,22 @@ private JourneyResponse getJourneyResponse(
             AuditEventUser auditEventUser =
                     new AuditEventUser(userId, ipvSessionId, govukSigninJourneyId, ipAddress);
 
-            var vcs = getVerifiableCredentials(userId, clientOAuthSessionItem.getEvcsAccessToken());
+            var evcsAccessToken = clientOAuthSessionItem.getEvcsAccessToken();
+            var vcs = getVerifiableCredentials(userId, evcsAccessToken);
+            var pendingReturnVcs = vcs.stream().filter(vc -> vc.isPendingReturn()).toList();
```
Joe-Edwards-GDS: I feel a bit uncomfortable about this - this won't work if the VCs have been read from anywhere except EVCS. In the current state it'll work because we only care about the result if EVCS reads are enabled, but it seems hard to reason about.

I wonder if we actually want EVCS to return some other data structure when you read, e.g. a `Map<EvcsVCState, List<VerifiableCredential>>`, or some other wrapper class? That way it's really clear which set of VCs you're looking at?

Otherwise I think we need to make it clear that the `isPending()` method _only works_ if you're using VCs read from EVCS (e.g. failing hard if you try and use it in any other scenario).

Joe-Edwards-GDS (commented):
Nothing to do as part of this review, but a reminder to myself... We cannot enable this until at least 10 days after starting EVCS writes.

Otherwise we'll get users that may
- Start a F2F journey before writes are enabled
- Go to the post office after writes are enabled
- Return with _only_ a F2F VC in EVCS (rather than a complete pending identity)

If we wanted to be more cautious, we could add another feature flag, or perhaps identify this case specifically as a non-migrated user.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: Yes ok I like the EVCS Map idea - will look into it :)

sam803: I've added a combo of he Map function and an inner class to keep track of the properites (hasF2f / hasPending)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -185,6 +190,23 @@ public CheckExistingIdentityHandler() {
         VcHelper.setConfigService(this.configService);
     }
 
+    private class VerifiableCredentials extends ArrayList<VerifiableCredential> {
```
Joe-Edwards-GDS: Extending `ArrayList` seems like a pretty weird thing to do...

I'd much prefer a wrapper class here, perhaps something like
```
private record VerifiableCredentialBundle(
  List<VerifiableCredential> credentials,
  boolean isF2FIdentity,
  boolean isPendingEvcsIdentity) {}
```
(Also clarifying that the pending only applies to EVCS)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -310,16 +339,23 @@ private JourneyResponse getJourneyResponse(
...
         if (configService.enabled(EVCS_READ_ENABLED)) {
-            var vcs = evcsService.getVerifiableCredentials(userId, evcsAccessToken, CURRENT);
-            if (vcs != null && !vcs.isEmpty()) {
-                return vcs;
+            var vcs =
+                    evcsService.getVerifiableCredentialsByState(
+                            userId, evcsAccessToken, CURRENT, PENDING_RETURN);
+            var pendingReturnVcs = vcs.get(PENDING_RETURN);
+            // use pending return vcs to determine identity if available
+            if (pendingReturnVcs != null && !pendingReturnVcs.isEmpty()) {
```
Joe-Edwards-GDS: (You can use `CollectionUtils.isNullOrEmpty` if you prefer)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -66,15 +68,30 @@ public void storePendingIdentity(
...
                 var jwt = SignedJWT.parse(vc.vc());
                 var cri = configService.getCriByIssuer(jwt.getJWTClaimsSet().getIssuer());
-                credentials.add(VerifiableCredential.fromValidJwt(userId, cri.getId(), jwt));
+                var credential = VerifiableCredential.fromValidJwt(userId, cri.getId(), jwt);
+                var current = credentials.get(vc.state());
+                if (current == null) {
+                    current = new ArrayList<VerifiableCredential>();
+                }
+                current.add(credential);
+                credentials.put(vc.state(), current);
```
Joe-Edwards-GDS: Not important, but you could save a couple of lines here:
```
if (!credentials.containsKey(vc.state())) {
  credentials.put(vc.state(), new ArrayList<VerifiableCredential>());
}
credentials.get(vc.state()).add(credential);
```

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reuse-existing-identity.yaml
```diff
@@ -23,8 +23,33 @@ states:
       next:
         targetState: UPDATE_DETAILS_PAGE
 
+  UPDATE_IDENTITY_START:
```
Joe-Edwards-GDS: Minor, but I wonder if it would be more accurate to describe this as store identity, perhaps:
- `REUSE_WITH_STORE_IDENTITY` for this state
- `STORE_NEW_IDENTITY` for the one below

(They may not have a prior identity to 'update')

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/reuse-existing-identity.yaml
```diff
@@ -23,8 +23,33 @@ states:
...
+        targetState: UPDATE_IDENTITY_BEFORE_IDENTITY_REUSE_PAGE
+
   # Journey states
 
+  UPDATE_IDENTITY_BEFORE_IDENTITY_REUSE_PAGE:
+    response:
+      type: process
+      lambda: store-identity
+      lambdaInput:
+        identityType: UPDATE
```
Joe-Edwards-GDS: This is a bit TBD, but at the moment this will always be a `NEW` identity - there aren't any update journeys that support F2F.

(If we start supporting it in future, we'll need to consider this more carefully..!)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -488,7 +524,8 @@ private JourneyResponse buildReuseResponse(
                 auditEventUser.getSessionId(),
                 false);
         migrateCredentialsToEVCS(userId, vcs, evcsAccessToken);
```
Joe-Edwards-GDS: I think we need to consider the interaction with migration here as well - if the user is returning from F2F then we shouldn't be running the migration process on the pending VCs (that should be covered by the new store identity step below).

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: I forgot about your preference for composition! It was mainly so I wouldn't have to replace anything that uses the list with list.credentials but happy either way :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -185,6 +190,13 @@ public CheckExistingIdentityHandler() {
         VcHelper.setConfigService(this.configService);
     }
 
+    private record VerifiableCredentialBundle(
+            List<VerifiableCredential> credentials, boolean isPendingEvcsIdentity) {
+        private boolean isF2fIdenity() {
```
Joe-Edwards-GDS: typo `idenity`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -484,18 +508,25 @@ private JourneyResponse buildReuseResponse(
...
-                VcHelper.filterVCBasedOnProfileType(vcs, attainedVot.getProfileType()),
+                VcHelper.filterVCBasedOnProfileType(
+                        vcBundle.credentials, attainedVot.getProfileType()),
                 auditEventUser.getSessionId(),
                 false);
-        migrateCredentialsToEVCS(userId, vcs, evcsAccessToken);
-        return JOURNEY_REUSE;
+        migrateCredentialsToEVCS(userId, vcBundle, evcsAccessToken);
+
+        return vcBundle.isPendingEvcsIdentity && vcBundle.isF2fIdenity()
```
Joe-Edwards-GDS: Do we need `vcBundle.isF2fIdentity` here? If you've got a valid pending identity, then I think we should store it - regardless of whether it's F2F (in practice it always will be!)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -484,18 +508,25 @@ private JourneyResponse buildReuseResponse(
...
 
     private void migrateCredentialsToEVCS(
-            String userId, List<VerifiableCredential> credentials, String evcsAccessToken)
+            String userId, VerifiableCredentialBundle vcBundle, String evcsAccessToken)
             throws EvcsServiceException, VerifiableCredentialException {
-        if (configService.enabled(EVCS_WRITE_ENABLED)) {
-            evcsMigrationService.migrateExistingIdentity(userId, credentials, evcsAccessToken);
+        if (configService.enabled(EVCS_WRITE_ENABLED)
+                && !vcBundle.isPendingEvcsIdentity
+                && !vcBundle.isF2fIdenity()) {
```
Joe-Edwards-GDS: Similar to above - we can probably just check `isPendingEvcsIdentity`.

In actual fact, we really want to check `isEvcsIdentity` - if it's the current EVCS identity then we shouldn't migrate either (although it will be a no-op in that case).

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: good spot :)

Joe-Edwards-GDS (approved):
Thanks!

PR #2037, opened 2024-06-19
PYIC-6710: exclude secrets from process cri callback

thebauSoftwire:
### What changed
- excluding secrets from `process-cri-callback`

### Why did it change

The .secrets.baseline file was being updated with new secrets every time a "secret" was detected. Instead, we should use rules to avoid false positives being added to the baseline. This means we'll have less conflicts as the only changes made to the baseline should be if the filters are updated.

### Issue tracking

- [PYIC-6710](https://govukverify.atlassian.net/browse/PYIC-6710)

[PYIC-6710]: https://govukverify.atlassian.net/browse/PYIC-6710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -244,45 +244,6 @@
         "line_number": 306
       }
     ],
-    "lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/pact/ticfCri/ContractTest.java": [
-      {
-        "type": "Base64 High Entropy String",
-        "filename": "lambdas/call-ticf-cri/src/test/java/uk/gov/di/ipv/core/callticfcri/pact/ticfCri/ContractTest.java",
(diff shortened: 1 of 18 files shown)


.secrets.baseline
```diff
@@ -115,7 +115,8 @@
       "pattern": [
         "(?i)dummyapikey",
         "(?i)test-secret",
-        "^x-api-key$"
+        "^x-api-key$",
+        "^1f9d73167e2166b707c6$"
```
thebauSoftwire, line 119: Same comment as here: https://github.com/govuk-one-login/ipv-core-back/pull/2036#discussion_r1647681199

sam803 (dismissed):

Joe-Edwards-GDS:
Superseded by other PRs

PR #2040, opened 2024-06-20
PYIC-6868: Deploy internal test API to build

Wynndow:
### What changed

Deploy internal test API to build

### Why did it change

This updates the conditional in the CF template to deploy the testing API and associated resources into the build env, as well as the dev env. This will allow us to run the API tests in the build environment.

### Issue tracking

- [PYIC-6868](https://govukverify.atlassian.net/browse/PYIC-6868)

[PYIC-6868]: https://govukverify.atlassian.net/browse/PYIC-6868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -104,6 +104,9 @@ Conditions:
   IsDev01: !Equals [ !Ref AWS::AccountId, "130355686670"]
   IsDev02: !Equals [ !Ref AWS::AccountId, "175872367215"]
   IsNotDevelopment: !Not [ !Condition IsDevelopment ]
+  IsDevOrBuild: !Or
+    - !Condition IsDevelopment
+    - !Equals [ !Ref Environment, build ]
```
Joe-Edwards-GDS: FWIW I think this might be slightly less brittle if it's named `HasInternalTestApi` or something, but it's not a big deal.

(In future you might change which environments the test api is deployed to, but you might not want to change other things conditional on 'dev or build')

Joe-Edwards-GDS (dismissed):

deploy/template.yaml
Wynndow: Yeah, I did think about that, and then didn't do it, for no really good reason. I'll fix it up.

Joe-Edwards-GDS (approved):

PR #2045, opened 2024-06-20
PYIC-6059 Added audit event on migration when EVCS_WRITE is enabled.

rpayal:
## Proposed changes
Audit event when migrating

### What changed
Added audit event on migration when EVCS_WRITE is enabled.
- set identity_type as 'update' as its an existing identity.

### Issue tracking

- [PYIC-6059](https://govukverify.atlassian.net/browse/PYIC-6059)

### Other considerations

[PYIC-6059]: https://govukverify.atlassian.net/browse/PYIC-6059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -528,6 +544,23 @@ private void sendAuditEvent(
...
+            Vot vot,
+            String deviceInformation)
+            throws SqsException {
+        auditService.sendAuditEvent(
+                AuditEvent.createWithDeviceInformation(
+                        auditEventTypes,
+                        configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                        auditEventUser,
+                        new AuditExtensionIdentityType(
+                                IdentityType.UPDATE, vot.equals(P0) ? null : vot),
```
Joe-Edwards-GDS: I think we need to agree what extensions should be included on this event - sorry I thought I'd marked the ticket as Blocked while we were working it out.

I don't think identity type really makes sense here - it's none of the three categories, it's migrating an existing identity. Instead, can we include the signatures of the VCs being recorded?

I'll update the ticket but I'm thinking something like:
```
"extensions": {
  "credentialSignatures": ["abc", "def"]
}
```

EVCS also records the signature of every VC stored, so we can cross-check between them to confirm that it's all working as expected.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -515,17 +522,28 @@ private JourneyResponse buildReuseResponse(
...
-            String userId, VerifiableCredentialBundle vcBundle, String evcsAccessToken)
-            throws EvcsServiceException, VerifiableCredentialException {
-        if (configService.enabled(EVCS_WRITE_ENABLED) && !vcBundle.isEvcsIdentity) {
+            AuditEventUser auditEventUser,
+            String deviceInformation,
+            Vot vot,
+            VerifiableCredentialBundle vcBundle,
+            String evcsAccessToken)
+            throws EvcsServiceException, VerifiableCredentialException, SqsException {
+        if (configService.enabled(EVCS_WRITE_ENABLED)) {
```
Joe-Edwards-GDS: You've lost a condition on `vcBundle.isEvcsIdentity` in the merge resolution here

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -109,6 +104,13 @@ public void storePendingVc(VerifiableCredential credential) throws EvcsServiceEx
                                 credential.getVcString(), PENDING_RETURN, null, null)));
     }
 
+    @Tracing
```
Joe-Edwards-GDS: I wonder if this tracing is at the right level - having tracing on the EVCS client methods is quite useful, but we probably don't need it on all the private methods in this class...

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -109,6 +104,13 @@ public void storePendingVc(VerifiableCredential credential) throws EvcsServiceEx
...
+    private List<EvcsGetUserVCDto> getEvcsGetUserVCDtos(
```
Joe-Edwards-GDS: `getUserVCs`?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: Added vc signatures to audit extension.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: Also added a scenario when its evcsIdentity (with f2f vc and pending_return) then do f2f completion vcs updation pending_return -> current.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
rpayal: done

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -515,17 +516,40 @@ private JourneyResponse buildReuseResponse(
...
                 auditEventUser.getSessionId(),
                 false);
-        migrateCredentialsToEVCS(userId, vcBundle, evcsAccessToken);
+
+        if (!vcBundle.isEvcsIdentity) {
+            migrateCredentialsToEVCS(
+                    auditEventUser, deviceInformation, vcBundle.credentials, evcsAccessToken);
+        } else if (vcBundle.isF2fIdentity() && vcBundle.isPendingEvcsIdentity) {
+            updateIdentityOnF2fCompletion(
+                    auditEventUser.getUserId(), evcsAccessToken, vcBundle.credentials);
```
Joe-Edwards-GDS: This appears to duplicate @sam803 's work on F2F journeys - see https://github.com/govuk-one-login/ipv-core-back/pull/2033

I think you can just put the condition back into the condition below, and leave this PR as just adding the audit event.

rpayal: F2F completion update - before the above change migration call is taking care of this. After this change we are only migrating for only vc's not in evcs (!vcBundle.isEvcsIdentity). So f2f vc in pending_return in evcs will not get updated on completion.
Also in this case we don't want to call migration and raise migration event.

rpayal: Got it ...I think u mean now in case of pending we are doing new journey /reuse-with-store.  which will take care of that.  (not seeing target state - UPDATE_IDENTITY_BEFORE_IDENTITY_REUSE_PAGE)

Joe-Edwards-GDS: Yep exactly :)

Joe-Edwards-GDS (approved):
Thanks :)

PR #2049, opened 2024-06-20
BAU: Use AWS CRT Client

Joe-Edwards-GDS:
AWS documentation now recommends using the AWS common-runtime client by default on lambda: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration.html#http-clients-recommend

This is also available in an async version, if/when we require it.


local-running/src/main/java/uk/gov/di/ipv/coreback/sqs/SqsReader.java
```diff
@@ -27,7 +27,7 @@ public SqsReader(RequestHandler<SQSEvent, SQSBatchResponse> sqsHandler) {
         this.sqs =
                 SqsClient.builder()
                         .region(EU_WEST_2)
-                        .httpClientBuilder(UrlConnectionHttpClient.builder())
+                        .httpClientBuilder(AwsCrtHttpClient.builder())
```
Joe-Edwards-GDS, line 30: (@Wynndow we should probably check if this client actually works in local-running mode - I think it should...)

sam803 (approved):

PR #2054, opened 2024-06-21
PYIC-6805: log evcs-tactical vc mismatches

sam803:
## Proposed changes

Adds some logs to sanity check the EVCS credentials

### What changed

If evcs writes are enabled:
 - get evcs credentials from the store and compare them with the tactical store
 - if there are any evcs credentials that do not exist in the tactical store - log a warning
 - if there are any tactical credentials that do that not exist in the relevant pending/current credentials, then log a warning

https://govukverify.atlassian.net/browse/PYIC-6805

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -23,6 +23,7 @@ import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.JourneyResponse;
 import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
+import uk.gov.di.ipv.core.library.enums.EvcsVCState;
 import uk.gov.di.ipv.core.library.enums.OperationalProfile;
 import uk.gov.di.ipv.core.library.enums.Vot;
 import uk.gov.di.ipv.core.library.exception.EvcsServiceException;
@@ -62,6 +63,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static com.amazonaws.util.CollectionUtils.isNullOrEmpty;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EVCS_READ_ENABLED;
@@ -75,6 +77,7 @@ import static uk.gov.di.ipv.core.library.domain.ProfileType.GPG45;
 import static uk.gov.di.ipv.core.library.domain.ProfileType.OPERATIONAL_HMRC;
 import static uk.gov.di.ipv.core.library.domain.VocabConstants.VOT_CLAIM_NAME;
 import static uk.gov.di.ipv.core.library.enums.EvcsVCState.CURRENT;
+import static uk.gov.di.ipv.core.library.enums.EvcsVCState.PENDING;
 import static uk.gov.di.ipv.core.library.enums.EvcsVCState.PENDING_RETURN;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_VOT;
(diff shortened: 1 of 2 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +333,75 @@ private JourneyResponse getJourneyResponse(
...
+        // if we have pending vcs just check those
+        var evcsToCheck = evcsEvcs.get(PENDING);
+        if (!isNullOrEmpty(evcsToCheck)) {
+            evcsToCheck = evcsEvcs.get(CURRENT);
+        }
+        // filter out unmigrated vcs from the tactical vcs
+        tacticalVcs = tacticalVcs.stream().filter(vc -> vc.getMigrated() != null).toList();
+
+        // check all the selected evcs vcs are in the tactical store
+        if (hasIdentityMismatch(evcsToCheck, tacticalVcs)) {
```
sam803: I'm not sure if I have this the right way round, do we:
 - check all the tactical vcs exist in the pending/current evcs
 - check all the pending/current evcs exist in the tactical store
 - both

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +333,76 @@ private JourneyResponse getJourneyResponse(
...
             var vcs =
                     evcsService.getVerifiableCredentialsByState(
-                            userId, evcsAccessToken, CURRENT, PENDING_RETURN);
-            var pendingReturnVcs = vcs.get(PENDING_RETURN);
-            // use pending return vcs to determine identity if available
-            if (!isNullOrEmpty(pendingReturnVcs)) {
-                return new VerifiableCredentialBundle(pendingReturnVcs, true, true);
+                            userId, evcsAccessToken, CURRENT, PENDING_RETURN, PENDING);
+
+            if (configService.enabled(EVCS_WRITE_ENABLED)) {
```
Joe-Edwards-GDS: We probably don't need this `if` - it can probably just happen if either reads/writes are enabled. In practice I don't think we will ever have reads enabled and not writes.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +333,76 @@ private JourneyResponse getJourneyResponse(
...
+            List<VerifiableCredential> tacticalVcs,
+            Map<EvcsVCState, List<VerifiableCredential>> evcsEvcs) {
+
+        // if we have pending vcs just check those
+        var evcsToCheck = evcsEvcs.get(PENDING);
+        if (isNullOrEmpty(evcsToCheck)) {
+            evcsToCheck = evcsEvcs.get(CURRENT);
+        }
+        // filter out unmigrated vcs from the tactical vcs
+        tacticalVcs = tacticalVcs.stream().filter(vc -> vc.getMigrated() != null).toList();
```
Joe-Edwards-GDS: Do you think we can roll this into the check, rather than filtering them out? I think there are basically two valid options:
- The tactical VC list matches the EVCS list, and all the VCs are marked as migrated
- The EVCS list is empty, and the VCs are not marked as migrated

If we see an 'extra' un-migrated VC, that's probably worth flagging.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +333,78 @@ private JourneyResponse getJourneyResponse(
...
+
+        // check all the evcs vcs are in the tactical store
+        var listOfAllEvcsVcs =
+                evcsEvcs.values().stream().flatMap(List<VerifiableCredential>::stream).toList();
+        if (hasIdentityMismatch(listOfAllEvcsVcs, tacticalVcs)) {
+            LOGGER.warn("Failed to find corresponding tactical credential for evcs credential");
+        }
+    }
+
+    @ExcludeFromGeneratedCoverageReport
```
Joe-Edwards-GDS: Why is this excluded? Feels like a pretty testable method

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +333,78 @@ private JourneyResponse getJourneyResponse(
...
+                                        return vcString != null
+                                                && vcString.equals(vc.getVcString());
+                                    })
+                            .toList();
+            if (isNullOrEmpty(match)) {
+
+                return true;
+            }
+        }
+        return false;
```
Joe-Edwards-GDS: This feels like a pretty verbose way of checking two sets are equal... Could you do something like:
```
var setA = listA.stream().map(VerifiableCredential::getVcString).collect(Collectors.toSet());
var setB = listB.stream().map(VerifiableCredential::getVcString).collect(Collectors.toSet());
return setA.equals(setB);
```

If you want to have different messages one way or the other, you can also use `setA.containsAll(setB)`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: I was struggling to get the required test coverage in sonar cloud - do you have a recomendation to test private methods in java?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: thanks, will try that - still getting used to streams :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +334,79 @@ private JourneyResponse getJourneyResponse(
...
         }
-        return new VerifiableCredentialBundle(
-                verifiableCredentialService.getVcs(userId), false, false);
+        return new VerifiableCredentialBundle(tacticalVcs, false, false);
+    }
+
+    @ExcludeFromGeneratedCoverageReport
+    private void logIdentityMismatches(
+            List<VerifiableCredential> tacticalVcs,
+            Map<EvcsVCState, List<VerifiableCredential>> evcsEvcs) {
```
Joe-Edwards-GDS: I think this should be `evcsVcs` 😄

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +334,79 @@ private JourneyResponse getJourneyResponse(
...
+        }
+
+        // check all the tactical vcs are in the selected evcs vcs
+        if (!hasUnmigratedVcs && !evcsVcStrings.containsAll(migratedTacticalVcStrings)) {
+            LOGGER.warn(
+                    "Failed to find corresponding evcs credential for migrated tactical credential");
+        }
+
+        // check all the evcs vcs are in the tactical store
+        if (!hasUnmigratedVcs && !migratedTacticalVcStrings.containsAll(allEvcsVcStrings)) {
```
Joe-Edwards-GDS: I think this check should just use `evcsVcStrings` as well - the tactical VCs should all be in the correct status in EVCS (i.e. pending or current, as appropriate)?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -331,22 +334,79 @@ private JourneyResponse getJourneyResponse(
...
+        var evcsToCheck = evcsEvcs.get(PENDING);
+        if (isNullOrEmpty(evcsToCheck)) {
+            evcsToCheck = evcsEvcs.get(CURRENT);
+        }
+        var allTacticalVcStrings =
+                tacticalVcs.stream()
+                        .map(VerifiableCredential::getVcString)
+                        .collect(Collectors.toSet());
+        var evcsVcStrings =
+                isNullOrEmpty(evcsToCheck)
```
Joe-Edwards-GDS: Minor but it might be neater to sort out `evcsToCheck` and avoid this null check. you can use `Optional` if you want - although it's obviously better:
```
var evcsToCheck = Optional.of(vcs.get(PENDING_RETURN))
        .or(() -> Optional.of(vcs.get(CURRENT)))
        .orElse(List.of());
```
(Java monads are all pretty rubbish 😞 )

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Sorry forgot to reply to this, _in general_ either:
- The private method is an implementation detail, and should be be exercised via the public methods of the class
- It's not an implementation detail, so pull it out into a helper (or make it public) and test it that way

This is a bit of a weird case, because it only does logging, which we're not well set up to test. It's still possible (you can mock out the logger if you use reflection, or pull out a helper that gives you the 'result' to log, with the logging itself separate), but the effort might not outweigh the cost.

Joe-Edwards-GDS (approved):
Looks sensible I think - if we start getting lots of these, we might need to add some more debugging information (e.g. CRI and issue date for the VCs, or something), but it's hard to know what we'll need at the moment.

PR #2056, opened 2024-06-21
Pyic 6763 low confidence journey choice

DanCorderIPV:
### What changed

Update check existing identity to start new journeys or reproving journeys for P1

This PR is restricted to new identities and reproving identities. Further changes will be needed to cope with other journeys that are affected by the presence of P1 in the vtr. 

### Why did it change

We are adding support for low confidence profiles

### Issue tracking

- [PYIC-6763](https://govukverify.atlassian.net/browse/PYIC-6763)

[PYIC-6763]: https://govukverify.atlassian.net/browse/PYIC-6763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -221,7 +226,7 @@ void setUpEach() {
     }
 
     @Test
-    void shouldReturnJourneyResetGpg45Identity() throws Exception {
```
DanCorderIPV, line 227: I assume `Reset` in the original test name was a typo

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -1136,9 +1158,7 @@ void shouldRemoveOperationalProfileIfMatchingGpg45ProfileIsPresent() throws Exce
     }
 
     @Test
-    void
-            shouldReturnMitigationJourneyStepResponseIfResetIdentityTrueCiMitigationJourneyStepPresentAndMitigationJourneyStepPresent()
```
DanCorderIPV, line 1168: I am more concerned by this test - it looked like it was doing what the original name said because the mitigation journey response set on line 1152 happened to be the reprove response which is also returned by a different code path.
Hopefully the original test name is wrong/out of date

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -0,0 +1,1275 @@
+name: New P1 Identity
```
DanCorderIPV: Aside from the first few lines this is just a copy/paste of the existing P2 journey map as a placeholder until the relevant ticket is done.

DanCorderIPV:
I'm a little concerned that there's nothing here (e.g. feature flag) that stops this code from being executed if an RP sends us a VTR containing P1. I thought that I had seen a feature flag ticket, but I can't find one now - do we need a flag as part of this ticket?

blakeyp:
> I'm a little concerned that there's nothing here (e.g. feature flag) that stops this code from being executed if an RP sends us a VTR containing P1. I thought that I had seen a feature flag ticket, but I can't find one now - do we need a flag as part of this ticket?

Hmm I'd like to think Orchestration have this flagged and wouldn't send us P1 requests yet but we probably can't rely on that, so yes we probably do want it behind a feature flag and to do so here

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
blakeyp: I was wondering if we could just reuse the p2 journey map at least for now (possibly renaming it `new-identity.yaml`) - the P1 journey is practically identical other than the no photo id route so I'm a bit nervous about having a duplicate to maintain. Or maybe we need to think about a way of packaging up the routes so we can reuse them

DanCorderIPV: Sorry, I misunderstood what you meant by "re-use" :)

blakeyp (dismissed):

blakeyp (approved):

PR #2057, opened 2024-06-21
PYIC-5759: add parsed claimsSet property to VerifiableCredential class

thebauSoftwire:
### What changed
- adds parsed claimsSet property to VerifiableCredential class which uses the new VerifiableCredentialParser
- updates some unit tests to use stub data that is more representative of real VCs

### Why did it change
Making a property available for the parsed claimsSet will make it easier to access the properties within it as currently, they are only accessible as JSON objects. Using the parser means we can enforce stricter type checking on the VCs and easier access to the properties.

### Issue tracking


libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -910,7 +910,6 @@ static VerifiableCredential vcF2fBrp() {
                                 List.of(
                                         TestVc.TestEvidence.builder()
                                                 .txn("some-uuid")
-                                                .type(CRI_STUB_CHECK_EVIDENCE_TYPE)
```
thebauSoftwire, line 928: Not sure why this was overriding the default (`IDENTITY_CHECK_EVIDENCE_TYPE`). Using the default type specified on the TestVc.TestEvidence class passes the test (same w below test). Could be that this stub data just wasn't representative but should double-check

thebauSoftwire:
Some things that need to be further investigated + fixed:
- Passport CRI consumer tests in the `process-cri-callback` lambda are failing
- Failing unit tests: `buildScoreShouldReturnCorrectScoreForDcmawCredential` and `getContraIndicatorsVCJwtThrowsErrorWhenNoVcBlock` which I think is the reason test-java is failing int he pipeline

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/VerifiableCredential.java
```diff
@@ -15,28 +20,48 @@
...
                     "Failed to get jwt claims to construct verifiable credential", e);
         }
     }
 
+    private uk.gov.di.model.VerifiableCredential parseCredential(JWTClaimsSet claimsSet)
+            throws CredentialParseException {
+        try {
+            return VerifiableCredentialParser.parseCredential(claimsSet);
+        } catch (CredentialParseException e) {
+            if ("production".equals(System.getenv(EnvironmentVariable.ENVIRONMENT.name()))) {
```
Joe-Edwards-GDS, line 54: This isn't ideal, but it's quite hard to get a config service here and I'm trying to be particularly paranoid about it!

I expect to remove this in the first PR that actually uses the parsed credential.

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/service/VerifiableCredentialServiceTest.java
```diff
@@ -74,34 +61,6 @@ void expectedSuccessWhenSaveCredentials() throws Exception {
         assertEquals(PASSPORT_NON_DCMAW_SUCCESSFUL_VC.getVcString(), vcStoreItem.getCredential());
     }
 
-    @Test
-    void expectedSuccessWithoutExpWhenSaveCredentials() throws Exception {
```
Joe-Edwards-GDS, line 78: (The test above already contains a VC with no expiry)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VerifiableCredentialParser.java
```diff
@@ -24,7 +24,9 @@ public class VerifiableCredentialParser {
     private static final Logger LOGGER = LogManager.getLogger();
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final ObjectMapper LAX_OBJECT_MAPPER =
-            new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+            new ObjectMapper()
+                    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                    .configure(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_AS_NULL, true);
```
Joe-Edwards-GDS, line 29: The only known place where we have unknown enum values is in one of our contract tests (@DanCorderIPV is following up with Lime on this), but it probably makes sense to be forgiving on this anyway.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/validator/VerifiableCredentialValidator.java
```diff
@@ -106,15 +104,10 @@ public VerifiableCredential parseAndValidate(
 
             var vc = VerifiableCredential.fromValidJwt(userId, criId, vcJwt);
 
-            if (configService.enabled(CoreFeatureFlag.PARSE_VC_CLASSES)) {
```
Wynndow, line 109: I think this feature flag isn't used anymore - can it be dropped from CoreFeatureFlags?

Wynndow (approved):

PR #2061, opened 2024-06-24
PYIC-5875: Test harness for API tests

Wynndow:
### What changed

Test harness for API tests

### Why did it change

This adds the basics of a new cucumber testing project for running API tests through core. It includes one feature that runs a happy path journey through an app journey.

It includes a GHA workflow to run the tests pre-merge.

Lots of details in the commit messages for anyone interested.

### Issue tracking

- [PYIC-5875](https://govukverify.atlassian.net/browse/PYIC-5875)

[PYIC-5875]: https://govukverify.atlassian.net/browse/PYIC-5875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/src/config/config.ts
Joe-Edwards-GDS, line 1: As discussed - I think it would be preferable to set up a `.env` file for this, which will make it easier to manage configurations as well as making it independent from other setup.

You can use https://www.npmjs.com/package/dotenv to load it programmatically when initialising the tests. I'd suggest a similar pattern of having a `.env.template` and `.env`.

The template can also replace all the defaults in `constants.ts`, and properly separate the config from code. 😄

These support comments, so you can do things like:
```
# Build configuration
SOME_VALUE=build-value
SOME_OTHER_VALUE=boop

# Dev configuration
# SOME_VALUE=dev-value
# SOME_OTHER_VALUE=beep
```

api-tests/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "apitests",
+  "version": "1.0.0",
+  "description": "API tests for core-back",
+  "type": "module",
+  "scripts": {
+    "test": "cucumber-js",
+    "lint": "eslint . && prettier . --check",
```
Joe-Edwards-GDS, line 8: I get style issues on the cucumber html report when I run this - probably worth excluding that if we can!

api-tests/README.md
Joe-Edwards-GDS, line 1: Could be worth recommending the Cucumber VSCode plugin (if you're using VSCode?) and how to configure it. I had to update the settings to let it pick up the glue (aka step definitions), as by default it looks for `features/**/*.ts`

api-tests/tsconfig.json
Joe-Edwards-GDS, line 1: Did this come from somewhere? Looks reasonable (although I wouldn't generally bother including all the config options like this - they'll get out of date when we update TS)

api-tests/src/utils/jar-generator.ts
```diff
@@ -0,0 +1,43 @@
...
+);
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+export const generateJar = async (
+  subject: string,
+  journeyType: string,
+): Promise<string> => {
+  const payloadData = JSON.parse(
+    fs.readFileSync(
```
Joe-Edwards-GDS: Can we use the async method if possible? If you change your import to `node:fs/promises` then you'll get a promise-based version, which is probably what you want :)

api-tests/src/utils/request-body-generators.ts
```diff
@@ -0,0 +1,95 @@
...
+    `client_id=${ORCHESTRATOR_CLIENT_ID}&` +
+    `client_assertion_type=urn%3Aietf%3Aparams%3Aoauth%3Aclient-assertion-type%3Ajwt-bearer&` +
+    `client_assertion=${await createSignedJwt({ sub: ORCHESTRATOR_CLIENT_ID, iss: ORCHESTRATOR_CLIENT_ID })}`
+  );
+};
+
+const readJsonFile = (criId: string, scenario: string, jsonType: JsonType) => {
+  return JSON.stringify(
+    JSON.parse(
+      fs.readFileSync(
```
Joe-Edwards-GDS: (ditto about async)

api-tests/src/utils/request-body-generators.ts
```diff
@@ -0,0 +1,95 @@
...
+    throw new Error("code not received in redirect URL");
+  }
+
+  return (
+    `grant_type=authorization_code&` +
+    `code=${code}&` +
+    `redirect_uri=${encodeURI(config.ORCHESTRATOR_REDIRECT_URL)}&` +
+    `client_id=${ORCHESTRATOR_CLIENT_ID}&` +
+    `client_assertion_type=urn%3Aietf%3Aparams%3Aoauth%3Aclient-assertion-type%3Ajwt-bearer&` +
+    `client_assertion=${await createSignedJwt({ sub: ORCHESTRATOR_CLIENT_ID, iss: ORCHESTRATOR_CLIENT_ID })}`
```
Joe-Edwards-GDS: You could use `URLSearchParams` here - that would do all the escaping for you :)
```
const params = new URLSearchParams();
params.set("grant_type", "authorization_code")
params.set("client_assertion_type", "urn:ietf:params:oauth:client-assertion-type:jwt-bearer")
// etc.
return params.toString();
```
(You might need to prepend a `?` if that's important)

api-tests/src/utils/request-body-generators.ts
```diff
@@ -0,0 +1,95 @@
...
+  );
+};
+
+const readJsonFile = (criId: string, scenario: string, jsonType: JsonType) => {
+  return JSON.stringify(
+    JSON.parse(
```
Joe-Edwards-GDS: This appears to parse and then immediately stringify - is that needed?

api-tests/src/clients/core-back-external-client.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return await response.json();
+};
+
+export const getIdentity = async (
+  tokenResponse: TokenResponse,
+): Promise<UserIdentity> => {
+  const response = await fetch(
+    config.CORE_BACK_EXTERNAL_API_URL + `/user-identity`,
```
Joe-Edwards-GDS: If you're using backticks here, I think you meant `${config.CORE_BACK_EXTERNAL_API_URL}/user-identity`?

api-tests/src/clients/core-back-internal-client.ts
Joe-Edwards-GDS, line 1: If we do anything more complex here, it's probably worth introducing something like `Axios` to help with the API calls - that lets you set up clients etc. with the correct base URL and headers.

Effectively a streamlined version of the core-front implementation https://github.com/govuk-one-login/ipv-core-front/blob/main/src/services/coreBackService.js

api-tests/src/interfaces/journey-engine-response.ts
Joe-Edwards-GDS, line 1: This isn't very idiomatic - at the moment you've declared that all responses have all these properties! If we want a single type to capture everything, you should mark them as optional:
```
export interface JourneyEngineResponse {
  page?: string;
  journey?: string;
  cri?: CriResponse;
  client?: ClientResponse;
}
```

However, if you want to be even more precise, you can do something like:
```
export type JourneyEngineResponse =
  | JourneyResponse
  | PageResponse
  | CriResponse
  | ClientResponse;

interface JourneyResponse {
  journey: string;
}

interface PageResponse {
  page: string;
}

interface CriResponse {
  cri: {
    id: string;
    redirectUrl: string;
  };
}

interface ClientResponse {
  client: {
    redirectUrl: string;
  };
}
```

This is a bit more precise, as you can only get each specific combination of properties, but it does mean you need to use a type guard to know which one you actually have, e.g.:
```
const isPageResponse = (
  journeyResponse: JourneyEngineResponse,
): journeyResponse is PageResponse => !!(journeyResponse as PageResponse).page;

const journeyResponse: JourneyEngineResponse = getJourneyResponse();

if (isPageResponse(journeyResponse)) {
  console.log(journeyResponse.page); // happy
  console.log(journeyResponse.cri); // Propert 'cri' does not exist on type 'PageResponse'
}
```

api-tests/src/interfaces/process-cri-callback-request.ts
```diff
@@ -0,0 +1,6 @@
+export interface ProcessCriCallbackRequest {
+  authorizationCode: string | null;
+  credentialIssuerId: string | null;
+  redirectUri: string | null;
+  state: string | null;
```
Joe-Edwards-GDS: are these really nullable?

api-tests/src/interfaces/process-cri-callback-response.ts
```diff
@@ -0,0 +1,3 @@
+export interface ProcessCriCallbackResponse {
```
Joe-Edwards-GDS: I think it technically returns a `JourneyEngineResponse` - I think if there's an error it'll return a page id rather than a journey step.

api-tests/src/interfaces/redirect-params.ts
Joe-Edwards-GDS, line 1: unused?

api-tests/src/steps/ipv-steps.ts
```diff
@@ -0,0 +1,95 @@
...
+      this.ipvSessionId,
+    );
+    this.lastJourneyEngineResponse = await internalClient.sendJourneyEvent(
+      journeyResponse.journey,
+      this.ipvSessionId,
+    );
+  },
+);
+
+Then("I get a client Oauth response", function (this: World): void {
```
Joe-Edwards-GDS: (I think I'd prefer OAuth 😄 )

Joe-Edwards-GDS (commented):
I might go over this again later, but it's looking pretty good I think!

I'm sure we'll find some tricky cases lurking in the existing test scenarios, but this is pretty much what I expected as a foundation. 🙂

api-tests/src/interfaces/journey-engine-response.ts
Joe-Edwards-GDS: Worth noting that if we have control over the types, you'd usually prefer to have a discriminator on the type, which avoids the need to write a dedicated type guard.

If we add a `type` property to our responses, then it gets much cleaner:
```
export type JourneyEngineResponse =
  | PageResponse
  | CriResponse;

interface PageResponse {
  type: 'page';
  page: string;
}

interface CriResponse {
  type: 'cri';
  cri: {
    id: string;
    redirectUrl: string;
  };
}

const journeyResponse: JourneyEngineResponse = getJourneyResponse();

if (journeyResponse.type === 'page') {
  console.log(journeyResponse.page); // happy
  console.log(journeyResponse.cri); // Propert 'cri' does not exist on type 'PageResponse'
}
```
Typescript is clever enough to realise that if you check the type is `page`, then it's a `PageResponse` 🎉 

(I wouldn't be against making that change just to make it easier to use in TS)

.github/workflows/pre-merge-checks.yml
```diff
@@ -130,3 +130,34 @@ jobs:
...
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: ./api-tests/.nvmrc
+          cache: npm
+          cache-dependency-path: api-tests/package-lock.json
+          registry-url: "https://npm.pkg.github.com"
+
+      - name: Install
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.PAT_READ_PACKAGES_SCOPE_CHRIS_WYNNE }}
```
Joe-Edwards-GDS: This probably shouldn't have your name on it...!

The reads seem to work for the gradle build using `secrets.GITHUB_TOKEN`, but I guess you tried that. If we need configuration at the data-vocab end, then we can reach out to one of the data architects.

In any case, we probably can't run these tests pre-merge yet, we'll need to set up an environment for them to run against (it's no good running against build if they rely on parallel code changes!)

See https://govukverify.atlassian.net/browse/PYIC-5878

api-tests/package.json
Wynndow: That's weird, I don't. And I remember when I set it up that it's supposed to ignore stuff in .gitignore. Is eslint or prettier failing?

api-tests/README.md
Wynndow: Good shout - I'm on intellij but had to do the same.

api-tests/tsconfig.json
Wynndow: Yeah, running `npx tsc --init` gave it to me.
I thought it was useful to see eveything, but tbh it's too much. I'll trim it out.

api-tests/package.json
Joe-Edwards-GDS:
```
$ npm run lint

> apitests@1.0.0 lint
> eslint . && prettier . --check

Checking formatting...
[warn] cucumber-report.html
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```
Looks like prettier

Wynndow: And your .gitignore contains `api-tests/cucumber-report.html`?

api-tests/src/utils/jar-generator.ts
Wynndow: Yep can do - but in this scenario why does it matter? We're going to just wait for it anyway?

api-tests/package.json
Joe-Edwards-GDS: Yep... Idk if it's something about the working directory maybe? I'm running this inside the `api-tests` directory

Wynndow: Yeah same. I'll have a think

api-tests/src/utils/request-body-generators.ts
Wynndow: Oh nice.

api-tests/src/utils/jar-generator.ts
Joe-Edwards-GDS: It's a good habit because there is pretty much never a good reason to use synchronous I/O in node.

Although you might think `await asyncCall()` is equivalent to doing it synchronously (and in Java it almost is), in Node this will block the entire event loop - meaning no other code can execute while waiting for it, whereas using async/await does not. In a webserver this is pretty fatal (all requests are blocked!), but even here, we have no idea what other processes cucumber might be trying to run in the background.

Wynndow: Ah, interesting. Thanks

api-tests/src/utils/request-body-generators.ts
Wynndow: Nope!

api-tests/src/interfaces/journey-engine-response.ts
Wynndow: The way it currently is - why does it work? If I've not marked them as nullable, and we're getting null values (presumably), why hasn't it blown up? I guess because the type system doesn't exist at runtime?

Joe-Edwards-GDS: Yep exactly - the types are all erased at runtime, so only get checked if you write some JS code to check them.

.github/workflows/pre-merge-checks.yml
Wynndow: Yeah, i tried with the GITHUB_TOKEN but no dice, unless I did something else wrong. 
But yeah I guess there's no point running against build - will drop this change.

api-tests/package.json
Wynndow: Fixed it - prettier does in fact NOT use gitignore. Had to add an ignore file

api-tests/src/interfaces/process-cri-callback-request.ts
Wynndow: no - as discussed I've removed them as I was already doing a null check on them

api-tests/src/interfaces/redirect-params.ts
Wynndow: unused.

api-tests/src/steps/ipv-steps.ts
Wynndow: Took me longer than it should have to spot the difference

api-tests/.env.template
```diff
@@ -0,0 +1,17 @@
+# Build env config
+CORE_BACK_COMPONENT_ID="https://identity.build.account.gov.uk"
+CORE_BACK_INTERNAL_API_URL="https://internal-test-api.identity.build.account.gov.uk"
+#CORE_BACK_INTERNAL_API_KEY=Provided dynamically if running in a pipeline. Provide to run against build env locally.
+CORE_BACK_EXTERNAL_API_URL="https://api.identity.build.account.gov.uk"
+CORE_BACK_PUBLIC_ENCRYPTION_KEY='{"kty":"RSA","e":"AQAB","n":"nel7ibmSTaXWhwEAdqKTiEVcxsYgv6CdXaz90aVN7IorlaCeNj0j06OsA4zdmWEjj21wEZULsxPoZo5N_tsQ7NtOnOkcnDc-g_Nbpt0jelzJSbFRkx3kwXy8YIYKR_myNbiHNTTc7S6GkQRg0N1MPWtzoEKYJs41AN4onrsvUzgpCypWwPy2-ppsaDvms_11YA7A7x3zHj9oKCPJ_uk_0MV3vZAxCxbiPb9ABGWcoGQ5QKGfv40ylBsEdOhE3w-3SAAQIrrHyMRGGiPxcNO161XVL-lOnYt93FgEe16LgpfE22UdENfHnG0UQaTgph1Dm24oqn7qpPTY2DfER5HCKQ"}' # pragma: allowlist secret
+ORCHESTRATOR_REDIRECT_URL="https://orch.stubs.account.gov.uk/callback"
+JAR_SIGNING_KEY='{"kty":"EC","d":"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU","crv":"P-256","x":"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM","y":"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04"}' # pragma: allowlist secret
```
Joe-Edwards-GDS, line 9: (I wonder if it's worth calling this `JAR_PUBLIC_SIGNING_KEY` just to stop anyone having a heart attack)

api-tests/src/config/config.ts
```diff
@@ -0,0 +1,14 @@
+import "dotenv/config";
```
Joe-Edwards-GDS, line 1: Do you need to add this to package.json? I get 'could not find package dotenv'

api-tests/src/types/journey-engine-response.ts
Joe-Edwards-GDS, line 1: I don't think it's worth separating types and interfaces - I'd probably just put them all under `types` (the distinction isn't important, the point is that this is where all the type definitions are).

I'd also combine `journey-engine-response.ts` and `journey-engine-responses.ts` into a single file - they naturally belong together 😄 

(You could also put all the 'internal API types' in a single file, should you wish)

api-tests/src/utils/type-guards.ts
Joe-Edwards-GDS, line 1: This is probably too generic a filename - I'd make it `journey-response-helpers.ts` or something.

(Alternatively, you could bundle them in with the type definitions)

api-tests/src/interfaces/journey-engine-responses.ts
Joe-Edwards-GDS, line 1: Technically I think there's also `ErrorResponse` - it's the same as `PageResponse` but with an added `statusCode`.

(Might be better to just add an optional `statusCode` onto `PageResponse`!)

.pre-commit-config.yaml
```diff
@@ -32,3 +32,11 @@ repos:
...
     hooks:
       - id: spotlessCheck
+
+  - repo: local
+    hooks:
+      - id: run-npm-lint
+        name: Lint API tests
+        entry: bash -c 'cd api-tests && npm run lint'
+        language: system
+        pass_filenames: false
```
Joe-Edwards-GDS, line 43: Is it possible to filter this so it only runs if you actually change something in `api-tests`?

api-tests/.env.template
Wynndow: Might be a bit confusing as this a private key - it's the test key we use all over the place and is already in the public domain. Could add a comment to make that obvious?

api-tests/src/config/config.ts
Wynndow: How weird. My cmd history shows I installed with save-dev, must have lost the changes to package.json somehow. Will add it in.

Joe-Edwards-GDS: save yourself some clicks `npm i -D dotenv` :P

api-tests/src/interfaces/journey-engine-responses.ts
Wynndow: Ah that's true. It does feel like a special case of a Page response, so I'll add an optional

.pre-commit-config.yaml
Wynndow: Turns out yes!

api-tests/src/config/config.ts
Wynndow: If only my poor aching fingers knew about this sooner!

.gitignore
```diff
@@ -53,3 +53,8 @@ local-running/bin/
 
 # User-specific secrets
 secrets.gradle
+
+api-tests/cucumber-report.html
+api-tests/.npmrc
+api-tests/.env
+api-tests/src/**/*.js
```
Joe-Edwards-GDS: (This addition looks a little suspicious...)

Joe-Edwards-GDS (commented):
Looks good to me - we can always revise things as we develop more test cases etc.

Let's wait until someone from QA has had a look before merging.

.gitignore
Wynndow: I added this as I ran `npx tsc` to try compiling stuff with typecript directly, rather than using ts-node. And it generated all the js files, which we don't want. Is there a better way to deal with this?

Joe-Edwards-GDS: You can add `noEmit: true` to your tsconfig.json to prevent it from emitting JS files - then `tsc` will just do type-checking.

Joe-Edwards-GDS:
One additional thought - can we add this to `dependabot.yml`?

Joe-Edwards-GDS (approved):

PR #2064, opened 2024-06-24
PYIC-6755: Match P1 identities

MikeCollingwood:
### What changed

- Add `getRequestedVotsByStrength` to find `Vot`s list from `vtr`
- Flatmap `supportedGpg45Profiles` for use in determining matched profile
  - Matching process already flexible to different profile strengths
- Add `Vot.fromProfile` so can set Vot in ipv session after matching profile determined
- Moved `SUPPORTED_VOTS_BY_DESCENDING_STRENGTH` to Vot, where it's most relevant

### Why did it change

- To allow users to match to P1 identities

### Issue tracking

- [PYIC-6755](https://govukverify.atlassian.net/browse/PYIC-6755)

[PYIC-6755]: https://govukverify.atlassian.net/browse/PYIC-6755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -12,9 +12,12 @@
 @ExcludeFromGeneratedCoverageReport
 public enum Vot {
     P0(List.of(), null, GPG45),
+    P1(List.of(Gpg45Profile.L1A), null, GPG45),
     P2(List.of(Gpg45Profile.M1A, Gpg45Profile.M1B, Gpg45Profile.M2B), null, GPG45),
     PCL250(null, List.of(OperationalProfile.PCL250), OPERATIONAL_HMRC),
     PCL200(null, List.of(OperationalProfile.PCL250, OperationalProfile.PCL200), OPERATIONAL_HMRC);
+    public static final List<Vot> SUPPORTED_VOTS_BY_DESCENDING_STRENGTH =
+            List.of(Vot.P2, Vot.PCL250, Vot.PCL200, Vot.P1, Vot.P0);
```
DanCorderIPV: Are you sure that adding `P0` is safe?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -40,4 +43,12 @@ public List<OperationalProfile> getSupportedOperationalProfiles() {
     public ProfileType getProfileType() {
         return this.profileType;
     }
+
+    public static Vot fromGpg45Profile(Gpg45Profile profile) {
+        return SUPPORTED_VOTS_BY_DESCENDING_STRENGTH.stream()
+                .filter(vot -> GPG45.equals(vot.profileType))
+                .filter(vot -> vot.getSupportedGpg45Profiles().contains(profile))
+                .findFirst()
```
DanCorderIPV: Apparently java doesn't have `findSingle()` which is disappointing. We are expecting exactly one match here so it would be good if we also threw an exception if more than one match is found.
It might be nice to adapt this solution to accept an error message and us it here: https://stackoverflow.com/a/22695031 (I've seen other places where we use `first` but mean `single` so I think it would be more generally useful.)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -198,7 +202,8 @@ private boolean hasMatchingGpg45Profile(
                                 vcs,
                                 ipAddress,
                                 deviceInformation));
-                ipvSessionItem.setVot(Vot.P2);
+
+                ipvSessionItem.setVot(Vot.fromGpg45Profile(matchedProfile.get()));
```
DanCorderIPV, line 206: This isn't your code, so just a comment - this is a big side effect to be hiding in this method

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandlerTest.java
```diff
@@ -486,6 +495,77 @@ void shouldReturnJourneyMetIfCheckRequiresAdditionalEvidenceResponseFalse() thro
...
+                        evaluateGpg45ScoresHandler.handleRequest(request, context),
+                        JourneyResponse.class);
+
+        assertEquals(JOURNEY_MET.getJourney(), response.getJourney());
+        verify(ipvSessionItem).setVot(Vot.P2);
+    }
+
+    @ParameterizedTest
+    @MethodSource("vcsMatchingVtr")
+    void shouldReturnTrueForMatchingGpg45Profile(List<String> vtr, List<Gpg45Profile> profiles)
```
DanCorderIPV: I'm not sure this test name describes what this is testing?
It looks like it's testing that the correct set of profiles are matched against for a given vtr

DanCorderIPV: Also can we test this without making `hasMatchingGpg45Profile` public?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/enums/VotTest.java
```diff
@@ -0,0 +1,31 @@
...
+        assertEquals(associatedVot, result);
+    }
+
+    private static Stream<Arguments> votProfiles() {
+        return Stream.of(
+                Arguments.of(L1A, Vot.P1),
+                Arguments.of(M1A, Vot.P2),
+                Arguments.of(M1B, Vot.P2),
+                Arguments.of(M2B, Vot.P2));
+    }
```
DanCorderIPV, line 36: Should have a test for no matches too

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/persistence/item/ClientOauthSessionItemTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.List;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class ClientOauthSessionItemTest {
+    @ParameterizedTest
+    @MethodSource("vtrRequestedVotsByStrength")
+    void getRequestedVotsByStrengthShould(List<String> vtr, List<Vot> expectedRequestedVots) {
```
DanCorderIPV: Should what?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/enums/VotTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static uk.gov.di.ipv.core.library.gpg45.enums.Gpg45Profile.L1A;
+import static uk.gov.di.ipv.core.library.gpg45.enums.Gpg45Profile.M1A;
+import static uk.gov.di.ipv.core.library.gpg45.enums.Gpg45Profile.M1B;
+import static uk.gov.di.ipv.core.library.gpg45.enums.Gpg45Profile.M2B;
+
+class VotTest {
+    @ParameterizedTest
+    @MethodSource("votProfiles")
+    void shouldGetCorrectStrengthVotWithProfile(Gpg45Profile profile, Vot associatedVot) {
+        var result = Vot.fromGpg45Profile(profile);
```
DanCorderIPV, line 24: I know it's a very short test, but please add //act and //assert

DanCorderIPV (changes requested):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
MikeCollingwood: With SPOT I'm assuming it's impossible for a P0 to be misinterpreted as a P1/ P2. Either way, I will remove it to be safe

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
MikeCollingwood: Could someone specify M1B as satisfying a P1 as well? Depends how the mapping is to be used but in theory we'd say we'd want the strongest in any scenario. If you feel that's an unrealistic scenario then I'll add this custom collector helper method to throw if not a singleton list

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandlerTest.java
MikeCollingwood: > I'm not sure this test name describes what this is testing?

I'll rename to `shouldReturnTrueForHasMatchingProfileGivenProfilesMatchVtr`?

> Also can we test this without making hasMatchingGpg45Profile public?

I already have but it wasn't as clear as I'd hoped, since the `handleRequest` tests have so to them already, so I wanted to clarify with a more explicit test. I'll remove if you think it's best.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
MikeCollingwood: I wrote the `fromGpg45Profile` method if that's what you mean? Or are you saying the ipv session VoT has a large side effect

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
DanCorderIPV: They maybe could, but it's not how things are set up at the moment. I think that's exactly the kind of thing we'd want the exception to highlight so someone has to explicitly think about whether this code is doing the right thing with the new configuration.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
DanCorderIPV: I mean that silently setting an important session property in method whose name reads like a simple test is ringing alarm bells.

lambdas/evaluate-gpg45-scores/src/test/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandlerTest.java
DanCorderIPV: > I'll rename to shouldReturnTrueForHasMatchingProfileGivenProfilesMatchVtr?

Calling it `shouldReturnTrue...` suggests that we should be checking that the method returns true, but the assert section is testing that the correct profiles were supplied to `getFirstMatchingProfile` - no `true` in sight :)

MikeCollingwood: Ah good point I've removed the check for the result. I'll rename the method source as well

`shouldUseCorrectGpg45ProfilesFromVtrToCheckMatchingVcs`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/CollectionHelper.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class CollectionHelper {
+    public static <T> Collector<T, ?, T> toSingletonOrNullIfEmpty() {
+        return Collectors.collectingAndThen(
+                Collectors.toList(),
+                list -> {
+                    if (list.size() > 1) {
+                        throw new IllegalStateException("List too large to be singleton");
+                    }
+                    if (list.isEmpty()) {
```
DanCorderIPV: Why not throw an exception for 0 results too? That's also unexpected isn't it?

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -174,21 +173,26 @@ private Map<String, Object> buildJourneyErrorResponse(ErrorResponse errorRespons
...
+
+            var requestedVotsByStrength = clientOAuthSessionItem.getRequestedVotsByStrength();
+            var supportedGpg45ProfilesByVotStrength =
+                    requestedVotsByStrength.stream()
+                            .flatMap(vot -> vot.getSupportedGpg45Profiles().stream())
+                            .toList();
+            var matchedProfile =
                     gpg45ProfileEvaluator.getFirstMatchingProfile(
-                            gpg45Scores, Vot.P2.getSupportedGpg45Profiles());
+                            gpg45Scores, supportedGpg45ProfilesByVotStrength);
```
blakeyp, line 193: If the `vtr` is ["P1", "P2"] does this match in descending strength, so if a P2 profile is matched it will be returned first?

MikeCollingwood: Yes so `getRequestedVotsByStrength` returns (supported) Vots in order of descending strength. Please see the test `shouldReturnJourneyMetForMeetingMediumAndLowConfidences` for that example. It references `P1_and_P2_profiles` which is `List.of(M1A, M1B, M2B, L1A)` (strength-ordered)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -174,21 +173,26 @@ private Map<String, Object> buildJourneyErrorResponse(ErrorResponse errorRespons
     }
 
     @Tracing
-    private boolean hasMatchingGpg45Profile(
+    public boolean hasMatchingGpg45Profile(
```
blakeyp: Has this been made public so it can be tested - can it not be tested indirectly through its caller?

MikeCollingwood: Yep, I've just made it private again. I wanted to have more obvious testing but Dan made the same comment earlier

blakeyp (dismissed):

DanCorderIPV (dismissed):

blakeyp (approved):

PR #2065, opened 2024-06-24
BAU: added test to catch invalid target state in journey config

sam803:
## Proposed changes

Adds a test to catch invalid journey map configurations where the target journey or target state do not exist

### Why did it change

To improve confidence when updating journey maps and prevent broken configurations in production environment

### DIFF
--- a/lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
+++ b/lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
@@ -143,10 +143,15 @@ class JourneyMapTest {
                             var basicEventStateMachineKeys = basicEventStateMachine.keySet();
                             assertTrue(
                                     basicEventStateMachineKeys.contains(
-                                            basicEvent.getTargetState()));
+                                            basicEvent.getTargetState()),
+                                    "Unknown target state %s"
+                                            .formatted(basicEvent.getTargetState()));
 
                         } else if (basicEvent.getTargetState() != null) {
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -122,6 +122,38 @@ void shouldMatchNestedJourneyEntryEvents(IpvJourneyTypes journeyType) throws IOE
...
+                        if (basicEvent.getTargetJourney() != null) {
+                            var basicEventStateMachine =
+                                    new StateMachineInitializer(
+                                                    IpvJourneyTypes.valueOf(
+                                                            basicEvent.getTargetJourney()))
+                                            .initialize();
+                            var basicEventStateMachineKeys = basicEventStateMachine.keySet();
+                            assertTrue(
+                                    basicEventStateMachineKeys.contains(
+                                            basicEvent.getTargetState()));
```
Joe-Edwards-GDS: What's the error message if this fails? Might be worth passing a custom message if there's no more suitable assertion to use, e.g.:
```
assertTrue(stateMachineKeys.contains(basicEvent.getTargetState()),
        "Unknown target state %s".formatted(basicEvent.getTargetState()));
```

sam803: We just get an assertion failed message - will add this :)

Joe-Edwards-GDS (approved):
Thanks for this!

PR #2076, opened 2024-06-26
PYIC-6819: new txma events

sam803:
## Proposed changes
New audit events for account intervention

### What changed

Two new events added:
 - IPV_ACCOUNT_INTERVENTION_START when a user has been identified as needing to reprove their identity and starts to do so
 - IPV_ACCOUNT_INTERVENTION_END on failed or successful completion of a user reproving their identity 

### Why did it change
So we can start tracking account interventions

### Issue tracking
https://govukverify.atlassian.net/browse/PYIC-6819

### DIFF
--- a/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
+++ b/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
@@ -183,8 +183,7 @@ public class BuildClientOauthResponseHandler
                             new AuditRestrictedDeviceInformation(input.getDeviceInformation())));
 
             var isReproveIdentity = clientOAuthSessionItem.getReproveIdentity();
-            if (Boolean.TRUE.equals(isReproveIdentity)
-                    && List.of(Vot.P0, Vot.P2).contains(ipvSessionItem.getVot())) {
+            if (Boolean.TRUE.equals(isReproveIdentity)) {
                 auditService.sendAuditEvent(
                         AuditEvent.createWithoutDeviceInformation(
                                 AuditEventTypes.IPV_ACCOUNT_INTERVENTION_END,
--- a/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
+++ b/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
@@ -21,11 +21,13 @@ import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport
 import uk.gov.di.ipv.core.library.auditing.AuditEvent;
 import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
 import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionAccountIntervention;
 import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestrictedDeviceInformation;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
 import uk.gov.di.ipv.core.library.domain.ErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
 import uk.gov.di.ipv.core.library.domain.JourneyRequest;
+import uk.gov.di.ipv.core.library.enums.Vot;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.SqsException;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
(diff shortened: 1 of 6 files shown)


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionAccountIntervention.java
```diff
@@ -0,0 +1,36 @@
...
+    @JsonProperty private final String type;
+
+    private AuditExtensionAccountIntervention(String type) {
+        this.type = type;
+    }
+
+    private AuditExtensionAccountIntervention(String type, Boolean success) {
+        this.type = type;
+        this.success = success;
+    }
```
Joe-Edwards-GDS, line 21: This would be fractionally more concise as a record class:
```
public record AuditExtensionAccountIntervention(String type, Boolean success) implements AuditExtensions {
    private static final String REPROVE_IDENTITY_TYPE = "reprove_identity";

    public static AuditExtensionAccountIntervention
            newReproveIdentityAuditExtensionAccountIntervention() {
        return new AuditExtensionAccountIntervention(REPROVE_IDENTITY_TYPE, null);
    }

    public static AuditExtensionAccountIntervention
            newReproveIdentityAuditExtensionAccountIntervention(boolean success) {
        return new AuditExtensionAccountIntervention(REPROVE_IDENTITY_TYPE, success);
    }
}
```
(Not a big deal)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionAccountIntervention.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static AuditExtensionAccountIntervention
+            newReproveIdentityAuditExtensionAccountIntervention() {
```
Joe-Edwards-GDS: You probably don't need the duplicate `AuditExtensionAccountIntervention` since the class qualifies it already. E.g.:
```
AuditExtensionAccountIntervention.reproveIdentity();
```
(Feel free to add a verb if you like)

lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
```diff
@@ -180,6 +182,20 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
                             auditEventUser,
                             new AuditRestrictedDeviceInformation(input.getDeviceInformation())));
 
+            var isReproveIdentity = clientOAuthSessionItem.getReproveIdentity();
+            if (Boolean.TRUE.equals(isReproveIdentity)
+                    && ipvSessionItem.getVot() != null
+                    && List.of(Vot.P0, Vot.P2).contains(ipvSessionItem.getVot())) {
```
Joe-Edwards-GDS: What's this check for? I'm not sure it's possible to reach here with a null VOT, but if they do, it's probably a failure.

lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
```diff
@@ -180,6 +182,20 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
+                auditService.sendAuditEvent(
+                        AuditEvent.createWithoutDeviceInformation(
+                                AuditEventTypes.IPV_ACCOUNT_INTERVENTION_END,
+                                configService.getSsmParameter(ConfigurationVariable.COMPONENT_ID),
+                                auditEventUser,
+                                AuditExtensionAccountIntervention
+                                        .newReproveIdentityAuditExtensionAccountIntervention(
+                                                ipvSessionItem.getErrorCode() == null)));
```
Joe-Edwards-GDS: I think this should be checking the VOT. For now P2 = success, anything else = failure.

(However in future this will need to capture P1 as well - cc @blakeyp )

Joe-Edwards-GDS (commented):
Thanks, I think this is almost right, but need to tweak the success criterion.

I wasn't very clear on the ticket sorry!

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionAccountIntervention.java
sam803: The only reason I did a class was to have a private constructor and enforce the use of the static functions to create the extension so that it controls what types can be created

lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
sam803: Mainly to make the existing tests pass - but I can change them instead :)

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/auditing/extension/AuditExtensionAccountIntervention.java
Joe-Edwards-GDS: Ah of course, you can't hide the default constructor with a record 🤔 

TBH I don't think that's a massive deal (if you want to enforce safety, you could make the `type` an Enum, perhaps) - but happy for you to take either approach.

lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
Joe-Edwards-GDS: Do we need the VOT check here at all? I think it should count as an unsuccessful intervention if they reach here with a `null` VOT... (which should be captured by the P2 check below).

We might also need to consider inherited identity VOTs, i.e. whether those are acceptable as a means of clearing an account intervention - you could try asking Simon Parsons, but otherwise fine to assume not for now.

lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
sam803: Ah ok thanks - should we not check the error code at all then?

Joe-Edwards-GDS: Happy to leave in the error code check, but I don't think it's needed.

The VOT shouldn't be set to P2 until we've confirmed that they have successfully proved their identity. In practice the only error code you can get after that is a timeout on the success screen - but it doesn't hurt to 'fail safe'.

sam803: thanks, will leave it as it is then :)

Joe-Edwards-GDS (approved):

PR #2081, opened 2024-06-27
BAU: Run spotless in pre-commit with local hook

Wynndow:
### What changed

Run spotless in pre-commit with local hook

### Why did it change

Rather than using a remote plugin, we can just run it locally and remove that dependance on a personal repo. Should have done it this way the first time around.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,11 @@ repos:
         verbose: true
         args: [--soft-fail]
 
-  - repo: https://github.com/Wynndow/spotless-pre-commit.git
-    rev: 'v0.2.0'
+  - repo: local
     hooks:
       - id: spotlessCheck
+        name: Run spotless check
+        entry: ./gradlew spotlessCheck
+        language: system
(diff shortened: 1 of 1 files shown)


sam803 (dismissed):

.pre-commit-config.yaml
```diff
@@ -28,7 +28,10 @@ repos:
...
 
-  - repo: https://github.com/Wynndow/spotless-pre-commit.git
-    rev: 'v0.2.0'
+  - repo: local
     hooks:
       - id: spotlessCheck
+        name: Run spotless check
+        entry: ./gradlew spotlessCheck
+        language: system
+        pass_filenames: false
```
Joe-Edwards-GDS, line 37: For bonus points, can we restrict this to only run if you modify a java file? That's 99% of commits, but it's quite slow for that 1% 😄

Wynndow: Every Little Helps

Joe-Edwards-GDS (approved):

PR #2083, opened 2024-06-27
PYIC-6897: handle partially migrated vcs

sam803:
## Proposed changes
Handle partially migrated evcs vcs

### What changed

If there are partially migrated evcs credentials, fall back to use the tactical vs

### Why did it change
After turning on evcs writes, we will have 10 days in which we may have only partially migrated vcs due to the async nature of face to face journeys

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6897

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -388,26 +388,26 @@ public class CheckExistingIdentityHandler
         if (isF2f && evcsVcs.size() == 1) {
             return true;
         }
-        // Tactical contains an extra non-F2F VC
-        if (tacticalVcs.size() == evcsVcs.size() + 1) {
-            return true;
-        }
         // Tactical contains the same as pending EVCS, with one extra F2F VC
-        var extraF2fVcs =
-                tacticalVcs.stream()
-                        .filter(
-                                credential ->
-                                        F2F.getId().equals(credential.getCriId())
-                                                && evcsVcs.stream()
-                                                        .noneMatch(
-                                                                evcsVC ->
-                                                                        evcsVC.getVcString()
-                                                                                .equals(
-                                                                                        credential
-                                                                                                .getVcString())))
-                        .toList();
+        if (tacticalVcs.size() == evcsVcs.size() + 1) {
 
-        return extraF2fVcs.size() == 1;
+            var extraF2fVcs =
+                    tacticalVcs.stream()
+                            .filter(
+                                    credential ->
+                                            F2F.getId().equals(credential.getCriId())
+                                                    && evcsVcs.stream()
+                                                            .noneMatch(
+                                                                    evcsVC ->
+                                                                            evcsVC.getVcString()
+                                                                                    .equals(
+                                                                                            credential
+                                                                                                    .getVcString())))
+                            .toList();
+
+            return extraF2fVcs.size() == 1;
+        }
+        return false;
     }
 
     @ExcludeFromGeneratedCoverageReport
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -362,6 +362,7 @@ public class CheckExistingIdentityHandler
                     if (vcBundle.isPendingEvcsIdentity()) {
                         hasPartiallyMigratedVcs = hasPartiallyMigratedVcs(tacticalVcs, vcBundle);
                     }
+                    // only use these vcs if they have been fully migrated
                     if (!hasPartiallyMigratedVcs) {
                         return vcBundle;
                     }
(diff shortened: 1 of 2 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -341,31 +341,80 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+                hasPartiallyMigratedVcs =
+                        hasPartiallyMigratedVcs(
+                                tacticalVcs,
+                                evcsIdentityVcs,
+                                isPendingEvcs,
+                                vcBundle.isF2fIdentity());
+            }
+            logIdentityMismatches(tacticalVcs, evcsVcs, hasPartiallyMigratedVcs);
+            // only use these evcs vcs if they exist and have been fully migrated
+            if (vcBundle != null && !hasPartiallyMigratedVcs) {
```
Joe-Edwards-GDS, line 371: I think there's a problematic case here (somewhat my fault...). This is likely to trigger a migration, but in that case there actually _will_ be VCs in EVCS.

I've checked with Taher and EVCS will produce a 409 Conflict if we attempt to store a VC that already exists.

The neatest solution might be to return a bundle with `hasEvcsIdentity = true` and `isPendingEvcsIdentity = true` - essentially 'filling in' the missing pending VC from the tactical store? Happy to hear other suggestions though.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -341,31 +341,80 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+
     @ExcludeFromGeneratedCoverageReport
     private void logIdentityMismatches(
             List<VerifiableCredential> tacticalVcs,
-            Map<EvcsVCState, List<VerifiableCredential>> evcsVcs) {
+            Map<EvcsVCState, List<VerifiableCredential>> evcsVcs,
+            boolean hasPartiallyMigratedVcs) {
 
+        if (hasPartiallyMigratedVcs) {
+            return;
```
Joe-Edwards-GDS, line 421: Can we log an INFO here, just to keep track of it?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -341,31 +341,80 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+                        .filter(
+                                credential ->
+                                        F2F.getId().equals(credential.getCriId())
+                                                && evcsVcs.stream()
+                                                        .noneMatch(
+                                                                evcsVC ->
+                                                                        evcsVC.getVcString()
+                                                                                .equals(
+                                                                                        credential
+                                                                                                .getVcString())))
```
Joe-Edwards-GDS: (thanks spotless...)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -341,31 +341,80 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+                        .toList();
+
+        return extraF2fVcs.size() == 1;
```
Joe-Edwards-GDS: Do you also need to check `tacticalVcs.size() == evcsVcs.size() + 1`, in case they differ in some non-F2F VCs?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: Sure, I had a warning initially but thought it might be a bit noisy

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: I thought the change I made to the `migrateCredentialsToEVCS` function would cover this (only migrates vcs that have migrated == null)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: Yes can do - wasn't sure if that was likely to happen

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -388,6 +388,10 @@ private boolean hasPartiallyMigratedVcs(
         if (isF2f && evcsVcs.size() == 1) {
             return true;
         }
+        // Tactical contains an extra non-F2F VC
+        if (tacticalVcs.size() == evcsVcs.size() + 1) {
```
Joe-Edwards-GDS, line 392: I think this needs to be combined with the check below - i.e. there is exactly one extra VC _and_ it is F2F.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -413,6 +417,7 @@ private void logIdentityMismatches(
             boolean hasPartiallyMigratedVcs) {
 
         if (hasPartiallyMigratedVcs) {
+            LOGGER.info("found partially migrated vcs");
```
Joe-Edwards-GDS: It's not critical, but can we try and use `LogHelper.buildLogMessage` for all these messages? It helps keep the format consistent.

(Otherwise we have some where `message` is a plain string, and some where `message` is an object with several properties)

Joe-Edwards-GDS (approved):

PR #2085, opened 2024-06-28
PYIC-5759 gpg45 evaluator

Joe-Edwards-GDS:
Updated GPG45 profile evaluator to use the parsed vocab classes

### DIFF
--- a/libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
+++ b/libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
@@ -1,35 +1,27 @@
 package uk.gov.di.ipv.core.library.gpg45;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.type.CollectionType;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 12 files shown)


libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -54,152 +44,72 @@ public Optional<Gpg45Profile> getFirstMatchingProfile(
...
-        processFraudWithActivityItems(evidenceMap);
+    public Gpg45Scores buildScore(List<VerifiableCredential> vcs) {
+
+        var identityChecks =
+                vcs.stream()
+                        .filter(vc -> vc.getCredential() instanceof IdentityCheckCredential)
+                        .flatMap(
+                                vc ->
+                                        ((IdentityCheckCredential) vc.getCredential())
+                                                .getEvidence().stream())
```
Joe-Edwards-GDS, line 55: Do we need a null-check on evidence here? I don't think it should ever be null for an `IdentityCheckCredential`...

thebauSoftwire: According to [this](https://vocab.account.gov.uk/v1/json-schemas/IdentityCheckCredential.json), evidence is required so I don't think we need to do a null check here

thebauSoftwire (approved):
Nice!

PR #2089, opened 2024-06-28
PYIC-6903: Store the evcs access token regardless if EVCS read/writes are enabled

sam803:
## Proposed changes

Store the evcs access token regardless if EVCS read/writes are enabled

### Why did it change

We currently only store the EVCS access token if one of the read/write feature flags is set, however we need to start storing it ahead of time as well, to cover the case where evcs is disabled at the start of the journey but enabled at the end of the journey

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6903


Joe-Edwards-GDS (approved):
Looks very safe :)

PR #2095, opened 2024-07-01
PYIC-6816: replace String cri in Verifiable credential with Cri enum

sam803:
## Proposed changes

Use the CRI enum rather than a string in VerifiableCredential

### What changed
 - VerifiableCredential class
 - Assoicated method calls and tests that use the CRI 

### Why did it change

Since the CRI enum was introduced in an earlier PR, we can use it in the Verifiable Credential to ensure that all passed CRI ids are valid

### Issue tracking

### DIFF
--- a/lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
+++ b/lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
@@ -75,6 +75,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.MFA_RESET;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.TICF_CRI_BETA;
+import static uk.gov.di.ipv.core.library.domain.Cri.ADDRESS;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_GET_CREDENTIAL;
 import static uk.gov.di.ipv.core.library.domain.IpvJourneyTypes.INITIAL_JOURNEY_SELECTION;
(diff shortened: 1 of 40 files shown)


lambdas/build-user-identity/src/test/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandlerTest.java
```diff
@@ -189,7 +190,7 @@ void shouldReturnCredentialsWithCiMitVCOnSuccessfulUserInfoRequest() throws Exce
                 .thenReturn(
                         VerifiableCredential.fromValidJwt(
                                 TEST_USER_ID,
-                                "test-cri-id",
+                                ADDRESS,
```
Joe-Edwards-GDS: This should really be `Cri.CIMIT` - you might need to add it to the enum.

It's a slightly unusual case, in that we don't get them from a CRI and we never store them (but the same is true of TICF). I'm not sure if you'll need to exclude it from any usages of `.values()`.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/Cri.java
```diff
@@ -38,6 +38,15 @@ public enum Cri {
         this.isNonEvidenceCri = isNonEvidence;
     }
 
+    public static Cri valueById(String id) {
```
Joe-Edwards-GDS: Maybe `fromId`?

Joe-Edwards-GDS (commented):
Before merging this I might run a scan in prod to make sure we don't have any VCs that don't have an id in this set.

(Should be unlikely!)

Joe-Edwards-GDS (approved):
We could try running a scan tomorrow morning - although I can't think of any reasons why it wouldn't be fine.

sam803:
> We could try running a scan tomorrow morning - although I can't think of any reasons why it wouldn't be fine.

Thanks, that would be good :)

PR #2100, opened 2024-07-02
PYIC-6816: use enum in CriCallbackRequest

sam803:
## Proposed changes

Use the CRI enum rather than a string in the CriCallbackRequest object

### What changed 
 - CriCallbackRequest class
 - Assoicated method calls and tests that use the CriCallbackRequest object
 - Added serializer and deserializer for Cri so that the Cri ID is used rather than the enum name

### Why did it change

Since the CRI enum was introduced in an earlier PR, we can use it in the code where a string CRI ID is used, to ensure that all passed CRI ids are valid

### Issue tracking

### DIFF
--- a/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/helpers/AsyncCriResponseHelper.java
+++ b/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/helpers/AsyncCriResponseHelper.java
@@ -2,9 +2,6 @@ package uk.gov.di.ipv.core.processasynccricredential.helpers;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.processasynccricredential.domain.BaseAsyncCriResponse;
 import uk.gov.di.ipv.core.processasynccricredential.domain.ErrorAsyncCriResponse;
 import uk.gov.di.ipv.core.processasynccricredential.domain.SuccessAsyncCriResponse;
@@ -13,9 +10,6 @@ import uk.gov.di.ipv.core.processasynccricredential.dto.CriResponseMessageDto;
 import static uk.gov.di.ipv.core.library.domain.Cri.F2F;
 
 public class AsyncCriResponseHelper {
-    private static final Logger LOGGER = LogManager.getLogger();
-    public static final String DEFAULT_CREDENTIAL_ISSUER = F2F.getId();
-
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
     private AsyncCriResponseHelper() {}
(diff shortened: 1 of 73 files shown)


lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/helpers/AsyncCriResponseHelper.java
```diff
@@ -33,14 +34,14 @@ public static BaseAsyncCriResponse getAsyncResponseMessage(String criResponseMes
         }
         if (criResponseMessageDto.getError() == null) {
             return SuccessAsyncCriResponse.builder()
-                    .credentialIssuer(criResponseMessageDto.getCredentialIssuer())
+                    .credentialIssuer(Cri.fromId(criResponseMessageDto.getCredentialIssuer()))
```
Joe-Edwards-GDS: I'm pretty suspicious of this tbh - @Wynndow do you know where this is spec'd (I can ask David/Kiwi if not).

The property in the the DTO is `iss`, which is strongly indicative of an 'issuer' field (i.e. the URI-formatted component id), rather than Cri ID (i.e. 'f2f').

As far as I can tell, F2F isn't actually supplying this field at all, and it defaults to F2F (as per logic above), so it might all be fine, but it would be good to work out what it's supposed to be and handle it properly.

Wynndow: I don't I'm afraid - i'm not too clued up on the async return stufff

Joe-Edwards-GDS (commented):
This looks broadly fine, but I'd like to check what's going on with the issuer in the async callback - I'm concerned if F2F start passing back a proper 'issuer' value we might break.

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/helpers/AsyncCriResponseHelper.java
Joe-Edwards-GDS: I can't see any reference to a `iss` field in the [ADR](https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3834875841/ADR+0058+CRI-+gt+IPV+Core+Out-of-band+Result+Communication#Asynchronous-success-response-message) or [Kiwi code](https://github.com/govuk-one-login/ipv-cri-f2f-api/blob/6883e5ab305e0400719a1912e328cc9c450652e8/src/utils/IPVCoreEvent.ts) and we get logs full of 'Credential Issuer not set', so I don't think this field actually exists...

Can we remove it from the `CriResponseMessageDto`? For now, it should be fine to default this to F2F (since that's the current behaviour anyway), but ideally we'd refactor to use `ConfigService#getCriByIssuer`, using the VC's issuer field.

cc @blakeyp - we'll need to change this if we start consuming any other CRIs asynchronously, but it should be obvious if that happens.

sam803: Thanks @Joe-Edwards-GDS - I've removed it and defaulting to F2F

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriDeserializer.java
```diff
@@ -0,0 +1,23 @@
...
+        this(null);
+    }
+
+    public CriDeserializer(Class<Cri> c) {
+        super(c);
+    }
+
+    @Override
+    public Cri deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+        return Cri.fromId(p.getText());
```
Joe-Edwards-GDS: Does this need to handle `null`? I notice that the `CriCheckingService` checks for a null/empty value. It may well be better to throw here (if it's an error anyway), but worth making sure we get a useful error message.

(We could also handle `null` in `Cri.fromId`, but that might itself be confusing!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CriCallbackRequest.java
```diff
@@ -15,7 +17,10 @@
 @Builder
 public class CriCallbackRequest {
     private String authorizationCode;
-    private String credentialIssuerId;
+
+    @JsonProperty("credentialIssuerId")
+    private Cri credentialIssuer;
```
Joe-Edwards-GDS: (I don't _love_ having our model classes different to the JSON structure, but I can appreciate the convenience)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriDeserializer.java
sam803: We should get an IllegalArgumentException if its null, but can change to something that must be caught and handled instead?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/CriCallbackRequest.java
sam803: same, I can change back to a string then provide an extra method, getCri to parse it into the enum?

Joe-Edwards-GDS: Yeah that might be cleaner - and saves you the serialisation classes (and you can `@JsonIgnore` it)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriDeserializer.java
Joe-Edwards-GDS: I'm not too fussed, so long as the message makes it obvious what's gone wrong (i.e. is not just `NullPointerException: null`).

If you're moving it to a separate method, you could just make that method return `null` if the `credentialIssuerId` is `null` and rely on the existing validationg logic.

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2109, opened 2024-07-02
PYIC-6961: Return build-client-oauth-response before checking session

Wynndow:
### What changed

Return build-client-oauth-response before checking session

### Why did it change

If a user has lost their session when returning from a CRI (very possibly because the redirect has opened a different browser), then we show them a recoverable error page with a continue button. That button fires a `build-client-oauth-response` event into core-back.

Core-back should see that event and return a
`build-client-oauth-response` journey, which the step function uses to invoke the lambda to actually close the OAuth session with the client.

Instead we've been throwing an error due to the ipvSessionId not existing.

This changes the logic to return that special case response before we try to check the session ID.

### Issue tracking

- [PYIC-6961](https://govukverify.atlassian.net/browse/PYIC-6961)

[PYIC-6961]: https://govukverify.atlassian.net/browse/PYIC-6961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -69,9 +69,9 @@ public class ProcessJourneyEventHandler
     public static final String CURRENT_PAGE = "currentPage";
     private static final String CORE_SESSION_TIMEOUT_STATE = "CORE_SESSION_TIMEOUT";
     private static final String NEXT_EVENT = "next";
-    private static final String END_SESSION_EVENT = "build-client-oauth-response";
-    private static final StepResponse END_SESSION_RESPONSE =
-            new ProcessStepResponse(END_SESSION_EVENT, null);
+    private static final String BUILD_CLIENT_OAUTH_RESPONSE_EVENT = "build-client-oauth-response";
+    private static final StepResponse BUILD_CLIENT_OAUTH_RESPONSE =
+            new ProcessStepResponse(BUILD_CLIENT_OAUTH_RESPONSE_EVENT, null);
     public static final String BACK_EVENT = "back";
     private final IpvSessionService ipvSessionService;
     private final AuditService auditService;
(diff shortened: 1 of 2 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -69,9 +69,9 @@ public class ProcessJourneyEventHandler
     public static final String CURRENT_PAGE = "currentPage";
     private static final String CORE_SESSION_TIMEOUT_STATE = "CORE_SESSION_TIMEOUT";
     private static final String NEXT_EVENT = "next";
-    private static final String END_SESSION_EVENT = "build-client-oauth-response";
-    private static final StepResponse END_SESSION_RESPONSE =
-            new ProcessStepResponse(END_SESSION_EVENT, null);
+    private static final String BUILD_CLIENT_OAUTH_RESPONSE_EVENT = "build-client-oauth-response";
+    private static final StepResponse BUILD_CLIENT_OAUTH_RESPONSE =
+            new ProcessStepResponse(BUILD_CLIENT_OAUTH_RESPONSE_EVENT, null);
```
Joe-Edwards-GDS, line 74: I see someone has 'fixed' this - I deliberately made this a separate constant, because the one above is for the journey event, and this one is for the journey response type (they just happen to be the same!)

Fine to leave..!

Joe-Edwards-GDS (approved):

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: Not sure I follow?

Joe-Edwards-GDS: There are two separate constants here:
- The journey event that triggers this special behaviour (`"build-client-oauth-response"`)
- The response emitted - a process step with `{ lambda: "build-client-oauth-response" }`

The two strings match, but they are not conceptually 'the same' - e.g. if we rename the lambda, one will change but the other won't. Just me being grumpy and pedantic :)

Joe-Edwards-GDS: If we changed the journey event to something explicit like `recovery-return-to-rp` then it'd be a lot less confusing

Wynndow: Yeah agreed. I think lets get this out and then look at fixing it up - will need co-ordination with the frontend

PR #2110, opened 2024-07-02
PYIC-5759: update user identity service to use new VerifiableCredential field

thebauSoftwire:
### What changed
The methods in `UserIdentityService` have been updated to use the new vc parser:
- generatePassportClaim
- generateAddressClaim
- generateNinoClaim
- generateDrivingPermitClaim
- isEvidenceVc
- getIdentityClaim
Updated UserIdentity to return classes instead of json nodes.

| User info JSON | before | after |
|-|-|-|
| passport | Screenshot 2024-07-07 at 18 48 50 | Screenshot 2024-07-07 at 18 49 36 |
| address | Screenshot 2024-07-07 at 18 50 39 | Screenshot 2024-07-07 at 18 50 55 |
| driving permit | Screenshot 2024-07-07 at 18 52 44 | Screenshot 2024-07-07 at 18 54 55 |

Name and birthdate changes as done as part of this PR: https://github.com/govuk-one-login/ipv-core-back/pull/2122
### Why did it change

We want to use the classes provided by vocab instead of the current json handling. 

### Issue tracking


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -508,52 +500,29 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
-            return OBJECT_MAPPER
-                    .readTree(SignedJWT.parse(credential).getPayload().toString())
-                    .path(VC_CLAIM)
-                    .path(node);
-        } catch (JsonProcessingException | ParseException e) {
-            throw new CredentialParseException(
-                    "Encountered a parsing error while attempting to parse VC store item: "
-                            + e.getMessage());
-        }
+    private List<NameParts> getNamePartsFromCredentialSubjectName(uk.gov.di.model.Name name) {
```
Joe-Edwards-GDS: Can we get rid of our `Name` class and replace it all with `uk.gov.di.model.Name`? We could do this separately if it's too ingrained, but I think that's where a lot of the real value will come :)

For example, that would make this method redundant!

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -508,52 +500,29 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
-                        BIRTH_DATE_PROPERTY_NAME,
-                        OBJECT_MAPPER
-                                .getTypeFactory()
-                                .constructCollectionType(List.class, BirthDate.class));
+        var names =
+                credentialSubject.getName() != null
+                        ? credentialSubject.getName().stream()
+                                .map(name -> new Name(getNamePartsFromCredentialSubjectName(name)))
+                                .toList()
+                        : new ArrayList<Name>();
```
Joe-Edwards-GDS: Prefer `List.of()` or `Collections.emptyList()` (unless you need it to be mutable).

If you make the change above, this could also use the Objects helper, if you think it's neater:
```
var names = requireNonNullElse(credentialSubject.getName(), List.of());
```
(the explicit null check is also fine though)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -508,52 +500,29 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
+
+        var birthDates =
+                credentialSubject.getBirthDate() != null
+                        ? credentialSubject.getBirthDate().stream()
+                                .map(bd -> new BirthDate(bd.getValue()))
+                                .toList()
```
Joe-Edwards-GDS: (similarly, if we can change `IdentityClaim` to use the vocab `BirthDate` then this mapping isn't needed)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -571,23 +540,53 @@ private Optional<JsonNode> generateAddressClaim(List<VerifiableCredential> vcs)
...
         }
 
-        var addressNode =
-                extractSubjectDetailFromVc(
-                        ADDRESS_PROPERTY_NAME,
-                        addressVc.get(),
-                        "Error while parsing Address CRI credential",
-                        ErrorResponse.FAILED_TO_GENERATE_ADDRESS_CLAIM);
+        var credentialSubject =
+                ((IdentityCheckCredential) addressVc.get().getCredential()).getCredentialSubject();
```
Joe-Edwards-GDS: The address VC should be an `AddressCredential`, not `IdentityCheckCredential`!

Also, it's a bit safer to do this with a checked cast, e.g.:
```
if (addressVc.get().getCredential() instanceof AddressCredential addressCredential) {
    var address = addressCredential.getCredentialSubject().getAddress();
    ...
} else {
    // error - not an address credential!
}
```

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -571,23 +540,53 @@ private Optional<JsonNode> generateAddressClaim(List<VerifiableCredential> vcs)
...
+                            ErrorResponse.CREDENTIAL_SUBJECT_MISSING.getMessage()));
+            throw new HttpResponseExceptionWithErrorBody(
+                    500, ErrorResponse.FAILED_TO_GENERATE_ADDRESS_CLAIM);
+        }
+
+        var address = credentialSubject.getAddress();
 
-        if (addressNode.isMissingNode()) {
-            LOGGER.error(LogHelper.buildLogMessage("Address property is missing from address VC"));
+        if (address == null || address.isEmpty()) {
```
Joe-Edwards-GDS: this is perfectly clear, but you can also use `software.amazon.awssdk.utils.CollectionUtils.isNullOrEmpty` if you prefer

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -620,42 +625,54 @@ private Optional<JsonNode> generateNinoClaim(
...
         }
+        var mappedNino = nino.stream().map(this::mapNinoToSocialSecurityRecord).toList();
 
-        return Optional.of(ninoNode);
+        return Optional.of(mappedNino);
     }
 
-    private Optional<JsonNode> generatePassportClaim(List<VerifiableCredential> vcs)
-            throws HttpResponseExceptionWithErrorBody {
+    private SocialSecurityRecord mapNinoToSocialSecurityRecord(SocialSecurityRecordDetails nino) {
```
Joe-Edwards-GDS: (Ditto again - it would be nice to move to using the vocab type and remove this mapping :) )

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -508,52 +500,29 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
-            return OBJECT_MAPPER.treeToValue(propertyNode, valueType);
-        } catch (JsonProcessingException e) {
-            LOGGER.error(LogHelper.buildErrorMessage("Failed to parse VC JWT", e));
-            throw new HttpResponseExceptionWithErrorBody(
-                    500, ErrorResponse.FAILED_TO_GENERATE_IDENTIY_CLAIM);
-        }
-    }
+    private IdentityClaim getIdentityClaim(VerifiableCredential vc) {
+        var credentialSubject =
+                ((IdentityCheckCredential) vc.getCredential()).getCredentialSubject();
```
Joe-Edwards-GDS: As below - prefer doing this in a checked cast:
```
if (vc.getCredential() instanceof IdentityCheckCredential credential) {
    var name = credential.getCredentialSubject().getName();
    ...
} else {
    // error - not an identity check credential!
}
```

However, I think you also want this to work for other types of credential? In fact, I think you can do something like:
```
if (vc.getCredential().getCredentialSubject() instanceof PersonWithIdentity person) {
    var name = person.getName();
    ...
} else {
    // No applicable subject in this credential
}
```
This should work for both `IdentityCheckCredential` and `IdentityAssertionCredential`

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -620,42 +625,54 @@ private Optional<JsonNode> generateNinoClaim(
...
         }
 
-        var passportNode =
-                extractSubjectDetailFromVc(
-                        PASSPORT_PROPERTY_NAME,
-                        passportVc.get(),
-                        "Error while parsing Passport CRI credential",
-                        ErrorResponse.FAILED_TO_GENERATE_PASSPORT_CLAIM);
+        var credentialSubject =
+                ((IdentityCheckCredential) passportVc.get().getCredential()).getCredentialSubject();
```
Joe-Edwards-GDS: (ditto)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -665,33 +682,46 @@ private Optional<JsonNode> generateDrivingPermitClaim(
...
 
-        var drivingPermitNode =
-                extractSubjectDetailFromVc(
-                        DRIVING_PERMIT_PROPERTY_NAME,
-                        drivingPermitVc.get(),
-                        "Error while parsing Driving Permit CRI credential",
-                        ErrorResponse.FAILED_TO_GENERATE_DRIVING_PERMIT_CLAIM);
+        var credentialSubject =
+                ((IdentityCheckCredential) drivingPermitVc.get().getCredential())
+                        .getCredentialSubject();
```
Joe-Edwards-GDS: (ditto)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -707,33 +737,24 @@ private Optional<VerifiableCredential> findVc(
...
-            String errorLog,
-            ErrorResponse errorResponse)
-            throws HttpResponseExceptionWithErrorBody {
-        try {
-            return getVcClaimNode(vc.getVcString(), VC_CREDENTIAL_SUBJECT).path(detailName);
-        } catch (CredentialParseException e) {
-            LOGGER.error(LogHelper.buildErrorMessage(errorLog, e));
-            throw new HttpResponseExceptionWithErrorBody(SC_SERVER_ERROR, errorResponse);
+    private boolean isEvidenceVc(VerifiableCredential vc) throws CredentialParseException {
+        var vcEvidence = ((IdentityCheckCredential) vc.getCredential()).getEvidence();
```
Joe-Edwards-GDS: (ditto)

Joe-Edwards-GDS (commented):
This is good - but two classes of comment:
- We can use checked casts to make sure we behave appropriately when things aren't the right type
- We want to replace the address/dob/name/nino types with the vocab ones, but happy to leave that to a separate PR if it'll add too much complexity to this one

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -105,7 +105,8 @@ public enum ErrorResponse {
     UNKNOWN_RESET_TYPE(1088, "unknown resetType received"),
     ERROR_CALLING_DCMAW_ASYNC_CRI(1089, "Error calling the DCMAW Async CRI"),
     FAILED_TO_UPDATE_IDENTITY(1090, "Failed to update identity"),
-    INVALID_SIGNING_KEY_TYPE(1091, "Signing key is invalid type. Must be EC or RSA");
+    INVALID_SIGNING_KEY_TYPE(1091, "Signing key is invalid type. Must be EC or RSA"),
+    INVALID_CREDENTIAL_TYPE(1092, "Invalid credential type");
```
Joe-Edwards-GDS: What's this for?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -446,16 +447,22 @@ static VerifiableCredential vcMissingPassportProperty() {
                         .build());
     }
 
+    private static VerifiableCredential generateAddressVc(TestVc.TestCredentialSubject subject) {
```
Joe-Edwards-GDS, line 460: (Hopefully this is a clean merge with mine, but it might conflict slightly)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -506,25 +508,37 @@ private List<NameParts> getNamePartsFromCredentialSubjectName(uk.gov.di.model.Na
...
+                                    .toList()
+                            : List.of();
+
+            return new IdentityClaim(names, birthDates);
 
-        return new IdentityClaim(names, birthDates);
+        } else {
+            LOGGER.error(
+                    LogHelper.buildLogMessage(
+                            "Address property missing from VC or empty address property "));
```
Joe-Edwards-GDS: I think this message isn't quite right

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -540,30 +554,36 @@ private Optional<List<Address>> generateAddressClaim(List<VerifiableCredential>
...
+            }
 
-        var address = credentialSubject.getAddress();
+            var address = credentialSubject.getAddress();
 
-        if (address == null || address.isEmpty()) {
-            LOGGER.error(
-                    LogHelper.buildLogMessage(
-                            "Address property missing from VC or empty address property "));
+            if (software.amazon.awssdk.utils.CollectionUtils.isNullOrEmpty(address)) {
```
Joe-Edwards-GDS: You can do a static import for this, so it's just `if (isNullOrEmpty(address)) {`

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: ah good spot!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
thebauSoftwire: this was for isEvidenceVc but I realised if it's not of type IdentityCheck then it should just return false instead of an error so have removed it

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -507,61 +502,50 @@ private List<ReturnCode> mapCisToReturnCodes(ContraIndicators contraIndicators)
...
+                            : List.of();
+
+            return new IdentityClaim(names, birthDates);
+
+        } else {
+            LOGGER.error(
+                    LogHelper.buildLogMessage(
+                            "Credential subject must be of type PersonWithIdentity."));
             throw new HttpResponseExceptionWithErrorBody(
                     500, ErrorResponse.FAILED_TO_GENERATE_IDENTIY_CLAIM);
```
Joe-Edwards-GDS: I'm not sure if this should be an error actually - will we definitely only pass credentials with an appropriate identity to this method?

It might be more consistent to just return an `IdentityClaim` with an empty list of names/birthDates?

Joe-Edwards-GDS (commented):
There are a few Sonar warnings that it would be good to clean up - generally places where we no longer need to throw exceptions, because we're using the already-parsed structure 😄

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: I've separated it out into this PR: https://github.com/govuk-one-login/ipv-core-back/pull/2122 for name and birthDate

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: ah you're right! Address is passed in too (though it feels like if this method is only for getting an identity claim it should only be used for VCs with an identity claim? This seems like a not-so obvious side-effect and maybe worth a refactor in future?)

Joe-Edwards-GDS: Yeah it feels a bit odd - perhaps it should return `Optional.empty()` or `null` for non-identity VCs to make it a bit clearer - or force the caller to pass in an `IdentityCheckCredential`.

Perhaps better left as a separate refactor?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: done as part of: https://github.com/govuk-one-login/ipv-core-back/pull/2122

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: done as part of: https://github.com/govuk-one-login/ipv-core-back/pull/2122

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -632,29 +611,35 @@ private Optional<JsonNode> generatePassportClaim(List<VerifiableCredential> vcs)
...
+                LOGGER.warn(mapMessage);
+                return Optional.empty();
+            }
 
-        return Optional.of(passportNode);
+            return Optional.of(passport);
+        } else {
+            LOGGER.error(LogHelper.buildLogMessage(MUST_BE_IDENTITYCHECK_MESSAGE));
+            throw new HttpResponseExceptionWithErrorBody(
+                    500, ErrorResponse.FAILED_TO_GENERATE_PASSPORT_CLAIM);
```
Joe-Edwards-GDS: Sorry one small thing - I think these all used to be warnings, but now they are errors...

Can we turn this into a warning log and return `Optional.empty` instead?

local-running/compose.yaml
```diff
@@ -118,7 +118,7 @@ services:
     environment:
       <<: *cri-config
       CREDENTIAL_ISSUER_NAME: Address (Stub)
-      CREDENTIAL_ISSUER_TYPE: USER_ASSERTED
+      CREDENTIAL_ISSUER_TYPE: ADDRESS
```
Joe-Edwards-GDS, line 121: Thanks! Sorry I missed this

Joe-Edwards-GDS (approved):
Thanks, big refactors are always a bit painful, but I hope it was somewhat satisfying!

PR #2113, opened 2024-07-03
PYIC-6897: handle case where one f2f in evcs

sam803:
## Proposed changes

In checkExistingIdentity lambda, use the tactical vcs if there are partially migrated credentials, with the evcs flags set so that the next store lambda is called next

### Why did it change

So that partially migrated credentials do not get migrated again (resulting in conflict in evcs service)

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6897

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -385,12 +385,12 @@ public class CheckExistingIdentityHandler
                 if (hasPartiallyMigratedVcs) {
                     // use tactical vcs but with evcs flags so that the store-identity lambda is
                     // called next and updates the evcs pending one
-                    return new VerifiableCredentialBundle(tacticalVcs, true, true);
+                    vcBundle = new VerifiableCredentialBundle(tacticalVcs, true, true);
                 }
             }
             logIdentityMismatches(tacticalVcs, evcsVcs, hasPartiallyMigratedVcs);
             // only use these evcs vcs if they exist and have been fully migrated
-            if (vcBundle != null && !hasPartiallyMigratedVcs) {
+            if (vcBundle != null) {
                 return vcBundle;
             }
         }
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -381,6 +381,21 @@ public class CheckExistingIdentityHandler
                                 evcsIdentityVcs,
                                 isPendingEvcs,
                                 vcBundle.isF2fIdentity());
+
+                if (hasPartiallyMigratedVcs) {
+                    var vcs = vcBundle.credentials;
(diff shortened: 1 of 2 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -381,6 +381,21 @@ private VerifiableCredentialBundle getVerifiableCredentials(
                                 evcsIdentityVcs,
                                 isPendingEvcs,
                                 vcBundle.isF2fIdentity());
+
+                if (hasPartiallyMigratedVcs) {
+                    var vcs = vcBundle.credentials;
+                    if (evcsIdentityVcs.size() == 1
+                            && F2F.equals(evcsIdentityVcs.get(0).getCri())) {
+                        // get the evcs f2f and add it to the tactical vcs if it does not exist
```
rpayal: tactical vcs will always have all the vc's including f2f one.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -381,6 +381,21 @@ private VerifiableCredentialBundle getVerifiableCredentials(
                                 evcsIdentityVcs,
                                 isPendingEvcs,
                                 vcBundle.isF2fIdentity());
+
+                if (hasPartiallyMigratedVcs) {
+                    var vcs = vcBundle.credentials;
```
rpayal: will here vcs be initialised to tacticalVcs ?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -381,6 +381,21 @@ private VerifiableCredentialBundle getVerifiableCredentials(
                                 evcsIdentityVcs,
                                 isPendingEvcs,
                                 vcBundle.isF2fIdentity());
+
+                if (hasPartiallyMigratedVcs) {
```
Joe-Edwards-GDS, line 385: I was thinking we could just do:
```
  return new VerifiableCredentialBundle(tacticalVcs, true, true);
}
```
i.e. force it to use the tactical pending VCs if the EVCS set is incomplete.

I think that's effectively what you're trying to do, but keeps the 'partial' logic separate, and hopefully clearer too?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: depends if EVCS_READ_ENABLED is true. Maybe I should be using tacticalVCs instead

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
sam803: Yeh, I think I got a bit confused :) now updated

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2117, opened 2024-07-03
Pyic 3907 vc success

Joe-Edwards-GDS:
Replace `CredentialEvidenceItem` and the `isSuccessful` logic with a streamlined version using the data-vocab classes


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/Cri.java
```diff
@@ -3,41 +3,36 @@
 import lombok.Getter;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
-import java.util.Arrays;
-import java.util.List;
-
 @Getter
 @ExcludeFromGeneratedCoverageReport
 public enum Cri {
-    PASSPORT("ukPassport"),
```
Joe-Edwards-GDS, line 12: (sorted alphabetically)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -434,70 +435,52 @@ static VerifiableCredential vcPassportMissingPassport() {
                         .build());
     }
 
+    private static VerifiableCredential generateAddressVc(TestVc.TestCredentialSubject subject) {
+        return generateVerifiableCredential(
+                TEST_SUBJECT,
+                Cri.ADDRESS,
+                TestVc.builder()
+                        .type(new String[] {VERIFIABLE_CREDENTIAL_TYPE, ADDRESS_CREDENTIAL_TYPE})
```
Joe-Edwards-GDS, line 443: (Address credentials should have type `AddressCredential`)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45IdentityCheckValidator.java
Joe-Edwards-GDS, line 1: This is the most important logic to check, as we rely on this for various reasons:
- Identifying a `fail-with-no-ci` response from a CRI in absence - e.g. thin file
- Excluding VCs from correlation and identity logic - i.e. 'bad' VCs that we shouldn't include in a generated identity

lambdas/process-async-cri-credential/src/test/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandlerTest.java
```diff
@@ -385,22 +376,17 @@ private static void verifyBatchResponseFailures(
...
+    private static OauthCriConfig createTestOauthCriConfig() throws URISyntaxException {
         return OauthCriConfig.builder()
                 .tokenUrl(new URI(""))
                 .credentialUrl(new URI(""))
                 .authorizeUrl(new URI(""))
                 .clientId("ipv-core")
                 .signingKey(EC_PRIVATE_KEY_JWK)
                 .encryptionKey(null)
-                .componentId(componentId)
+                .componentId(ProcessAsyncCriCredentialHandlerTest.TEST_COMPONENT_ID)
```
Wynndow:
```suggestion
                .componentId(TEST_COMPONENT_ID)
```

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -434,70 +435,52 @@ static VerifiableCredential vcPassportMissingPassport() {
...
-                    TestVc.builder()
-                            .credentialSubject(
-                                    TestVc.TestCredentialSubject.builder()
-                                            .address(List.of(ADDRESS_1))
-                                            .build())
-                            .build());
+            generateAddressVc(
+                    TestVc.TestCredentialSubject.builder().address(List.of(ADDRESS_1)).build());
 
     static VerifiableCredential vcAddressTwo() {
```
Wynndow, line 455: The mix of fields and static methods we have in this class confuses me sometimes. And can make the tests behaving in unexpected ways (timestamps and equalities)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
```diff
@@ -538,7 +521,7 @@ static VerifiableCredential vcExperianFraudFailed() {
                         .build();
         return generateVerifiableCredential(
                 TEST_SUBJECT,
-                Cri.EXPERIAN_KBV,
+                EXPERIAN_FRAUD,
```
Wynndow, line 524: Should I be worried that there were no broken tests with this the wrong way around?

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45IdentityCheckValidator.java
```diff
@@ -0,0 +1,56 @@
...
+            case BAV, DRIVING_LICENCE, PASSPORT -> isEvidenceSuccessful(identityCheck);
+            case EXPERIAN_FRAUD -> isFraudCheckSuccessful(identityCheck);
+            case DWP_KBV, EXPERIAN_KBV, HMRC_KBV -> isVerificationSuccessful(identityCheck);
+            case DCMAW, DCMAW_ASYNC, F2F, HMRC_MIGRATION -> isEvidenceSuccessful(identityCheck)
+                    && isVerificationSuccessful(identityCheck);
+            case NINO -> isNinoSuccessful(identityCheck);
+            case ADDRESS,
+                    CIMIT,
+                    CLAIMED_IDENTITY,
+                    TICF -> false; // These CRIs should not perform identity checks
```
Wynndow: They're not "unsuccessful". Makes we wonder if the method should be called something more specific to GPG45? `isGpg45Success`? Or something.

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45IdentityCheckValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class Gpg45IdentityCheckValidator {
+    @ExcludeFromGeneratedCoverageReport
+    private Gpg45IdentityCheckValidator() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    public static boolean isSuccessful(IdentityCheck identityCheck, Cri cri) {
+        return switch (cri) {
+            case BAV, DRIVING_LICENCE, PASSPORT -> isEvidenceSuccessful(identityCheck);
```
Wynndow, line 24: This is way better

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -204,7 +209,7 @@ void areVCsCorrelatedReturnFalseWhenNamesDiffer() throws Exception {
                 List.of(
                         TestFixtures.createVerifiableCredential(
                                 USER_ID_1,
-                                ADDRESS.getId(),
+                                PASSPORT.getId(),
```
Wynndow, line 212: Why make the VCs all from the same issuer? I don't think it actually matters for these tests, but having different issuers might be more realistic and potentially catch some weird edge case?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -51,54 +40,37 @@ public class VcHelper {
...
-        try {
-            return isValidEvidence(
-                    OBJECT_MAPPER.treeToValue(evidenceArray, CREDENTIAL_EVIDENCE_ITEM_LIST_TYPE));
-        } catch (JsonProcessingException e) {
-            throw new CredentialParseException("Unable to create credential evidence list", e);
+            return evidence.stream()
+                    .anyMatch(
+                            check -> Gpg45IdentityCheckValidator.isSuccessful(check, vc.getCri()));
         }
+        return true;
```
Wynndow, line 65: So anything that isn't an IdentityCheckCredential will be true? Such as an AddressCredential - which we won't run the validator on. But if we did run the validator on an address VC it'd be false?

lambdas/process-async-cri-credential/src/test/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandlerTest.java
Joe-Edwards-GDS: ah I think this must have been an automatic refactor 🤦

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
Joe-Edwards-GDS: Yes, it's not quite as uniform as I'd like. If you'd like to address it, you could pick up https://govukverify.atlassian.net/browse/PYIC-6973 (possibly after the first wave of vocab refactoring!)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/VcFixtures.java
Joe-Edwards-GDS: Maybe... I think it indicates how few parts of our code actually care about the VC type.

(It broke because the evidence validator now uses this explicitly, rather than trying to guess the VC type based on the evidence scores...)

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45IdentityCheckValidator.java
Joe-Edwards-GDS: Yeah, perhaps this should just be an error instead - if one of these _is_ presenting an IdentityCheck, that's an issue!

(I just got a bit nervous about tightening validation...)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: To be a bit more precise, an AddressCredential (from any CRI) will always be 'successful', but an IdentityCheckCredential from the Address CRI will be 'unsuccessful'.

As above, perhaps this should be an error instead...

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/validators/Gpg45IdentityCheckValidator.java
Joe-Edwards-GDS: I think I'd prefer to:
- If we want to make it an error, validate in the `VerifiableCredentialValidator` (SPOT-style)
- Leave this as a non-throwing check, but with a warning log

Wynndow: Yeah. that seems fair

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Wynndow: I think that't the clarification I was missing. The warning log makes sense.

Wynndow (approved):

PR #2123, opened 2024-07-04
PYIC-6883: Use async client for Audit service

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-core-common-infra/pull/995**

### What changed

Use async client for Audit service

### Why did it change

This adds the option to use an async SQS client for the Audit Service.

Inititally we want this behind a feature flag, in case we find any issues with it. This requires having two clients ready. Hopefully this won't add too much overhead as they'll only be initialized when used for the first time.

### Issue tracking

- [PYIC-6883](https://govukverify.atlassian.net/browse/PYIC-6883)

[PYIC-6883]: https://govukverify.atlassian.net/browse/PYIC-6883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ spotless {
 subprojects {
 	task allDeps(type: DependencyReportTask) {}
 	configurations.all {
+		exclude group: 'software.amazon.awssdk', module: 'netty-nio-client'
 		exclude group: 'software.amazon.awssdk', module: 'apache-client'
 	}
 }
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,6 @@ spotless {
 subprojects {
 	task allDeps(type: DependencyReportTask) {}
 	configurations.all {
-		exclude group: 'software.amazon.awssdk', module: 'netty-nio-client'
 		exclude group: 'software.amazon.awssdk', module: 'apache-client'
 	}
 }
--- a/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
+++ b/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
@@ -91,7 +91,7 @@ public class ProcessAsyncCriCredentialHandler
         this.configService = new ConfigService();
         this.verifiableCredentialValidator = new VerifiableCredentialValidator(configService);
         this.verifiableCredentialService = new VerifiableCredentialService(configService);
-        this.auditService = new AuditService(AuditService.getSqsClient(), configService);
+        this.auditService = new AuditService(AuditService.getSqsClients(), configService);
         this.ciMitService = new CiMitService(configService);
         this.criResponseService = new CriResponseService(configService);
         this.evcsService = new EvcsService(configService);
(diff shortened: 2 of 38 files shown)


lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -102,43 +102,50 @@ public ProcessAsyncCriCredentialHandler() {
...
-            } catch (VerifiableCredentialException e) {
-                LOGGER.error(
-                        new StringMapMessage()
-                                .with(
-                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
-                                        "Failed to process VC response message.")
-                                .with(LOG_ERROR_DESCRIPTION.getFieldName(), e.getErrorResponse()));
-                failedRecords.add(new SQSBatchResponse.BatchItemFailure(message.getMessageId()));
+            for (SQSMessage message : event.getRecords()) {
+                failedRecords.addAll(processOrReturnItemFailure(message));
```
Wynndow, line 110: Split this into a separate method to avoid nested try/catch blocks

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandlerTest.java
```diff
@@ -60,12 +74,22 @@ class ProcessCriCallbackHandlerTest {
...
+    @AfterEach
+    void checkAuditEventWait() {
+        InOrder auditInOrder = inOrder(mockAuditService);
+        auditInOrder.verify(mockAuditService).awaitAuditEvents();
+        auditInOrder.verifyNoMoreInteractions();
+    }
+
     @Test
-    void getJourneyResponseShouldReturnNextWhenAllChecksPassForCreatedVcs() throws Exception {
+    void shouldReturnNextWhenAllChecksPassForCreatedVcs() throws Exception {
```
Wynndow, line 88: Most of these changes are updating the tests to call the handler, and not just the `getJourneyResponse` method within it. And then asserting the outputs of the handler. Which is probably what we want.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
```diff
@@ -2,46 +2,90 @@
...
-                .region(EU_WEST_2)
-                .httpClientBuilder(UrlConnectionHttpClient.builder())
-                .build();
+    public static SqsClients getSqsClients() {
+        return new SqsClients(
+                SqsClient.builder()
+                        .region(EU_WEST_2)
+                        .httpClientBuilder(UrlConnectionHttpClient.builder())
+                        .build(),
+                SqsAsyncClient.builder().region(EU_WEST_2).build());
```
Wynndow: Deliberately using the deafult async client (netty), as we've removed use of the crt client. We should switch if/when we bring the crt client back.

build.gradle
```diff
@@ -31,7 +31,6 @@ spotless {
 subprojects {
 	task allDeps(type: DependencyReportTask) {}
 	configurations.all {
-		exclude group: 'software.amazon.awssdk', module: 'netty-nio-client'
```
Wynndow: This was originally excluded as a suggested best practice for Lambdas to not include httpclients (or any package really) that you're not using. It's brought in by the sdk automatically. We're using it now.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
Joe-Edwards-GDS: It might be worth trying the async CRT client, in case it's more reliable - the Netty client has a slower startup time which might wipe out any gains from the async calls..!

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandlerTest.java
Joe-Edwards-GDS: Probably, although it's a bit inconvenient that we have to do a load of additional JSON serialisation/deserialisation.

I think the real moral here is that we should keep our handlers small (so all the actual processing logic is tested separately)!

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/AuditService.java
Joe-Edwards-GDS: I can't reproduce the segfault with the async client so I'd like to try it in lower environments and see...

rpayal (approved):
LGTM
its behind the FS.

PR #2125, opened 2024-07-04
PYIC-6978 Updated configService featureFlag enabled check, log and return false in case of exception from ssm.

rpayal:
## Proposed changes
Handle exception from SSM for feature flag enabled check. Return false in case of featureFlag missing for the env.

### What changed
Updated configService featureFlag enabled check.
- log warning message.
- return false in case of featureFlag missing for the env.

### Issue tracking

- [PYIC-6978](https://govukverify.atlassian.net/browse/PYIC-6978)

### Other considerations

[PYIC-6978]: https://govukverify.atlassian.net/browse/PYIC-6978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -265,13 +265,22 @@ public class ConfigService {
     }
 
     public boolean enabled(FeatureFlag featureFlag) {
-        return Boolean.parseBoolean(
-                getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlag.getName()));
+        return enabled(featureFlag.getName());
     }
 
     public boolean enabled(String featureFlagValue) {
-        return Boolean.parseBoolean(
-                getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlagValue));
+        try {
+            return Boolean.parseBoolean(
+                    getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlagValue));
+        } catch (ParameterNotFoundException ex) {
+            LOGGER.warn(
+                    (new StringMapMessage())
+                            .with(
+                                    LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                    "SSM parameter not found for feature flag:"
+                                            + featureFlagValue));
+            return false;
+        }
     }
 
     public String getCoreSecretValue(ConfigurationVariable secretName) {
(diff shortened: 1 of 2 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -265,13 +266,24 @@ public Map<String, List<MitigationRoute>> getCimitConfig() throws ConfigExceptio
...
+                            getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlagValue));
+        } catch (SsmException ex) {
+            LOGGER.warn(
+                    (new StringMapMessage())
+                            .with(
+                                    LOG_MESSAGE_DESCRIPTION.getFieldName(),
+                                    "SSM parameter not found for feature flag:"
+                                            + featureFlagValue));
+        }
+        return enabled;
```
Joe-Edwards-GDS: minor thing, but I think it'd be clearer to get rid of the variable and explicitly `return false` inside the catch

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -265,13 +266,24 @@ public Map<String, List<MitigationRoute>> getCimitConfig() throws ConfigExceptio
...
 
     public boolean enabled(String featureFlagValue) {
-        return Boolean.parseBoolean(
-                getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlagValue));
+        boolean enabled = false;
+        try {
+            enabled =
+                    Boolean.parseBoolean(
+                            getSsmParameter(ConfigurationVariable.FEATURE_FLAGS, featureFlagValue));
+        } catch (SsmException ex) {
```
Joe-Edwards-GDS: Is there anything more precise we can/should filter on here (i.e. so we only catch 'parameter not found' errors, and not other exceptions)?

rpayal: yeah thought so.....can use
ssmProvider.get - ParameterNotFoundException 
ssmProvider.getMultiple - InvalidFilterKeyException 

should be ok just to use ParameterNotFoundException.

rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
rpayal: done

Joe-Edwards-GDS (approved):

PR #2128, opened 2024-07-05
PYIC-6957: Retry EVCS request if 429 status code returned

sam803:
## Proposed changes

Retry the EVCS http requests if a 429 (throttled) status code is returned

### What changed

- Added utility retry class 
- Using retry class in any code that uses the same logic (EmailService, IPVSessionService)
- If we received a 429 from evcs service, we should retry the requests with a backoff

### Why did it change

To improved resilience of the evcs service

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6957


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -46,19 +47,30 @@ public class EvcsClient {
     public static final String VC_STATE_PARAM = "state";
     private static final Logger LOGGER = LogManager.getLogger();
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final List<Integer> RETRYABLE_STATUS_CODES = Arrays.asList(429);
```
Joe-Edwards-GDS: (Not important here, but `List.of(..)` is slightly preferable for immutable lists)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
Joe-Edwards-GDS, line 1: There's an implementation of exponential backoff in `IpvSessionService#getIpvSessionByAccessToken` - it might not be, but do you think there's any sensible opportunity for a shared utility?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -200,13 +212,32 @@ private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
...
+            throws EvcsServiceException {
         LOGGER.info(LogHelper.buildLogMessage("Sending HTTP request to EVCS"));
         try {
             HttpResponse<String> response =
                     httpClient.send(evcsHttpRequest, HttpResponse.BodyHandlers.ofString());
+
+            var statusCode = response.statusCode();
+            if (RETRYABLE_STATUS_CODES.contains(statusCode) && numAttempts < MAX_RETRIES) {
+
+                var delay = (long) Math.pow(2, numAttempts++) * RETRY_DELAY_MILLIS;
```
Joe-Edwards-GDS: I'd usually prefer not to mutate a method parameter, so I might make this `numAttempts` and then pass `numAttempts + 1` to the recursive call.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/Sleeper.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.ipv.core.library.client;
+
+public interface Sleeper {
+    public void sleep(long millis) throws InterruptedException;
```
Joe-Edwards-GDS: (Given that you're mocking this out anyway, this could just be a real implementation)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
sam803: @Joe-Edwards-GDS ah yes its pretty similar! It would be better in terms of code duplication and if we think we might use this pattern elsewhere in the future, I'll have a go at it

Joe-Edwards-GDS: We might (the second option on https://govukverify.atlassian.net/browse/PYIC-6972) - have a go and see how clunky it turns out.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -203,13 +213,40 @@ private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
...
+
         try {
-            HttpResponse<String> response =
-                    httpClient.send(evcsHttpRequest, HttpResponse.BodyHandlers.ofString());
+            var response =
+                    Retry.runTaskWithBackoff(
+                            sleeper,
+                            MAX_RETRIES,
+                            RETRY_DELAY_MILLIS,
+                            new RetryableTask<HttpResponse<String>>() {
```
Joe-Edwards-GDS: This is a 'functional interface', which means you can replace with a lambda rather than an anonymous class, i.e.
```
Retry.runTaskWithBackoff(
    sleeper,
    MAX_RETRIES,
    RETRY_DELAY_MILLIS,
    (isLastAttempt) -> {
        // implementation
    }
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -203,13 +213,40 @@ private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
...
+                                        var res =
+                                                httpClient.send(
+                                                        evcsHttpRequest,
+                                                        HttpResponse.BodyHandlers.ofString());
+                                        var statusCode = res.statusCode();
+                                        if (!isLastAttempt
+                                                && RETRYABLE_STATUS_CODES.contains(statusCode)) {
+                                            return Optional.empty();
+                                        }
+                                        return Optional.ofNullable(res);
```
Joe-Edwards-GDS: This should be `Optional.of(res)`, unless you think it might genuinely be null?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/retry/Retry.java
```diff
@@ -0,0 +1,33 @@
...
+            if (res.isPresent()) {
+                return res.get();
+            }
+            if (isLastAttempt) {
+                throw new RetryException("max retries reached for task: " + maxAttempts);
+            }
+            var backoff = (long) (waitInterval * Math.pow(2, count++));
+            sleeper.sleep(backoff);
+        }
+        throw new RetryException("max retries reached for task: " + maxAttempts);
```
Joe-Edwards-GDS: (I think conventionally exception messages are capitalised properly - i.e. `Max retries reached for task`)

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
```diff
@@ -203,13 +213,40 @@ private void checkResponseStatusCode(HttpResponse<String> evcsHttpResponse)
...
+                                    try {
+                                        var res =
+                                                httpClient.send(
+                                                        evcsHttpRequest,
+                                                        HttpResponse.BodyHandlers.ofString());
+                                        var statusCode = res.statusCode();
+                                        if (!isLastAttempt
+                                                && RETRYABLE_STATUS_CODES.contains(statusCode)) {
+                                            return Optional.empty();
+                                        }
```
Joe-Edwards-GDS: This feels a bit weird - we're missing any logging etc. in the case that we retry (but I think we probably want to know if it fails the first time).

Can we move the `checkResponseStatusCode` check inside the retryable task instead?

The fact all the exceptions have to be checked makes handling them cleanly a bit of a mess, although if there's only one type per use-case it is possible to make it a generic parameter:
```
public interface RetryableTask<T, E extends Throwable> {
    /*
     * Interface for a retryable task - the implementation should return
     * an Optional.empty() to indicate that the task should be retried and
     * throw a RetryException if any errors occur
     */
   Optional<T> run(boolean isLastAttempt) throws E, InterruptedException;
}
```

I'm also wondering if there's a model where we distinguish `Retryable` from `NonRetryable` exceptions? e.g. rather than returning an `Optional` you:
- Return a value for success
- Throw `RetryableException(e)` for an exception that may be retried
- Throw `NonRetryableException(e)` for an exception that may not be retried

That way you can put the logic around when to retry inside the `Retry` helper... We might have to do some dodgy exception throwing somewhere though. It is possible to trick the compiler (e.g. https://projectlombok.org/features/SneakyThrows) if we think it'll make it much more readable.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/retry/Retry.java
```diff
@@ -0,0 +1,36 @@
...
+        }
+        var count = 0;
+        while (count < maxAttempts) {
+            var isLastAttempt = (count + 1) >= maxAttempts;
+            var res = task.run(isLastAttempt);
+            if (res.isPresent()) {
+                return res.get();
+            }
+            if (isLastAttempt) {
+                throw new RetryException("max attempts reached for task: " + maxAttempts);
```
Joe-Edwards-GDS: This feels a bit counter-intuitive as it duplicates the throw below, which is actually unreachable

I'd probably move the wait inside an `if` instead, and probably just convert this to a `for` loop:
```
for (var attempt = 1; attempt <= maxAttempts; attempt++) {
    var res = task.run(attempt == maxAttempts);
    if (res.isPresent()) {
        return res.get();
    }
    if (attempt < maxAttempts) {
        var backoff = (long) (waitInterval * Math.pow(2, attempt - 1));
        sleeper.sleep(backoff);
    }
}
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/client/EvcsClient.java
sam803: thanks @Joe-Edwards-GDS  - I liked the different exceptions idea, have updated (also with a sneaky throw)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -58,32 +66,34 @@ public Optional<IpvSessionItem> getIpvSessionByAuthorizationCode(String authoriz
...
+                    LogHelper.buildLogMessage(
+                            "getIpvSessionByAccessToken() backoff and retry sleep was interrupted"));
+            Thread.currentThread().interrupt();
+        } catch (NonRetryableException e) {
+            LOGGER.warn(
+                    LogHelper.buildErrorMessage(
+                            "getIpvSessionByAccessToken() exception occurred retrying getItemByIndex",
+                            e));
         }
         return Optional.ofNullable(ipvSessionItem);
```
Joe-Edwards-GDS: Can we just return the retry result as a (non-optional) `IpvSession`?

And when catching the Non-Retryable exception, we need to do something similar to `EvcsClient` where we pass through a wrapped `UnknownAccessException` and wrap other errors appropriately.

Joe-Edwards-GDS (commented):
I think this is looking better - it's still a bit awkward, but I don't think we can do much better with Java exception constraints :(

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
sam803: Yes, I was just trying to keep the method signature the same to minimise changes

For the exception, I think we are only logging it so is there any need to handle the different types?

sam803: So do you mean to throw the error and let the callers handle it?

Joe-Edwards-GDS: Yeah - if you look at the calling code, it checks if it's empty and then throws its own `UnknownAccessTokenException` (which you can pull down into here)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -65,37 +65,39 @@ public Optional<IpvSessionItem> getIpvSessionByAuthorizationCode(String authoriz
...
+            if (e.getCause() instanceof UnknownAccessTokenException uatException) {
+                throw uatException;
+            }
             LOGGER.warn(
                     LogHelper.buildErrorMessage(
                             "getIpvSessionByAccessToken() exception occurred retrying getItemByIndex",
                             e));
         }
-        return Optional.ofNullable(ipvSessionItem);
+        return null;
```
Joe-Edwards-GDS: This should probably throw an exception as well (and perhaps be inside the catch)? Otherwise we'll just NPE somewhere down the line...

Joe-Edwards-GDS (commented):
Looks good except for one missing exception perhaps

Joe-Edwards-GDS (dismissed):
A bit awkward, but probably still better than reimplementing each time?

Might be worth checking that the build pipeline is clear before merging

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2130, opened 2024-07-05
PYIC-6738: Use vtr to determine ci threshold

MikeCollingwood:
### What changed

- Updated `CimitUtilityService`
  - New `isScoreBreachingThreshold` to commonise logic checking threshold
    - This now incorporates vtr, to reference config at `self/ciScoringThreshold/%s`
  - New `checkCiLevel` method to wrap the commonly-used logic:
    - Threshold check followed by getMitigationJourney...
    - Added in unit tests
- Pass through `vtr` from `clientOauthSessionItem`'s

### Why did it change

- To enable vot-specific ci threshold checking

### Issue tracking

- [PYIC-6738](https://govukverify.atlassian.net/browse/PYIC-6738)

### Notes

Also, need to wait for this config to go in first:
- https://github.com/govuk-one-login/ipv-core-common-infra/pull/997

After merging this core-back PR, remove redundant config code:
- https://github.com/govuk-one-login/ipv-core-common-infra/pull/998

Also, as a result of this work, an opportunity has become clear to implement a Vtr domain class to handle vtr-specific logic, here is the ticket written up for it: https://govukverify.atlassian.net/browse/PYIC-6984

[PYIC-6738]: https://govukverify.atlassian.net/browse/PYIC-6738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -169,16 +168,9 @@ private Map<String, Object> callTicfCri(IpvSessionItem ipvSessionItem, ProcessRe
                         clientOAuthSessionItem.getGovukSigninJourneyId(),
                         request.getIpAddress());
 
-        if (ciMitUtilityService.isBreachingCiThreshold(cis)) {
-            LOGGER.info(
-                    LogHelper.buildLogMessage(
-                            "CI score is breaching threshold - setting VOT to P0"));
-            ipvSessionItem.setVot(Vot.P0);
```
MikeCollingwood, line 181: Is removing this OK?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
```diff
@@ -8,7 +8,7 @@ public enum ConfigurationVariable {
     CIMIT_CONFIG("cimit/config"),
     CIMIT_SIGNING_KEY("cimit/signingKey"),
     CI_CONFIG("self/ci-config"),
-    CI_SCORING_THRESHOLD("self/ciScoringThreshold"),
+    CI_SCORING_THRESHOLD("self/ciScoringThreshold/%s"),
```
MikeCollingwood: Update this to be ciScoringThresholdByVot

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
DanCorderIPV: I'm not sure it is safe, so I've re-instated it

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
```diff
@@ -41,9 +45,30 @@ public List<String> getScopeClaims() {
...
 
-    public List<Vot> getRequestedVotsByStrength() {
+    public List<Vot> getRequestedVotsByStrengthDescending() {
         return Vot.SUPPORTED_VOTS_BY_DESCENDING_STRENGTH.stream()
                 .filter(vot -> vtr.contains(vot.name()))
                 .toList();
     }
+
+    // Refactor this out in PYIC-6984
+    public Vot getLowestStrengthRequestedGpg45Vot(ConfigService configService) {
```
blakeyp: @shivanshuit914 probably want to use this method in #2154 ?

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -116,17 +116,23 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             var vcs = sessionCredentialsService.getCredentials(ipvSessionId, userId);
 
-            UserIdentity userIdentity =
+            // PYIC-6901 there's probably a better way of getting the target Vot rather than
+            // re-calculating it here. Can we store it in the session in CheckExistingIdentity?
+            // We only actually need this to calculate CI breaches for return codes and the audit
+            // message
+            var targetVot = clientOAuthSessionItem.getLowestStrengthRequestedVot(configService);
```
DanCorderIPV: As the comment says here I'm not very happy with this. Is it worth fixing as part of this ticket or leave it for later?

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -702,12 +707,18 @@ void shouldNoMatchStrongestVotAndAlsoVCsDoNotCorrelate() throws Exception {
     @ParameterizedTest
     @MethodSource("votAndVtrCombinationsThatShouldStartIpvJourney")
     void shouldReturnJourneyIpvGpg45MediumResponseIfNoProfileAttainsVot(
-            Map<String, Object> votAndVtr) {
+            List<String> vtr, Optional<Vot> vot) throws Exception {
         when(ipvSessionService.getIpvSession(TEST_SESSION_ID)).thenReturn(ipvSessionItem);
         when(criResponseService.getFaceToFaceRequest(TEST_USER_ID)).thenReturn(null);
+
+        List<VerifiableCredential> credentials = new ArrayList<>();
```
DanCorderIPV, line 774: This section seems to have been accidentally deleted in the move to the VC class

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -464,9 +470,9 @@ private boolean isBirthDateEmpty(List<BirthDate> birthDates) {
                 || birthDates.stream().map(BirthDate::getValue).allMatch(StringUtils::isEmpty);
     }
 
-    private List<ReturnCode> getFailReturnCode(ContraIndicators contraIndicators)
+    private List<ReturnCode> getFailReturnCode(ContraIndicators contraIndicators, Vot vot)
```
MikeCollingwood: Could Vot be renamed here so it can be made sense of wrt return codes? "requestedVot" maybe?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -303,8 +309,8 @@ private void buildUserIdentityBasedOnProfileType(
                 successfulVcs.add(vc);
             }
         }
-        if (Vot.P0.equals(vot)) {
-            userIdentityBuilder.returnCode(getFailReturnCode(contraIndicators));
+        if (Vot.P0.equals(achievedVot)) {
```
MikeCollingwood, line 286: Is it valid that we say we have a target vot for the user and anything else should be considered a failure? If so:
- Could we replace the achievedVot with the boolean isTargetVotAchieved? I think it makes the function of the code more obvious
- We could reduce other methods such as `getStrongestAttainedVotForVtr`

(question not a suggestion)

Though one suggestion is achieved is spelt acheived in a couple of places (not here, but above e.g.)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -233,10 +229,15 @@ public JourneyResponse checkVcResponse(
                             clientOAuthSessionItem.getGovukSigninJourneyId(),
                             callbackRequest.getIpAddress());
 
-            if (ciMitUtilityService.isBreachingCiThreshold(cis)) {
-                return ciMitUtilityService
-                        .getCiMitigationJourneyResponse(cis)
-                        .orElse(JOURNEY_FAIL_WITH_CI);
+            // Check CI levels against the lowest confidence identity requested so we don't send the
```
MikeCollingwood:
```suggestion
            // Check CI levels for the lowest confidence identity requested so we don't send the
```
Might clarify the comment slightly

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -265,24 +261,21 @@ private JourneyResponse getJourneyResponse(
...
-            var preferredNewIdentityLevel = Vot.P2;
-            if (configService.enabled(P1_JOURNEYS_ENABLED)
-                    && clientOAuthSessionItem.getVtr().contains(Vot.P1.name())) {
-                preferredNewIdentityLevel = Vot.P1;
-            }
+            // If we want to prove or mitigate CIs for an identity we want to go for the lowest
+            // strength that is acceptable to the caller. We can only prove/mitigate GPG45
+            // identities
+            var lowestGpg45ConfidenceRequested =
+                    clientOAuthSessionItem.getLowestStrengthRequestedGpg45Vot(configService);
```
MikeCollingwood: We should probably add the result to the session and initialise in InitialiseIpvSessionHandler.

1. Because the we don't want to have to re-calculate it in BuildUserIdentityHandler
2. Since there wants to be a targetVot default for cases where vtr not set

lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
MikeCollingwood: I think this method isn't the same since it isn't filtering by gpg45 profile types. I don't think two session-stored attributes would be appropriate: `targetVot` & `targetGpg45Vot`, but think maybe we only need `targetGpg45Vot`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
```diff
@@ -41,9 +45,46 @@ public List<String> getScopeClaims() {
...
+        var requestedGpg45VotsByStrengthDescending =
+                getRequestedVotsByStrengthDescending().stream()
+                        .filter(vot -> vot.getProfileType() == ProfileType.GPG45)
+                        .toList();
+
+        var lowestStrengthRequestedGpg45Vot =
+                requestedGpg45VotsByStrengthDescending.get(
+                        requestedGpg45VotsByStrengthDescending.size() - 1);
```
MikeCollingwood: Do we want to check these indexes exist, or are we relying on a runtime exception? Maybe a catch block would be best:
```java
catch (ArrayIndexOutOfBoundsException e) {
```

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -325,7 +323,7 @@ private JourneyResponse getJourneyResponse(
                             auditEventUser,
                             deviceInformation,
                             contraIndicators)
-                    : buildNoMatchResponse(contraIndicators, preferredNewIdentityLevel);
+                    : buildNoMatchResponse(contraIndicators, lowestGpg45ConfidenceRequested);
```
MikeCollingwood: Could buildNoMatchResponse have the parameter defined be renamed to lowestGpg45ConfidenceRequested

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
DanCorderIPV: I believe we want to leave open the option of someone requesting a P1 but actually getting a P2 and treating that as success.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
DanCorderIPV: Change to `targetVot` for consistency

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: We now use the target vot in `BuildUserIdentityHandler`
The vtr currently isn't set for reverification jounrneys, so we do now set target vot to 'P2' for those journeys until they are fixed.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ClientOAuthSessionItem.java
DanCorderIPV: This shouldn't ever be a problem and as you say if it does cause an issue we will know about it from the out of bounds exception that gets thrown. We could add more code here, but I think all it would get us is a nicer error message in the logs (which is often worthwhile for errors that we think might happen).

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: In `buildNoMatchResponse` the parameter is called `preferredNewIdentityLevel` which seems to match what it represents in that function quite well?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/ClientOAuthSessionDetailsServiceTest.java
```diff
@@ -123,60 +123,6 @@ void shouldCreateClientOAuthSessionItem() throws ParseException {
                 clientOAuthSessionItemArgumentCaptor.getValue().getVtr());
     }
 
-    @Test
```
DanCorderIPV, line 124: This is a test for temporary code deleted above

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -295,6 +291,11 @@ private JourneyResponse getJourneyResponse(
                 return JOURNEY_REPROVE_IDENTITY_GPG45_MEDIUM;
             }
 
+            // PYIC-6901 Confirm that we only want to compare CI scores against the lowest requested
+            // VOT
```
blakeyp: Is this comment still needed? we said new/custom mitigation journeys are out of scope for MVP

DanCorderIPV: Maybe we should get rid of the comment, but this needs to be changed too I think. On 6901 you said that if a user needs to mitigate a P2 but in the meantime comes back to us for a P1 we should continue the P2 mitigation, but I don't think that will happen if they are breaching a P2 but not a P1.

DanCorderIPV: I'm not sure how we would know what mitigation they are in the middle of

blakeyp: That sounds reasonable for now, I'd say any special handling of it beyond what we already have is out of scope for MVP - but good to keep a reference I guess if we want to come back to this in future

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
```diff
@@ -175,16 +172,24 @@ private Map<String, Object> callTicfCri(IpvSessionItem ipvSessionItem, ProcessRe
...
-        if (ciMitUtilityService.isBreachingCiThreshold(cis)) {
+        // We need to know what vot to check for mitigations against.
+        // If the user has achieved a profile we should use that, if they haven't then the session
+        // Vot will still be P0
+        // and we should use the target Vot when looking for mitigations.
+        var achievedVot = ipvSessionItem.getVot();
+        var targetVot = ipvSessionItem.getTargetVot();
+        if (achievedVot != Vot.P0) {
+            targetVot = achievedVot;
+        }
```
Joe-Edwards-GDS: Might be clearer as a ternary...
```
var ciVot = ipvSessionItem.getVot() == Vot.P0
    ? ipvSessionItem.getTargetVot()
    : ipvSessionItem.getVot()
```
And if we do similar logic elsewhere, it might be worth pulling out a helper somewhere?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -848,8 +843,11 @@ private boolean achievedWithGpg45Profile(
                                 gpg45Scores, requestedVot.getSupportedGpg45Profiles())
                         : Optional.empty();
 
+        var isBreaching =
+                ciMitUtilityService.isBreachingCiThreshold(contraIndicators, requestedVot);
+
         // Successful match
-        if (matchedGpg45Profile.isPresent()) {
+        if (matchedGpg45Profile.isPresent() && !isBreaching) {
```
Joe-Edwards-GDS, line 849: (note that by pulling this out, it won't be able to short-circuit)

blakeyp (dismissed):

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -132,21 +145,32 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
                 return JOURNEY_VCS_NOT_CORRELATED.toObjectMap();
             }
 
-            boolean hasMatchingGpg45Profile =
-                    hasMatchingGpg45Profile(
+            var contraIndicators =
+                    ciMitService.getContraIndicators(userId, govukSigninJourneyId, ipAddress);
```
Joe-Edwards-GDS: Is it safe to skip this if there is only a single VOT in the VTR*? - This can add a good 500-1000ms to this execution, for a check that we don't expect to be used for MVP.

In future we may want to think about other more complex approaches - e.g. caching the CI scores in some way - but at the moment it feels ok to optimise for the single-VTR case.

(*or more precisely, if there is only a single CI threshold for all the VOTs in the VTR)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
Joe-Edwards-GDS: (I think it's not so much about the mitigation journey specifically, as much as blocking the user with a CI fail)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -235,10 +231,17 @@ public JourneyResponse checkVcResponse(
...
+            // user on an unnecessary mitigation journey.
+            var lowestConfidenceRequested =
+                    clientOAuthSessionItem
+                            .getParsedVtr()
+                            .getLowestStrengthRequestedGpg45Vot(configService);
```
Joe-Edwards-GDS: Is this just `ipvSession.getTargetVot()`?

Is there a reason for the inconsistency - could we either always use a value on the session or some logic to compute it from the VTR when needed (since it's a cheap calculation).

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
Joe-Edwards-GDS: Yeah, if possible we'd like to be able to handle a VTR of `[P1, P2]` correctly, even if it's not going to be part of the MVP.

In that scenario, it could be possible to have `targetVot = P1` but `achievedVot = P2`. However the inverse should _not_ be possible - i.e. we should never see `targetVot = P2` and `achievedVot = P1` (you should drop to `P0` if you don't achieve the target).

Joe-Edwards-GDS (commented):
This looks fine, but two main questions:

1. Would we be better off with a helper method that returns the 'minimum CI threshold' or something.

Following logic like:
- If vot > P0, use the threshold for that vot
- If reverification, use P2 threshold
- Otherwise use the minimum threshold for VOTs in the VTR

2. Can we avoid the extra call to CIMIT? Or is it fundamentally a new requirement

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
DanCorderIPV: I spotted this one and just fixed it :)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: It seemed clearer this was and the short circuit probably won't save a lot of time?

Joe-Edwards-GDS: It's likely to be negligible yes, although not necessarily as trivial as it sounds. The calculation does make several cacheable requests to parameter store, including some JSON deserialisation.

(Fine to leave)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
DanCorderIPV: At the moment yes, I think it would be safe to skip here because we'd have failed scoring earlier in the journey. But is that something we want to rely on remaining true in the future, and would we notice if it broke?

lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/CallTicfCriHandler.java
DanCorderIPV: Commonised in IpvSessionItem

Joe-Edwards-GDS (approved):

PR #2131, opened 2024-07-05
PYIC-6855 Change made to update DataStore to allow delete multiple items in a batch request.

rpayal:
## Proposed changes
Allow deletion of multiple items in a batch request to dynamoDB.

### What changed
Change made to DataStore to allow delete multiple items in a batch request.

### Issue tracking

- [PYIC-6855](https://govukverify.atlassian.net/browse/PYIC-6855)

### Other considerations

[PYIC-6855]: https://govukverify.atlassian.net/browse/PYIC-6855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
@@ -157,10 +157,12 @@ public class DataStore<T extends DynamodbItem> {
         return table.deleteItem(key);
     }
 
+    @ExcludeFromGeneratedCoverageReport
     public void deleteAllByPartition(String partitionValue) {
         delete(getItems(partitionValue));
     }
 
+    @ExcludeFromGeneratedCoverageReport
     public int delete(List<T> items) {
         if (!items.isEmpty()) {
             BatchWriteResult batchWriteResult =
@@ -181,6 +183,7 @@ public class DataStore<T extends DynamodbItem> {
         return items.size();
     }
 
+    @ExcludeFromGeneratedCoverageReport
     private WriteBatch createWriteBatchForDeleteItems(List<T> items) {
         WriteBatch.Builder<T> builder =
                 WriteBatch.builder(this.typeParameterClass).mappedTableResource(this.table);
@@ -190,6 +193,7 @@ public class DataStore<T extends DynamodbItem> {
         return builder.build();
     }
 
+    @ExcludeFromGeneratedCoverageReport
     private BatchWriteResult processBatchWrite(WriteBatch writeBatch) {
         return getClient()
                 .batchWriteItem(
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
@@ -9,9 +9,12 @@ import software.amazon.awssdk.enhanced.dynamodb.DynamoDbTable;
 import software.amazon.awssdk.enhanced.dynamodb.Expression;
 import software.amazon.awssdk.enhanced.dynamodb.Key;
 import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
+import software.amazon.awssdk.enhanced.dynamodb.model.BatchWriteItemEnhancedRequest;
+import software.amazon.awssdk.enhanced.dynamodb.model.BatchWriteResult;
 import software.amazon.awssdk.enhanced.dynamodb.model.PutItemEnhancedRequest;
 import software.amazon.awssdk.enhanced.dynamodb.model.QueryConditional;
 import software.amazon.awssdk.enhanced.dynamodb.model.QueryEnhancedRequest;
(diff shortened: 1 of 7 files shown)


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
```diff
@@ -147,7 +147,7 @@ public void deleteSessionCredentialsForCri(String ipvSessionId, Cri cri)
                     LogHelper.buildLogMessage(
                             String.format(
                                     "Deleted %d credentials for %s from session credentials table",
-                                    deleted.size(), criId)));
+                                    deleted, criId)));
```
Joe-Edwards-GDS: It feels a bit weird to return an `int` from `delete` just for this - you could separate `dataStore.getItemsBySortKeyPrefix(ipvSessionId, criId)` into a separate variable and use the size of that instead?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -154,14 +157,49 @@ public T delete(String partitionValue, String sortValue) {
...
+                    processBatchWrite(createWriteBatchForDeleteItems(items));
+            // 'unprocessedDeleteItemsForTable()' returns keys for delete requests that did not
+            // process.
+            List<Key> unprocessedItems =
+                    batchWriteResult.unprocessedDeleteItemsForTable(this.table);
+            if (!unprocessedItems.isEmpty()) {
+                unprocessedItems.forEach(
+                        key ->
+                                LOGGER.warn(
+                                        "Delete fail for item key: %s".formatted(key.toString())));
```
Joe-Edwards-GDS: Do you think this is a warning or an error? In the 'end of journey' case, a warning is fine, but I think for the in-session reset case, this is an error - attempting to continue might cause unexpected results.

I'd probably just throw an exception?

Joe-Edwards-GDS (commented):
Thanks - looks plausible to me, one question about the error handling

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java
rpayal: Earlier delete(List<T>) was returning List<T> of Items and then we were puting in log how many items got deleted.
That is why now with batchWrite delete we can't return List of items deleted so returning back how many got deleted.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
rpayal: If we throw an exception (as needed by in-session reset cse)) here then I suppose this delete() method don't need to return int.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -172,15 +173,12 @@ public int delete(List<T> items) {
             List<Key> unprocessedItems =
                     batchWriteResult.unprocessedDeleteItemsForTable(this.table);
             if (!unprocessedItems.isEmpty()) {
-                unprocessedItems.forEach(
-                        key ->
-                                LOGGER.warn(
-                                        "Delete fail for item key: %s".formatted(key.toString())));
+                String errMessage = "Failed during batch deletion.";
+                LOGGER.error(errMessage);
```
Joe-Edwards-GDS: Can we use `LogHelper.buildLogMessage` here?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -172,15 +173,12 @@ public int delete(List<T> items) {
...
+                throw new BatchDeleteException(errMessage);
             }
-            return items.size() - unprocessedItems.size();
         }
         LOGGER.info("No items to delete");
```
Joe-Edwards-GDS: This might need to be in an `else` now? Also ditto with `LogHelper`.

Joe-Edwards-GDS (commented):
Thanks this looks good, just one small log issue

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
rpayal: done

Joe-Edwards-GDS (approved):

PR #2134, opened 2024-07-08
PYIC-6943: journey map - use a select for journey selection

sam803:
### What changed

Use a `<select>` in the journey map to allow users to select the journey

Screenshot 2024-07-09 at 10 02 52

Screenshot 2024-07-08 at 09 32 18

### Why did it change

Due to the increasing number of journeys, the current method is becoming unusable

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6943


journey-map/public/index.mjs
```diff
@@ -101,18 +101,16 @@ const switchJourney = async (targetJourney, targetState) => {
 };
 
 const setupHeader = () => {
```
Joe-Edwards-GDS, line 113: I wonder if we should still include some of the 'frequent' options in the header - i.e. `NEW_P2_IDENTITY`? I'd quite like to avoid extra clicks where possible...

Do you think it'd also be worth making the `h1` (IPV Core Journey Map) a link back to the homepage?

journey-map/public/style.css
```diff
@@ -189,3 +189,11 @@ form fieldset p {
     stroke-width: 3px;
     filter: drop-shadow(1px 0 0 #ffdd00) drop-shadow(-1px 0 0 #ffdd00) drop-shadow(0 1px 0 #ffdd00) drop-shadow(0 -1px 0 #ffdd00);
 }
+#header-actions {
+    position: fixed;
+    right: 30px;
+}
+#journey-select {
```
Joe-Edwards-GDS, line 201: I think this (and the search box) could have a slightly larger font size, e.g. `1rem`?

journey-map/public/index.html
```diff
@@ -14,7 +14,12 @@ <h1>
                     </svg>
                     <span>IPV Core Journey Map</span>
                 </h1>
-                <input id="state-search" type="search" placeholder="Search for state...">
+                <div id="header-actions">
+                    <h4>Current Journey
```
Joe-Edwards-GDS: I don't think this should be an `h4` - if it's a heading it's an `h2` (since it follows an `h1`), and otherwise it's a `label` for the select

journey-map/public/style.css
```diff
@@ -189,3 +189,11 @@ form fieldset p {
     stroke-width: 3px;
     filter: drop-shadow(1px 0 0 #ffdd00) drop-shadow(-1px 0 0 #ffdd00) drop-shadow(0 1px 0 #ffdd00) drop-shadow(0 -1px 0 #ffdd00);
 }
+#header-actions {
+    position: fixed;
+    right: 30px;
```
Joe-Edwards-GDS: You can do this more naturally with `margin-left: auto` - that way it'll flow correctly if we start putting other items in the header.

journey-map/public/style.css
```diff
@@ -189,3 +189,11 @@ form fieldset p {
...
+}
+#journey-select {
+    margin-right: 30px;
+    margin-left: 3px;
+}
```
Joe-Edwards-GDS, line 208: At the moment the select and search input are slightly different heights... It's not super important, but it'd be nice to make them the same..! (Few options for doing this, easiest might be to make `header-actions` `display: flex` and add `height: 100%`)

Joe-Edwards-GDS (commented):
Does what it says on the tin, but I have a few suggestions to try and help elevate :)

journey-map/public/style.css
sam803: Yes looks better

journey-map/public/index.mjs
```diff
@@ -23,6 +23,10 @@ const JOURNEY_TYPES = {
     REVERIFICATION: 'Reverification',
 };
 
+const COMMON_JOURNEY_TYPS = [
```
Joe-Edwards-GDS: (small typo here)

journey-map/public/index.mjs
```diff
@@ -102,18 +108,27 @@ const switchJourney = async (targetJourney, targetState) => {
...
-    // Add header entries for each journey
+    // Add option to journey select for each journey
     Object.entries(JOURNEY_TYPES).forEach(([id, label]) => {
-        const link = document.createElement('a');
-        link.href = getJourneyUrl(id);
-        link.innerText = label;
-        link.onclick = async (e) => {
-            e.preventDefault();
-            await switchJourney(id, null);
+        if (COMMON_JOURNEY_TYPS.indexOf(id) != -1) {
```
Joe-Edwards-GDS: (it'd probably be easier to iterate over `COMMON_JOURNEY_TYPES` and then look up the label in `JOURNEY_TYPES`)

Joe-Edwards-GDS (dismissed):
Looks good, with a couple of minor nits - sorry I've given you another conflict :(

Joe-Edwards-GDS (approved):

PR #2135, opened 2024-07-08
BAU: fix tests reliant on static age

sam803:
## Proposed changes
Fix tests which are reliant on static age value


Joe-Edwards-GDS (approved):
What an absolute clanger - not sure how it was missed the the first time round, but thanks for fixing.

PR #2139, opened 2024-07-08
PYIC-6990 Local DataStore

Joe-Edwards-GDS:
### What changed

Extracted an interface for `DataStore` and added a new `InMemoryDataStore` implementation alongside the existing one, renamed as `DynamoDataStore`.

### Why did it change

To run our code locally, we would like to eliminate AWS dependencies. This won't be a perfect representation of behaviour, but should generally be 'good enough', and will have a much quicker dev-cycle than deploying an environment into AWS.

### Issue tracking

- [PYIC-6990](https://govukverify.atlassian.net/browse/PYIC-6990)

[PYIC-6990]: https://govukverify.atlassian.net/browse/PYIC-6990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS, line 1: One important consideration - at the moment we are not serialising/deserialising the records (we're just storing the Java objects as-is).

Since our objects are mutable, this means it's possible for code to 'update' the records by mutating them (without explicitly saving the result), which would break with a real store.

The alternatives are probably:
- Serialise/deserialise via some internal DynamoDB method (if one exists)
- Serialise/deserialise via JSON - relying on this being roughly equivalent to the DynamoDB mapping
- Make our record classes immutable, and create new ones when updating the records

Joe-Edwards-GDS: For the last one (assuming it doesn't break DynamoDB) Lombok provides useful annotations for this purpose:
- `@Value` turns it into a read-only (immutable) class
- `@With` adds 'withXX' methods to create a new instance with a modified field

So e.g. when updating the ipvSession we can do something like:
```
updateIpvSession(
  ipvSessionItem
    .withAccessToken(accessToken)
    .withAccessTokenMetadata(accessTokenMetadata)
);
```

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public List<T> getItems(String partitionValue) {
+        return records.values().stream()
+                .filter(i -> getPartitionKey(i).equals(partitionValue))
+                .toList();
+    }
+
+    @SuppressWarnings(
+            "java:S3011") // We rely on field access to avoid needing to search for a getter
```
Joe-Edwards-GDS, line 121: Otherwise we have to take the field name and turn it into a 'plausible' property accessor - e.g. `foo` -> `isFoo` or `getFoo`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS, line 1: We could also use a local dynamodb docker container, rather than this...
- More realistic DynamoDB implementation
- Requires more orchestration to run, including stuff like creation of the relevant tables (can't just 'press play')

local-running/runLocalStack.sh
```diff
@@ -62,4 +62,4 @@ fi
 
 export ENVIRONMENT=$env
 export DEV_ACCOUNT_NUM=$dev_no
-aws-vault exec $profile -- docker compose up $attach
```
Wynndow: Not sure we want to build everything ever time we start it up

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Wynndow: How does option 3 help with the lack of serialisation? We'll still be missing that step?

FWIW, and we talked about this offline, I think that lack of serialisation is probably fine, as there are a bunch of other tests and environments to get through before getting anywhere near prod.

Wynndow: I think this might be overkill. And would be annoying to manage.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class InMemoryDataStore<T extends PersistenceItem> implements DataStore<T> {
+    private static final String KEY_SEPARATOR = "/";
+    private static final ConcurrentMap<String, Map<String, ?>> TABLES = new ConcurrentHashMap<>();
+
+    private final ConcurrentMap<String, T> records;
```
Wynndow, line 23: Why ConcurrentMap? Is there a particular need for thread safety here?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+    private final Class<T> klass;
+    private final Method partitionKeyMethod;
+    private final Method sortKeyMethod;
+    private final boolean hasSortKey;
+
+    @SuppressWarnings("unchecked") // Rely on callers to use consistent types per table
+    public InMemoryDataStore(String tableName, Class<T> klass) {
+        Method partitionKey = null;
+        Method sortKey = null;
```
Wynndow: Do you need these intermediate vars? Can you just assign to the fields in the for loop?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public T getItem(String partitionValue, boolean warnOnNull) {
+        return getItem(partitionValue);
+    }
+
+    @Override
+    public T getItemByIndex(String indexName, String value) {
+        Method indexMethod = null;
+        for (var method : klass.getMethods()) {
```
Wynndow, line 89: Part of me is wondering if it's worth checking for secondary index's in the constructor, and creating an index map for them if they do. Then keeping them up to date when writing. But we're probaby never going to have "loads" of items when running locally so it might be overkill.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @SuppressWarnings(
+            "java:S3011") // We rely on field access to avoid needing to search for a getter
+    @Override
+    public List<T> getItemsWithBooleanAttribute(String partitionValue, String name, boolean value) {
+        try {
+            var field = klass.getDeclaredField(name);
+            field.setAccessible(true);
+            return records.values().stream()
```
Wynndow, line 127: This doesn't seem to be using the partition value - we'd want to filter by that first I think. Otherwise you'll end up with a bunch of items you're not interested in?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    @Override
+    public List<T> getItemsBySortKeyPrefix(String partitionValue, String sortPrefix) {
+        return records.values().stream()
+                .filter(
+                        i ->
+                                getPartitionKey(i).equals(partitionValue)
+                                        && getSortKey(i).startsWith(sortPrefix))
```
Wynndow, line 154: `getSortKey()` can return null, which will blow this up? Hopefully we wouldn't use this method on a table that doesn't have a sort key, but might be worth addressing.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: Option 3 makes the records immutable, so avoids the mutability problem entirely - you'd have to update the store with a new (updated) copy of the record to change anything.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: This is effectively the dynamodb 'table', which is stored statically in `TABLES` (but stored as an instance variable as well, for convenience) - this might be accessed by multiple requests in parallel, so I think we need thread safety.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Wynndow: Gotcha

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Wynndow: OK, because we're departing from running in the lambda runtime. If we were to run the tests suites against local we could end up with concurrent requests.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: If you don't use the intermediate vars then you'd need to initialise the instance variables to null and make them non-final, which feels a bit unsatisfactory.

We could extract the method search to avoid it though...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: Yeah... I think it's probably not worth it, unless we need this to be super performant. Reflection is 'slow', but not _that_ slow (it'll be orders of magnitude faster than an actual call to dynamodb!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: Good spot!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStore.java
Joe-Edwards-GDS: Yeah, it almost feels like a good thing to blow up if we try to use a sort key on a table without one! But we could throw a more useful exception I suppose.

local-running/runLocalStack.sh
Joe-Edwards-GDS: Are you sure? Without this I wasn't sure how to get my docker container to reflect the latest code :/

It still uses build caching - so e.g. it won't rebuild the stubs (unless they've changed).

Wynndow: I've been building the things I'm interested in with `docker-compose build <service>` before running this script. There are perhaps better ways to do this, but I'm not sure building everything ever time is right. Even with the build caching it's going to do a bunch of stuff for ~14 services.

Joe-Edwards-GDS: I see... It seems to be pretty fast when using all cached builds, but will add a gradle compilation step if they change. I'd be a bit more worried about inadvertently running an old stub and having to debug issues tbh.

I can revert, given that you use it more than me... Hopefully once local running works properly we won't need a massive docker-compose anyway :)

Wynndow: I think always building stuff won't avoid running an old stub - as if you don't pull the latest from main you'll still have an old version locally

Wynndow (approved):
Nice!

PR #2148, opened 2024-07-09
PYIC-6943: fix title home link when running with a url path

sam803:
## Proposed changes

This PR https://github.com/govuk-one-login/ipv-core-back/pull/2134 introduced a link to the title that should take the user to the initial home page, but this does not work on the github page since the url include a path

### What changed

Dynamically set the url for the title link based on the window.location.href

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6943


journey-map/public/index.mjs
```diff
@@ -67,6 +67,8 @@ const stateSearch = document.getElementById('state-search');
 const journeyDesc = document.getElementById('journeyDesc');
 const journeyName = document.getElementById('journeyName');
 
+document.getElementById('header-title').href = window.location.href.split('?')[0];
```
Joe-Edwards-GDS: I'd prefer this to match the mechanism used in our other links - we can leave the href fixed as `/`, but override the click behaviour.

Perhaps something like:
```
link.onclick = async (e) => {
  e.preventDefault();
  await switchJourney(DEFAULT_JOURNEY_TYPE);
};
```

I've also been trying to put all the selectors in the block above, and any 'initialisation logic' in the `initialize` method at the bottom - that makes it easier to keep track of all the bits and pieces.

Joe-Edwards-GDS (commented):
Thanks for spotting!

journey-map/public/index.mjs
sam803: @Joe-Edwards-GDS thanks, done :)

Joe-Edwards-GDS (approved):
Thanks!

PR #2149, opened 2024-07-09
PYIC-5759: update vc helper to use parsed VCs

thebauSoftwire:
### What changed
Updates `VcHelper` to use parsed VCs + updates unit tests.

### Why did it change

We want to use the classes provided by vocab instead of the current json handling.

### Issue tracking

### DIFF
--- a/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
+++ b/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
@@ -8,6 +8,7 @@ import uk.gov.di.ipv.core.library.enums.Vot;
 import uk.gov.di.ipv.core.library.exceptions.UnrecognisedVotException;
 import uk.gov.di.ipv.core.library.gpg45.validators.Gpg45IdentityCheckValidator;
 import uk.gov.di.ipv.core.library.service.ConfigService;
+import uk.gov.di.model.BirthDate;
 import uk.gov.di.model.DrivingPermitDetails;
 import uk.gov.di.model.IdentityCheck;
 import uk.gov.di.model.IdentityCheckCredential;
@@ -108,9 +109,10 @@ public class VcHelper {
 
     public static Integer extractAgeFromCredential(VerifiableCredential vc) {
         if (vc.getCredential().getCredentialSubject() instanceof PersonWithIdentity person) {
-            var birthDates = person.getBirthDate();
+            var birthDate =
+                    extractFromFirstElementOfList(person.getBirthDate(), BirthDate::getValue);
 
-            return isNullOrEmpty(birthDates) ? null : getAge(birthDates.get(FIRST).getValue());
+            return birthDate == null ? null : getAge(birthDate);
         }
         return null;
     }
--- a/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
+++ b/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
@@ -9,11 +9,13 @@ import uk.gov.di.ipv.core.library.exceptions.UnrecognisedVotException;
 import uk.gov.di.ipv.core.library.gpg45.validators.Gpg45IdentityCheckValidator;
 import uk.gov.di.ipv.core.library.service.ConfigService;
 import uk.gov.di.model.DrivingPermitDetails;
+import uk.gov.di.model.IdentityCheck;
 import uk.gov.di.model.IdentityCheckCredential;
 import uk.gov.di.model.PassportDetails;
 import uk.gov.di.model.PersonWithDocuments;
 import uk.gov.di.model.PersonWithIdentity;
 import uk.gov.di.model.ResidencePermitDetails;
+import uk.gov.di.model.RiskAssessment;
 import uk.gov.di.model.RiskAssessmentCredential;
 
 import java.text.ParseException;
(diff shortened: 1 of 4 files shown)


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -77,54 +70,80 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
+            if (vc.getCredential() instanceof IdentityCheckCredential identityCheckCredential) {
+                var identityChecks = identityCheckCredential.getEvidence();
+
+                if (!isNullOrEmpty(identityChecks)) {
+                    txnIds.add(identityChecks.get(ONLY).getTxn());
+                }
+                continue;
+            }
+
+            if (vc.getCredential() instanceof RiskAssessmentCredential riskAssessmentCredential) {
```
Joe-Edwards-GDS: (In practice this will never happen, given where this method is called, but doesn't hurt to leave it in)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -77,54 +70,80 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
+                var riskAssessments = riskAssessmentCredential.getEvidence();
+
+                if (!isNullOrEmpty(riskAssessments)) {
+                    txnIds.add(riskAssessments.get(ONLY).getTxn());
+                }
             }
         }
         return txnIds;
```
Joe-Edwards-GDS: This is nice and readable, but if you wanted a 'stream'-y approach, you could do something like:
```
return vcs.stream()
        .map(vc -> vc.getCredential())
        .filter(credential -> credential instanceof IdentityCheckCredential)
        .flatMap(credential -> ((IdentityCheckCredential) credential).getEvidence().stream())
        .map(evidence -> evidence.getTxn())
        .toList();
```
...except it'd have to be a bit less beautiful, because we need to add in some null checks and also handle risk assessments. Would be easier in Typescript 😢

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -77,54 +70,80 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
-                            .valueToTree(vc.getClaimsSet().getClaim(VC_CLAIM))
-                            .path(VC_EVIDENCE);
-            if (evidenceArray.isArray()
-                    && !evidenceArray.isEmpty()) { // not all VCs have an evidence block
-                txnIds.add(evidenceArray.get(ONLY).path(VC_EVIDENCE_TXN).asText());
+            if (vc.getCredential() instanceof IdentityCheckCredential identityCheckCredential) {
+                var identityChecks = identityCheckCredential.getEvidence();
+
+                if (!isNullOrEmpty(identityChecks)) {
+                    txnIds.add(identityChecks.get(ONLY).getTxn());
```
Joe-Edwards-GDS: This feels a bit weird - I know it's only matching existing behaviour, but feels like it should probably be `identityChecks.forEach(ic -> txnIds.add(ic.getTxn()));` (or an equivalent with `addAll`)

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -77,54 +70,80 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
-        var birthDateArr =
-                OBJECT_MAPPER
-                        .valueToTree(vc.getClaimsSet().getClaim(VC_CLAIM))
-                        .path(VC_CREDENTIAL_SUBJECT)
-                        .path(VC_BIRTH_DATE);
-        if (birthDateArr.isMissingNode() || birthDateArr.isEmpty()) {
+        if (vc.getCredential().getCredentialSubject() instanceof PersonWithIdentity person) {
+            var birthDates = person.getBirthDate();
+
+            return isNullOrEmpty(birthDates) ? null : getAge(birthDates.get(ONLY).getValue());
```
Joe-Edwards-GDS: Again, not your code, but can we rename `ONLY` to `FIRST` perhaps? We're not actually checking that there's a single element

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -77,54 +70,80 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
+        }
+        return null;
+    }
+
+    private static String getIcaoIssuerFromPassport(List<PassportDetails> passports) {
+        if (isNullOrEmpty(passports)) {
+            return null;
+        }
+        return passports.get(ONLY).getIcaoIssuerCode();
+    }
```
Joe-Edwards-GDS: I wonder if you can extract a helper for this sort of thing, e.g.:
```
private static <T, R> R extractFromList(List<T> list, Function<T, R> extractor) {
    return isNullOrEmpty(list) ? null : extractor.apply(list.get(0));
}
```
```
var icaoIssuerCode = extractFromList(
        person.getResidencePermit(),
        ResidencePermitDetails::getIcaoIssuerCode);
```

Joe-Edwards-GDS (commented):
Looks good - a few little optional suggestions.

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -75,56 +73,74 @@ public static List<VerifiableCredential> filterVCBasedOnProfileType(
...
-                OBJECT_MAPPER
-                        .valueToTree(vc.getClaimsSet().getClaim(VC_CLAIM))
-                        .path(VC_CREDENTIAL_SUBJECT)
-                        .path(VC_BIRTH_DATE);
-        if (birthDateArr.isMissingNode() || birthDateArr.isEmpty()) {
-            return null;
+        if (vc.getCredential().getCredentialSubject() instanceof PersonWithIdentity person) {
+            var birthDates = person.getBirthDate();
+
+            return isNullOrEmpty(birthDates) ? null : getAge(birthDates.get(FIRST).getValue());
```
Joe-Edwards-GDS: Can you use your new method here?

thebauSoftwire: good spot!

Joe-Edwards-GDS (approved):

PR #2153, opened 2024-07-10
PYIC-6816: use cri enum in config service

sam803:
## Proposed changes
Change config service methods that take a string cri id as a parameter to use the Cri enum instead

### What changed

 - Changes to config service methods
 - Changes to any code/tests that use the config service methods
 - Changes to BuildCriOauthRequestHandler.java where string cri is passed

### Issue tracking

### DIFF
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -150,15 +150,14 @@ public class BuildCriOauthRequestHandler
             String ipAddress = getIpAddress(input);
             configService.setFeatureSet(getFeatureSet(input));
 
-            var criId = getCriIdFromJourney(input.getJourneyUri().getPath());
-            if (criId == null) {
+            var cri = getCriFromJourney(input.getJourneyUri().getPath());
+            if (cri == null) {
                 return new JourneyErrorResponse(
                                 JOURNEY_ERROR_PATH,
                                 SC_BAD_REQUEST,
-                                ErrorResponse.MISSING_CREDENTIAL_ISSUER_ID)
+                                ErrorResponse.INVALID_CREDENTIAL_ISSUER_ID)
                         .toObjectMap();
             }
-            var cri = Cri.fromId(criId);
             LogHelper.attachCriIdToLogs(cri);
 
             String criContext = getJourneyParameter(input, CONTEXT);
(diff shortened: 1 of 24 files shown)


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -157,14 +158,15 @@ public Map<String, Object> handleRequest(JourneyRequest input, Context context)
...
             String criContext = getJourneyParameter(input, CONTEXT);
             EvidenceRequest criEvidenceRequest =
                     EvidenceRequest.fromBase64(getJourneyParameter(input, EVIDENCE_REQUESTED));
-            String connection = configService.getActiveConnection(criId);
+            String connection = configService.getActiveConnection(cri);
             OauthCriConfig criConfig =
-                    configService.getOauthCriConfigForConnection(connection, criId);
+                    configService.getOauthCriConfigForConnection(connection, cri);
 
             if (criConfig == null) {
```
Joe-Edwards-GDS: You might need to pull this up (or replace it) - since it's effectively being superseded by`Cri.fromId`.

I wonder if it's worth pushing inside `getCriFromJourney`, and that can do all the relevant null checking/validity checking in one place.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -184,7 +184,8 @@ public void validateCallbackRequest(
                 || !state.equals(criOAuthSessionItem.getCriOAuthSessionId())) {
             throw new InvalidCriCallbackRequestException(ErrorResponse.INVALID_OAUTH_STATE);
         }
-        if (configService.getOauthCriActiveConnectionConfig(criId) == null) {
+        if (configService.getOauthCriActiveConnectionConfig(callbackRequest.getCredentialIssuer())
+                == null) {
             throw new InvalidCriCallbackRequestException(
                     ErrorResponse.INVALID_CREDENTIAL_ISSUER_ID);
         }
```
Joe-Edwards-GDS, line 192: Similarly - is this check still valid?

Joe-Edwards-GDS (commented):
Spotted a couple of cases where we're implicitly validating CRI ID, where previously we validated explicitly.

For user input I think it's a bit more important to have proper validation, since it's not necessarily an internal value we can rely.

(I may have missed some other cases...)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
sam803: I'm not sure if that would ever return null tbh - wouldn't we get a `NoConfigForConnectionException` if the config wasn't found for the CRI?

Joe-Edwards-GDS: probably! - even more reason to use the CRI enum as the validation mechanism :)

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
sam803: I did initially have it in there but wasn't able to distinguish between a `MISSING_CREDENTIAL_ISSUER_ID` and `INVALID_CREDENTIAL_ISSUER_ID` errors - can we just use `INVALID_CREDENTIAL_ISSUER_ID` for both?

Joe-Edwards-GDS: Sounds fine to me

Joe-Edwards-GDS (approved):

PR #2156, opened 2024-07-10
PYIC-6987: New audit events for f2f cancel

sam803:
## Proposed changes

Adds two new audit events for when a user decides to cancel their f2f visit

### What changed

 - IPV_F2F_USER_CANCEL_START fired when user clicks confirm to delete their details
 - IPV_F2F_USER_CANCEL_END fired after the details are deleted

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-6987


Joe-Edwards-GDS (approved):
Love how simple this is with the journey map events :D


...next task is to get these to show up in the journey map visualisation in some sensible way!

PR #2157, opened 2024-07-10
PYIC-7003 Suppress EVCS errors if reads are not enabled

Joe-Edwards-GDS:
### What changed

Suppress EVCS errors that occur while EVCS reads are not enabled

### Why did it change

Until EVCS reads are enabled, the EVCS records are not being used as a source of truth, so we can ignore failures and fall back to existing behaviour.

### Issue tracking

- [PYIC-7003](https://govukverify.atlassian.net/browse/PYIC-7003)

[PYIC-7003]: https://govukverify.atlassian.net/browse/PYIC-7003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -355,71 +355,74 @@ public class CheckExistingIdentityHandler
         var tacticalVcs = verifiableCredentialService.getVcs(userId);
 
         if (configService.enabled(EVCS_WRITE_ENABLED) || configService.enabled(EVCS_READ_ENABLED)) {
-            try {
-                var evcsVcs =
-                        evcsService.getVerifiableCredentialsByState(
-                                userId, evcsAccessToken, CURRENT, PENDING_RETURN);
-
-                // Use pending return vcs to determine identity if available
-                var evcsIdentityVcs = evcsVcs.get(PENDING_RETURN);
-                var isPendingEvcs = true;
-                var hasPartiallyMigratedVcs = false;
-                VerifiableCredentialBundle vcBundle = null;
-                if (isNullOrEmpty(evcsIdentityVcs)) {
-                    evcsIdentityVcs = evcsVcs.get(CURRENT);
-                    isPendingEvcs = false;
-                }
-                if (!isNullOrEmpty(evcsIdentityVcs)) {
-                    vcBundle =
-                            new VerifiableCredentialBundle(
-                                    configService.enabled(EVCS_READ_ENABLED)
-                                            ? evcsIdentityVcs
-                                            : tacticalVcs,
-                                    true,
-                                    isPendingEvcs);
-
-                    hasPartiallyMigratedVcs =
-                            hasPartiallyMigratedVcs(
-                                    tacticalVcs,
-                                    evcsIdentityVcs,
-                                    isPendingEvcs,
-                                    vcBundle.isF2fIdentity());
-
-                    if (hasPartiallyMigratedVcs) {
-                        // use tactical vcs but with evcs flags so that the store-identity lambda is
-                        // called next and updates the evcs pending one
-                        vcBundle = new VerifiableCredentialBundle(tacticalVcs, true, true);
-                    }
-                }
-                logIdentityMismatches(tacticalVcs, evcsVcs, hasPartiallyMigratedVcs);
-                // only use these evcs vcs if they exist and have been fully migrated
-                if (vcBundle != null) {
-                    return vcBundle;
-                }
-            } catch (EvcsServiceException e) {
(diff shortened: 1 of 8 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -355,63 +355,74 @@ private VerifiableCredentialBundle getVerifiableCredentials(
         var tacticalVcs = verifiableCredentialService.getVcs(userId);
 
         if (configService.enabled(EVCS_WRITE_ENABLED) || configService.enabled(EVCS_READ_ENABLED)) {
```
rpayal, line 357: As pointed earlier we don't need EVCS_WRITE_ENABLED check as only in case of EVCS_READ_ENABLED only evcs will be source of truth for VCs.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -355,63 +355,74 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+
+            if (!isNullOrEmpty(evcsIdentityVcs)) {
                 if (hasPartiallyMigratedVcs) {
                     // use tactical vcs but with evcs flags so that the store-identity lambda is
                     // called next and updates the evcs pending one
-                    vcBundle = new VerifiableCredentialBundle(tacticalVcs, true, true);
+                    return new VerifiableCredentialBundle(tacticalVcs, true, true);
+                } else {
+                    return new VerifiableCredentialBundle(
+                            configService.enabled(EVCS_READ_ENABLED)
```
rpayal, line 395: if EVCS_WRITE_ENABLED is removed then we don't need this check.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -355,63 +355,74 @@ private VerifiableCredentialBundle getVerifiableCredentials(
...
+                            true,
+                            isPendingEvcs);
                 }
             }
-            logIdentityMismatches(tacticalVcs, evcsVcs, hasPartiallyMigratedVcs);
-            // only use these evcs vcs if they exist and have been fully migrated
-            if (vcBundle != null) {
-                return vcBundle;
+        } catch (EvcsServiceException e) {
+            if (configService.enabled(EVCS_READ_ENABLED)) {
```
rpayal, line 403: same goes here.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Not quite - this check would still need to be `EVCS_READ_ENABLED` while the one above would be simplified to `EVCS_WRITE_ENABLED`.

I'm not sure it's that valuable to simplify the double condition though - if we can easily support both combinations in a rational way, why wouldn't we?

rpayal: We will only reach here for evcs VCs when EVCS_READ_ENABLED

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
rpayal: When only EVCS_WRITE_ENABLED then source of truth is tactical store....then no need to fetch VCs from EVCS.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: At the moment that's not the case, we're still reading from the tactical store for a couple of reasons:
- checking that we're correctly keeping them in sync (i.e. `logIdentityMismatches`)
- knowing whether to migrate the tactical store identity

We aren't using at as the source of truth though, which is what this check here is for (so that if EVCS reads are not enabled, we use tactical VCs regardless)

We could skip some of these checks, but we'd need to read from EVCS anyway to safely perform the migration, so it doesn't save us loads.

rpayal (approved):

PR #2165, opened 2024-07-12
Pyic 3904  vc success tidy

Joe-Edwards-GDS:
### What changed

Tidy up some VC success and address parsing code in BuildProvenUserIdentity

### Why did it change

We are using new methods that rely on the vocab classes instead, to save duplicated effort.


Wynndow (approved):
Looks good!

PR #2170, opened 2024-07-15
BAU: Fix reprove identity flag check

MikeCollingwood:
### What changed

- Change flag check from OR -> AND


blakeyp (changes requested):
Not sure what this is trying to fix (?) but there's a feature flag/set to force clearing a user's identity (used in some of the tests), which is distinct from the actual reprove identity account intervention, hence the OR.

PR #2171, opened 2024-07-15
PYIC-6959: Build test images

Wynndow:
### What changed

Build test images

### Why did it change

[PYIC-6959: Downgrade eslint](https://github.com/govuk-one-login/ipv-core-back/pull/2171/commits/d69bf456f8e5c1d17610162b2da43820a8461ec0)

Dependabot recently bumped eslint from 8 to 9, but this causes a
dependency resolution issue with typescript-eslint, which needs eslint
8.

[PYIC-6959: Build api-tests image](https://github.com/govuk-one-login/ipv-core-back/pull/2171/commits/c8b67d4d1f763c14968573479b073847781a0c81)

This adds Dockerfiles and a GHA workflow to build the api-tests images
for the pipeline.

It will only build the image if certain files have changed, to avoid
building it on every core-back merge, which would be unneccessary.

### Issue tracking

- [PYIC-6959](https://govukverify.atlassian.net/browse/PYIC-6959)

[PYIC-6959]: https://govukverify.atlassian.net/browse/PYIC-6959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/secure-post-merge.yml
+++ b/.github/workflows/secure-post-merge.yml
@@ -11,7 +11,45 @@ on:
       - '.github/workflows/secure-post-merge-notags.yml'
 
 jobs:
+  check-if-api-tests-changed:
+    runs-on: ubuntu-latest
+    outputs:
+      api-tests-changed: ${{ steps.get-changed-files.outputs.any_changed }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get changed api-tests files
+        id: get-changed-files
+        uses: tj-actions/changed-files@6b2903bdce6310cfbddd87c418f253cf29b2dec9 # v44.5.6
+        with:
+          files: |
+            api-tests/**
+            .github/workflows/secure-pipeline-api-tests-image.yml
+
+      - name: List changed files
+        env:
+          CHANGED_FILES: ${{ steps.get-changed-files.outputs.all_changed_files }}
+        run: |
+          for file in ${CHANGED_FILES}; do
+            echo "$file was changed"
+          done
+
+  build-test-images-if-needed:
+    needs: check-if-api-tests-changed
+    if: ${{ needs.check-if-api-tests-changed.outputs.api-tests-changed == 'true' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/secure-post-merge.yml
```diff
@@ -11,7 +11,45 @@ on:
...
+      api-tests-changed: ${{ steps.get-changed-files.outputs.any_changed }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get changed api-tests files
+        id: get-changed-files
+        uses: tj-actions/changed-files@v44
```
Joe-Edwards-GDS: Sad that there's no out of the box solution for this :(

A cursory google also reveals https://github.com/dorny/paths-filter which appears to be a bit more focused on this use-case, but I don't have a clear idea which is best.

Might be worth flagging to @huwd and perhaps @TobySaundersGDS that we'd like to introduce a 3rd-party action here. There's a formal process for npm dependencies, but I'm not sure about GHAs (https://govukverify.atlassian.net/wiki/spaces/DIWAY/pages/4335697997/Library+approval+process)

I do think it's quite valuable to add this though!

Joe-Edwards-GDS (commented):
This looks reasonable to me, but might be worth asking for @TobySaundersGDS (or someone else from dev-platform) for a review as well

.github/workflows/secure-post-merge.yml
Wynndow: Have pinged them a message, see what they think. The two actions look pretty similar in functionality, just the syntax seems marginally different. Both look like well used and active repo's. ¯\_(ツ)_/¯

Wynndow: @Joe-Edwards-GDS Have checked and it's ok - @TobySaundersGDS suggested pinning to a particular SHA though, which I've done.

Joe-Edwards-GDS (approved):
Looks plausible - obviously worth keeping an eye on when we merge, as I'm sure you will :)

PR #2176, opened 2024-07-17
Pyic 6738 set target vot

DanCorderIPV:
### What changed

Set a target Vot value in the IPV session

### Why did it change

We will need to use the target vot property in an upcoming PR but we need to ensure that all users part way through using the site have this property set before we merge the following PR.

### Issue tracking

- [PYIC-6738](https://govukverify.atlassian.net/browse/PYIC-6738)

[PYIC-6738]: https://govukverify.atlassian.net/browse/PYIC-6738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -272,6 +272,11 @@ public class CheckExistingIdentityHandler
                             .getParsedVtr()
                             .getLowestStrengthRequestedGpg45Vot(configService);
 
+            // As almost all of our journeys are proving or mitigating a GPG45 vot we set the
+            // target vot here as a default value. It will be overridden for identity reuse.
+            ipvSessionItem.setTargetVot(lowestGpg45ConfidenceRequested);
+            ipvSessionService.updateIpvSession(ipvSessionItem);
+
             var contraIndicators =
                     ciMitService.getContraIndicators(
                             clientOAuthSessionItem.getUserId(), govukSigninJourneyId, ipAddress);
@@ -309,6 +314,12 @@ public class CheckExistingIdentityHandler
                             vcs,
                             areGpg45VcsCorrelated);
             if (profileMatchResponse.isPresent()) {
+                // We are re-using an existing Vot so it might not be a GPG45 vot
+                ipvSessionItem.setTargetVot(
+                        clientOAuthSessionItem
+                                .getParsedVtr()
+                                .getLowestStrengthRequestedVot(configService));
+                ipvSessionService.updateIpvSession(ipvSessionItem);
                 return profileMatchResponse.get();
             }
 
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -68,7 +68,6 @@ import static com.amazonaws.util.CollectionUtils.isNullOrEmpty;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EVCS_READ_ENABLED;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EVCS_WRITE_ENABLED;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.INHERITED_IDENTITY;
-import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.P1_JOURNEYS_ENABLED;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.REPEAT_FRAUD_CHECK;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.RESET_IDENTITY;
 import static uk.gov.di.ipv.core.library.domain.Cri.EXPERIAN_FRAUD;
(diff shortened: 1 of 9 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/Vtr.java
```diff
@@ -0,0 +1,58 @@
...
+                && !configService.enabled(P1_JOURNEYS_ENABLED)) {
+            lowestStrengthRequestedGpg45Vot =
+                    requestedGpg45VotsByStrengthDescending.get(
+                            requestedGpg45VotsByStrengthDescending.size() - 2);
+        }
+
+        return lowestStrengthRequestedGpg45Vot;
+    }
+
+    // Refactor this out in PYIC-6984
```
blakeyp: Are these comments still relevant?

DanCorderIPV: Good point, not these two

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -120,8 +120,13 @@ public IpvSessionItem generateIpvSession(
         if (errorObject == null) {
             if (isReverification) {
                 ipvSessionItem.pushState(new JourneyState(REVERIFICATION, START_STATE));
+                // Currently reverifcation journeys don't check the user's existing profile so we
+                // have to hard code this to P2 here. Eventually this can be set in the first step
+                // of the reverification journey.
+                ipvSessionItem.setTargetVot(Vot.P2);
```
blakeyp, line 126: @shivanshuit914 added something to [ClientOAuthSessionDetailsService.java](https://github.com/govuk-one-login/ipv-core-back/pull/2168/files#diff-0384087424c1800866db67985f333eed3e5ab7c662b4f0af4da055d24cea30f7) to set a hardcoded vot when vtr is empty - can this now be removed?

DanCorderIPV: Hmm, I don't think so. It does look like it caused by the same root though.

DanCorderIPV: I'm checking with Joe, perhaps we can remove Shiv's changes in the follow up PR

blakeyp (approved):

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -310,6 +314,12 @@ private JourneyResponse getJourneyResponse(
                             vcs,
                             areGpg45VcsCorrelated);
             if (profileMatchResponse.isPresent()) {
+                // We are re-using an existing Vot so it might not be a GPG45 vot
```
MikeCollingwood, line 317: Imo, I feel it makes more sense to start with targetVot for all types then, on understanding it must be gpg45 (inverse of this if statement), update it to be gpg45 type.

(Could then remove the comment above setTargetVot in CheckExistingIdentity since the usage feels clearer)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -120,8 +120,13 @@ public IpvSessionItem generateIpvSession(
...
             } else {
                 ipvSessionItem.pushState(new JourneyState(INITIAL_JOURNEY_SELECTION, START_STATE));
+                // For non-reverification journeys targetVot is set in CheckExistingIdentity
```
MikeCollingwood, line 129: I think this comment implies it's not done in the other case. Maybe an update to the comment above indicating it "updates" or "resets" targetVot?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -265,13 +264,18 @@ private JourneyResponse getJourneyResponse(
...
+            // identities
+            var lowestGpg45ConfidenceRequested =
+                    clientOAuthSessionItem
+                            .getParsedVtr()
+                            .getLowestStrengthRequestedGpg45Vot(configService);
+
+            // As almost all of our journeys are proving or mitigating a GPG45 vot we set the
+            // target vot here as a default value. It will be overridden for identity reuse.
+            ipvSessionItem.setTargetVot(lowestGpg45ConfidenceRequested);
+            ipvSessionService.updateIpvSession(ipvSessionItem);
```
MikeCollingwood, line 278: Could this code be put in with ipv session initialisation? There's no difference functionally and makes it easier to see how the ipv session is created

DanCorderIPV: I did try this initially, I don't remember exactly why, but it's not simple to make it work that way.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: Logically I kind of agree with that, but practically there are lots of exit points in this method and you'd have to remember to reset the target vot at every one of them except this one which felt a lot more fragile.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
DanCorderIPV: That implication is true - I guess these comments are confusing in some way then.

The other branch above is only for MFA reset (reverification) journeys and on those journeys the branch above is the only place that target vot is set.

What would you change in the comments to avoid the misunderstanding?

MikeCollingwood: Ah, we put people onto the reverification journey from InitialiseIpvSession. I was in the mindset we were setting the vot there (but I see your comment on that being difficult) so got confused

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
MikeCollingwood: Tried to think of a way of making it less fragile but think I have to agree with this. It's a shame but practical

PR #2187, opened 2024-07-19
PYIC-7018: Separate common journeys

MikeCollingwood:
### What changed

- Separated out 2 journeys from new p2 identity because will be used in P1 and simplifies 

new-p2-identity:
Screenshot 2024-07-19 at 11 50 36

f2f-hand-off:
Screenshot 2024-07-19 at 11 52 58

evaluate-scores:
Screenshot 2024-07-19 at 11 53 17

### Why did it change

- P1 will use these as well and they are identical in both. It will reduce code and could make the functionality simpler to understand

### Issue tracking

- [PYIC-7018](https://govukverify.atlassian.net/browse/PYIC-7018)

### Notes

- Apart from redundancy, this will be easily checked with e2e tests

[PYIC-7018]: https://govukverify.atlassian.net/browse/PYIC-7018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -290,57 +290,10 @@ states:
         targetJourney: INELIGIBLE
         targetState: INELIGIBLE_SKIP_MESSAGE
 
-  EVALUATE_GPG45_SCORES:
-    response:
-      type: process
-      lambda: evaluate-gpg45-scores
-    events:
-      met:
-        targetState: STORE_IDENTITY_BEFORE_SUCCESS
-        checkFeatureFlag:
-          ticfCriBeta:
-            targetState: CRI_TICF_BEFORE_SUCCESS
-      unmet:
-        targetJourney: FAILED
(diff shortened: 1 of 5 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -329,94 +290,11 @@ states:
...
-    response:
-      type: process
-      lambda: call-ticf-cri
-    parent: CRI_TICF_STATE
-    events:
-      next:
-        targetState: CRI_F2F
-      enhanced-verification:
-        targetState: CRI_F2F
-
```
blakeyp: To stage out this change into prod we might want to keep these old states initially and only remove them once deployed (in case it breaks any journeys that people are halfway through) ?

MikeCollingwood: Ah good catch, will do

blakeyp (approved):

PR #2191, opened 2024-07-22
Pyic 6900 low verification kbvs

DanCorderIPV:
### What changed

Add verification score to KBV requests

### Why did it change

So we can specify a lower score for low confidence journeys

### Issue tracking

- [PYIC-6900](https://govukverify.atlassian.net/browse/PYIC-6900)

[PYIC-6900]: https://govukverify.atlassian.net/browse/PYIC-6900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV:
I've talked to the Orange team and they say that their work to use the verification score is already in place in staging (but not prod) so this should just work as long as we don't send a low confidence score in prod.

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelperTest.java
```diff
@@ -167,8 +168,8 @@ void shouldCreateSignedJWTWithGivenParameters(
         for (Map.Entry<String, Object> entry : expectedClaims.entrySet()) {
             var actual = result.getJWTClaimsSet().getClaim(entry.getKey());
             assertEquals(
-                    entry.getValue(),
-                    actual,
+                    objectMapper.writeValueAsString(entry.getValue()),
+                    objectMapper.writeValueAsString(actual),
```
DanCorderIPV, line 172: This test used to check that the object references were the same. This now checks that the objects serialise to the same string.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/EvidenceRequest.java
```diff
@@ -10,19 +10,41 @@
...
+
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
-    @JsonInclude(JsonInclude.Include.NON_NULL)
     private final String scoringPolicy;
+    private final Integer strengthScore;
+    private final Integer verificationScore;
+
+    // The JSON serialiser used by the Nimbus JWT library includes null values within claims so we
+    // need to have a way of giving it the values without including nulls. This is far from ideal.
```
MikeCollingwood: Could do by serialising and unserialising with Jackson?

```java
    public Map<String, Object> toMap() {
        String jsonString = objectMapper.writeValueAsString(obj);
        return objectMapper.readValue(jsonString, new TypeReference<Map<String, Object>>() {});
    }
```

Maybe we could bake this into a base class designed to parse into a map? (probably overkill unless you can think of other uses?)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/domain/EvidenceRequestTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class EvidenceRequestTest {
+    @SuppressWarnings("java:S125") // Commented out code
+    // {"scoringPolicy":"gpg45","strengthScore":2}
```
MikeCollingwood, line 15: Do you know why other comments don't need to be suppressed? Would this work?
```suggestion
    // Origin: {"scoringPolicy":"gpg45","strengthScore":2}
```

DanCorderIPV: SonarCloud thinks that this is commented out code rather than a real comment so it complains
I tried `Data: {"scoringPolicy":"gpg45","strengthScore":2}` SonarCloud still didn't like it :(

blakeyp (dismissed):

blakeyp (approved):

PR #2196, opened 2024-07-23
PYIC-6991 local config

Joe-Edwards-GDS:
### What changed

Added a config service that reads config from a local YAML file, which is used in local dev.

Also set up a template YAML file to help make setup easier, in lieu of a more sophisticated link to dev-deploy.

### Why did it change

We want to be able to run core-back locally without needing to set up config in parameter store.

### Issue tracking

- [PYIC-6991](https://govukverify.atlassian.net/browse/PYIC-6991)

[PYIC-6991]: https://govukverify.atlassian.net/browse/PYIC-6991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.secrets.baseline
```diff
@@ -127,10 +127,15 @@
         "(?i)^apikey$",
         "(?i)^test_api_key$",
         "(?i)^test_basic_auth_secret$",
-        "(?i)^test_authorisation_code$"
+        "(?i)^test_authorisation_code$",
+        "E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM",
+        "KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04",
+        "RBXnILIdExUEWUJMlYeD6agE8u9gGgA3InKrd5TKhhY",
+        "kKtt9v_xq9oqvv5_E8AHcV77IYQfyNwaTQyTYxdO_UM",
+        "vyapkvJXLwpYRJjbkQD99V2gcPEUKrO3dwjcAA9TPkLucQEZvYZvb7-wfSHxlvJlJcdS20r5PKKmqdPeW3Y4ir3WsVVeiht2iOZUreUO5O3V3o7ImvEjPS_2_ZKMHCwUf51a6WGOaDjO87OX_bluV2dp01n-E3kiIl6RmWCVywjn13fX3jsX0LMCM_bt3HofJqiYhhNymEwh39oR_D7EE5sLUii2XvpTYPa6L_uPwdKa4vRl4h4owrWEJaJifMorGcvqhCK1JOHqgknN_3cb_ns9Px6ynQCeFXvBDJy4q71clkBq_EZs5227Y1S222wXIwUYN8w5YORQe3M-pCIh1Q"
```
Joe-Edwards-GDS, line 135: These are the high-entropy components of the test keys - this would probably let us remove a bunch of allowlist comments, but there's no real urgency to do so.

local-running/core.local.params.template.yaml
Joe-Edwards-GDS, line 1: This could get quite annoying to keep updated, particularly for things like feature flags (that are updated frequently).

It might be nice to set up some kind of script that generates this config from core-common-infra or similar. It uses the `ipv-core-dev` client for CRIs (which usually corresponds to dev02) - but it doesn't particularly matter since ultimately it won't be linked to any AWS environment.

local-running/setConfigForLocalOrCloudRunning.py
Joe-Edwards-GDS, line 1: This now only sets up the local SQS client, which I haven't worked out a plan for yet...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public YamlConfigService(File parametersFile, File secretsFile) {
+        try {
+            var paramsYaml = YAML_OBJECT_MAPPER.readTree(parametersFile).get(CORE);
+            var secretsYaml = YAML_OBJECT_MAPPER.readTree(secretsFile).get(CORE);
+
+            addConfig(parameters, paramsYaml);
+            addConfig(secrets, secretsYaml);
+
+            paramsYaml
```
Wynndow: This'll have already been mapped into the `parameters` map I think - do we need to maintain a separate map? The override could just be looked up in the parameters map?

local-running/compose.yaml
```diff
@@ -72,250 +68,11 @@ services:
...
-      USER_ISSUED_CREDENTIALS_TABLE_NAME: user-issued-credentials-v2-${ENVIRONMENT}
       LAMBDA_TASK_ROOT: handler
-      LOCAL_DEV: true
-      ENVIRONMENT:
+      ENVIRONMENT: local
       JAVA_TOOL_OPTIONS: -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5002
     ports:
       - "3002:3002"
       - "5002:5002"
-
```
Wynndow, line 92: Nice

local-running/setConfigForLocalOrCloudRunning.py
```diff
@@ -4,240 +4,6 @@
 from collections import namedtuple
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
-# A script to set the required param values in SSM for local or cloud running.
```
Wynndow, line 7: Also nice

Wynndow (dismissed):
Looks good - just one question

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
Joe-Edwards-GDS: Yeah, in my head it was to avoid iterating through the parameters, but I don't think we need to.

Wynndow (approved):

PR #2199, opened 2024-07-23
PYIC-7016: add nino only context for the p1 journey multiple doc page

shivanshuit914:
## Proposed changes

- Add nino only context for the p1 journey multiple doc page

### What changed

- Updated P1 journey map

### Issue tracking

- [PYIC-7016](https://govukverify.atlassian.net/browse/PYIC-7016)

[PYIC-7016]: https://govukverify.atlassian.net/browse/PYIC-7016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
@@ -138,7 +138,7 @@ states:
         targetJourney: TECHNICAL_ERROR
         targetState: ERROR
       end:
-        targetState: MULTIPLE_DOC_F2F_CHECK_PAGE
+        targetState: MULTIPLE_DOC_F2F_NINO_CHECK_PAGE
         checkIfDisabled:
           f2f:
             targetState: MULTIPLE_DOC_CHECK_PAGE
@@ -167,7 +167,7 @@ states:
     response:
       type: page
       pageId: prove-identity-no-photo-id
-      context: nino-only
+      context: nino
     events:
       next:
         targetState: CRI_CLAIMED_IDENTITY_NO_PHOTO_ID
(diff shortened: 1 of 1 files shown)


MikeCollingwood (commented):
Could we merge the two states to be something like this?
```yaml
  MULTIPLE_DOC_CHECK_PAGE:
    ...
    events:
      ...
      end:
        targetState: F2F_PYI_POST_OFFICE
        checkIfDisabled:
          f2f:
            targetJourney: INELIGIBLE
            targetState: INELIGIBLE_SKIP_MESSAGE
```

EDIT: We don't want f2f routing anyway, so this is not valid ^

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -237,6 +207,10 @@ states:
         targetState: NINO_START_PAGE
       end:
         targetState: F2F_PYI_POST_OFFICE
+        checkIfDisabled:
```
MikeCollingwood: Were we saying we didn't need any routing for f2f? If so then we should probably just send them to:
```
targetJourney: INELIGIBLE
targetState: INELIGIBLE_SKIP_MESSAGE
```

shivanshuit914: Hm possible, @blakeyp  Can you pls confirm if we need this check ? and need to send user to diff target State for f2f ?

blakeyp: For now we do still need to handle F2F being disabled same as the P2 journey

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -223,11 +223,11 @@ states:
         targetJourney: INELIGIBLE
         targetState: INELIGIBLE_SKIP_MESSAGE
 
-  MULTIPLE_DOC_F2F_CHECK_PAGE:
+  MULTIPLE_DOC_F2F_NINO_CHECK_PAGE:
     response:
       type: page
-      context: f2f
-      pageId: page-multiple-doc-check # PYIC-7016: Update so can produce nino event
+      context: nino-only
```
blakeyp: Just a thought just `nino` might be better as the context value? `nino-only` could imply that it only has the NINO option on the page 😅

shivanshuit914: Yes makes sense

blakeyp (approved):

PR #2201, opened 2024-07-23
PYIC-6960 local audit service

Joe-Edwards-GDS:
### What changed

Added local audit service that stores up audit events and makes them available to the local running process

### Why did it change

When running locally we don't want to use AWS resources, and we would like to be able to make assertions on the audit events in API tests.

### Issue tracking

- [PYIC-6960](https://govukverify.atlassian.net/browse/PYIC-6960)

[PYIC-6960]: https://govukverify.atlassian.net/browse/PYIC-6960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/src/steps/ipv-steps.ts
```diff
@@ -119,3 +125,18 @@ When(
 Then("I get a {string} identity", function (this: World, vot: string): void {
   assert.equal(vot, this.identity.vot);
 });
+
+Then(
+  "a(n) {string} audit event was recorded",
```
Wynndow: As we can only run this step locally, i'm wonderin if the step anem should reflect that. If this is running in the build environment and I see it go green, I might assume that this has actually been checked.

api-tests/src/steps/ipv-steps.ts
```diff
@@ -119,3 +125,18 @@ When(
...
+  async function (this: World, eventName: string): Promise<void> {
+    if (config.LOCAL_AUDIT_EVENTS) {
+      const auditEvents = await getAuditEvents(this.journeyId);
+      const event = auditEvents.find((e) => e.event_name === eventName);
+      if (!event) {
```
Wynndow, line 135: Could this be a bit more specific somehow? In future, if we were checking that for example we'd received a `IPV_VC_RECEIVED` event, that could have come from a bunch of places.

api-tests/src/types/audit-events.ts
```diff
@@ -0,0 +1,13 @@
...
+  event_timestamp_ms: number;
+  component_id: string;
+  user: {
+    user_id: string;
+    session_id: string;
+    govuk_signin_journey_id: string;
+    ip_address: string;
+  };
+  extensions?: unknown;
+  restricted?: unknown;
```
Wynndow, line 12: What benefit does `unknown` give us over `any`?

api-tests/src/steps/ipv-steps.ts
```diff
@@ -4,6 +4,7 @@ import { World } from "../types/world.js";
 import * as internalClient from "../clients/core-back-internal-client.js";
 import * as externalClient from "../clients/core-back-external-client.js";
 import * as criStubClient from "../clients/cri-stub-client.js";
+import { getAuditEvents } from "../clients/local-audit-client.js";
```
Wynndow: It's probably fine, but all the other clients are imported with a `*`, so that we call a method on them in the code. Which makes it obvious which client it was. Dunno how you feel about that.

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/LocalAuditService.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public void sendAuditEvent(AuditEvent auditEvent) throws AuditException {
+        pendingAuditEvents.add(auditEvent);
+        LOGGER.info(LogHelper.buildLogMessage("Sending audit event").with("event", auditEvent));
+    }
+
+    @Override
+    public synchronized void awaitAuditEvents() {
+        AUDIT_EVENTS.addAll(pendingAuditEvents);
```
Wynndow, line 36: nice

api-tests/src/steps/ipv-steps.ts
Joe-Edwards-GDS: Yeah fair enough - I sort of dislike `*` imports because they can unexpectedly pollute your namespace (same rationale as Java) and inhibit tree-shaking (hard to know which imports you actually need).

However - neither of those are really a concern here, and it might make it a bit clearer as you say.

api-tests/src/steps/ipv-steps.ts
Joe-Edwards-GDS: Yeah, it looks like you can actually mark the step as 'skipped' by returning the literal `'skipped'` so I'll try that.

What do you think? It wouldn't tell you by reading the scenario, but would show up in the test report.

api-tests/src/steps/ipv-steps.ts
Joe-Edwards-GDS: Yeah, I think we'd want to extend this to write proper audit event tests (this was really just a POC)...

Either something generic `a(n) {string} audit event was recorded with property '{string}' = '{string}'` or we could add a few bespoke ones `an IPV_VC_RECEIVED event from {string} CRI`

api-tests/src/types/audit-events.ts
Joe-Edwards-GDS: `unknown` forces you to explicitly assert the type, whereas `any` allows you to treat it as any type (making it easier to make mistakes).

For example:
```
function fn(unknownParam: unknown, anyParam: any) {
  unknownParam.foo; // not allowed
  anyParam.foo; // ok

  (unknownParam as SomeTypeWithFoo).foo; // ok (but an unchecked cast)

  if (isSomeTypeWithFoo(unknownParam)) {
    unknownParam.foo; // ok
  }
}
```

libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/LocalAuditService.java
Joe-Edwards-GDS: Yeah - I thought this was a neat way to test that we call await!

api-tests/src/steps/ipv-steps.ts
Wynndow: Yeah, that sounds good. Would be good to not run the step when not needed. And if it's obvious from looking at the report that's enough

Joe-Edwards-GDS: Ah this doesn't quite work, it ends up skipping the entire scenario :( That might be fine if we have separate audit scenarios, but it'd be kinda convenient to combine them.

I suppose I can add `(local only)` or something to the name - a bit rubbish, but if you've got decent autocomplete it shouldn't make things much harder.

Joe-Edwards-GDS: Updated - used square brackets so it doesn't conflict with cucumber syntax

Wynndow (approved):

PR #2205, opened 2024-07-24
PYIC-7112 Update to have ipvSession call from CheckExistingIdentity behind retry and exponential backoff.

rpayal:
## Proposed changes
Resolve errors seen in log during load testing 
- 170 errors where the journey handler couldn't find the session

### What changed
Update to have ipvSession call from CheckExistingIdentity behind retry and exponential backoff.

### Issue tracking

- [PYIC-7112](https://govukverify.atlassian.net/browse/PYIC-7112)

### Other considerations

[PYIC-7112]: https://govukverify.atlassian.net/browse/PYIC-7112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/GetIpvSessionException.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.di.ipv.core.library.exceptions;
+
+public class GetIpvSessionException extends RuntimeException {
```
Joe-Edwards-GDS: I don't think these should be `RuntimeException`s - we want to catch them!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -51,52 +52,54 @@ public IpvSessionService(ConfigService configService) {
...
-    }
-
-    public IpvSessionItem getIpvSessionByAccessToken(String accessToken)
-            throws UnknownAccessTokenException, GetAccessTokenException {
-        try {
-            return Retry.runTaskWithBackoff(
-                    sleeper,
-                    7,
-                    10,
+    public IpvSessionItem getIpvSession(String ipvSessionId, boolean withRetry)
```
Joe-Edwards-GDS: Is this ever called with `withRetry = false`? Should we just rename it `getIpvSessionWithRetry`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/UnknownAccessTokenException.java
Joe-Edwards-GDS, line 1: (Should this also be combined with `GetIpvSessionException`?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -51,52 +52,54 @@ public IpvSessionService(ConfigService configService) {
...
     }
 
+    public Optional<IpvSessionItem> getIpvSessionByAuthorizationCode(String authorizationCode)
+            throws GetIpvSessionException {
+        return callRunTaskWithBackoff(
+                () -> {
+                    IpvSessionItem ipvSessionItem =
+                            dataStore.getItemByIndex(
+                                    "authorizationCode", DigestUtils.sha256Hex(authorizationCode));
+                    return Optional.ofNullable(ipvSessionItem);
```
Joe-Edwards-GDS: Will this ever retry? It looks like it'll just return `Optional.empty` if it can't find it the first time...

It probably needs to look a bit more like `getIpvSessionByAccessToken` and return a non-optional `IpvSessionItem`.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -174,4 +177,23 @@ public void updateIpvSession(IpvSessionItem updatedIpvSessionItem) {
...
+            return Retry.runTaskWithBackoff(sleeper, 7, 10, task);
+        } catch (InterruptedException e) {
+            LOGGER.warn(LogHelper.buildLogMessage("backoff and retry sleep was interrupted"));
+            Thread.currentThread().interrupt();
+            throw new GetIpvSessionException("Failed to get ipv session", e);
+        } catch (NonRetryableException e) {
+            LOGGER.warn(
+                    LogHelper.buildErrorMessage(
+                            "exception occurred while retrying to get ipv session", e));
+            if (e.getCause() instanceof UnknownAccessTokenException uatException) {
```
Joe-Edwards-GDS: Should we have a similar block here for `e.getCause() instanceof GetIpvSessionException`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/GetIpvSessionException.java
Joe-Edwards-GDS: I'd probably prefer `IpvSessionNotFoundException` or similar - feels a bit more specific

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/UnknownAccessTokenException.java
rpayal: thought so initially....done.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/GetIpvSessionException.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
rpayal: called frm original method - getIpvSession(String ipvSessionId)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/GetIpvSessionException.java
rpayal: Was catching it where needed in handler where retry is needed. 
Yeah was avoiding changes in other places where retry is not used for get Ipv session. 
Done.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
Joe-Edwards-GDS: Yeah, but that's the only place, so instead you could do:
```
public IpvSessionItem getIpvSession(String ipvSessionId) {
    return dataStore.getItem(ipvSessionId);
}

public IpvSessionItem getIpvSessionWithRetry(String ipvSessionId)
        throws IpvSessionNotFoundException {
    return callRunTaskWithBackoff(...);
}
```
You still only have one copy of each implementation, and the non-retried version doesn't need to declare an exception that it'll never throw...

Joe-Edwards-GDS (commented):
Looks like one of the contract tests has broken as well

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -50,51 +49,59 @@ public IpvSessionService(ConfigService configService) {
         this.sleeper = new Sleeper();
     }
 
-    public IpvSessionItem getIpvSession(String ipvSessionId) {
-        return dataStore.getItem(ipvSessionId);
+    public IpvSessionItem getIpvSession(String ipvSessionId) throws IpvSessionNotFoundException {
+        IpvSessionItem ipvSessionItem = dataStore.getItem(ipvSessionId);
+        if (ipvSessionItem == null) {
+            throw new IpvSessionNotFoundException(
+                    "The session not found in the database for the supplied session Id");
```
Joe-Edwards-GDS, line 56: Careful, this is now a breaking change if any clients were expecting to handle `null` being returned. For example in `ProcessJourneyEventHandler` there's some code that handles `null` (but also just throws an error).

Hopefully you can now remove that code, and can you make sure that there's nowhere similar?

Joe-Edwards-GDS (commented):
This looks like an improvement, but want to make sure we haven't inadvertently changed any behaviour.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
rpayal: only here we are checking for null, removed it as not needed as handled in exception.

Joe-Edwards-GDS (approved):

PR #2210, opened 2024-07-25
PYIC-6992 local key config

Joe-Edwards-GDS:
### What changed

- Turned `KmsEs256SignerFactory` into a more generic `SignerFactory`, which creates signers either using KMS or a locally provided key
- Extracted a `JWEDecrypterFactory` which creates decrypters either using KMS or a locally provided key
- Updated template local config with non-KMS keys for signing and decryption

### Why did it change

To run core-back without a KMS dependency

### Issue tracking

- [PYIC-6992](https://govukverify.atlassian.net/browse/PYIC-6992)

[PYIC-6992]: https://govukverify.atlassian.net/browse/PYIC-6992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/service/KmsRsaDecrypter.java
```diff
@@ -74,6 +73,8 @@ public byte[] decrypt(
                     AlgorithmSupportMessage.unsupportedJWEAlgorithm(alg, supportedJWEAlgorithms()));
         }
 
+        var keyId = configService.getParameter(JAR_KMS_ENCRYPTION_KEY_ID);
```
Joe-Edwards-GDS, line 76: This means that the decrypter 'knows' about the config parameter for performing decryption (so it would be less reusable if we needed to decrypt other JWEs with a different key), but...
- We don't need to do this currently
- The previous method meant we needed to be aware of the different types of decrypter (calling code needs to know about KMS key ids)

A 'better' solution might involve a factory being passed around (like with the signer), but it doesn't feel like a valuable refactoring right now.

rpayal (dismissed):
LGTM

rpayal (approved):

PR #2212, opened 2024-07-25
PYIC-6021: update CIMIT service to use CIMIT internal api

thebauSoftwire:
### What changed
- updates policies for relevant resources to have access to secret api key
- make http request when `cimitApiGateway` feature flag is enabled
- adds new unit tests for the above

### Why did it change

The CIMIT interface is being updated to use a private API Gateway. We want to update the CIMIT service in core-back to be able to send requests to this new interface.

### Issue tracking

- [PYIC-6021](https://govukverify.atlassian.net/browse/PYIC-6021)

[PYIC-6021]: https://govukverify.atlassian.net/browse/PYIC-6021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/cimit/domain/PostCiPrivateApiRequest.java
Joe-Edwards-GDS, line 1: We can probably drop the 'Private' from these classes - i.e. just `PostCiApiRequest.java`

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -72,72 +104,137 @@ public CiMitService(
...
+                var uri = getUriBuilderWithBaseApiUrl(POST_CI_ENDPOINT).build();
+
+                var response = sendPostHttpRequest(uri, govukSigninJourneyId, ipAddress, payload);
+                var parsedResponse =
+                        OBJECT_MAPPER.readValue(response.body(), PrivateApiResponse.class);
+
+                if (FAILED_RESPONSE.equals(parsedResponse.result())) {
+                    logApiRequestError(parsedResponse);
+                    throw new CiPutException(FAILED_API_REQUEST);
+                }
```
Joe-Edwards-GDS: Could parsing the response and detecting failures live inside `sendPostHttpRequest` (perhaps renamed to `sendCimitApiRequest`)? Looks the same for each method

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -72,72 +104,137 @@ public CiMitService(
...
-
-        LOGGER.info(LogHelper.buildLogMessage("Sending VC to CIMIT."));
-        var response = lambdaClient.invoke(invokeRequest);
-
-        if (lambdaExecutionFailed(response)) {
-            logLambdaExecutionError(response, CI_STORAGE_PUT_LAMBDA_ARN);
+            throw new CiPutException("Failed to serialize payload for post CI request.");
+        } catch (CiPutException e) {
+            throw e;
+        } catch (Exception e) {
```
Joe-Edwards-GDS: We very rarely want to catch blanket `Exception`s - can we be more specific about which exceptions you expect?

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -72,72 +104,137 @@ public CiMitService(
...
-                        .build();
-
-        LOGGER.info(LogHelper.buildLogMessage("Sending VC to CIMIT."));
-        var response = lambdaClient.invoke(invokeRequest);
-
-        if (lambdaExecutionFailed(response)) {
-            logLambdaExecutionError(response, CI_STORAGE_PUT_LAMBDA_ARN);
+            throw new CiPutException("Failed to serialize payload for post CI request.");
+        } catch (CiPutException e) {
+            throw e;
```
Joe-Edwards-GDS: (Once you make the block below more specific, this shouldn't be required!)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -281,4 +401,62 @@ private void logLambdaExecutionError(
...
+        return requestBuilder;
+    }
+
+    private HttpResponse<String> sendHttpRequest(HttpRequest cimitHttpRequest)
+            throws NonRetryableException {
+        try {
+            LOGGER.info(LogHelper.buildLogMessage("Sending HTTP request to CiMit."));
+            return httpClient.send(cimitHttpRequest, HttpResponse.BodyHandlers.ofString());
+        } catch (IOException e) {
+            throw new NonRetryableException(e);
```
Joe-Edwards-GDS: I don't think `NonRetryableException` is right here... I think you either want to let the `IOException` propagate up to the handler methods and catch it there, or catch here and consolidate all the CIMIT exceptions into a more general `CimitException`...

(I don't think we ever really need to distinguish `CiPutException` from `CiRetrievalException`, for example)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -72,72 +104,137 @@ public CiMitService(
...
-                            new PutCiRequest(govukSigninJourneyId, ipAddress, vc.getVcString()));
-
+            if (configService.enabled(CIMIT_API_GATEWAY_ENABLED)) {
+                var payload =
+                        OBJECT_MAPPER.writeValueAsString(
+                                new PostCiPrivateApiRequest(vc.getVcString()));
+
+                var uri = getUriBuilderWithBaseApiUrl(POST_CI_ENDPOINT).build();
+
+                var response = sendPostHttpRequest(uri, govukSigninJourneyId, ipAddress, payload);
```
Joe-Edwards-GDS: Would it be easier to just pass the relative URL here and let the common method set up the base URL?

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -157,57 +255,79 @@ public ContraIndicators getContraIndicators(VerifiableCredential vc)
...
-
-        InvokeResponse response;
         try {
-            response = lambdaClient.invoke(invokeRequest);
+            if (configService.enabled(CIMIT_API_GATEWAY_ENABLED)) {
+                var uriBuilder = getUriBuilderWithBaseApiUrl(GET_VCS_ENDPOINT);
+
+                var uri = uriBuilder.addParameter(USER_ID_PARAMETER, userId).build();
+
+                var response = sendGetHttpRequest(uri, govukSigninJourneyId, ipAddress);
```
Joe-Edwards-GDS: (similarly - maybe we should just add `userId` as a parameter to this method and pass along with `GET_VCS_ENDPOINT`, so the URL is built in one place)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/ConfigurationVariable.java
```diff
@@ -7,6 +7,8 @@ public enum ConfigurationVariable {
     CIMIT_COMPONENT_ID("cimit/componentId"),
     CIMIT_CONFIG("cimit/config"),
     CIMIT_SIGNING_KEY("cimit/signingKey"),
+    CIMIT_INTERNAL_API_BASE_URL("cimit/apiBaseUrl"),
+    CIMIT_INTERNAL_API_KEY("cimitInternalApi"),
```
Joe-Edwards-GDS: (No need to call this `INTERNAL`)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -72,72 +104,137 @@ public CiMitService(
...
+        LOGGER.info(LogHelper.buildLogMessage("Sending VC to CIMIT."));
         try {
-            payload =
-                    OBJECT_MAPPER.writeValueAsString(
-                            new PutCiRequest(govukSigninJourneyId, ipAddress, vc.getVcString()));
-
+            if (configService.enabled(CIMIT_API_GATEWAY_ENABLED)) {
+                var payload =
+                        OBJECT_MAPPER.writeValueAsString(
+                                new PostCiPrivateApiRequest(vc.getVcString()));
```
rpayal: Seem it'll be PUT api call here.

thebauSoftwire: I think for the API we use a post whereas using the lambdas directly we use a put

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -157,57 +236,76 @@ public ContraIndicators getContraIndicators(VerifiableCredential vc)
...
+                    throw new CiRetrievalException(FAILED_LAMBDA_MESSAGE);
+                }
+                return response.payload().asUtf8String();
+            }
         } catch (LambdaException e) {
             LOGGER.error(LogHelper.buildErrorMessage("AWSLambda client invocation failed.", e));
             throw new CiRetrievalException(FAILED_LAMBDA_MESSAGE);
+        } catch (CiRetrievalException e) {
+            throw e;
+        } catch (Exception e) {
```
Joe-Edwards-GDS: Same comments about catching `Exception` :)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -281,4 +379,70 @@ private void logLambdaExecutionError(
...
+    }
+
+    private HttpRequest.Builder buildHttpRequest(
+            URI uri, String govukSigninJourneyId, String ipAddress) {
+        var requestBuilder =
+                HttpRequest.newBuilder()
+                        .uri(uri)
+                        .header(GOVUK_SIGNIN_JOURNEY_ID_HEADER, govukSigninJourneyId)
+                        .header(CONTENT_TYPE, ContentType.APPLICATION_JSON.toString())
+                        .header(X_API_KEY_HEADER, configService.getSecret(CIMIT_API_KEY));
```
Joe-Edwards-GDS: (I think this might need to be optional as well, so it's not supplied if there is no API key)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -281,4 +379,70 @@ private void logLambdaExecutionError(
...
+    }
+
+    private HttpResponse<String> sendHttpRequest(HttpRequest cimitHttpRequest)
+            throws NonRetryableException, IOException {
+        try {
+            LOGGER.info(LogHelper.buildLogMessage("Sending HTTP request to CiMit."));
+            return httpClient.send(cimitHttpRequest, HttpResponse.BodyHandlers.ofString());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new NonRetryableException(e);
```
Joe-Edwards-GDS: I don't think we should be using `Retryable`/`NonRetryableException` anywhere here

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
thebauSoftwire: ah missed this! Thanks!

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CiMitService.java
```diff
@@ -430,9 +432,16 @@ private HttpRequest.Builder buildHttpRequest(
...
-        return httpClient.send(cimitHttpRequest, HttpResponse.BodyHandlers.ofString());
+            throws CimitHttpRequestException {
+        try {
+            LOGGER.info(LogHelper.buildLogMessage("Sending HTTP request to CiMit."));
+            return httpClient.send(cimitHttpRequest, HttpResponse.BodyHandlers.ofString());
+        } catch (InterruptedException | IOException e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            throw new CimitHttpRequestException(FAILED_API_REQUEST);
```
Joe-Edwards-GDS, line 443: N.B. this will lose any useful information in the `IOException`, but I think this might have already been the case...

Perhaps merge this PR and then we can look at fixing it up after. In general, whenever you catch an exception you should capture the details, by _either_:
- Log the exception and raise a new one
- Include the exception as a `cause` of the new one
- (Explicitly decide that you don't care about the details of the error)

Joe-Edwards-GDS (approved):

PR #2223, opened 2024-07-29
PYIC-6083: Store inherited identities in evcs

Wynndow:
### What changed

Store inherited identities in evcs

### Why did it change

[PYIC-6083: Store inherited identities in EVCS](https://github.com/govuk-one-login/ipv-core-back/commit/480522d065f0831ab939755e42b9c6c528eccc68)

This updates the initalise session lambda to store inherited identities
in EVCS, rather than the tactical store.

We're not going to be going live with inherited identities before we're
migrated over to EVCS, so we don't need to worry about migrating
anything.

Adds a new method to the EVCS service to do the updating of an existing
inherited ID to historic, and storing a new one as current.

[PYIC-6083: Do not remove inherited ID when P2 attained](https://github.com/govuk-one-login/ipv-core-back/commit/aa6c604bb4819f527ae4198b9d2cc2805ef8f329)

We no longer want to supersede an inherited identity when a P2
idenentity is achieved - they should be treated separately. This means
not deleting an inherited identity VC if a P2 identity is attained.

[PYIC-6083: Don't update inherited identity when storing P2](https://github.com/govuk-one-login/ipv-core-back/commit/07ae5bd396cc04fdeebcd69019b1f1e141c08583)

We want a user to be able to have an inherited identity along side a P2
identity. This means that if a user with an existing inherited identity
attains a P2, we don't want to lose the inherited identity VC.

This adds metadata to the inherited identity VC in EVCS, which we can
then use when storing completed identities to ensure we don't update the
inherited identity from `CURRENT` to `HISTORIC`.

There are other options for how to do this. We could parse the VCs and
check the issuer, but this is a bit messy and doesn't play nicely with
the streams we're using.

The metadata itself could be in a different shape. This uses the key
"inheritedIdentity" with a value of the Cri ID. This may be more useful
in future that just a boolean flag.

[PYIC-7022: Include current inherited identity VC in bundle](https://github.com/govuk-one-login/ipv-core-back/pull/2223/commits/3cce4ce86f799bb4783499dc211cafc561c42c86)

When we store an inherited identity VC in EVCS it has a status of
`CURRENT`. If a user is returning from a F2F journey we were only using
the VCs with a `PENDING_RETURN` status. We need to ensure we're not
throwing away the current inherited identity VC, so that we can match
the correct VOT in the VTR for inherited identity journeys.

### Issue tracking

- [PYIC-6083](https://govukverify.atlassian.net/browse/PYIC-6083)

[PYIC-6083]: https://govukverify.atlassian.net/browse/PYIC-6083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -321,29 +329,98 @@ private static boolean isListEmpty(List<String> list) {
...
+                            "Too many current inherited identities found - %d",
+                            existingInheritedIdentity.size()));
+        }
+
+        if (existingInheritedIdentity.isEmpty()) {
+            LOGGER.info("No existing inherited identity found - storing new one");
+
+            evcsService.storeInheritedIdentity(userId, inheritedIdentityVc, null);
+            ipvSessionItem.setInheritedIdentityReceivedThisSession(true);
+            ipvSessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS, line 388: One concern I've just thought of here... EVCS writes are not synchronous, so we cannot rely on this being immediately readable in `CheckExistingIdentity`.

This might not be easy to solve... We could add it to the IPV Session as well/instead (i.e. `inheritedIdentityReceivedThisSession` goes from a boolean to the VC itself), but then the logic in CheckExistingIdentity becomes fiddly. Perhaps we just park it for now and create a ticket to track later.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -321,29 +329,98 @@ private static boolean isListEmpty(List<String> list) {
...
+            evcsService.storeInheritedIdentity(
+                    userId, inheritedIdentityVc, existingInheritedIdentity.get(0));
+            ipvSessionItem.setInheritedIdentityReceivedThisSession(true);
+            ipvSessionService.updateIpvSession(ipvSessionItem);
+        } else {
+            LOGGER.info("Existing inherited identity has stronger VOT - discarding new one");
+        }
+    }
+
+    @Tracing
```
Joe-Edwards-GDS: (Do we need tracing on this method? Is it expected to be significant to performance?)

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -321,29 +329,98 @@ private static boolean isListEmpty(List<String> list) {
...
+
+        if (existingInheritedIdentity.size() > 1) {
+            throw new InheritedIdentityException(
+                    String.format(
+                            "Too many current inherited identities found - %d",
+                            existingInheritedIdentity.size()));
+        }
+
+        if (existingInheritedIdentity.isEmpty()) {
+            LOGGER.info("No existing inherited identity found - storing new one");
```
Joe-Edwards-GDS: Can we use `LogHelper` when building messages, so we have a consistent JSON structure

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -321,29 +329,98 @@ private static boolean isListEmpty(List<String> list) {
...
+                                userId, clientOAuthSessionItem.getEvcsAccessToken(), CURRENT)
+                        .stream()
+                        .filter(vc -> HMRC_MIGRATION.equals(vc.getCri()))
+                        .toList();
+
+        if (existingInheritedIdentity.size() > 1) {
+            throw new InheritedIdentityException(
+                    String.format(
+                            "Too many current inherited identities found - %d",
+                            existingInheritedIdentity.size()));
```
Joe-Edwards-GDS: Does this need to be an error?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
```diff
@@ -220,6 +249,10 @@ private void updateExistingUserVCs(
             var existingCurrentUserVcsNotInSessionToUpdate =
                     existingEvcsUserVCs.stream()
                             .filter(vc -> vc.state().equals(CURRENT))
+                            .filter(
+                                    vc ->
+                                            vc.metadata().get("inheritedIdentity")
+                                                    == null) // Don't update inherited identity VCs
```
Joe-Edwards-GDS: Do we need a null-check on `metadata` here?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -395,6 +395,12 @@ private VerifiableCredentialBundle getEvcsCredentialBundle(
             if (isNullOrEmpty(evcsIdentityVcs)) {
                 evcsIdentityVcs = evcsVcs.get(CURRENT);
                 isPendingEvcs = false;
+            } else {
+                // Ensure we keep any inherited ID VCs in the bundle
```
Joe-Edwards-GDS, line 399: Good spot! I wonder if, when we revisit inherited ids properly, we should try and store them as a separate status 🤔 - something to worry about later.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -321,29 +329,98 @@ private static boolean isListEmpty(List<String> list) {
...
             throws RecoverableJarValidationException, ParseException, CredentialParseException {
         try {
+            var userId = clientOAuthSessionItem.getUserId();
             var inheritedIdentityVc =
                     validateHmrcInheritedIdentity(userId, inheritedIdentityJwtClaim);
             sendInheritedIdentityReceivedAuditEvent(
                     inheritedIdentityVc, auditEventUser, deviceInformation);
-            if (!isHmrcInheritedIdentityWithStrongerVotPresent(inheritedIdentityVc, userId)) {
-                storeHmrcInheritedIdentity(claimsSet, ipvSessionItem, inheritedIdentityVc);
+            if (configService.enabled(EVCS_READ_ENABLED)
```
Joe-Edwards-GDS: (Do we need EVCS read enabled here? In practice it won't matter - since we expect inherited ids to come significantly later)

Joe-Edwards-GDS (commented):
Looks like a sensible starting point for inherited identities in EVCS, thanks.

I expect there'll be some rough edges to tidy up when we revisit them 'properly', but that's expected.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: Ooo interesting. It would be good to have a proper understanding of what is an acceptable wait between writing and reading. Happy to park it though until we're closer to inherited identities going live

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: No, I think I was just trying to be consistent - but we don't trace all the private methods in here anyway. Will take it out

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: It's an unexpected situation - and how would you recover from it? Which inherited identity VC from EVCS would you choose to use?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/service/EvcsService.java
Wynndow: Probably a good idea

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: This took me a lot longer than I would have liked to find. I find the logic around EVCS and fetching VCs in check-existing-identity hard to grok!

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Wynndow: I think I was just being overly cautious. I did contemplate not including any evcs flags here at all and just letting the inherited identity flag do all the work, if we're sure we're not going to use inherited identities before evcs. I'll drop the read check.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: It's all a little bit fiddly...

Should be simpler once we get rid of the tactical store code, but that's a long way off still.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Joe-Edwards-GDS: Yeah, I think we want to trace 'operations' that might be useful in diagnosing latency issues, but probably not every private helper method

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Joe-Edwards-GDS: Yeah, I guess it'd be nice to recover gracefully from it if we can.

In any case we should certainly log a warning, but it feels as though there are two reasonable behaviours afterwards:
- Just handle the first one and ignore any others (if the others are stronger they'll pop up in the matching algorithm if relevant anyway)
- Evaluate whether the new id is weaker than _all_ the existing inherited ids (and mark all of them as historic if they are superseded)

Wynndow: Yeah, that seems reasonable. I think I'll implement option 2 - if the new identity is equal to or stronger than all of them, I'll make them historic

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
Joe-Edwards-GDS: It's not super precise atm, in 'most' cases it'll be very small, but during their load test it occasionally creeps up to a few seconds.

I think the rule of thumb is that it's probably ok if there's some user interaction between the two, but likely to be a problem if it's all internal processing.

Joe-Edwards-GDS (dismissed):
Looks good, but you've acquired a conflict with your own PR 😬

Joe-Edwards-GDS (approved):

PR #2242, opened 2024-07-31
PYIC-7047: Standardise secret names

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-core-common-infra/pull/1025**

### What changed

Standardise secret names

### Why did it change

This updates the names of our secrets in secrets manager to be in line with how we name our other parameters.

This has allowed us to drop the specific getApiKeySecret from the config service, as all the api keys are now stored as plaintext rather than in a json blob.

The new secrets have all been added in the higher envs, and there is a separate PR on the config repo to add them in the lower envs. This should not go in before that.

### Issue tracking

- [PYIC-7047](https://govukverify.atlassian.net/browse/PYIC-7047)

[PYIC-7047]: https://govukverify.atlassian.net/browse/PYIC-7047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/pre-merge-checks.yml
```diff
@@ -146,8 +146,8 @@ jobs:
       - name: Setup app secrets
         run: |
           cp ../local-running/core.local.secrets.template.yaml ../local-running/core.local.secrets.yaml && \
-          sed -i s/CIMIT_API_KEY/${{ secrets.CIMIT_API_KEY }}/ ../local-running/core.local.secrets.yaml && \
-          sed -i s/EVCS_API_KEY/${{ secrets.EVCS_API_KEY }}/ ../local-running/core.local.secrets.yaml
+          sed -i s/CIMIT_API_KEY/${{ secrets.API_KEY_CIMIT }}/ ../local-running/core.local.secrets.yaml && \
+          sed -i s/EVCS_API_KEY/${{ secrets.API_KEY_EVCS }}/ ../local-running/core.local.secrets.yaml
```
Joe-Edwards-GDS, line 150: is there a reason for this change?

Obviously you'll need to update the secrets in GitHub config... Note that you'll also need to update them in the 'dependabot' config as well

local-running/core.local.secrets.template.yaml
```diff
@@ -7,22 +7,22 @@ core:
...
       {"ci":"breaching","detectedScore":20,"checkedScore":-20,"returnCode":"breaching"},
       {"ci":"breaching-p1-only","detectedScore":7,"checkedScore":-7,"returnCode":"breaching-p1-only"},
       {"ci":"needs-enhanced-verification","detectedScore":20,"checkedScore":-20,"returnCode":"needs-enhanced-verification"},
       {"ci":"needs-alternate-doc","detectedScore":20,"checkedScore":-20,"returnCode":"needs-alternate-doc"},
       {"ci":"always-required","detectedScore":1,"checkedScore":-1,"returnCode":"always-required"}
     ]'
-    gov-uk-notify:
-      api-key: "NOTIFY_API_KEY"
+  govUkNotify:
+    apiKey: "NOTIFY_API_KEY" # pragma: allowlist secret
```
Joe-Edwards-GDS, line 19: Funny that `apiKey` is bad but `api-key` is fine 🙃

Joe-Edwards-GDS (approved):
Love a good tidy up

.github/workflows/pre-merge-checks.yml
Wynndow: Yeah, if I wanted the pre-merge checks to pass for this PR, and not break pre-merge checks for other PRs I needed to add a new secret - the value is just plaintext, not json format. TBF, I think I could have gotten away with just updating the EVCS token, but then the format would have been out of sync.

Have added the new secrets to the dependabot config - not sure why they're needed though.

local-running/core.local.secrets.template.yaml
Wynndow: so weird

Wynndow:
Cross those fingers...

PR #2243, opened 2024-07-31
PYIC-7048: read bearer token ttl from config

sam803:
### What changed

Read bearer token ttl from the config rather than an env var

### Why did it change

To be consistent with other related config variables

### Issue tracking

- https://govukverify.atlassian.net/browse/PYIC-7048

### Other considerations

- [X] Update [README](./blob/main/README.md) with any new instructions or tasks

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -31,14 +31,10 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
-
-import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.BEARER_TOKEN_TTL;
 
 public abstract class ConfigService {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final Logger LOGGER = LogManager.getLogger();
-    private static final long DEFAULT_BEARER_TOKEN_TTL_IN_SECS = 3600L;
 
     @Getter @Setter private static boolean local = false;
 
@@ -72,6 +68,11 @@ public abstract class ConfigService {
         return Boolean.parseBoolean(getParameter(configurationVariable, pathProperties));
     }
 
+    public long getLongParameter(
+            ConfigurationVariable configurationVariable, String... pathProperties) {
+        return Long.parseLong(getParameter(configurationVariable, pathProperties));
+    }
+
     public List<String> getStringListParameter(
             ConfigurationVariable configurationVariable, String... pathProperties) {
         return Arrays.asList(getParameter(configurationVariable, pathProperties).split(","));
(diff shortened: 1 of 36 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -114,11 +112,14 @@ private String formatPath(String path, String... pathProperties) {
...
-                .map(Long::valueOf)
-                .orElse(DEFAULT_BEARER_TOKEN_TTL_IN_SECS);
+        try {
+            return Long.parseLong(getParameter(ConfigurationVariable.BEARER_TOKEN_TTL));
+        } catch (ConfigParameterNotFoundException | NumberFormatException e) {
+            LOGGER.warn(
+                    LogHelper.buildErrorMessage("Failed to get bearer token ttl from config", e));
+            return DEFAULT_BEARER_TOKEN_TTL_IN_SECS;
+        }
     }
```
Joe-Edwards-GDS: I think I'd prefer removing this method entirely, and either doing `Long.parseLong(getParameter(...))` where it's used, or adding a more generic `getLongParameter(..)` which does the parsing for you.

As a general rule, I think we should avoid adding specific methods for each bit of config (or we should add specific methods for _all_ config values, but that's very tedious).

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -114,11 +112,14 @@ private String formatPath(String path, String... pathProperties) {
...
     public long getBearerAccessTokenTtl() {
-        return Optional.ofNullable(getEnvironmentVariable(BEARER_TOKEN_TTL))
-                .map(Long::valueOf)
-                .orElse(DEFAULT_BEARER_TOKEN_TTL_IN_SECS);
+        try {
+            return Long.parseLong(getParameter(ConfigurationVariable.BEARER_TOKEN_TTL));
+        } catch (ConfigParameterNotFoundException | NumberFormatException e) {
+            LOGGER.warn(
+                    LogHelper.buildErrorMessage("Failed to get bearer token ttl from config", e));
+            return DEFAULT_BEARER_TOKEN_TTL_IN_SECS;
```
Joe-Edwards-GDS: Do we actually need this default any more? Can we just add this as the configured value in all environments

sam803: Yeh I guess I was being a bit cautious but have added them here https://github.com/govuk-one-login/ipv-core-common-infra/pull/1026 , will remove

Joe-Edwards-GDS (dismissed):
Thanks - didn't realise it'd be so noisy but looks good :)

Obviously don't merge before the config itself has been deployed!

Joe-Edwards-GDS (approved):
Nice - thanks for updating the other ones as well!

Joe-Edwards-GDS (changes requested):
Ok looks like you need to add the config to the local running params as well, but hopefully obvious how to do that...

local-running/core.local.params.yaml
```diff
@@ -1,4 +1,4 @@
-# Local running config params
+œ# Local running config params
```
Joe-Edwards-GDS: 👀

sam803: thanks, not sure where that came from :)

Joe-Edwards-GDS (approved):

PR #2249, opened 2024-08-01
PYIC-7135 Create new lambda, to determine makeup of identities in Tactical Store

rpayal:
## Proposed changes
- Create and config setup for the required new lambda
- Scan read dynamoDb table "user-issued-credentials-v2-${Environment}" in batches
- Apply business logic to user and all its VC’s
  userID, identity (P0/P1/P2/PCL200/PCL250), constituteCriDocumentType
   e.g urn:uuid:9a60d8a0-d0b8-4366-9057-8341fab0f04f, P2, [drivingLicence, address, fraud, kbv]
-Store above create result.

### What changed
- Created new lambda
- Business logic to check user identity and its constituent

### Issue tracking

- [PYIC-7135](https://govukverify.atlassian.net/browse/PYIC-7135)

### Other considerations

[PYIC-7135]: https://govukverify.atlassian.net/browse/PYIC-7135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStore.java
```diff
@@ -137,6 +147,25 @@ public List<T> getItems(String partitionValue) {
                 .toList();
     }
 
+    @Override
+    public List<T> getItems() {
```
Joe-Edwards-GDS: We will need to optimise this quite a bit more - as written this would collect and return the entire VC store into a single object (which is several gigabytes)!

The first major optimisation is to limit the data to only the attributes required - in the initial pass to collect the user ids we _only_ need to read `userId`, not the entire VC body.

The second optimisation is to process the pages in the response one-by-one, rather than trying to fetch them all up-front. We should also look at logging the `LastEvaluatedKey` so that we can resume the scan in a subsequent lambda invocation, in case it takes too long or encounters an error.

This will require exposing the DynamoDB implementation outside of the `DataStore`, so I'd also advise removing these methods from the `DataStore` interface and explicitly constructing a `DynamoDataStore` in the report handler. (Or avoid `DataStore` entirely and handle it separately).

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,196 @@
...
+                            .toList();
+            List<ReportUserIdentityItem> reportUserIdentities =
+                    userIds.stream()
+                            .map(
+                                    usrId ->
+                                            new ReportUserIdentityItem(
+                                                    usrId,
+                                                    Vot.P0.name(),
+                                                    Collections.emptyList(),
+                                                    false))
```
Joe-Edwards-GDS: We should initialise the record with `null` in the fields we haven't computed yet - otherwise it's hard to distinguish from the actual values

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,196 @@
...
+                            .toList();
+            reportUserIdentityDataStore.createOrUpdate(reportUserIdentities);
+
+            processUsersToFindLOCAndUpdateDb(userIds);
```
Joe-Edwards-GDS: (as mentioned, we probably want to do this as a separate step)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,196 @@
...
+        List<ReportUserIdentityItem> reportUserIdentityItems = new ArrayList<>();
+        for (String userId : userIds) {
+            var tacticalVcs = verifiableCredentialService.getVcs(userId);
+            if (!userIdentityService.areVcsCorrelated(tacticalVcs)) {
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                String.format("User (%s) VCs not correlated.", userId)));
+                continue;
+            }
+            boolean migrated = tacticalVcs.stream().allMatch(vc -> vc.getMigrated() != null);
```
Joe-Edwards-GDS: We should probably log a warning here if they don't all match - it would indicate an issue with some of our assumptions.

The only case I can foresee is anyone who completed a F2F journey at the PO but did _not_ return to IPV Core - their F2F VC could have been 'migrated' (as PENDING_RETURN), but the rest weren't. Should be rare/nonexistent though!

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,196 @@
...
+
+            reportUserIdentityItems.add(
+                    new ReportUserIdentityItem(
+                            userId,
+                            votAttained.orElse(Vot.P0).name(),
+                            reportUserIdentityService.getIdentityConstituent(tacticalVcs),
+                            migrated));
+        }
+        LOGGER.info(LogHelper.buildLogMessage("Updating processed user's identity."));
+        reportUserIdentityDataStore.createOrUpdate(reportUserIdentityItems);
```
Joe-Edwards-GDS: Note that this `createOrUpdate` has the same semantics as the other `create` methods - it will _overwrite_ existing records, not update them

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,196 @@
...
+                        .getItems(ATTR_NAME_TO_SUMMARISE_ON, Vot.PCL200.name())
+                        .size();
+        long totalP1 =
+                reportUserIdentityDataStore
+                        .getItems(ATTR_NAME_TO_SUMMARISE_ON, Vot.P1.name())
+                        .size();
+        long totalP0 =
+                reportUserIdentityDataStore
+                        .getItems(ATTR_NAME_TO_SUMMARISE_ON, Vot.P0.name())
+                        .size();
```
Joe-Edwards-GDS: As with the first scan, this is very inefficient - this is doing 5 full table scans which each return a complete list of records.

Instead, I think we need to:
- Perform a single table scan, returning all relevant details
- Aggregate each page before fetching the next

That way we can calculate all the stats in one scan, and without needing to hold the full dataset in memory.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/item/ReportUserIdentityItem.java
```diff
@@ -0,0 +1,37 @@
...
+@ExcludeFromGeneratedCoverageReport
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
+public class ReportUserIdentityItem implements PersistenceItem {
+    private String userId;
+    private String identity;
+
+    @JsonProperty("constitute")
+    private List<String> constituteCriDocumentType;
```
Joe-Edwards-GDS: Should this be something like `constituentVcs` rather than `constitute`?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/item/ReportUserIdentityItem.java
Joe-Edwards-GDS, line 1: A couple of additional asks from the data architects:
- Can we make the partition key a hash of the user id, to allow us to use it in logs more freely. We'll still need to store the plaintext user id in a separate entry
- Can we add `vcCount` (or similar) to indicate the number of VCs held by the user (without needing to count the constituent VCs field)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+        return false;
+    }
+
+    public List<String> getIdentityConstituent(List<VerifiableCredential> tacticalVcs) {
+        return tacticalVcs.stream()
+                .map(
+                        vc -> {
+                            if (vc.getCri().equals(Cri.DCMAW)) {
```
Joe-Edwards-GDS: Does this apply to F2F as well?

Joe-Edwards-GDS (commented):
Thanks - this looks like the right approach, but we'll need a few changes before it's suitable for prod

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStore.java
rpayal: Yeah was working on that...implemented the changes.
Also added request object to give option to restart. The LastEvaluatedKey is provided in report result and also in logs.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: added a log statement.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Yeah the item get added if table doesn't contain item with primary key. 
If the table contains an item with the same primary key, it will be replaced with this item.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: reduce it to one paginated call to gather all users identity for aggregation.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/item/ReportUserIdentityItem.java
rpayal: done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/item/ReportUserIdentityItem.java
rpayal: done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
rpayal: missed earlier...added.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -88,28 +94,65 @@ public ReportUserIdentityHandler() {
...
-            reportUserIdentityDataStore.createOrUpdate(reportUserIdentities);
+            ReportProcessingRequest reportProcessingRequest =
+                    objectMapper.readValue(inputStream, ReportProcessingRequest.class);
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Request received-" + reportProcessingRequest.toString()));
+            List<String> userIds = new ArrayList<>();
+            if ((!reportProcessingRequest.continueUserScan()
+                            && (reportProcessingRequest.tacticalStoreLastEvaluatedKey() == null))
+                    || reportProcessingRequest.continueUserScan()) {
```
Joe-Edwards-GDS: Can be simplified to `reportProcessingRequest.continueUserScan() || reportProcessingRequest.tacticalStoreLastEvaluatedKey() == null`

But why wouldn't we supply `continueUserScan: true` even if there is no continuation key?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -88,28 +94,65 @@ public ReportUserIdentityHandler() {
...
+                    List<ReportUserIdentityItem> reportUserIdentities =
+                            userIds.stream()
+                                    .map(
+                                            usrId ->
+                                                    new ReportUserIdentityItem(
+                                                            usrId,
+                                                            null,
+                                                            0,
+                                                            Collections.emptyList(),
+                                                            false))
```
Joe-Edwards-GDS: Can we put `null` in all of these fields - otherwise it'll happily aggregate invalid rows

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -88,28 +94,65 @@ public ReportUserIdentityHandler() {
...
-                                                    Collections.emptyList(),
-                                                    false))
-                            .toList();
-            reportUserIdentityDataStore.createOrUpdate(reportUserIdentities);
+            ReportProcessingRequest reportProcessingRequest =
+                    objectMapper.readValue(inputStream, ReportProcessingRequest.class);
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Request received-" + reportProcessingRequest.toString()));
+            List<String> userIds = new ArrayList<>();
```
Joe-Edwards-GDS: It looks like you're still attempting to build up a list of all user ids in this list? That's probably tolerable within memory limits, but should be unnecessary - internally we only need to inspect one page of scan results at a time.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -88,28 +94,65 @@ public ReportUserIdentityHandler() {
...
+                                                            0,
+                                                            Collections.emptyList(),
+                                                            false))
+                                    .toList();
+                    reportUserIdentityDataStore.createOrUpdate(reportUserIdentities);
+                    if (tableScanResult.lastEvaluatedKey() != null) {
+                        exclusiveStartKey = tableScanResult.lastEvaluatedKey();
+                        reportProcessingResult.tacticalStoreLastEvaluatedKey(exclusiveStartKey);
+                    }
+                } while (exclusiveStartKey != null);
```
Joe-Edwards-GDS: How fast does this scan execute? We might need a mechanism that allows us to stop processing if the lambda is about to time out, rather than powering through everything.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -88,28 +94,65 @@ public ReportUserIdentityHandler() {
...
+                                            usrId ->
+                                                    new ReportUserIdentityItem(
+                                                            usrId,
+                                                            null,
+                                                            0,
+                                                            Collections.emptyList(),
+                                                            false))
+                                    .toList();
+                    reportUserIdentityDataStore.createOrUpdate(reportUserIdentities);
+                    if (tableScanResult.lastEvaluatedKey() != null) {
```
Joe-Edwards-GDS: How does this loop terminate? We execute until `exclusiveStartKey` is null, but only update it to non-null values?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -141,56 +184,80 @@ private void processUsersToFindLOCAndUpdateDb(List<String> userIds)
                                 String.format("User (%s) VCs not correlated.", userId)));
                 continue;
             }
-            boolean migrated = tacticalVcs.stream().allMatch(vc -> vc.getMigrated() != null);
+            boolean anyVCsMigrated = tacticalVcs.stream().anyMatch(vc -> vc.getMigrated() != null);
+            boolean allVCsMigrated = tacticalVcs.stream().allMatch(vc -> vc.getMigrated() != null);
+
+            if (anyVCsMigrated && !allVCsMigrated) {
+                LOGGER.info(
```
Joe-Edwards-GDS, line 189: can be a warning log?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -141,56 +184,80 @@ private void processUsersToFindLOCAndUpdateDb(List<String> userIds)
...
 
-    private ReportProcessingResult buildReportProcessingResult() {
+    private ReportProcessingResult.ReportProcessingResultBuilder buildReportProcessingResult(
+            ReportProcessingRequest reportProcessingRequest,
+            ReportProcessingResult.ReportProcessingResultBuilder reportProcessingResult) {
         LOGGER.info(LogHelper.buildLogMessage("Building report processing summary result."));
+        Map<String, AttributeValue> exclusiveStartKey = null;
+        if (reportProcessingRequest.userIdentitylastEvaluatedKey() != null) {
+            exclusiveStartKey = reportProcessingRequest.userIdentitylastEvaluatedKey();
+        }
```
Joe-Edwards-GDS, line 301: Can this just be `var exclusiveStartKey = reportProcessingRequest.userIdentitylastEvaluatedKey()`?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -141,56 +184,80 @@ private void processUsersToFindLOCAndUpdateDb(List<String> userIds)
...
+        List<ReportUserIdentityItem> userIdentities = new ArrayList<>();
```
Joe-Edwards-GDS: This is still aggregating the entire table into a single object. We need to process the aggregate metrics on a page-by-page basis, something like:
```
var totalP2 = 0L;
var totalP1 = 0L;
// etc.
do {
    var tableScanResult =
            reportUserIdentityDataStore.getItems(exclusiveStartKey, "identity", "migrated");
    for (var item : tableScanResult.items()) {
        switch (item.getIdentity()) {
            case "P2" -> totalP2++;
            case "P1" -> totalP1++;
        }
        // etc.
    }
    exclusiveStartKey = tableScanResult.lastEvaluatedKey();
} while (exclusiveStartKey != null);
```
That way each scan result can be garbage-collected after it's processed, and we only store the aggregate statistics at the end.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/persistence/DataStore.java
Joe-Edwards-GDS, line 1: I think we can still leverage some of the existing `DynamoDataStore`...

We should either put the new methods in the existing `DynamoDataStore`, or make this an `ScanningDynamoDataStore extends DynamoDataStore` (or something similar).

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Continue till we get null for the lastEvaluatedKey from scan.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Done ...but it would've worked as identity value was null (will not have picked that record on aggregation).

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: yeah missed that....updated the logic for processUsersToFindLOCAndUpdateDb() to pick unique users paginated scan from new table.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: done..will appear in next commit :)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Handle it in processUsersToFindLOCAndUpdateDb()

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Changed to just
 if (reportProcessingRequest.continueUniqueUserScan())

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+                                .map(
+                                        usrId ->
+                                                new ReportUserIdentityItem(
+                                                        usrId, null, null, null, null))
+                                .toList();
+                reportUserIdentityScanDynamoDataStore.createOrUpdate(reportUserIdentities);
+
+                checkLambdaRemainingExecutionTime(context);
+
+                if (tableScanResult.lastEvaluatedKey() != null) {
```
Joe-Edwards-GDS: Have you tested this with multiple pages? I don't think you want this condition, otherwise the loop will never terminate...

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+                List<ReportUserIdentityItem> reportUserIdentityItems = new ArrayList<>();
+                for (String userId : userIds) {
+                    var tacticalVcs = verifiableCredentialService.getVcs(userId);
+                    if (!userIdentityService.areVcsCorrelated(tacticalVcs)) {
+                        LOGGER.info(
+                                LogHelper.buildLogMessage(
+                                        String.format(
+                                                "User (%s) VCs not correlated.",
+                                                ReportUserIdentityItem.getUserHash(userId))));
+                        continue;
```
Joe-Edwards-GDS: Should we `continue` here? We still want to record the stats for these users, they just don't have an identity.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+        return reportProcessingResult.summary(
+                new ReportSummary(
+                        reportSummaryItem.getTotalP2(),
+                        reportSummaryItem.getTotalP2Migrated(),
+                        reportSummaryItem.getTotalP1(),
+                        reportSummaryItem.getTotalP0()));
+    }
+
+    private void checkLambdaRemainingExecutionTime(Context context)
```
Joe-Edwards-GDS, line 314: Nice - we might want to tune the threshold when we start using big datasets (1s is pretty tight) but that'll be easy to do.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Added a check for lambda timeout before starting a next pagination scan and gracefully exit if time left is less then 1sec.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Request received- " + reportProcessingRequest.toString()));
+            // Step-1
+            scanToExtractUniqueUserIdFromTacticalStore(
+                    reportProcessingResult, reportProcessingRequest, context);
+            // Step-2
+            processUsersToFindLOCAndUpdateDb(
+                    reportProcessingRequest, reportProcessingResult, context);
+            // Step-3
```
Joe-Edwards-GDS, line 117: I was thinking the three steps would be:

1. Scan VC table to collect user ids (as you've done)
    - Medium speed - needs to scan all 2.5M VCs and update the other table
2. Scan user table and update with identity details
    - Slow speed - needs to fetch all VCs for each user in turn, calculate identity and store the result
3. Scan user table and aggregate identity details
    - Fast(ish) speed - needs to scan 0.5-1M user records and aggregate the data

Perhaps it's worth revising the input class to reflect these three steps, so you can choose which to execute?

At the moment I think we've also pushed some of 3 into 2 - with the extra report table. I'm not sure if this is useful, it slightly depends how quick step 3 is with another full table scan.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,335 @@
...
+                var userIds =
+                        tableScanResult.items().stream()
+                                .map(ReportUserIdentityItem::getUserId)
+                                .toList();
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "Checking user identity for total (%s) users.",
+                                        userIds.size())));
+                List<ReportUserIdentityItem> reportUserIdentityItems = new ArrayList<>();
```
Joe-Edwards-GDS: This is still trying to store a record for every user in the table, which is going to be huge - do we actually need this for anything?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: The loop will end when there the scan result null for lastEvaluatedKey.

For table scan - If there is not a LastEvaluatedKey in the result, then there are no more items to be retrieved.

Joe-Edwards-GDS: Maybe I'm missing something, but if `tableScanResult.lastEvaluatedKey() = null` then we won't set `exclusiveStartKey` in this block, so `exclusiveStartKey` will stay non-null and the loop will continue?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Done - updated those users with P0

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: you got it right....Sorry was not looking properly.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: With new changes it'll process unique users in one table scan page.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Step-2 we are aggregating result and pesisting it to result table which is used in step-3 to avoid full table scan of all unique users frm report-user-identity table.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DataStore.java
```diff
@@ -28,6 +29,9 @@ static <T extends PersistenceItem> DataStore<T> create(
 
     void createIfNotExists(T item) throws ItemAlreadyExistsException;
 
+    @ExcludeFromGeneratedCoverageReport
```
Joe-Edwards-GDS: Do we need to annotate the interface method? There's no implementation to cover

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStore.java
```diff
@@ -174,23 +188,49 @@ public T delete(String partitionValue, String sortValue) {
...
+        processBatchOperation(items, true);
+    }
+
+    @Override
+    public void deleteAllByPartition(String partitionValue) throws BatchDeleteException {
+        delete(getItems(partitionValue));
+    }
+
+    @ExcludeFromGeneratedCoverageReport
+    private void processBatchOperation(List<T> items, boolean isForDeletion)
```
Joe-Edwards-GDS, line 200: I think it would be more natural for this to take a `WriteBatch` argument, rather than the boolean, but it's not vital.

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/persistence/InMemoryDataStoreTest.java
```diff
@@ -17,7 +17,7 @@
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
-class InMemoryStoreTest {
+class InMemoryDataStoreTest {
```
Joe-Edwards-GDS, line 20: (thanks!)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
```diff
@@ -0,0 +1,129 @@
...
+    public Optional<Vot> getStrongestAttainedVotForCredentials(List<VerifiableCredential> vcs)
+            throws ParseException {
+        for (Vot votToCheck : SUPPORTED_VOTS_BY_DESCENDING_STRENGTH) {
+            boolean requestedVotAttained;
+            if (votToCheck.getProfileType().equals(GPG45)) {
+                requestedVotAttained =
+                        achievedWithGpg45Profile(
+                                votToCheck, VcHelper.filterVCBasedOnProfileType(vcs, GPG45));
+            } else {
+                requestedVotAttained = hasOperationalProfileVc(votToCheck, vcs);
```
Joe-Edwards-GDS: We can probably just skip this check - there won't be any operational profiles in prod, so it'll be a waste of effort (leave a comment if you want to warn that this won't capture operational profiles in future).

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            if (requestedVotAttained) {
+                return Optional.of(votToCheck);
+            }
+        }
+        return Optional.empty();
+    }
+
+    public List<String> getIdentityConstituent(List<VerifiableCredential> tacticalVcs) {
```
Joe-Edwards-GDS, line 55: (Note this is very similar to `Gpg45ProfileEvaluator#getVcDocumentType`, but it's fine to leave them separate)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
```diff
@@ -0,0 +1,129 @@
...
+                                    }
+                                }
+                            }
+                            return cri.getId();
+                        })
+                .toList();
+    }
+
+    private boolean achievedWithGpg45Profile(Vot votToCheck, List<VerifiableCredential> vcs) {
+        Gpg45Scores gpg45Scores = gpg45ProfileEvaluator.buildScore(vcs);
```
Joe-Edwards-GDS: You could pull this up a level and then you don't need to re-calculate it for P1 (if P2 fails)?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+        LOGGER.info(LogHelper.buildLogMessage("Start processing report."));
+
+        ReportProcessingResult.ReportProcessingResultBuilder reportProcessingResult =
+                ReportProcessingResult.builder();
+        try {
+            ReportProcessingRequest reportProcessingRequest =
+                    objectMapper.readValue(inputStream, ReportProcessingRequest.class);
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Request received- " + reportProcessingRequest.toString()));
```
Joe-Edwards-GDS: Will this request contain any PII (i.e. user ids) in the continuation token? I think we can probably just drop this log tbh.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+                    LogHelper.buildLogMessage(
+                            "Processing unique users to update their aggregate identities."));
+
+            List<ReportUserIdentityItem> totalReportUserIdentityItems = new ArrayList<>();
+            for (var page :
+                    reportUserIdentityScanDynamoDataStore.getItems(
+                            reportProcessingRequest.userIdentitylastEvaluatedKey(),
+                            ATTR_NAME_USER_ID,
+                            "identity",
+                            "migrated")) {
```
Joe-Edwards-GDS: Why do we need to project `identity` and `migrated` - we don't appear to use them? (And they'll likely be null anyway)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+                var userIds = page.items().stream().map(ReportUserIdentityItem::getUserId).toList();
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "Checking user identity for total (%s) users.",
+                                        userIds.size())));
+                List<ReportUserIdentityItem> reportUserIdentityItems = new ArrayList<>();
+                for (String userId : userIds) {
+                    evaluateUserIdentityDetail(reportUserIdentityItems, userId);
+                }
```
Joe-Edwards-GDS, line 198: Are you confident this will always evaluate within the lambda timeout (1s)? We might need to adjust the timeout to capture this _or_ move the remaining execution time check inside here.

If you're not sure, we can see how it performs in build and adjust from there.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+            }
+            LOGGER.info(LogHelper.buildLogMessage("User identity check scan completed."));
+            reportProcessingResult.userIdentitylastEvaluatedKey(null);
+            reportProcessingResult.users(totalReportUserIdentityItems);
+        }
+    }
+
+    private void evaluateUserIdentityDetail(
+            List<ReportUserIdentityItem> reportUserIdentityItems, String userId)
+            throws CredentialParseException, HttpResponseExceptionWithErrorBody, ParseException {
```
Joe-Edwards-GDS: (Feels like it'd be more natural for this method to return the evaluated identity, rather than append to a list?)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+            ReportProcessingResult.ReportProcessingResultBuilder reportProcessingResult,
+            Context context)
+            throws CredentialParseException, HttpResponseExceptionWithErrorBody, ParseException,
+                    BatchDeleteException, StopBeforeLambdaTimeoutException {
+        if (reportProcessingRequest.continueUserIdentityScan()) {
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Processing unique users to update their aggregate identities."));
+
+            List<ReportUserIdentityItem> totalReportUserIdentityItems = new ArrayList<>();
```
Joe-Edwards-GDS: This is still trying to collect a list of all user identities in-memory - can we remove this.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+                            ATTR_NAME_USER_ID,
+                            "identity",
+                            "migrated")) {
+                var userIds = page.items().stream().map(ReportUserIdentityItem::getUserId).toList();
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "Checking user identity for total (%s) users.",
+                                        userIds.size())));
+                List<ReportUserIdentityItem> reportUserIdentityItems = new ArrayList<>();
```
Joe-Edwards-GDS: We can and should initialise this to the correct size (i.e. `userIds.size()`) - resizing the array as we add new items is pretty inefficient.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportProcessingResult.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@ExcludeFromGeneratedCoverageReport
+public record ReportProcessingResult(
+        ReportSummary summary,
+        List<ReportUserIdentityItem> users,
```
Joe-Edwards-GDS: I don't think there's any value in trying to return individual user info here - it'll either be too small to be of use, or too big to handle appropriately

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+    private void evaluateUserIdentityDetail(
+            List<ReportUserIdentityItem> reportUserIdentityItems, String userId)
+            throws CredentialParseException, HttpResponseExceptionWithErrorBody, ParseException {
+        var tacticalVcs = verifiableCredentialService.getVcs(userId);
+        if (!userIdentityService.areVcsCorrelated(tacticalVcs)) {
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            String.format(
+                                    "User (%s) VCs not correlated.",
+                                    ReportUserIdentityItem.getUserHash(userId))));
```
Joe-Edwards-GDS: It'll be easier to process these logs if the message is fixed and the user hash is included as a separate field (e.g. `.with("user_hash", ...)`)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+        List<ReportUserIdentityItem> totalP2Identities =
+                items.stream().filter(ui -> Vot.P2.name().equals(ui.getIdentity())).toList();
+        reportSummaryItem.setTotalP2(reportSummaryItem.getTotalP2() + totalP2Identities.size());
+        reportSummaryItem.setTotalP2Migrated(
+                reportSummaryItem.getTotalP2Migrated()
+                        + totalP2Identities.stream()
+                                .filter(item -> (Boolean.TRUE.equals(item.getMigrated())))
+                                .toList()
+                                .size());
```
Joe-Edwards-GDS: `.toList().size()` => `.count()`

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -0,0 +1,328 @@
...
+                                .filter(ui -> Vot.P1.name().equals(ui.getIdentity()))
+                                .toList()
+                                .size());
+        reportSummaryItem.setTotalP0(
+                reportSummaryItem.getTotalP0()
+                        + items.stream()
+                                .filter(ui -> Vot.P0.name().equals(ui.getIdentity()))
+                                .toList()
+                                .size());
+        reportSummaryScanDynamoDataStore.update(reportSummaryItem);
```
Joe-Edwards-GDS: One thing I'm concerned about here is whether we'll get any race-conditions due to eventual consistency. Is it possible that we will update the count and then the next page will try and update the count before the previous one has finished?

The best option is probably to turn this update into an update expression, which would do something like:
`SET total_p0 = total_p0 + :newp0` (etc.)

(Or go back to aggregating the user report table in a separate scan)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportSummary.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Builder;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@Builder
+@ExcludeFromGeneratedCoverageReport
+public record ReportSummary(
+        @JsonProperty("Total P2") long totalP2Identities,
+        @JsonProperty("Total P2 migrated") long totalP2IdentitiesMigrated,
+        @JsonProperty("Total P1") long totalP1Identities,
+        @JsonProperty("Total P0") long totalP0Identities) {}
```
Joe-Edwards-GDS: Is it easy to include the counts by other info too - i.e. the consitituents?

Perhaps we should test it in build first, and can iterate on the output later

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
```diff
@@ -0,0 +1,129 @@
...
+                                    } else if (isNotEmpty(person.getResidencePermit())) {
+                                        return cri.getId() + "-residencePermit";
+                                    } else if (isNotEmpty(person.getSocialSecurityRecord())) {
+                                        return cri.getId() + "-socialSecurityRecord";
+                                    }
+                                }
+                            }
+                            return cri.getId();
+                        })
+                .toList();
```
Joe-Edwards-GDS, line 81: It might be worth sorting these, unless you're confident that they will already be? i.e. so we don't end up mixing up `dcmaw,address,fraud` and `address,dcmaw,fraud`.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/persistence/item/ReportUserIdentityItem.java
```diff
@@ -0,0 +1,53 @@
...
+        this.vcCount = vcCount;
+        this.constituentVcs = constituentVcs;
+        this.migrated = migrated;
+    }
+
+    private String hashUserId;
+    @JsonIgnore private String userId;
+    private String identity;
+    private Integer vcCount;
+    private List<String> constituentVcs;
```
Joe-Edwards-GDS: This can probably just be a comma-separated string - it'll be a lot more compact in the table (and more user-readable as well)

Joe-Edwards-GDS (commented):
I think the core structure is looking good now, but a bunch of small improvements

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/persistence/item/ReportUserIdentityItem.java
rpayal: Done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
rpayal: Done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportSummary.java
rpayal: Not sure we needed that.  Do u mean total VC's involved to create those total P2 indenties.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
rpayal: Cleaned up.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Another scan will be same additional scan as in step-2. 

Also with current approach there is chance of corrupt summary data...... i.e if user run the the step-2 again with same lastEvaluatedKey then while doing step-2 we'll currupt summary data.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportSummary.java
Joe-Edwards-GDS: something like:
```
{
  "dcmaw-passport,address,fraud": 123,
  "dcmaw-drivingPermit,address,fraud": 456,
  "ukPassport,address,fraud,kbv": 789,
   ...
}
```

Joe-Edwards-GDS: This will be harder if you store it in a record, and not just aggregate in a single scan - we may be able to work that out later.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
Joe-Edwards-GDS: What's your suggestion?

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/persistence/DataStore.java
rpayal: Done-created local ScanningDynamoDataStore to this lambda which extends DynamoDataStore.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportSummary.java
rpayal: Done req. change to implement this.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: if table scan is quick then we can separate the step-3 from step-2.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStore.java
rpayal: Will be little complex to implement as in processBatchOperation we are creating writeBatch for 25 items in one go.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/service/ReportUserIdentityService.java
rpayal: Done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: Not sure, will need to observe in build. Changed timeout to (1min)

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: done

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/domain/ReportProcessingResult.java
rpayal: Removed

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
rpayal: done

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStore.java
Joe-Edwards-GDS: ok sure

Joe-Edwards-GDS (dismissed):
Assuming the API tests are about to pass, and you just got caught up in the stub breakage

Joe-Edwards-GDS (approved):

PR #2263, opened 2024-08-05
PYIC-7058: dedupe vcs by document type in EvaluateGpg45ScoresHandler

sam803:
### What changed

De-duplicate vcs by document type when evaluating gpg45 scores
### Why did it change
Currently it is possible through an edge-case journey to obtain an M2B profile using the same document for two different CRIs 

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7058


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -276,4 +285,38 @@ private AuditEvent buildProfileMatchedAuditEvent(
...
+    private List<VerifiableCredential> deduplicateVcsByDocumentType(
+            List<VerifiableCredential> vcs) {
+        var deduplicatedVcs = new HashMap<String, VerifiableCredential>();
+        for (var vc : vcs) {
+            deduplicatedVcs.putIfAbsent(getVcDocumentType(vc), vc);
+        }
+        return new ArrayList<>(deduplicatedVcs.values());
+    }
+
+    private String getVcDocumentType(VerifiableCredential vc) {
```
Joe-Edwards-GDS: I think we can change this to use the VC classes, and avoid messing about with JSON mappers, e.g.:
```
private String getVcDocumentType(VerifiableCredential vc) {
    if (vc.getCredential().getCredentialSubject() instanceof IdentityCheckSubject subject) {
        if (!isNullOrEmpty(subject.getDrivingPermit())) {
            return "drivingPermit";
        }
        if (!isNullOrEmpty(subject.getPassport())) {
            return "passport";
        }
        // etc.
    }
    return null;
}
```
Note that there are quite a few more docs than just passport and DL - bank account, NINO, BRP, ID card. The methods on the class should help you :)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -276,4 +285,38 @@ private AuditEvent buildProfileMatchedAuditEvent(
                         VcHelper.extractTxnIdsFromCredentials(credentials)),
                 new AuditRestrictedDeviceInformation(deviceInformation));
     }
+
+    private List<VerifiableCredential> deduplicateVcsByDocumentType(
+            List<VerifiableCredential> vcs) {
+        var deduplicatedVcs = new HashMap<String, VerifiableCredential>();
+        for (var vc : vcs) {
+            deduplicatedVcs.putIfAbsent(getVcDocumentType(vc), vc);
```
Joe-Edwards-GDS: I think we need to be a bit cleverer here - we don't just want to take the first VC of each type, we want to take the 'strongest'.

IMO this logic should probably belong in `Gpg45ProfileEvaluator#getEvidences`, and when collecting the 'evidence'  (aka strength + validity) scores, we should only take the maximum from each type.

This could theoretically be ambiguous, if you had 'strength 2, validity 1' and 'strength 1, validity 2', but in practice this won't happen. For the purposes of our logic, we can take the strongest by strength and then validity.

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -45,9 +49,9 @@ public Optional<Gpg45Profile> getFirstMatchingProfile(
     }
 
     public Gpg45Scores buildScore(List<VerifiableCredential> vcs) {
-
+        var deduplicatedVcs = deduplicateVcsByDocumentType(vcs);
         var identityChecks =
-                vcs.stream()
+                deduplicatedVcs.stream()
```
Joe-Edwards-GDS: This has slightly different semantics, as it'll exclude _all_ scores from common document types, not just the 'evidence' scores.

The DCMAW VC also includes a verification score, which we don't want to lose, and there's also some chance we'll be splitting the DCMAW VC into two, so we'd have:
- one DL VC with strength + validity scores
- one DL VC with a verification score
- (perhaps some other DL VC with a strength + validity score)

In this scenario we only want to allow one 'evidence' item (i.e. strength and validity) but still accept the verification score.

You might have to keep the full `identityCheckCredential` here and then the doc type will be available when building the evidences (and you can push the deduplication there).

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -63,6 +67,91 @@ public Gpg45Scores buildScore(List<VerifiableCredential> vcs) {
...
+            if (isEmpty(docType)) {
+                result.add(vc);
+            } else {
+                var existing = deduplicatedVcs.get(docType);
+                if (existing == null) {
+                    deduplicatedVcs.put(docType, vc);
+                } else {
+                    deduplicatedVcs.put(
+                            docType, selectVerifiableCredentialWithSameDocumentType(existing, vc));
+                }
```
Joe-Edwards-GDS, line 85: If you manage to boil this down to Evidence (rather than VCs), you could streamline this a bit:
```
var existing = deduplicatedEvidence.get(docType);
if (existing == null || vc.compareTo(existing) > 0) {
  deduplicatedEvidence.put(docType, vc);
}
```
where you make `Evidence` implement the `Comparable` interface

Joe-Edwards-GDS (commented):
Sorry this is getting a bit fiddly!

libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
sam803: Ah ok thanks,  will take a look at this now

libs/gpg45-evaluator/src/test/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluatorTest.java
```diff
@@ -76,6 +82,25 @@ void buildScoreShouldReturnCorrectScoreForPassportCredential() throws Exception
         assertEquals(expectedScores, builtScores);
     }
 
+    @Test
+    void buildScoreShouldReturnCorrectScoreForDuplicateDrivingPermitCredential() throws Exception {
```
Joe-Edwards-GDS, line 86: You could consider adding another test case that covers a strength/validity + separate verification for the same doc, but fine as-is

Joe-Edwards-GDS (approved):

PR #2274, opened 2024-08-08
PYIC-7186: Update status codes of /async/credential PACT test interactions

MikeCollingwood:
### What changed

- Update status codes of /async/credential PACT test interactions

### Why did it change

- Mobile team send back different status codes as to what was initially suggested in the PACT tests we have with them

### Issue tracking

- [PYIC-7186](https://govukverify.atlassian.net/browse/PYIC-7186)

[PYIC-7186]: https://govukverify.atlassian.net/browse/PYIC-7186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sandy-may (approved):
@MikeCollingwood 

Just flagging that there isn't an interaction for the `401` status code on the `/credential` endpoint - up to you whether you include it or not.

Example would be that there is no Authorization Header

blakeyp (approved):

PR #2276, opened 2024-08-08
PYIC-7069: repeat fraud check api tests

thebauSoftwire:
### What changed
Add api tests for repeat fraud check

### Why did it change
Increase coverage of api tests to cover key journeys

### Issue tracking

- [PYIC-7069](https://govukverify.atlassian.net/browse/PYIC-7069)

[PYIC-7069]: https://govukverify.atlassian.net/browse/PYIC-7069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/repeat-fraud-check.feature
+++ b/api-tests/features/repeat-fraud-check.feature
@@ -18,9 +18,21 @@ Feature: Repeat fraud check journeys
     Then I get an OAuth response
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
-#    When I return using 'P2' identity
-#    Then I see 'confirm-your-details' with my details
-#    When I select update
-#    Then I see 'update-details' page response
-#    When I select Name change
-#    Then I should be successful in changing my name
+
+    # Repeat fraud check with update name
+    When I start a new 'medium-confidence' journey
+    Then I get a 'confirm-your-details' page response
+    When I submit a 'given-names-only' event
+    Then I get a 'page-update-name' page response
+    When I submit a 'update-name' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-changed-first-name-only-passport-valid' details to the CRI stub
+    Then I get a 'page-dcmaw-success' page response
+    When I submit a 'next' event
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-changed-first-name-only-score-2-with-expiry' details to the CRI stub
+    Then I get a 'page-ipv-success' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
--- a/api-tests/features/repeat-fraud-check.feature
+++ b/api-tests/features/repeat-fraud-check.feature
@@ -1,0 +1,26 @@
+Feature: Repeat fraud check journeys
+
+  @Build
+  Scenario: Fraud 6 Months Expiry + Name Update
+    Given I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
+    Then I get a 'page-dcmaw-success' page response
+    When I submit a 'next' event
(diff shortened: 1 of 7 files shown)


api-tests/src/steps/cri-steps.ts
```diff
@@ -48,12 +50,18 @@ When(
       throw new Error("Last journey engine response was not a CRI response");
     }
 
+    const matchedRegex = /^(.*)-with-expiry$/.exec(scenario);
+    const nbf = matchedRegex === null ? undefined : EXPIRED_NBF;
+    const extractedScenario =
+      matchedRegex === null ? scenario : matchedRegex[1];
```
thebauSoftwire: Wanted to re-use the stub data we already had so used regex to extract them but can get rid of this and just duplicate the data instead if this isn't as readable

api-tests/features/repeat-fraud-check.feature
```diff
@@ -0,0 +1,69 @@
+Feature: Repeat fraud check journeys
+
+  @Build
+  Scenario: Fraud 6 Months Expiry + Name Update
```
sam803: You should be able to use a [Background](https://cucumber.io/docs/gherkin/reference/#background)  to reduce duplication :)

thebauSoftwire: oh nice thanks didn't know about this :)

api-tests/src/steps/cri-steps.ts
```diff
@@ -14,6 +14,8 @@ import {
 } from "../types/internal-api.js";
 import { CriStubRequest } from "../types/cri-stub.js";
 
+const EXPIRED_NBF = 1658829758;
```
Joe-Edwards-GDS: It's often nice to add a comment to values like this to explain them - e.g.
```
const EXPIRED_NBF = 1658829758; // 26/07/2022
```

api-tests/src/steps/cri-steps.ts
Joe-Edwards-GDS: This is ok, but if you use a step for this, you might get better autocomplete and it'll read nicer...

The simplest approach is to pull out the action as a method and define two steps that use it:
```
const submitCriDetails = async (world: World, scenario: string, nbf?: number): Promise<void> => {
  
};

When(
  "I submit {string} details to the CRI stub",
  async function (this: World, scenario: string): Promise<void> {
    await submitCriDetails(this, scenario);
  },
);

When(
  "I submit expired {string} details to the CRI stub",
  async function (this: World, scenario: string): Promise<void> {
    await submitCriDetails(this, scenario, EXPIRED_NBF);
  },
);
```

But a cleverer approach might be to switch to a regex step:
```
When(
  /I submit (expired )?'([\w-]+)' details to the CRI stub/,
  async function (this: World, expired: string, scenario: string): Promise<void> {
    await submitCriDetails(this, scenario, expired ? EXPIRED_NBF : undefined);
  },
);
```
(`expired` will have the value `expired ` if included, or undefined)

api-tests/features/repeat-fraud-check.feature
```diff
@@ -0,0 +1,53 @@
...
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-changed-first-name-only-score-2-with-expiry' details to the CRI stub
+    Then I get a 'page-ipv-success' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+
+  @Build
+  Scenario: Fraud 6 Months Expiry + No Update
```
Joe-Edwards-GDS: A silly thing, but it feels more logical to me to put the simplest case first?

Joe-Edwards-GDS (commented):
This looks good, but a couple of suggestions that might make it neater

api-tests/src/steps/cri-steps.ts
thebauSoftwire: I tried using a regex too but wasn't as nice as this thanks! :)

Joe-Edwards-GDS (approved):
@sam803 there'll probably be some overlap with your PR as well (particularly in the stub data) but hopefully not too hard to resolve

PR #2280, opened 2024-08-09
PYIC-7073: reprove identity test

sam803:
### What changed

Add reprove identity journey to api tests

### Why did it change

To increase test coverage

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7073

### DIFF
--- a/api-tests/src/steps/ipv-steps.ts
+++ b/api-tests/src/steps/ipv-steps.ts
@@ -54,11 +54,11 @@ After(function (this: World, options: ITestCaseHookParameter) {
 });
 
 When(
-  "I start a new {string} journey{}",
+  /I start a new ?'([\w-]+)' journey( with reprove identity)?/,
   async function (
     this: World,
     journeyType: string,
-    extra: string,
+    reproveIdentity: string,
   ): Promise<void> {
     this.userId = this.userId ?? getRandomString(16);
     this.journeyId = getRandomString(16);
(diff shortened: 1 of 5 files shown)


api-tests/src/steps/ipv-steps.ts
```diff
@@ -54,15 +54,20 @@ After(function (this: World, options: ITestCaseHookParameter) {
...
+    reproveIdentity: string,
+  ): Promise<void> {
     this.userId = this.userId ?? getRandomString(16);
     this.journeyId = getRandomString(16);
     this.ipvSessionId = await internalClient.initialiseIpvSession(
       await generateInitialiseIpvSessionBody(
         this.userId,
         this.journeyId,
         journeyType,
+        reproveIdentity ? true : false,
```
Joe-Edwards-GDS, line 70: (`!!reproveIdentity` would be perfectly idiomatic IMO)

api-tests/src/utils/jar-generator.ts
```diff
@@ -29,6 +30,7 @@ export const generateJar = async (
   const payload = {
     ...payloadData,
     ...{
+      reprove_identity: reproveIdentity,
```
Joe-Edwards-GDS, line 33: Note that `false` and `undefined` are not (necessarily) equivalent from a JSON POV - but we should be treating them the same in core

Joe-Edwards-GDS (approved):
Once we've got a few of these, I'm going to consider whether there's a clever way to set up users with an existing identity (without running through a journey), but fine to leave for now - it's nice and explicit.

PR #2288, opened 2024-08-13
PYIC-6914: Add lng parameter to CRI redirectUri

MikeCollingwood:
### What changed

- Add lng parameter to cri redirect uri created in BuildCriOauthRequest
- Add language to step function, and referenced as CriJourneyRequest dto for BuildCriOauthRequest, so it can be passed from the frontend

### Why did it change

- DWP want the language choice of the user to be passed through from the frontend

### Issue tracking

- [PYIC-6914](https://govukverify.atlassian.net/browse/PYIC-6914)

[PYIC-6914]: https://govukverify.atlassian.net/browse/PYIC-6914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
@@ -1,19 +1,21 @@
 package uk.gov.di.ipv.core.library.domain;
 
-import lombok.Builder;
+import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
 @ExcludeFromGeneratedCoverageReport
 @EqualsAndHashCode(callSuper = true)
+@AllArgsConstructor
 @NoArgsConstructor
 @Data
+@SuperBuilder
 public class CriJourneyRequest extends JourneyRequest {
     private String language;
 
-    @Builder()
     public CriJourneyRequest(
             String ipvSessionId,
             String ipAddress,
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
@@ -13,7 +13,7 @@ import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport
 public class CriJourneyRequest extends JourneyRequest {
     private String language;
 
-    @Builder(builderMethodName = "criJourneyRequestBuilder")
+    @Builder()
     public CriJourneyRequest(
             String ipvSessionId,
             String ipAddress,
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
@@ -15,22 +15,4 @@ import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport
 @SuperBuilder
 public class CriJourneyRequest extends JourneyRequest {
     private String language;
-
-    public CriJourneyRequest(
-            String ipvSessionId,
-            String ipAddress,
-            String deviceInformation,
-            String clientOAuthSessionId,
-            String journey,
-            String featureSet,
-            String language) {
-        super(
-                ipvSessionId,
-                ipAddress,
-                deviceInformation,
-                clientOAuthSessionId,
-                journey,
-                featureSet);
-        this.language = language;
-    }
 }
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
@@ -5,23 +5,21 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
-import java.util.Map;
-
 @EqualsAndHashCode(callSuper = true)
 @NoArgsConstructor
 @Data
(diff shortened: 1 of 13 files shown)


local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
```diff
@@ -131,6 +133,18 @@ private JourneyRequest buildJourneyRequest(Context ctx, String journey) {
                 .build();
     }
 
+    private CriJourneyRequest buildFrontendJourneyRequest(Context ctx, String journey) {
```
MikeCollingwood, line 136: rename

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -106,7 +106,8 @@ public enum ErrorResponse {
     UNEXPECTED_CREDENTIAL_TYPE(1092, "Unexpected credential type"),
     IPV_SESSION_NOT_FOUND(
             1093, "Invalid IPV session id provided, corresponding session not found in db"),
-    CLIENT_OAUTH_SESSION_NOT_FOUND(1093, "Client OAuth session not found");
+    CLIENT_OAUTH_SESSION_NOT_FOUND(1093, "Client OAuth session not found"),
+    MISSING_LANGUAGE(1093, "Missing language choice from the frontend");
```
DanCorderIPV: Not sure what's going on here - I think these error codes should be unique, and probably not re-used even if the old error is no longer used. There 1093 is already being used above for IPV_SESSION_NOT_FOUND so not sure why it appears to also have been used for CLIENT_OAUTH_SESSION_NOT_FOUND

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+@Data
+public class CriJourneyRequest extends JourneyRequest {
+    private String language;
+
+    @Builder(builderMethodName = "criJourneyRequestBuilder")
```
DanCorderIPV: Not sure if we need to have a custom builder name here as it's only ever used as `CriJourneyRequest.builder()` so it's clear what type it is.

DanCorderIPV (changes requested):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
MikeCollingwood: Ah I saw when referencing builder, it references the JourneyRequest builder, which didn't work. But I see by keeping the annotation without the explicit naming it overrides the builder and does work

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
MikeCollingwood: Ah that's an error. I'll update it so it increments. I'll add a test for this to ensure the error responses don't overlap - I see the same with 1051

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
```diff
@@ -106,7 +104,10 @@ public enum ErrorResponse {
     UNEXPECTED_CREDENTIAL_TYPE(1092, "Unexpected credential type"),
     IPV_SESSION_NOT_FOUND(
             1093, "Invalid IPV session id provided, corresponding session not found in db"),
-    CLIENT_OAUTH_SESSION_NOT_FOUND(1093, "Client OAuth session not found");
+    CLIENT_OAUTH_SESSION_NOT_FOUND(1094, "Client OAuth session not found"),
+    MISSING_LANGUAGE(1095, "Missing language choice from the frontend"),
+    FAILED_TO_VALIDATE_VERIFIABLE_CREDENTIAL_RESPONSE(
+            1096, "Failed to validate verifiable credential response");
```
DanCorderIPV, line 110: @blakeyp I'm not sure what these error codes are used for in terms of reporting etc - it looks like 1051 has been duplicated for some time. Is there any problem we with deduplicating it now? (to a lesser extent 1093 too)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/CriJourneyRequest.java
```diff
@@ -1,19 +1,21 @@
...
+import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 
 @ExcludeFromGeneratedCoverageReport
 @EqualsAndHashCode(callSuper = true)
+@AllArgsConstructor
```
DanCorderIPV, line 12: Do you need this with the explicit constructor below?

MikeCollingwood: I did need it with the `@Builder` annotation (`Lombok @Builder needs a proper constructor for this class `), but it seems the `@SuperBuilder` can take either constructor. Either way, it seems the `@AllArgsConstructor` can replace the explicit constructor

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ErrorResponse.java
MikeCollingwood: Also, worth noting there are some gaps in the iterated codes

blakeyp: Not so worried about gaps but yeah let's deduplicate (we aren't relying on these specific error codes in any reporting)

DanCorderIPV (approved):

PR #2289, opened 2024-08-13
PYIC-7072: inherited identities

sam803:
### What changed

Add tests for inherited idenities

### Why did it change

To increase API test coverage

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7072


api-tests/src/utils/ipv-session.ts
```diff
@@ -0,0 +1,21 @@
+export class IPVSessionDetails {
```
Joe-Edwards-GDS: I generally try and avoid using `class` in Typescript unless I've got a really good reason...

Can this just be a plain old JS object?
```
interface IpvSessionDetails {
  subject: string;
  journeyId: string;
  journeyType: string;
  isReproveIdentity: boolean;
  inheritedIdentityId?: string;
}
```

api-tests/src/steps/ipv-steps.ts
```diff
@@ -58,20 +60,24 @@ After(function (this: World, options: ITestCaseHookParameter) {
 });
 
 When(
-  /I start a new ?'([\w-]+)' journey( with reprove identity)?/,
+  /I start a new ?'([\w-]+)' journey( with reprove identity)?(:? with inherited identity '([\w-]+)')?/,
```
Joe-Edwards-GDS: What's the `:?` for?

api-tests/src/steps/ipv-steps.ts
```diff
@@ -58,20 +60,24 @@ After(function (this: World, options: ITestCaseHookParameter) {
...
   async function (
     this: World,
     journeyType: string,
     reproveIdentity: string,
+    inheritedIdentityId: string,
```
Joe-Edwards-GDS: these last two should be `string | undefined`, I think?

api-tests/data/jar-requests/medium-confidence-pcl200-pcl250.json
Joe-Edwards-GDS, line 1: I'm slightly wondering if it's valuable storing these as separate JSON files - I suppose it makes it super-clear, but there is a lot of duplication when they only vary in the VTR field...

What do you think?

api-tests/features/inherited-identity.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Inherited Identity
```
Joe-Edwards-GDS, line 2: Might want to tag the Feature (not the scenarios) with `@Build` - assuming they work against build?

Joe-Edwards-GDS (commented):
Sorry I've given you a merge conflict!

api-tests/src/steps/ipv-steps.ts
sam803: A mistake - I intended for it to be a non-capturing group `?:`

api-tests/data/jar-requests/medium-confidence-pcl200-pcl250.json
sam803: Yes I did consider passing the vtr to `generateJar` function but would add more complexity to the [regex](https://github.com/govuk-one-login/ipv-core-back/blob/163602bc0840efcc3f481c42c87554da742ac960/api-tests/src/steps/ipv-steps.ts#L62) in the step so I opted for this approach - happy to change though :)

api-tests/src/utils/ipv-session.ts
sam803: Yes, sure (typescript novice here!)

api-tests/data/jar-requests/medium-confidence-pcl200-pcl250.json
sam803: I guess I could check the `journeyType` parameter for `pcl200` and `pcl250` then set them accordingly without having to alter the regex?

Joe-Edwards-GDS: I think this is ok for now, especially if we only have like 3 different variants, and we can fix it up later if need be.

If you did want to change I'd probably refactor that step as:
`Given I start a new 'P2' journey`
`Given I start a new 'P2, PCL250' journey`
etc.

In future we're going to also need P1 journeys, and to distinguish 'identity proving' from 'reverification' journeys - but that's about it.

api-tests/src/steps/ipv-steps.ts
```diff
@@ -104,7 +112,7 @@ When(
   ): Promise<void> {
     let attempt = 1;
     while (attempt <= MAX_ATTEMPTS) {
-      await startNewJourney(this, journeyType, false);
+      await startNewJourney(this, journeyType, false, undefined);
```
Joe-Edwards-GDS, line 115: If you make the last parameter optional (i.e. `journeyId?: string`) then you wouldn't need this - but this is nice and explicit

Joe-Edwards-GDS (approved):
Thanks, looks a bit more idiomatic now :)

PR #2303, opened 2024-08-15
BAU update AWS X-Ray tracing

Joe-Edwards-GDS:
### What changed

- Include AspectJ on libs that use tracing
- Remove Tracing from fast in-memory operations
- Add Tracing to external service calls
- Add TracingHttpClient to include more detailed information on external HTTP requests

### Why did it change

Missing AspectJ meant that the `@Tracing` annotations in libs were ignored. Moreover, tracing was applied inconsistently, often tracing useless information while leaving gaps where it would be useful.


Wynndow (approved):
Seems reasonable!

PR #2308, opened 2024-08-15
PYIC-7163: allow coi failure users to delete account

sam803:
### What changed

Changes to the update-name journey to route users to the new `update-details-failed` page (where they have the option to delete their account) if we get `fail-with-no-ci` or `access-denied` events from dcmaw (and the `updateDetailsAccountDeletion` is set)

### Why did it change
This change aims to provide users with clear information and options when they are unable to use the live/likeness application to prove their identity. By offering self-service options, users can either proceed to the service they intended to access or delete their account, improving user experience and reducing the need for call centre support.

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7163

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -43,20 +43,14 @@ states:
   FAILED_UPDATE_DETAILS:
     events:
       next:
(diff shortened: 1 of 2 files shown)


thebauSoftwire (approved):

Joe-Edwards-GDS (approved):
This looks ok, but we need to think how this is going to work when there's a variant for repeat fraud checks as well...

Happy for you to address that in the next ticket or now, as you prefer.

PR #2319, opened 2024-08-19
BAU: Fix MITIGATION_KBV_OPTIONS to MITIGATION_02_OPTIONS

MikeCollingwood:
### What changed

- MITIGATION_KBV_OPTIONS => MITIGATION_02_OPTIONS

### Why did it change

- MITIGATION_02_OPTIONS was the intended state to be referenced


DanCorderIPV (approved):
As discussed this isn't perfect, but better than an error and in a case that is pretty unlikely to actually happen in practice.
There will also be a ticket raised to fix this properly.

PR #2320, opened 2024-08-19
PYIC-7198: Make ticf request timeout configurable

Wynndow:
### What changed

Make ticf request timeout configurable

### Why did it change

This updates the ticfCriService to allow the request timeout to be configurable. A request should never take longer than 2 seconds to be processed by the TICF CRI, so we can reduce our own timeout.

The config value comes from the RestCriConfig for the TICF CRI. We may not want to explicitly set a timeout on other CRIs, so the value is optional, defaulting to 30 seconds. Currently the TICF CRI is the only CRI that uses the RestCriConfig.

### Issue tracking

- [PYIC-7198](https://govukverify.atlassian.net/browse/PYIC-7198)

[PYIC-7198]: https://govukverify.atlassian.net/browse/PYIC-7198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
@@ -12,8 +12,16 @@ import java.net.URI;
 @Getter
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
-@JsonIgnoreProperties(ignoreUnknown = false)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class RestCriConfig extends CriConfig {
+    private static final long DEFAULT_REQUEST_TIMEOUT = 30;
+
+    private Long requestTimeout;
     private URI credentialUrl;
     private boolean requiresApiKey;
+
+    public long getRequestTimeout() {
+        // This it to avoid having to define the request timeout in config if not desired
+        return requestTimeout == null ? DEFAULT_REQUEST_TIMEOUT : requestTimeout;
+    }
 }
--- a/lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
+++ b/lambdas/call-ticf-cri/src/main/java/uk/gov/di/ipv/core/callticfcri/service/TicfCriService.java
@@ -24,6 +24,8 @@ import java.io.IOException;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.net.http.HttpTimeoutException;
+import java.time.Duration;
 import java.util.List;
 
 import static uk.gov.di.ipv.core.library.domain.Cri.TICF;
@@ -40,6 +42,7 @@ public class TicfCriService {
     private final VerifiableCredentialValidator jwtValidator;
     private final SessionCredentialsService sessionCredentialsService;
 
+    @ExcludeFromGeneratedCoverageReport
     public TicfCriService(ConfigService configService) {
         this.configService = configService;
         this.httpClient = HttpClient.newHttpClient();
@@ -47,7 +50,7 @@ public class TicfCriService {
         this.sessionCredentialsService = new SessionCredentialsService(configService);
     }
 
-    @ExcludeFromGeneratedCoverageReport
+    // Used by contract tests
     public TicfCriService(
             ConfigService configService,
             VerifiableCredentialValidator jwtValidator,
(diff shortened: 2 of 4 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
```diff
@@ -12,8 +12,16 @@
 @Getter
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
-@JsonIgnoreProperties(ignoreUnknown = false)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
Wynndow, line 15: Making this true makes it easier to add new config values - the values can be set in config and then the code updated to read them

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
```diff
@@ -12,8 +12,16 @@
...
 public class RestCriConfig extends CriConfig {
+    private static final long DEFAULT_REQUEST_TIMEOUT = 30;
+
+    private long requestTimeout;
     private URI credentialUrl;
     private boolean requiresApiKey;
+
+    public long getRequestTimeout() {
+        // This it to avoid having to define the request timeout in config if not desired
+        return requestTimeout == 0 ? DEFAULT_REQUEST_TIMEOUT : requestTimeout;
```
Wynndow: If requestTimeout is not defined in the config json, the default value for a long is 0, which is an invalid argument for a timeout, so we have to override to something better. We need to handle this as we can't update the config first due to not ignoring unknown values.
Doing this is actually quite useful though as it allows the value to be optional in the config json.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
Joe-Edwards-GDS: Makes sense - not sure why it was false in the first place tbh

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
Joe-Edwards-GDS: You might want to consider making the field a `Long` and handling `null`, rather than 0? I think that conveys intent more clearly...

It would also allow you to make `null` => 'no timeout', rather than a default, should that be useful.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
```diff
@@ -12,8 +12,16 @@
 @Getter
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
-@JsonIgnoreProperties(ignoreUnknown = false)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class RestCriConfig extends CriConfig {
+    private static final long DEFAULT_REQUEST_TIMEOUT = 30;
+
+    private long requestTimeout;
```
Joe-Edwards-GDS: Perhaps `requestTimeoutSeconds`? This is probably ok - our other timeouts are generally seconds as well

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
Wynndow: Yep, that makes sense. I'll leave the default as a 30 second timeout I think, unless you think that having no timeout would be preferable?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
Wynndow: I think it should be reasonably clear from context that it's seconds - and it's nice to keep things terser.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/RestCriConfig.java
Joe-Edwards-GDS: It probably doesn't make much difference tbh - our default Lambda timeout is 40s, but API GW will time out after 29s (and arguably we should reduce both of those - nobody is going to wait that long for a page to load!)

The only benefit of having no timeout is then you don't have to argue about what the default is :D

Joe-Edwards-GDS (approved):

PR #2324, opened 2024-08-19
BAU: Retry CRI OAuth session retrieval

Joe-Edwards-GDS:
### What changed

- Added retries to CRI OAuth fetch
- Added retry to API tests in codepipeline

### Why did it change

Make the tests more resilient.

We have seen at least one instance where the new CRI OAuth session could not be retrieved, and these retries should address it. We always expect a CRI OAuth session to exist if we're receiving a CRI callback.

This shouldn't happen in production because there will generally be some time spent in the CRI, but it could be a risk for CRIs without any user interaction in future.

### DIFF
--- a/api-tests/secure-pipeline/run-tests.sh
+++ b/api-tests/secure-pipeline/run-tests.sh
@@ -12,7 +12,7 @@ export CORE_BACK_INTERNAL_API_KEY
 
 cd /api-tests
 
-npm run test:build
+npm run test:build -- --profile codepipeline
 
 api_tests_exit_code=$?
 cp reports/api-tests-cucumber-report.json "$TEST_REPORT_ABSOLUTE_DIR"
--- a/api-tests/src/config/config.ts
+++ b/api-tests/src/config/config.ts
@@ -37,8 +37,7 @@ const config = {
     name: getMandatoryConfig("ASYNC_QUEUE_NAME"),
     delaySeconds: parseInt(getMandatoryConfig("ASYNC_QUEUE_DELAY")),
   },
-  localAuditEvents:
-    getOptionalConfig("process.env.LOCAL_AUDIT_EVENTS") === "true",
+  localAuditEvents: getOptionalConfig("LOCAL_AUDIT_EVENTS") === "true",
   cimit: {
     managementCimitUrl: getMandatoryConfig("CIMIT_STUB_BASE_URL"),
     managementCimitApiKey: getMandatoryConfig("MANAGEMENT_CIMIT_STUB_API_KEY"),
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
@@ -5,9 +5,14 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.StringMapMessage;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
 import uk.gov.di.ipv.core.library.domain.Cri;
+import uk.gov.di.ipv.core.library.exceptions.CriOAuthSessionNotFoundException;
+import uk.gov.di.ipv.core.library.exceptions.NonRetryableException;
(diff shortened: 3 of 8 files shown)


api-tests/src/config/config.ts
```diff
@@ -37,8 +37,7 @@ const config = {
     name: getMandatoryConfig("ASYNC_QUEUE_NAME"),
     delaySeconds: parseInt(getMandatoryConfig("ASYNC_QUEUE_DELAY")),
   },
-  localAuditEvents:
-    getOptionalConfig("process.env.LOCAL_AUDIT_EVENTS") === "true",
+  localAuditEvents: getOptionalConfig("LOCAL_AUDIT_EVENTS") === "true",
```
Joe-Edwards-GDS, line 40: (oops!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
```diff
@@ -17,25 +22,47 @@
...
+                            throw new RetryableException(new CriOAuthSessionNotFoundException());
+                        }
+                        return criOauthSessionItem;
+                    });
+        } catch (InterruptedException | NonRetryableException e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            LOGGER.warn(LogHelper.buildErrorMessage("Could not find CRI OAuth session", e));
+            return null;
```
Joe-Edwards-GDS, line 64: It might be preferable to just throw the exception instead - but that would require unpicking a bunch of other code :(

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -35,6 +35,11 @@ public class IpvSessionService {
     private static final String START_STATE = "START";
     private static final String ERROR_STATE = "ERROR";
 
+    // Max sleeping time will be roughly WAIT_TIME * 2 ^ (MAX_ATTEMPTS - 1), plus execution time
+    // AWS say 'Consistency across all copies of data is usually reached within a second'
+    private static final int MAX_ATTEMPTS = 5;
+    private static final int WAIT_TIME_MILLIS = 50;
```
Joe-Edwards-GDS, line 41: More precisely, it's:
`WAIT_TIME * 2 ^ (MAX_ATTEMPTS - 1) - WAIT_TIME + MAX_ATTEMPTS * request time`

api-tests/secure-pipeline/run-tests.sh
```diff
@@ -12,7 +12,7 @@ export CORE_BACK_INTERNAL_API_KEY
 
 cd /api-tests
 
-npm run test:build
+npm run test:build -- --profile codepipeline
```
Wynndow, line 15: Had to look up what was going on here: https://github.com/cucumber/cucumber-js/blob/main/docs/profiles.md

Neat!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
```diff
@@ -17,25 +22,47 @@
 public class CriOAuthSessionService {
     private static final Logger LOGGER = LogManager.getLogger();
 
+    // Max sleeping time will be roughly WAIT_TIME * 2 ^ (MAX_ATTEMPTS - 1), plus execution time
+    // AWS say 'Consistency across all copies of data is usually reached within a second'
+    private static final int MAX_ATTEMPTS = 5;
+    private static final int WAIT_TIME_MILLIS = 50;
+
     private final DataStore<CriOAuthSessionItem> dataStore;
+    private final Sleeper sleeper;
```
Wynndow, line 31: Curious why we use this sleeper class, when it's just wrapping Thread.sleep()?

Wynndow (approved):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CriOAuthSessionService.java
Joe-Edwards-GDS: Just so we can stub it out in the tests I think

PR #2326, opened 2024-08-19
PYIC-7179: Use TICF CRI enabled setting

Wynndow:
### What changed

Use TICF CRI enabled setting

### Why did it change

We're currently using a feature flag to determine if we should be calling the TICF CRI or not. The CRI config for TICF includes an "enabled" value, similarly to other CRIs. We can use this rather than a flag to route journeys. This means we can remove that flag.

### Issue tracking

- [PYIC-7179](https://govukverify.atlassian.net/browse/PYIC-7179)

[PYIC-7179]: https://govukverify.atlassian.net/browse/PYIC-7179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
@@ -35,10 +35,10 @@ states:
       lambda: evaluate-gpg45-scores
     events:
       met:
-        targetState: STORE_IDENTITY_BEFORE_SUCCESS
-        checkFeatureFlag:
-          ticfCriBeta:
-            targetState: CRI_TICF_BEFORE_SUCCESS
+        targetState: CRI_TICF_BEFORE_SUCCESS
+        checkIfDisabled:
+          ticf:
+            targetState: STORE_IDENTITY_BEFORE_SUCCESS
       unmet:
         targetJourney: FAILED
         targetState: FAILED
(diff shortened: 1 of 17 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
```diff
@@ -35,10 +35,10 @@ states:
...
     events:
       met:
-        targetState: STORE_IDENTITY_BEFORE_SUCCESS
-        checkFeatureFlag:
-          ticfCriBeta:
-            targetState: CRI_TICF_BEFORE_SUCCESS
+        targetState: CRI_TICF_BEFORE_SUCCESS
+        checkIfDisabled:
+          ticf:
+            targetState: STORE_IDENTITY_BEFORE_SUCCESS
```
Joe-Edwards-GDS, line 41: I kinda like this as well, because it'll make TICF enabled the default - hopefully making us less likely to forget about it!

Joe-Edwards-GDS (approved):
I think you missed the refinement for this, but also interested to know whether you think it's a good idea!

The main rationale is so we can disable it later (without using a `beta` feature flag)

Wynndow:
> I think you missed the refinement for this, but also interested to know whether you think it's a good idea!
> 
> The main rationale is so we can disable it later (without using a `beta` feature flag)

Yeah I do think it's a good idea - although it would be nice to find a way to remove the branches in the journey map at some point if possible - they're a bit annoying.

PR #2327, opened 2024-08-19
PYIC-7207: Remove context for F2F enabled/disabled

DanCorderIPV:
DO NOT MERGE UNTIL WELSH TRANSLATIONS ARE DONE

### What changed

Change routing to page-multiple-doc-check so it is the same whether f2f is enabled or not. Change routing from page-multiple-doc-check so it goes to the PYI escape page if F2F is disabled. 

### Why did it change

New designs from UCD to simplify the flow

### Issue tracking

- [PYIC-7207](https://govukverify.atlassian.net/browse/PYIC-7207)

[PYIC-7207]: https://govukverify.atlassian.net/browse/PYIC-7207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -197,51 +191,28 @@ states:
...
-    events:
-      ukPassport:
-        targetState: CRI_UK_PASSPORT_J2
-      drivingLicence:
-        targetState: CRI_DRIVING_LICENCE_J3
-      end:
-        targetJourney: INELIGIBLE
-        targetState: INELIGIBLE_SKIP_MESSAGE
-
-  MULTIPLE_DOC_F2F_CHECK_PAGE:
```
blakeyp, line 235: We probably want to keep this is at first to protect any in-flight journeys while this deploys out?

DanCorderIPV: Good point, I guess someone can be in this state for a while that could straddle a deployment

blakeyp:
Looks good, HOLD until we add Welsh translations

AmritSidhu (commented):
Do we still need the check f2f disabled for `CRI_UK_PASSPORT_J2` & `CRI_DRIVING_LICENCE_J3`?

DanCorderIPV:
Yes, I think so. The change in this PR is to initially go to a page that works whether F2F is disabled or not. That page then routes the user based on whether F2F is disabled or not. So we're not removing the check, we're just moving it.

AmritSidhu (dismissed):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -191,29 +191,18 @@ states:
       next:
         targetState: POST_DCMAW_SUCCESS_PAGE
       not-found:
-        targetState: MULTIPLE_DOC_F2F_NINO_CHECK_PAGE
```
MikeCollingwood, line 194: Do similar changes need to be made on the output of nested journeys?

AmritSidhu (approved):

PR #2343, opened 2024-08-21
PYIC-7179: Use TICF CRI enabled setting

Wynndow:
### What changed

Use TICF CRI enabled setting

### Why did it change

[PYIC-7179: Use TICF CRI enabled setting](https://github.com/govuk-one-login/ipv-core-back/commit/99567c64fba72eb6cd04ccbef3d9fd0ce12f6c41)

We're currently using a feature flag to determine if we should be
calling the TICF CRI or not. The CRI config for TICF includes an
"enabled" value, similarly to other CRIs. We can use this rather than a
flag to route journeys. This means we can remove that flag.

### Issue tracking

- [PYIC-7179](https://govukverify.atlassian.net/browse/PYIC-7179)

[PYIC-7179]: https://govukverify.atlassian.net/browse/PYIC-7179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
+++ b/lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
@@ -44,7 +44,8 @@ import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredential
 
 import java.util.Map;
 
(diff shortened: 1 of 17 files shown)


Joe-Edwards-GDS (approved):
Thanks - appreciate the effort trying to tidy up the config, but I think this is ultimately neater 🙏

PR #2348, opened 2024-08-22
PYIC-6857 tweaks

Joe-Edwards-GDS:
### What changed

- Add extra parameter to control generation of the report
- Add extra parameter to configure max page size when scanning
- Return number of identities processed in intermediate scans
- Add ForkJoinPool to run identity checks in parallel

### Why did it change

We're only using a relatively small amount of DynamoDB capacity and the script takes a long time to run, so try a few simple changes to try and speed up the execution time.


lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -101,116 +106,124 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
+                            .map(VcStoreItem::getUserId)
+                            .distinct()
+                            .map(usrId -> new ReportUserIdentityItem(usrId, null, null, null, null))
+                            .toList());
 
-                reportProcessingResult.tacticalStoreLastEvaluatedKey(
-                        DynamoDbHelper.unmarshallLastEvaluatedKey(page.lastEvaluatedKey()));
+            reportProcessingResult.setTacticalStoreLastEvaluatedKey(
+                    DynamoDbHelper.unmarshallLastEvaluatedKey(page.lastEvaluatedKey()));
+            reportProcessingResult.addTacticalVcsEvaluated(page.items().size());
```
rpayal, line 172: Seems this will always have page size value. Is that u wanted or want to capture items processed in that run before break.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
```diff
@@ -101,116 +106,124 @@ public void handleRequest(InputStream inputStream, OutputStream outputStream, Co
...
+                                    userIds.size())));
+            reportUserIdentityScanDynamoDataStore.createOrUpdate(reportUserIdentityItems);
 
-                checkLambdaRemainingExecutionTime(context);
-            }
-            LOGGER.info(LogHelper.buildLogMessage("User identity check scan completed."));
-            reportProcessingResult.userIdentitylastEvaluatedKey(null);
+            reportProcessingResult.setUserIdentitylastEvaluatedKey(
+                    DynamoDbHelper.unmarshallLastEvaluatedKey(page.lastEvaluatedKey()));
+            reportProcessingResult.addUserIdentitiesEvaluated(page.items().size());
```
rpayal, line 221: same...Seems this will always have page size value.

lambdas/report-user-identity/src/main/java/uk/gov/di/ipv/core/reportuseridentity/ReportUserIdentityHandler.java
Joe-Edwards-GDS: This is adding the page size to the total, so should be capturing the total processed items (unless it fails due to an exception).

You're right, the page size will mostly be constant - unless the total size of items exceeds 1MB or we've reached the end of the scan.

rpayal (approved):

PR #2351, opened 2024-08-22
PYIC-7165: add a journey context so that we can decide routes based o…

sam803:
### What changed

 add a journey context so that we can decide routes based on this later in the journey
### Why did it change

We need to show different variants of the `sorry-we-could-not-update-your-detaiils` page depending on the initial journey context (either `repeat fraud check `or `reuse`), but since these journeys use a sub-journey (`update-address`) there is no way of telling the original journey context.

This adds a string context to the ipv session item which keeps a track of this and can then be used in routing specific contexts to different locations (alternative was to have lots of duplication in the journey map config)

### Issue tracking

https://govukverify.atlassian.net/browse/PYIC-7165
https://govukverify.atlassian.net/browse/PYIC-7166


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -40,6 +40,22 @@ states:
...
+  FAILED_CONFIRM_DETAILS_WITH_DELETION:
+    events:
+      next:
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION
+        auditEvents:
+          - IPV_USER_DETAILS_UPDATE_END
+        auditContext:
+          successful: false
+        checkFeatureFlag:
+          ticfCriBeta:
```
Joe-Edwards-GDS: (note you'll get a conflict on this from Chris's recent change)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
```diff
@@ -180,6 +180,10 @@ states:
       access-denied:
         targetJourney: FAILED
         targetState: FAILED_CONFIRM_DETAILS
+        checkJourneyContext:
+          rfc:
+            targetJourney: FAILED
+            targetState: FAILED_CONFIRM_DETAILS
```
Joe-Edwards-GDS: This is currently not doing anything? I wonder if the rfc/reuse check should move into failed.yaml instead...

So that this is unchanged:
- If the feature flag is disabled, it should route to the old `FAILED_CONFIRM_DETAILS`
- If the feature flag is enabled, it should route to the new `FAILED_UPDATE_DETAILS`
  - If reuse, that displays the reuse variant
  - If rfc, that displays the rfc variant

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -36,6 +36,7 @@ public class IpvSessionItem implements PersistenceItem {
     private String emailAddress;
     private ReverificationStatus reverificationStatus;
     private List<String> stateStack = new ArrayList<>();
+    private String journeyContext;
```
Joe-Edwards-GDS, line 45: The main thing I want to avoid if we introduce this is having fiddly bits of state that are hidden from the journey map and end up being missed. We've had several of bugs caused by a bit of state being set, and then becoming invalid - e.g. we set the VOT in CheckExistingIdentity, but forgot to reset it when the user asks to update their details.

Some things to consider:
- is it obvious what 'journey context' means?
  - does it need an explanatory comment?
- when does the journey context get reset?
  - If it should only be for things that apply to an entire journey, do we need to say so?
  - If that's too restrictive, we need to be very careful to identify when it no longer applies
- might we ever want to set two overlapping journey context values?
  - Should it be multivalued, and if so, how would that work - a list or a map?
  - (This might just add a bunch of complexity for no reason)

The `rfc` usage at the moment seems relatively risk-free at the moment... either a journey needs to perform a repeat fraud check or not, even if you decide to change your name. 🙂

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
Joe-Edwards-GDS, line 1: How are we going to display this in the journey map?

Joe-Edwards-GDS (commented):
Left a few things to think about

Seems like a roughly sensible idea though - as you say, we'd otherwise need to duplicate almost the entire journey!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
sam803: Thanks for the comments,
- where would be the best place to document the journey context? Readme/Source file? I think it probably does need some explanation
- The context does not get reset but can be overidden by a later non null assignation. We could add a feature to reset it if the necessity arises in the future (e.g. `resetJourneyContext`) or check for an empty string (currently it will only overwrite the current value if its not empty or null) 
- We could make it a list to account for this possibility if we think it may arise in the future? Not too much added complexity

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
sam803: I added so that if the `updateDetailsAccountDeletion` flag is enabled it wont drop into the FAILED_UPDATE_DETAILS for rfc journeys as the context takes precedence. But happy to switch the check to the failure if it makes more sense

Joe-Edwards-GDS: I think so - conceptually there are 3 changes in the fast-follow, even if it's not exactly how the tickets have been defined:
- Certain events from dcmaw (e.g. this `access-denied`) should display a new page, which we've called `FAILED_UPDATE_DETAILS`, while other events (e.g. `fail-with-ci`) still route to `FAILED_CONFIRM_DETAILS`
  - This is what the feature flag check below is doing
- `FAILED_CONFIRM_DETAILS` now has two variants, for reuse vs. repeat fraud check
  - I think this should probably just live in `failed.yaml`
- `FAILED_UPDATE_DETAILS` also has two variants, for reuse vs. repeat fraud check
  - similarly

So this file decides which type of failure it is (confirm vs. update - not great names..!), and the failures decide which variant of that you see.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
Joe-Edwards-GDS: - I think it should probably have some kind of comment in this file - but we could also refer to it in the team manual if there's an appropriate place.

- Yeah, we can worry about that later - probably either distinguishing null vs. empty, or using a magic 'delete' value

- I think the complexity with a list is knowing when to append vs replace... it's fine to leave as a string for now :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
sam803: I think we should be able to do this, but might be better in a separate ticket given my limited time left on the team :(  will create one

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
sam803: Done :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: This needs to be inverted (see `FAILED_CONFIRM_DETAILS`)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
```diff
@@ -454,6 +454,14 @@ states:
...
+        checkFeatureFlag:
+          updateDetailsAccountDeletion:
+            targetJourney: FAILED
+            targetState: FAILED_CONFIRM_DETAILS_WITH_DELETION
```
Joe-Edwards-GDS: I think I was expecting _all_ of these to go to `FAILED_CONFIRM_DETAILS`, and then the feature flag and context routing would happen there?

So conceptually this file just decides whether you see `FAILED_CONFIRM_DETAILS` or `FAILED_UPDATE_DETAILS`

sam803: My reasoning for doing this way was that I wasn't  sure that everything that uses `FAILED_CONFIRM_DETAILS` should have the delete option, for example a `not-found` reply from DCMAW, since I believe the ticket stated that only `access-denied `and `fail-with-no-ci` responses should have the delete option? see implementation notes from https://govukverify.atlassian.net/browse/PYIC-7163

Joe-Edwards-GDS: I think there are two screens, and all routes should go to one or the other:
- For `access-denied` and `fail-with-no-ci` - https://govukverify.atlassian.net/browse/PYIC-7163
    - This is a new page for the fast-follow work
- For everything else - https://govukverify.atlassian.net/browse/PYIC-7165
    - This is an update to an existing page for the fast-follow work

I think we've called the first one `FAILED_UPDATE_DETAILS` and the second one `FAILED_CONFIRM_DETAILS`...

sam803: Ok I have moved the logic to failed.yaml and its simpler :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -25,6 +25,20 @@ states:
             targetState: RETURN_TO_RP
 
   FAILED_CONFIRM_DETAILS:
+    events:
+      next:
+        targetJourney: FAILED
+        targetState: FAILED_CONFIRM_DETAILS_WITHOUT_DELETION
```
Joe-Edwards-GDS: This is pretty sneaky - 'switching journey' to a state within the same journey(!) This probably has a couple of unintended side-effects, like recording multiple SUBJOURNEY_START audit events and showing up weirdly in the journey map that we need to think through.

I think we can just do it in one state instead? It's long, but I think that's necessary - there are 4 different possible targets depending on the situation:
```
  FAILED_CONFIRM_DETAILS:
    events:
      next:
        targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
        auditEvents:
          - IPV_USER_DETAILS_UPDATE_END
        auditContext:
          successful: false
        checkIfDisabled:
          ticf:
            targetState: COULD_NOT_CONFIRM_DETAILS_PAGE
            auditEvents:
              - IPV_USER_DETAILS_UPDATE_END
            auditContext:
              successful: false
            checkFeatureFlag:
              updateDetailsAccountDeletion:
                targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION
                auditEvents:
                  - IPV_USER_DETAILS_UPDATE_END
                auditContext:
                  successful: false
        checkFeatureFlag:
          updateDetailsAccountDeletion:
            targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS_WITH_DELETION
            auditEvents:
              - IPV_USER_DETAILS_UPDATE_END
            auditContext:
              successful: false
```

Of course once we remove the feature flag it'll simplify back down again :)

Joe-Edwards-GDS (commented):
Sorry to keep going back and forth - I think it should be almost there now!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
sam803: ah ok - didn't realise you could combine the checkFeatureFlag and checkIfDisabled checks together, will work on this tomorrow :)

Joe-Edwards-GDS: No worries - it's not super obvious (or beautiful...) but I think conveys the intent relatively well.

One important thing to know is that disabled checks take precedence over feature flags (and otherwise it's in the order written).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
sam803: @Joe-Edwards-GDS https://govukverify.atlassian.net/browse/PYIC-7304

Joe-Edwards-GDS (approved):

PR #2358, opened 2024-08-23
PYIC-6857 configurable parallelism and logging

Joe-Edwards-GDS:
### What changed

- Ability to supply a 'parallelism' to the script, to control degree of parallelism
- Added logs to step 1, so activity is visible
- Removed profile matching logs from step 2

### Why did it change

- The parallelism makes a big difference to the speed of the lambda, and can be tuned to speed up the process.
- The existing logs for step 1 don't indicate it doing anything
- The existing logs for step 2 are very noisy - logging two GPG45 profile matched messages for every match


rpayal (approved):
LGTM

PR #2367, opened 2024-08-27
PYIC-7164 Update journey map for repeatFraudCheck fast follow.

rpayal:
## Proposed changes
Update journey map for repeatFraudCheck fast follow.

### What changed
Updated journey map

### Issue tracking

- [PYIC-7164](https://govukverify.atlassian.net/browse/PYIC-7164)

### Other considerations

[PYIC-7164]: https://govukverify.atlassian.net/browse/PYIC-7164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -310,8 +310,6 @@ states:
       pageId: update-details-failed
       context: repeatFraudCheck
     events:
-      continue:
-        targetState: RETURN_TO_RP
       delete:
         targetState: DELETE_HANDOVER_PAGE
       back:
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -76,23 +76,9 @@ states:
                   - IPV_USER_DETAILS_UPDATE_END
                 auditContext:
                   successful: false
-            checkFeatureFlag:
-              updateDetailsAccountDeletion:
-                targetState: COULD_NOT_UPDATE_DETAILS_PAGE
-                auditEvents:
-                  - IPV_USER_DETAILS_UPDATE_END
-                auditContext:
-                  successful: false
         checkJourneyContext:
           rfc:
             targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
-            auditEvents:
-              - IPV_USER_DETAILS_UPDATE_END
-            auditContext:
-              successful: false
-        checkFeatureFlag:
-          updateDetailsAccountDeletion:
-            targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS
             auditEvents:
               - IPV_USER_DETAILS_UPDATE_END
             auditContext:
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -71,11 +71,18 @@ states:
               successful: false
             checkJourneyContext:
               rfc:
-                targetState: COULD_NOT_CONFIRM_DETAILS_PAGE
+                targetState: COULD_NOT_UPDATE_DETAILS_PAGE
                 auditEvents:
                   - IPV_USER_DETAILS_UPDATE_END
                 auditContext:
                   successful: false
+                checkFeatureFlag:
+                  updateDetailsAccountDeletion:
+                    targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
+                    auditEvents:
+                      - IPV_USER_DETAILS_UPDATE_END
+                    auditContext:
+                      successful: false
         checkJourneyContext:
           rfc:
             targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
@@ -83,22 +90,13 @@ states:
               - IPV_USER_DETAILS_UPDATE_END
             auditContext:
               successful: false
-
-  FAILED_UPDATE_DETAILS_RFC:
-    events:
-      next:
-        targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC
-        auditEvents:
-          - IPV_USER_DETAILS_UPDATE_END
-        auditContext:
-          successful: false
-        checkIfDisabled:
-          ticf:
-            targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
-            auditEvents:
-              - IPV_USER_DETAILS_UPDATE_END
-            auditContext:
-              successful: false
+            checkFeatureFlag:
+              updateDetailsAccountDeletion:
+                targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC
+                auditEvents:
+                  - IPV_USER_DETAILS_UPDATE_END
+                auditContext:
+                  successful: false
 
   FAILED_BAV:
     events:
@@ -312,9 +310,8 @@ states:
     events:
       delete:
         targetState: DELETE_HANDOVER_PAGE
-      back:
-        targetJourney: INITIAL_JOURNEY_SELECTION
-        targetState: START
+      returnToRp:
+        targetState: RETURN_TO_RP
         auditEvents:
           - IPV_USER_DETAILS_UPDATE_ABORTED
 
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -98,6 +98,22 @@ states:
             auditContext:
               successful: false
 
+  FAILED_UPDATE_DETAILS_RFC:
+    events:
+      next:
+        targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC
+        auditEvents:
+          - IPV_USER_DETAILS_UPDATE_END
+        auditContext:
+          successful: false
+        checkIfDisabled:
+          ticf:
+            targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
+            auditEvents:
+              - IPV_USER_DETAILS_UPDATE_END
+            auditContext:
+              successful: false
+
   FAILED_BAV:
     events:
       next:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -85,7 +85,7 @@ states:
                   successful: false
         checkJourneyContext:
           rfc:
-            targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
+            targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC
```
sam803: Do we also need to check the feature flag `updateDetailsAccountDeletion` ?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -57,21 +57,21 @@ states:
   FAILED_UPDATE_DETAILS:
     events:
       next:
-        targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
+        targetState: CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS
```
sam803: I think this needs to stay the same CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS so that the default behavior is maintained for when the feature flag is not set

thebauSoftwire (dismissed):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
rpayal: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS  is default to state FAILED_CONFIRM_DETAILS

I suppose for state FAILED_UPDATE_DETAILS default will be CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
rpayal: not sure as we don't have any default page without delete option.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -57,42 +57,28 @@ states:
...
             auditContext:
               successful: false
             checkJourneyContext:
               rfc:
-                targetState: COULD_NOT_CONFIRM_DETAILS_PAGE
-                auditEvents:
-                  - IPV_USER_DETAILS_UPDATE_END
-                auditContext:
-                  successful: false
-            checkFeatureFlag:
```
thebauSoftwire: For my own learning, in what order is this evaluated e.g. if rfc is the journey context but the `updateDetailsAccountDeletion` feature flag is also enabled?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -57,42 +57,28 @@ states:
...
-              updateDetailsAccountDeletion:
-                targetState: COULD_NOT_UPDATE_DETAILS_PAGE
+                targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
                 auditEvents:
                   - IPV_USER_DETAILS_UPDATE_END
                 auditContext:
                   successful: false
         checkJourneyContext:
           rfc:
-            targetState: CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS
```
thebauSoftwire, line 88: might be my unfamiliarity with how the `checkJourneyContext` works (see [this](https://github.com/govuk-one-login/ipv-core-back/pull/2367/files#r1740647559) comment) but if `FAILED_UPDATE_DETAILS` only has two variants (reuse or rfc), would handling the `checkFeatureFlag` be better in the target state?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: I think Sam is right here - the new `failed update details` page should only be shown if the feature flag is set.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: The order is defined in `BasicEvent#resolve`, which currently looks like 
- Check for disabled CRIs
- Check for journey context
- Check for feature flag

It's recursive, so once any of these matches, it'll search inside. E.g. in this case if you have journey context and the feature flag, the journey context will match first.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -57,42 +57,28 @@ states:
...
               rfc:
-                targetState: COULD_NOT_CONFIRM_DETAILS_PAGE
-                auditEvents:
-                  - IPV_USER_DETAILS_UPDATE_END
-                auditContext:
-                  successful: false
-            checkFeatureFlag:
-              updateDetailsAccountDeletion:
-                targetState: COULD_NOT_UPDATE_DETAILS_PAGE
+                targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
```
Joe-Edwards-GDS: I think this also needs to be behind a feature flag for now

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: You could - and that would remove the need for a separate `CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC`. But you'd still need to check context here if TICF is disabled.

On the other hand, this approach means that all the context checking is in one place (this entry state) :shrug:

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: I think we do - otherwise we might reach this without it being enabled?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -283,6 +288,20 @@ states:
...
 
+  COULD_NOT_UPDATE_DETAILS_PAGE_RFC:
+    response:
+      type: page
+      pageId: update-details-failed
+      context: repeatFraudCheck
+    events:
+      delete:
+        targetState: DELETE_HANDOVER_PAGE
+      back:
```
Joe-Edwards-GDS: I'm not sure about overloading the 'back' event here - IMO we should try to reserve that for a 'back' button only

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -283,6 +288,20 @@ states:
...
+        targetJourney: INITIAL_JOURNEY_SELECTION
```
Joe-Edwards-GDS: Is this right? The content implies that this should return you to the RP (with no identity)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
thebauSoftwire: ah nice, thanks :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
thebauSoftwire: Here's the corresponding core-front PR: https://github.com/govuk-one-login/ipv-core-front/pull/1566

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -64,39 +64,39 @@ states:
           successful: false
         checkIfDisabled:
           ticf:
-            targetState: COULD_NOT_CONFIRM_DETAILS_PAGE
+            targetState: COULD_NOT_UPDATE_DETAILS_PAGE
```
Joe-Edwards-GDS: I think this should stay a `COULD_NOT_CONFIRM_DETAILS_PAGE`, as you'll reach here without the feature flag

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -64,39 +64,39 @@ states:
...
             auditEvents:
               - IPV_USER_DETAILS_UPDATE_END
             auditContext:
               successful: false
             checkJourneyContext:
```
Joe-Edwards-GDS: I think you need to invert the context and feature flag checks? - we only need to check the context if the feature flag is enabled

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -64,39 +64,39 @@ states:
...
                 auditContext:
                   successful: false
+                checkFeatureFlag:
+                  updateDetailsAccountDeletion:
+                    targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
+                    auditEvents:
+                      - IPV_USER_DETAILS_UPDATE_END
+                    auditContext:
+                      successful: false
         checkJourneyContext:
```
Joe-Edwards-GDS: As above, I think you need to invert the context and feature flag checks?

Joe-Edwards-GDS: So overall I think it's something like:

- `CRI_TICF_BEFORE_COULD_NOT_CONFIRM_DETAILS` (as existing)
  - If TICF is disabled
    - `COULD_NOT_CONFIRM_DETAILS_PAGE` (as existing)
    - If feature flag is enabled
      - `COULD_NOT_UPDATE_DETAILS_PAGE`
      - If rfc context
        - `COULD_NOT_UPDATE_DETAILS_PAGE_RFC`
  - If feature flag is enabled
    - `CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS`
    - If rfc context
      - `CRI_TICF_BEFORE_COULD_NOT_UPDATE_DETAILS_RFC`

So each possibility occurs once - there are just six of them D:

We should be able to simplify soon, fingers crossed...
- Once this goes live we can get rid of the non-feature-flagged versions
- I just created https://govukverify.atlassian.net/browse/PYIC-7383 to simplify some of the TICF cases

Joe-Edwards-GDS (commented):
Sorry, I think we're almost there but needs a tiny bit of tweaking, to make sure all the new pages are behind the feature flag.

(We'll also have to make some other changes in light of Simon's new ticket, but let's separate the PRs)

Joe-Edwards-GDS (approved):

PR #2368, opened 2024-08-27
PYIC-7034 nested journey visualisation

thebauSoftwire:
### What changed
- can now click on nested journey block to open in new page
- update nested journey blocks to be visually distinct from "other" blocks
- fixes visualisation of STRATEGIC_APP_TRIAGE nested journey
  - updates the STRATEGIC_APP_TRIAGE nested journey to return exit events for multiple doc check page instead of inheriting from parent journey

| Description | Screenshot |
|-|-|
| `ADDRESS_AND_FRAUD` nested journey |  Screenshot 2024-08-27 at 12 48 34 |
| `STRATEGIC_APP_TRIAGE` | Screenshot 2024-08-27 at 12 49 11 |
| updated nested journey key |  Screenshot 2024-08-27 at 12 49 37 |

### Why did it change
Clearer visualisation of nested journey when viewed separately instead of in-place

### Issue tracking

- [PYIC-7034](https://govukverify.atlassian.net/browse/PYIC-7034)

[PYIC-7034]: https://govukverify.atlassian.net/browse/PYIC-7034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


journey-map/public/index.mjs
```diff
@@ -93,12 +101,15 @@ const getPageUrl = (id, context) => {
...
-const getJourneyUrl = (id) => `?journeyType=${encodeURIComponent(id)}`;
+const getJourneyUrl = (id) => `?${JOURNEY_TYPE_SEARCH_PARAM}=${encodeURIComponent(id)}`;
+
+const getNestedJourneyUrl = (id) => `?${NESTED_JOURNEY_TYPES}=${encodeURIComponent(id)}`;
 
 const switchJourney = async (targetJourney, targetState) => {
     // Update URL
     const params = new URLSearchParams(window.location.search);
-    params.set('journeyType', targetJourney);
+    params.delete(NESTED_JOURNEY_TYPE_SEARCH_PARAM);
```
thebauSoftwire, line 106: when switching to a different journey, we want to remove the `nestedJourneyType` so that it doesn't just show the nested journey map (logic to handle which to show is on line 219)

journey-map/public/index.mjs
```diff
@@ -84,6 +91,7 @@ const loadJourneyMaps = async () => {
         const journeyResponse = await fetch(`./${encodeURIComponent(upperToKebab(journeyType))}.yaml`);
         journeyMaps[journeyType] = yaml.parse(await journeyResponse.text());
     }));
+
     const nestedResponse = await fetch('./nested-journey-definitions.yaml');
     nestedJourneys = yaml.parse(await nestedResponse.text());
```
thebauSoftwire, line 91: I did also try and separate the nested journeys into separate files instead (there's another ticket to extract a KBV nested journey so thought it might be worth doing to avoid a long file) but I wasn't sure if it was within the scope of this ticket to also update the states machine to look for these separate files instead or if that should be part of another ticket?

Joe-Edwards-GDS: Good question - feels like it should definitely be a separate PR, although I'm happy for you to do it as part of this ticket or a separate one as you prefer

thebauSoftwire: I think I'll do it as part of [7035](https://govukverify.atlassian.net/browse/PYIC-7035) to not block this one as it's purely a journey map ticket

Joe-Edwards-GDS (dismissed):

journey-map/public/render.mjs
```diff
@@ -35,20 +35,25 @@ export const getOptions = (journeyMaps) => {
 };
 
 // Expand out parent states
-const expandParents = (journeyMap) => {
+// Will also search 'otherStates' (e.g. for nested journeys)
```
Joe-Edwards-GDS, line 38: This allows us to see the failure states etc. inherited by `CRI_STATE` depending on the context.

In future we probably want to avoid this pattern - it's pretty confusing that the nested journey can inherit things from the journey it's in, but it works like that for now.

image

vs.

image

journey-map/public/render.mjs
```diff
@@ -240,36 +270,57 @@ const calcOrphanStates = (journeyMap) => {
...
+    ...Object.fromEntries(
+        Object.entries(nestedJourney.entryEvents)
+            .map(([event, def]) => [
+                `entry_${event}`.toUpperCase(),
+                {
+                    entryEvent: event,
+                    events: { [event]: def }
+                }
+            ]),
+    ),
```
thebauSoftwire, line 285: Nice! This looks way better and more helpful with the different entry events separated

thebauSoftwire (commented):
Nice! Thanks for picking this up for me :) I can't actually approve this PR myself bc I created it but happy with the changes

Wynndow (approved):

PR #2374, opened 2024-08-29
PYIC-7324: Write VCs directly to EVCS for reuse tests

Wynndow:
### What changed

Write VCs directly to EVCS for reuse tests

### Why did it change

API tests that required an existing identity would have to run a full journey through the system to generate it. This change allows for VCs generated by the CRI stubs to be written directly to EVCS for a user.

This should mean less boilerplate as well as a faster running test suite.

### Issue tracking

- [PYIC-7324](https://govukverify.atlassian.net/browse/PYIC-7324)

[PYIC-7324]: https://govukverify.atlassian.net/browse/PYIC-7324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/mfa-reset-journey.feature
+++ b/api-tests/features/mfa-reset-journey.feature
@@ -2,22 +2,11 @@
 Feature: MFA reset journey
 
   Scenario: MFA reset journey
-    Given I start a new 'medium-confidence' journey
-    Then I get a 'page-ipv-identity-document-start' page response
-    When I submit an 'appTriage' event
-    Then I get a 'dcmaw' CRI response
-    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
-    Then I get a 'page-dcmaw-success' page response
-    When I submit a 'next' event
-    Then I get an 'address' CRI response
-    When I submit 'kenneth-current' details to the CRI stub
-    Then I get a 'fraud' CRI response
-    When I submit 'kenneth-score-2' details to the CRI stub
-    Then I get a 'page-ipv-success' page response
-    When I submit a 'next' event
-    Then I get an OAuth response
-    When I use the OAuth response to get my identity
-    Then I get a 'P2' identity
+    Given the subject already has the following credentials
+      | CRI     | scenario                     |
+      | dcmaw   | kenneth-driving-permit-valid |
+      | address | kenneth-current              |
+      | fraud   | kenneth-score-2              |
 
     # Start MFA Reset journey for existing user
     When I start a new 'reverification' journey
--- a/api-tests/src/clients/cri-stub-client.ts
+++ b/api-tests/src/clients/cri-stub-client.ts
@@ -1,4 +1,9 @@
-import { CriStubRequest, CriStubResponse } from "../types/cri-stub.js";
+import {
+  CriStubGenerateVcRequest,
+  CriStubRequest,
+  CriStubResponse,
+} from "../types/cri-stub.js";
+import config from "../config/config.js";
 
 export const callHeadlessApi = async (
   redirectUrl: string,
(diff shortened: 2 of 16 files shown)


sam803 (approved):

api-tests/features/mfa-reset-journey.feature
```diff
@@ -2,22 +2,11 @@
...
-    Then I get an 'address' CRI response
-    When I submit 'kenneth-current' details to the CRI stub
-    Then I get a 'fraud' CRI response
-    When I submit 'kenneth-score-2' details to the CRI stub
-    Then I get a 'page-ipv-success' page response
-    When I submit a 'next' event
-    Then I get an OAuth response
-    When I use the OAuth response to get my identity
-    Then I get a 'P2' identity
+    Given the subject already has the following credentials
```
Joe-Edwards-GDS, line 5: I think there's perhaps some lack of clarity over who **I** am in these steps - are you 'the subject', or some entity operating core-back on behalf of 'the subject'.

Perhaps the latter interpretation is more technically correct, I think I just expected `Given I already have the following stored credentials`..!

api-tests/src/clients/cri-stub-client.ts
```diff
@@ -21,3 +26,25 @@ export const callHeadlessApi = async (
 
   return criStubResponse.json();
 };
+
+export const generateVc = async (
+  criId: string,
+  body: CriStubGenerateVcRequest,
+): Promise<string> => {
+  const response = await fetch(
+    `https://${criId}-cri.stubs.account.gov.uk/credentials/generate`,
```
Joe-Edwards-GDS, line 35: (This is a tiny bit fragile, but feels worth it)

api-tests/src/clients/cri-stub-client.ts
```diff
@@ -21,3 +26,25 @@ export const callHeadlessApi = async (
...
+    {
+      headers: {
+        "x-api-key": config.credentialIssuers.generateCredentialApiKey,
+      },
+      method: "POST",
+      body: JSON.stringify(body),
```
Joe-Edwards-GDS, line 41: (We don't do it above either, but technically if we're posting a body like this, we should include a `content-type` header)

api-tests/src/steps/cri-steps.ts
```diff
@@ -174,3 +178,33 @@ When(
...
+When(
+  /^the subject already has the following (expired )?credentials$/,
+  async function (
+    this: World,
+    expired: "expired " | undefined,
+    table: DataTable,
+  ): Promise<void> {
+    this.userId = this.userId ?? getRandomString(16);
+    const credentials: string[] = [];
+    for (const row of table.hashes()) {
```
Joe-Edwards-GDS, line 191: If you wanted to speed this up, you could do this in parallel with `Promise.all` - but there's no real need unless it's particularly slow (and perhaps harder to read/debug)

```
const credentials = await Promise.all(table.hashes()
  .map(async (row) => criStubClient.generateVc(
    row.CRI,
    await generateVcRequestBody(
      this.userId,
      row.CRI,
      row.scenario,
      expired ? EXPIRED_NBF : undefined,
    ),
  ),
));
```
It'll kick off all the promises in parallel and then wait for them all to complete

api-tests/src/steps/cri-steps.ts
```diff
@@ -174,3 +178,33 @@ When(
     );
   },
 );
+
+When(
+  /^the subject already has the following (expired )?credentials$/,
```
Joe-Edwards-GDS, line 183: (perhaps debatable if this belongs in `cri-steps` but sure!)

Joe-Edwards-GDS (commented):
This looks much neater, and should be faster too - thanks!

api-tests/src/steps/cri-steps.ts
Joe-Edwards-GDS: An easier to follow example without the loop would be:
```
// Await in parallel
const firstPromise = doAsyncThing();
const secondPromise = doAsyncThing();
const results = await Promise.all([firstPromise, secondPromise]);
```
vs.
```
// Await in series
const firstResult = await doAsyncThing();
const secondResult = await doAsyncThing();
const results = [firstResult, secondResult];
```

PR #2391, opened 2024-09-03
PYIC-7316: TICF api tests

thebauSoftwire:
### What changed
Adds TICF api tests
- adds `MANAGEMENT_TICF_API_KEY` and `TICF_STUB_BASE_URL` to env files

### Why did it change
Increase test coverage of API tests

### Issue tracking

- [PYIC-7316](https://govukverify.atlassian.net/browse/PYIC-7316)

[PYIC-7316]: https://govukverify.atlassian.net/browse/PYIC-7316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/pre-merge-checks.yml
```diff
@@ -165,6 +165,7 @@ jobs:
           MANAGEMENT_CIMIT_STUB_API_KEY: ${{ secrets.MANAGEMENT_CIMIT_STUB_API_KEY }}
           EVCS_STUB_API_KEY: ${{ secrets.API_KEY_EVCS }}
           CRI_STUB_GEN_CRED_API_KEY: ${{ secrets.CRI_STUB_GEN_CRED_API_KEY }}
+          MANAGEMENT_TICF_API_KEY: ${{ secrets.MANAGEMENT_TICF_API_KEY }}
```
thebauSoftwire, line 169: I don't have access to the secrets for this repo so this will need to get added in

Joe-Edwards-GDS: If you message me I can add the secrets

api-tests/features/ticf/failed-ticf-requests.feature
```diff
@@ -0,0 +1,45 @@
+@Build
+Feature: Failed TICF requests
+  Scenario Outline: TICF Management API request configured for 400 initially and no CI in reuse journey
+    Given there is an existing TICF record for the user with details
```
Joe-Edwards-GDS: I think this is a slightly funny way of wording it - it's not so much an existing 'record', as priming the stub.

Perhaps something like `Given TICF CRI will respond with`?

api-tests/features/ticf/failed-ticf-requests.feature
```diff
@@ -0,0 +1,45 @@
...
+      | responseDelay | 0                            |
+      | type          | RiskAssessment               |
+      | txn           | randomUuid                   |
+      | statusCode    | <statusCode>                          |
```
Joe-Edwards-GDS: Do we need all of these? It might be more obvious what you're interested in if you omit all the defaults
- `responseDelay` of zero is the default
- are `txn` and `type` ever anything else?

I'm torn as to whether it'd be cleaner just do drop the datatable and have some simpler steps (`Given TICF CRI will respond with a {number} statusCode` etc.) but it makes it more flexible this way.

api-tests/features/ticf/failed-ticf-requests.feature
```diff
@@ -0,0 +1,45 @@
+@Build
+Feature: Failed TICF requests
+  Scenario Outline: TICF Management API request configured for 400 initially and no CI in reuse journey
```
Joe-Edwards-GDS: '400' should be `<statusCode>` I think

It could probably be a bit more descriptive tbh - 'TICF CRI returning a <statusCode> error' is what this is trying to test. The interesting part of the test is not a reuse either, it's identity proving!

api-tests/features/ticf/failed-ticf-requests.feature
```diff
@@ -0,0 +1,45 @@
...
+    When I submit 'kenneth-current' details to the CRI stub
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-score-2' details to the CRI stub
+    Then I get a 'page-ipv-success' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity without a 'TICF' VC
+
+    Given there is an existing TICF record for the user with details
```
Joe-Edwards-GDS: What's this extra bit of the test checking?

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
```diff
@@ -0,0 +1,115 @@
+Feature: TICF failed journeys error scenarios
+
+  Rule: Via alternate doc route
+    Background: Start TICF alternate doc journey given user already has an existing TICF record
+      Given there is an existing TICF record for the user with details
+        | responseDelay | 0                            |
+        | type          | RiskAssessment               |
+        | txn           | randomUuid                   |
```
Joe-Edwards-GDS: This looks like the default TICF CRI behaviour?

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
```diff
@@ -0,0 +1,115 @@
...
+      When I submit 'kenneth-passport-valid' details to the CRI stub
+      Then I get an 'address' CRI response
+      When I submit 'kenneth-current' details to the CRI stub
+      Then I get a 'fraud' CRI response
+      When I submit 'kenneth-score-2' details to the CRI stub
+      Then I get a 'page-pre-experian-kbv-transition' page response
+      When I submit a 'next' event
+      Then I get a 'kbv' CRI response
+      When I submit 'kenneth-needs-enhanced-verification' details to the CRI stub
+      Then I get a 'pyi-suggest-other-options' page response
```
Joe-Edwards-GDS: The rule and background say alternate doc, but this is an enhanced verification journey?

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
```diff
@@ -0,0 +1,115 @@
...
+      When I submit 'kenneth-passport-invalid' details to the CRI stub
+      Then I get a 'pyi-no-match' page response
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P0' identity with a 'TICF' VC
+      And the TICF VC has properties
+        | cis  |                              |
+        | type | RiskAssessment               |
+        | txn  | randomUuid                   |
```
Joe-Edwards-GDS: (Default behaviour?)

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
```diff
@@ -0,0 +1,115 @@
...
+
+  Rule: Via post-office or no-photo-id
```
Joe-Edwards-GDS: Seems weird to combine these - should they just be separate scenarios?

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
Joe-Edwards-GDS, line 1: This seems like a slightly random set of journeys to test - @SreeBandla01 do you know what the rationale here is?

For context I think there are two things we want to test with TICF:
1. That when calling TICF CRI we handle a variety of responses appropriately - this includes things like a CI, bad status code or a timeout
2. That we call TICF CRI appropriately in each journey (i.e. always)

I think this file is covering a handful of journeys for 2. (and the scenarios for 1. implicitly cover some as well), but nowhere near all of them.

Now that we're starting to enable TICF CRI, I wonder if we should instead:
- Enable TICF CRI in dev and build
- Add a step (or update the existing `I get a .. identity` step) that confirms you receive a default TICF VC on most/all other scenarios.
- (get rid of these tests)

api-tests/features/ticf/ticf-journey.feature
```diff
@@ -0,0 +1,326 @@
+Feature: TICF journey
+  Rule: No TICF CI initially via app journey
+    Background: User starts a medium-confidence journey and has existing TICF record with no CI
```
Joe-Edwards-GDS: (similarly - wording is a bit off, there is no existing record, it's just returning a default response with no CIs)

api-tests/features/ticf/ticf-journey.feature
```diff
@@ -0,0 +1,326 @@
...
+        | type | RiskAssessment               |
+        | txn  | randomUuid                   |
+
+    @Build
+    Scenario: TICF CRI request - with no CI in separate session
+      Then I get a 'P2' identity with a 'TICF' VC
+      And the TICF VC has properties
+        | cis  |                              |
+        | type | RiskAssessment               |
+        | txn  | randomUuid                   |
```
Joe-Edwards-GDS: This seems to entirely supersede the previous scenario - do we need both?

api-tests/features/ticf/ticf-journey.feature
```diff
@@ -0,0 +1,326 @@
...
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P0' identity with a 'TICF' VC
+      And the TICF VC has properties
+        | cis  | BREACHING                    |
+        | type | RiskAssessment               |
+        | txn  | randomUuid                   |
+
+    @Build
+    @Build
```
Joe-Edwards-GDS: double annotation?

api-tests/features/ticf/ticf-journey.feature
```diff
@@ -0,0 +1,326 @@
+Feature: TICF journey
+  Rule: No TICF CI initially via app journey
```
Joe-Edwards-GDS: I think these scenarios are categorised in a slightly surprising way - I think the areas we're actually testing here are:
- TICF responses on a new identity proving journey
  - Covering the various types of responses
- TICF responses on a reuse journey
  - Covering the various types of responses
  - The fact it didn't get a CI in the initial part of the journey is largely irrelevant, and we could use Chris's VC setup step to shortcut

For some cases (e.g. the timeout) we could consider adding a scenario that only runs on one of the journeys (IPV/reuse)

api-tests/src/clients/ticf-management-api.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+};
+
+export const postUserToTicfManagementApi = async (
+  userId: string,
+  cis: string[] | undefined,
+  type: string,
+  responseDelay: number,
+  txn: string | undefined,
+  statusCode: string | undefined,
```
Joe-Edwards-GDS: Might be easier to just have this accept the object generated above? (Along with `userId`)

api-tests/src/clients/ticf-management-api.ts
```diff
@@ -0,0 +1,47 @@
...
+) => {
+  const statusCodeUrlParam = statusCode ? `/statuscode/${statusCode}` : "";
+  const url = `${config.ticf.managementTicfUrl}/management/user/${userId}${statusCodeUrlParam}`;
+
+  const response = await fetch(url, {
+    method: "POST",
+    headers: {
+      "x-api-key": config.ticf.managementTicfApiKey,
```
Joe-Edwards-GDS, line 33: (we've not done it anywhere else, but strictly speaking this should also include `content-type: application/json`)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -235,13 +266,27 @@ When(
   },
 );
 
-Then("I get a {string} identity", function (this: World, vot: string): void {
-  if (!this.identity) {
-    throw new Error("No identity found.");
-  }
+Then(
+  /I get a '(\w+)' identity(?: with(out)? a '([\w-]+)' VC)?/,
```
Joe-Edwards-GDS: Do you think this might be neater as a separate step, like the identity assertions in update details?
e.g.
```
Then I get a 'P2' identity
And my identity includes a 'TICF' credential
```
(or similar)

makes it easier to combine with others :)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -51,6 +59,22 @@ const describeResponse = (response: JourneyEngineResponse): string => {
   return `unknown ${JSON.stringify(response)}`;
 };
 
+const CREDENTIAL_ISSUERS: Record<string, string> = {
+  TICF: "https://ticf.stubs.account.gov.uk",
+};
+
+const checkForVc = (
```
Joe-Edwards-GDS: I might reframe this as an assertion, so the language is more consistent, something like:
`assertIdentityContainsVc(identity, vc, not)`

(but not very important!)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -334,3 +379,38 @@ Then(
     );
   },
 );
+
+When(
+  "there is an existing TICF record for the user with details",
```
Joe-Edwards-GDS: These could plausibly be extracted into a `ticf-steps` file (cautious about this growing into a monster!), but not crucial.

api-tests/src/steps/ipv-steps.ts
```diff
@@ -235,13 +266,27 @@ When(
...
+  ): void {
+    if (!this.identity) {
+      throw new Error("No identity found.");
+    }
+    assert.equal(this.identity.vot, vot);
 
-  assert.equal(this.identity.vot, vot);
-});
+    if (expectedVc) {
+      this.vcs = decodeCredentialJwts(
```
Joe-Edwards-GDS: We should probably set this every time - otherwise it's a bit surprising that you need to have added an 'expected' VC to call other steps that use `World.vcs`

Joe-Edwards-GDS: Alternatively, I wonder if it'd be better to move the check into the other step, so it's more like:
```
Then I get a 'P2' identity
And there is a TICF VC [with properties ...]
```

api-tests/src/utils/jwt-decoder.ts
```diff
@@ -0,0 +1,12 @@
+import * as jose from "jose";
+import { VcJwtPayload } from "../types/external-api.js";
+
+export const decodeCredentialJwts = (jwts: string[]) => {
+  const mappedJwts: Record<string, VcJwtPayload> = {};
+  jwts.forEach((jwt) => {
+    const claims = jose.decodeJwt(jwt) as VcJwtPayload;
+    mappedJwts[claims.iss] = claims;
+  });
```
Joe-Edwards-GDS, line 9: If we were using Node 21+ we could use a groupBy here :(

`return jwts.map(jose.decodeJwt).groupBy((vc) => vc.iss)`

api-tests/src/utils/jwt-decoder.ts
```diff
@@ -0,0 +1,12 @@
+import * as jose from "jose";
+import { VcJwtPayload } from "../types/external-api.js";
+
+export const decodeCredentialJwts = (jwts: string[]) => {
+  const mappedJwts: Record<string, VcJwtPayload> = {};
+  jwts.forEach((jwt) => {
+    const claims = jose.decodeJwt(jwt) as VcJwtPayload;
```
Joe-Edwards-GDS: IMO it would be safer and more natural to specify the generic parameter here: `const vc = jose.decodeJwt<VcJwtPayload>(jwt)`

Joe-Edwards-GDS (commented):
Sorry for all the comments D:

Looks good, but I'm keen to take this opportunity to get the tests into tidy shape, and not inherit lots of the problems from the existing ones!

api-tests/features/ticf/failed-ticf-requests.feature
thebauSoftwire: I'm not entirely sure, I guess whether a TICF VC is still returned for the same user despite failing the first time? I was reluctant on deleting parts of a test bc I didn't want to accidentally delete an edge case

api-tests/src/steps/ipv-steps.ts
thebauSoftwire: I've added it to the `/^I use the OAuth response to get my (identity|MFA reset result)$/` step since this is where we get the identity and to keep `there is a {VC} credential` more extendable

api-tests/src/utils/jwt-decoder.ts
thebauSoftwire: :')

api-tests/features/ticf/failed-ticf-requests.feature
thebauSoftwire: An undefined txn means the request has timed out but I have updated it to default to a random string unless it's specified in the table to be an empty string (which is the only case where we'd specify the txn) 

Also decided to keep the table for flexibility but it does mean that we'd still need to define an empty table like this: :')
```
Given TICF CRI will respond with default parameters
      | | |
```

Joe-Edwards-GDS: I don't think you need it at all if it's all defaults - the TICF CRI stub should respond with a 'default' success response if you don't prime it with anything :)

Joe-Edwards-GDS: (unless you specifically want to remove an existing setup)

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
thebauSoftwire: Have removed the type check but since this is an assertion step would it be better to explicit about what we expect instead of having defaults? E.g. when the scenarios specify no CI journeys

thebauSoftwire: Actually ignore this, it reduces a lot of duplication and can still specify certain properties if need be

api-tests/features/ticf/failed-ticf-requests.feature
Joe-Edwards-GDS: Yeah, doesn't seem massively useful to me - but happy to leave in if you're concerned. We should check that TICF works on a reuse journey, but that's covered by some of the scenarios in the main test cases already.

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
Joe-Edwards-GDS: Oh of course this is an assertion, not the setup. Never mind then - I agree

api-tests/features/ticf/failed-ticf-requests.feature
thebauSoftwire: ohh I see, thanks :)

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
thebauSoftwire: this makes sense instead of creating separate test-cases for all journeys. But would each of these journeys also need a test covering the different TICF responses too? Would it be fine to just have them on some journeys e.g. timeout, error

Joe-Edwards-GDS: I think that would be too many scenarios, and not massively valuable - unless the different journeys have different behaviour by the point they reach TICF

api-tests/features/ticf/ticf-successful-responses.feature
```diff
@@ -0,0 +1,115 @@
...
+      Then I get a 'P2' identity
+      And the TICF VC has properties
+        | cis  |                              |
+        | type | RiskAssessment               |
+
+  Rule: TICF request times out
+    @Build
+    Scenario: Via app - TICF request has a response delay less than 5s
+      Given TICF CRI will respond with default parameters
+        | responseDelay | 4         |
```
Joe-Edwards-GDS: This should be missing a txn id, rather than a responseDelay, I think? (Otherwise it's identical to the scenario above)

api-tests/features/cimit/alternate-doc.feature
```diff
@@ -27,6 +27,9 @@ Feature: CIMIT - Alternate doc
     Then I get an OAuth response
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
+    And the TICF VC has properties
+      | cis  |                              |
+      | type | RiskAssessment               |
```
Joe-Edwards-GDS: I think we can probably omit this, now that the step above checks for the presence of a TICF VC.

So that in most scenarios we simply confirm that we have called TICF and returned the VC, and in the TICF-specific scenarios, we can check that it contains appropriate content.

api-tests/features/ticf/ticf-failed-journeys-error-scenarios.feature
thebauSoftwire: PR to turn on TICF CRI is here: https://github.com/govuk-one-login/ipv-core-common-infra/compare/pyic-7316-turn-ticf-on-in-build?expand=1

api-tests/src/clients/ticf-management-api.ts
```diff
@@ -1,41 +1,41 @@
...
-  cis: string[] | undefined,
-  type: string,
-  responseDelay: number,
-  txn: string | undefined,
-  statusCode: string | undefined,
+  parsedTicfManagementParameters: {
+    statusCode: string;
+  } & TicfManagementParameters,
 ) => {
+  const { statusCode } = parsedTicfManagementParameters;
```
Joe-Edwards-GDS, line 26: This is a bit awkward isn't it :(

I'm not sure what the rationale behind making `statusCode` a url param (rather than body param like everything else)... Perhaps a tech debt ticket for us.

Joe-Edwards-GDS (commented):
Looks much better, thanks! Just a couple of things

api-tests/features/ticf/ticf-successful-responses.feature
Joe-Edwards-GDS: oh you've already fixed :D

local-running/core.local.secrets.template.yaml
```diff
@@ -29,7 +29,7 @@ core:
     ticf:
       connections:
         stub:
-          apiKey: "TICF_API_KEY" # pragma: allowlist secret
+          apiKey: "MANAGEMENT_TICF_API_KEY" # pragma: allowlist secret
```
Joe-Edwards-GDS: I don't think this is right - the key here should be for the main TICF API, not the management API

.github/workflows/pre-merge-checks.yml
```diff
@@ -157,7 +157,8 @@ jobs:
           sed -i s/CIMIT_API_KEY/${{ secrets.API_KEY_CIMIT }}/ ../local-running/core.local.secrets.yaml && \
           sed -i s/EVCS_API_KEY/${{ secrets.API_KEY_EVCS }}/ ../local-running/core.local.secrets.yaml && \
           sed -i s/ASYNC_QUEUE_API_KEY/${{ secrets.ASYNC_QUEUE_API_KEY }}/ ../local-running/core.local.secrets.yaml && \
-          sed -i s/ASYNC_QUEUE_NAME/${{ steps.extract_queue_name.outputs.queue_name }}/ ../local-running/core.local.secrets.yaml
+          sed -i s/ASYNC_QUEUE_NAME/${{ steps.extract_queue_name.outputs.queue_name }}/ ../local-running/core.local.secrets.yaml && \
+          sed -i s/MANAGEMENT_TICF_API_KEY/${{ secrets.MANAGEMENT_TICF_API_KEY }}/ ../local-running/core.local.secrets.yaml
```
Joe-Edwards-GDS: As below, I don't think the app should need the management API key, but it will need the main API key instead!

Joe-Edwards-GDS (approved):

PR #2397, opened 2024-09-05
PYIC-7263: Bulk EVCS migration script

Wynndow:
### What changed

Bulk EVCS migration script

### Why did it change

This adds a new lamdba to perform the bulk migration of VCs from the tactical store to EVCS.

It steals a lot of logic from the report lambda, and moves some classes to a shared lib.

### Issue tracking

- [PYIC-7263](https://govukverify.atlassian.net/browse/PYIC-7263)

[PYIC-7263]: https://govukverify.atlassian.net/browse/PYIC-7263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Joe-Edwards-GDS, line 1: As written, I think we've conflated the size of the batch (i.e. the group of users that we want to migrate in this run) and the size of the page (i.e. how many we fetch from DynamoDB at a time).

I think we'll probably want to be able to control these independently... so the request is probably more like:
```
{
  "batch": {
    "id:": "some-batch-id",
    "lastEvaluatedKey": "<last-user-from-previous-batch>",
    "remaining": 10000
  },
  "pageSize": 500,
  "parallelism": 4
}
```
So this would attempt to process 10,000 records from the given user, in pages of 500.

An open question might be whether we want batch sizes larger than what can be processed within a single lambda timeout... It might be easiest just to try and avoid it!

Otherwise, we'll need to calculate the remaining users after part-completing a particular batch.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+                                HASH_USER_ID,
+                                AttributeValue.builder()
+                                        .s(previousPageLastEvaluatedHashUserId)
+                                        .build());
+        LOGGER.info(
+                LogHelper.buildLogMessage(
+                        String.format(
+                                "Starting bulk migration with exclusiveStartKey: %s",
+                                exclusiveStartKey)));
+        int parallelism = request.parallelism() == 0 ? DEFAULT_PARALLELISM : request.parallelism();
```
Joe-Edwards-GDS: I would make `request.parallelism()` an `Integer` (not `int`) - that way it's much more explicit that 'not specified' => `DEFAULT_PARALLELISM`

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+        var forkJoinPool = new ForkJoinPool(parallelism);
```
Joe-Edwards-GDS: I think strictly we should call `forkJoinPool.shutdown()` in a `finally` block around this - so it's not leaked in the next invocation of the lambda.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+        LOGGER.info(LogHelper.buildLogMessage(String.format("Parallelism: %d", parallelism)));
+
+        for (var page :
+                reportUserIdentityScanDynamoDataStore.scan(
+                        exclusiveStartKey, request.batchSize(), HASH_USER_ID, USER_ID, IDENTITY)) {
```
Joe-Edwards-GDS: We could add filters to this scan, to pre-emptively filter out things like non-P2 and already-migrated identities. It probably won't add that much time to include them, but we shouldn't need to.

We'd just need to make sure we understand what we're adding up - i.e. the 'skipped' audit events are only for 'identities that were deleted/migrated since we ran the report'.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+                LOGGER.info(LogHelper.buildLogMessage("No lastEvaluatedKey - all batches run"));
+                report.setLastEvaluatedHashUserId("Null - all batches complete");
+            }
+        }
+
+        return report.finalise();
+    }
+
+    private void migrateIdentity(ReportUserIdentityItem reportItem, BatchSummary batchSummary) {
+        // Skip any non P2 identities
```
Joe-Edwards-GDS, line 208: Note that it is possible that the user has subsequently proved a P2 identity on the same account - technically this just means the user didn't have a P2 identity when we last ran the report. In practice I expect that's not very common.

Not a problem, since any new P2 identities will already be in EVCS.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/Report.java
```diff
@@ -0,0 +1,61 @@
...
+import lombok.Data;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+@Data
+public class Report {
+    private final Map<String, BatchSummary> batchSummaries;
```
Joe-Edwards-GDS: per above, this is probably page summaries, rather than batch - which feels less useful to expose

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/Report.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void addBatchSummary(BatchSummary summary) {
+        this.batchSummaries.put(summary.getBatchId(), summary);
+    }
+
+    public void setLastEvaluatedHashUserId(String lastEvaluatedHashUserId) {
+        this.lastEvaluatedHashUserId = lastEvaluatedHashUserId;
+    }
+
+    public Report finalise() {
```
Joe-Edwards-GDS: Can we just do this in `addBatchSummary` instead?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/Report.java
```diff
@@ -0,0 +1,61 @@
+package uk.gov.di.ipv.core.bulkmigratevcs.domain;
+
+import lombok.Data;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+@Data
```
Joe-Edwards-GDS: Should possibly be `@Getter` if you only want people to update it using the defined methods

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+            } catch (Throwable e) {
+                logError(
+                        "Migration failed - error writing to tactical",
+                        reportItem,
+                        e,
+                        batchSummary.getBatchId());
+                // Send failed audit event
+                batchSummary.incrementFailedTacticalWrite(reportItem.getHashUserId());
+            }
+        } else {
```
Joe-Edwards-GDS: Not that important, but I think it might be clearer to arrange so all the skips come first, and then the default is to migrate
- If < P2 -> skipped for being incomplete
- If no VCs -> skipped because no VCs
- If already migrated -> skipped because already migrated
- Otherwise good to go

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+
+    private String getSkippedReason(List<VerifiableCredential> vcs) {
+        return vcs.isEmpty() ? "no VCs to migrate" : "all VCs already migrated";
+    }
+
+    private String getBatchId(String lastEvaluatedKey, int pageSize) {
+        // The last evaluated key with the number of items in the page - should be enough to
+        // process the same batch in future
+        return String.format(
+                "%s:%d", lastEvaluatedKey == null ? "START" : lastEvaluatedKey, pageSize);
```
Joe-Edwards-GDS: I think it might be better to pass in the `batchId`, that way we can decide later what the format is?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/BatchSummary.java
```diff
@@ -0,0 +1,55 @@
...
+        this.failedTacticalReadHashUserIds = new ArrayList<>();
+        this.failedTacticalWriteHashUserIds = new ArrayList<>();
+    }
+
+    public synchronized void incrementMigrated() {
+        migrated++;
+        total++;
+    }
+
+    public synchronized void incrementSkipped() {
```
Joe-Edwards-GDS: If we're recording this (which seems like a good idea :) ) - perhaps we should also break down the different reasons for skipping, like we do with the failures?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    private void storeUnmigratedVcsInEvcs(
+            String userId, List<VerifiableCredential> vcs, String batchId, Instant timestamp)
+            throws EvcsServiceException {
+        evcsClient.storeUserVCs(
+                userId,
+                vcs.stream()
+                        .filter(vc -> vc.getMigrated() == null)
```
Joe-Edwards-GDS: This might need thinking about - it feels a bit iffy to migrate a partial identity as CURRENT, since that should never have happened.

The only valid scenario for an incomplete migration would have been a pending F2F identity, where one half got migrated but the other did not. FWIW we didn't see any of these when running the original report, so this may be an entirely moot point.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,271 @@
...
+    private void setMigratedOnTacticalStoreVcs(List<VerifiableCredential> vcs, Instant timestamp)
+            throws VerifiableCredentialException {
+        verifiableCredentialService.updateIdentity(
+                vcs.stream()
+                        .filter(vc -> vc.getMigrated() == null)
+                        .map(
+                                vc -> {
+                                    vc.setMigrated(timestamp);
+                                    return vc;
+                                })
```
Joe-Edwards-GDS: IntelliJ helpfully tells me this can be simplified with a `peek`:
```
vcs.stream()
        .filter(vc -> vc.getMigrated() == null)
        .peek(vc -> vc.setMigrated(timestamp))
        .toList());
```

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: I think I wasn't imagining to have such large batches. Having a page be the batch seemed like a natural fit. Does having more, smaller batches seem like a problem? It would mean that if something needed replaying or rolling back it might be a bit more targeted.
The page value being supplied here is actually the "limit" value - there doesn't seem to be hard upper limit, but not supplying it would mean the max value is used. I think it might be 1MB of data returned...

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: Yeah I figured this - but thought that since we're storing all new identities in EVCS it was a moot point.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: This goes back to the question at the top - my mental model is to give the script some parameters and let it decide the batches.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: We could add partiallly migrated identities to the skipped reasons, and keep track of them. That way if we do encounter any we can decide what to do after?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: That was my original implementation, but then sonar shouted at me for using a method that's supposed to be only meant for debugging. Apparently it can be skipped by the JVM for optimisation reasons. Although if that'd actually happen when you've got it doing "stuff" I'm not sure.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,367 @@
...
+                                    .with(
+                                            PAGE_EXCLUSIVE_START_KEY,
+                                            pageSummary.getExclusiveStartKey())
+                                    .with(PAGE_ITEM_COUNT, pageSummary.getCount()));
+                    report.finalise(
+                            "Lambda close to timeout", getPageLastEvaluatedHashUserId(page));
+                    throw new StopBeforeLambdaTimeoutException();
+                }
+
+                if (report.getTotalEvaluated() >= request.batch().size()) {
```
Joe-Edwards-GDS, line 148: Does this mean we might 'overshoot' a batch, if the fetched page sizes don't line up?

Do you think we need to worry about this, or at least warn if it happens? In practice, if we make sure page size is a divisor of batch size it should be ok 🤞

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/BatchReport.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public BatchReport(String batchId) {
+        this.batchId = batchId;
+        this.pageSummaries = new ArrayList<>();
+        this.allFailedTacticalReadHashUserIds = new ArrayList<>();
+        this.allFailedTacticalWriteHashUserIds = new ArrayList<>();
+        this.allFailedEvcsWriteHashUserIds = new ArrayList<>();
+    }
+
+    public synchronized void addPageSummary(PageSummary summary) {
```
Joe-Edwards-GDS: I don't think this needs to be synchronized - it looks like it's only called on the main thread

(unlike the page ones, which do)

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,367 @@
...
+                            LogHelper.buildLogMessage(
+                                            "Lambda close to timeout - stopping execution")
+                                    .with(LOG_BATCH_ID.getFieldName(), batchId)
+                                    .with(
+                                            PAGE_EXCLUSIVE_START_KEY,
+                                            pageSummary.getExclusiveStartKey())
+                                    .with(PAGE_ITEM_COUNT, pageSummary.getCount()));
+                    report.finalise(
+                            "Lambda close to timeout", getPageLastEvaluatedHashUserId(page));
+                    throw new StopBeforeLambdaTimeoutException();
```
Joe-Edwards-GDS: Could just `break` here? (or `return report`)?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -0,0 +1,367 @@
...
+                    LOGGER.info(
+                            LogHelper.buildLogMessage("All items in batch processed")
+                                    .with(LOG_BATCH_ID.getFieldName(), batchId)
+                                    .with(
+                                            PAGE_EXCLUSIVE_START_KEY,
+                                            pageSummary.getExclusiveStartKey())
+                                    .with(PAGE_ITEM_COUNT, pageSummary.getCount()));
+                    report.finalise(
+                            "All items in batch processed", getPageLastEvaluatedHashUserId(page));
+                    throw new BatchProcessedException();
```
Joe-Edwards-GDS: Ditto - could break/return here?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/BatchReport.java
```diff
@@ -0,0 +1,57 @@
...
+        totalEvaluated += summary.getTotal();
+
+        allFailedTacticalReadHashUserIds.addAll(summary.getFailedTacticalReadHashUserIds());
+        allFailedTacticalWriteHashUserIds.addAll(summary.getFailedTacticalWriteHashUserIds());
+        allFailedEvcsWriteHashUserIds.addAll(summary.getFailedEvcsWriteHashUserIds());
+
+        this.pageSummaries.add(summary);
+    }
+
+    public void finalise(String exitReason, String nextBatchExclusiveStartKey) {
```
Joe-Edwards-GDS: I would avoid `finalise` - it could be confused with the `finalize` method (that is called when an object is about to be garbage collected).

Could just be a `setExitReason` and `setNextBatchExclusiveStartKey`?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Joe-Edwards-GDS: It sounds like the Sonar rule is slightly misguided, the point is that streams are lazy, and intermediate operations will only occur when they are needed (rather than them being dropped arbitrarily). For example:
```
Stream.of("foo", "bar", "baz")
    .peek((v) -> System.out.println(v))
    .findFirst();
```
This will (probably) only print the first entry, but the same would apply if it was done in a `.map` as well! (They are both intermediate operations).

The additional warning applies to `peek`, because you're more likely to do something side-effecty. (With a `map`, this behaviour is probably a bit more expected)

Joe-Edwards-GDS: In this case, the optimal implementation is probably to use neither of these options:
```
vcs.forEach(vc -> vc.setMigrated(timestamp));
verifiableCredentialService.updateIdentity(vcs);
```
😄

Joe-Edwards-GDS (commented):
Looking pretty good I think :)

I imagine most further changes will come out of testing now...

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: I did consider this and thought that it probably doesn't matter all that much, we won't be orders of magnitude out. I also thought it's proobably very likely that the page and batch sizes will line up, unless we decide to do something a bit odd.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: More than one break and Sonar gets angry with you. Could just return report I suppose, which would be tidier

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/domain/BatchReport.java
Wynndow: I wanted to avoid two calls, but I suppose it doesn't really matter

Joe-Edwards-GDS: Yeah, you could always call `setNextBatchExclusiveStartKey` on every iteration (regardless of termination) if you wanted to reduce duplication, but either way is fine.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Joe-Edwards-GDS: Yeah, I think we can make it work.

(I'd just like to avoid having to explain 'we said batch 1 would be X users, but it actually ended up being Y', if possible!)

Joe-Edwards-GDS (approved):

PR #2398, opened 2024-09-05
PYIC-7383 TICF special routing

Joe-Edwards-GDS:
### What changed

Added a special case to the journey engine to skip TICF CRI steps if the CRI is disabled.

### Why did it change

Easier to manage than manually defining fallback states which just skip a state

### Issue tracking

- [PYIC-7383](https://govukverify.atlassian.net/browse/PYIC-7383)

[PYIC-7383]: https://govukverify.atlassian.net/browse/PYIC-7383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/public/render.mjs
+++ b/journey-map/public/render.mjs
@@ -19,7 +19,7 @@ const addDefinitionOptions = (definition, disabledOptions, featureFlagOptions) =
 
 // Traverse the journey map to collect the available 'disabled' and 'featureFlag' options
 export const getOptions = (journeyMaps) => {
-    const disabledOptions = [];
+    const disabledOptions = ['ticf'];
     const featureFlagOptions = [];
 
     Object.values(journeyMaps).forEach((journeyMap) => {
@@ -30,6 +30,9 @@ export const getOptions = (journeyMaps) => {
             });
         });
     });
+
+    disabledOptions.sort();
+    featureFlagOptions.sort();
 
     return { disabledOptions, featureFlagOptions };
 };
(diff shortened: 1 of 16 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS, line 1: As discussed, this does add some more special-casing into the journey engine (it now needs to know about special rules for TICF CRI), so we should only merge this if the benefit outweighs this!

journey-map/public/render.mjs
```diff
@@ -31,6 +31,9 @@ export const getOptions = (journeyMaps) => {
         });
     });
 
+    disabledOptions.sort();
+    featureFlagOptions.sort();
```
thebauSoftwire, line 35: Just curious, why do we need to sort these?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -62,27 +42,6 @@ states:
...
-                  - IPV_USER_DETAILS_UPDATE_END
-                auditContext:
-                  successful: false
-                checkJourneyContext:
-                  rfc:
-                    targetState: COULD_NOT_UPDATE_DETAILS_PAGE_RFC
-                    auditEvents:
-                      - IPV_USER_DETAILS_UPDATE_END
-                    auditContext:
-                      successful: false
```
thebauSoftwire, line 85: Nice!

thebauSoftwire (approved):
Nice! Just one comment for my own learning

journey-map/public/render.mjs
Joe-Edwards-GDS: We don't need to, but it makes them more consistent (otherwise it's just whatever order they happen to be discovered in the YAML)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -300,6 +304,23 @@ private State executeStateTransition(
             }
         }
 
+        // Special case to skip TICF CRI if it has been disabled,
+        // to save us defining lots of fallback routes in the journey map
+        if (result.state() instanceof BasicState basicState
+                && basicState.getResponse() instanceof ProcessStepResponse processResponse
+                && TICF_CRI_LAMBDA.equals(processResponse.getLambda())
+                && !configService.getBooleanParameter(
+                        CREDENTIAL_ISSUER_ENABLED, Cri.TICF.getId())) {
```
Wynndow, line 313: I'd be tempted to stick this in a well named private method

Wynndow (approved):
I like it, although I'm a bit concerned about how we make future us remember this is a thing.

PR #2402, opened 2024-09-06
PYIC-7341 Add JWKS API tests

Joe-Edwards-GDS:
### What changed

Add tests for JWKS API

### Why did it change

JWKS API is a critical part of core-back, as other services rely on it to function correctly

### Issue tracking

- [PYIC-7341](https://govukverify.atlassian.net/browse/PYIC-7341)

[PYIC-7341]: https://govukverify.atlassian.net/browse/PYIC-7341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.secrets.baseline
```diff
@@ -134,10 +134,17 @@
...
+        "8oxAsAsoYjwxkCLgCgolXOb4SOQevXntNGvcl_h_FUZOofw71PEcmrzRSukkUD8GDS8Cnb-GejtzKWpHHvkNUHnoWd_Fpo-1yBNXZrytXYh65Fn9H7j1j7-cVGolNvjWhccVc8QaZbMRdUOXBsncaj6IF4_TG_fWO7mVyZLoupM",
+        "30p3BamouZONQ5hXyN4qIdZTJoGjSxNGHCDf0EFjoqPsJ_gij1z8eOxWtk9YfP0-fpX5Ji5klYziB0H_UP0XFeBxYn3iAMWMCpamkR0CSVNdHj9ZADqW4huVynssHDLL3PbuG-btohuTnVBDikr9OVt185BitO_LDGPugad5NOM",
+        "zWMGxAwuvkDKl4twvA-rY-gSB2dkwH4m52UO_4Fl29YVcAXvhHpIBqAteonCPlGgBjnwtTQh5A_DjSZ54zhyldslbjWXYVwAStRWBqFvDO0tY4DwX5R4HKnc736S1OIShmX3PHirAcOI3QvHL_BvvWgGRi18TwqnVxvoQzEr8hE",
+        "rJAH28B6PLDjg1HAxeOr--p-6EtXvzVuAV8GmRNvxMmaQGlrjA_4rhs4CDzZzf6IxZtxqShBY7I1HBbNEfIGXUnPVqaieqYwzB4RLCSXF8PtOusKBTtCUAuxf4IhXMmE84eCDN4y2A0nz3OwPrglaqZcSZdUMTEAYtr4AAtFiLE",
+        "27L2sofeonh_IZ3G2YEEti-udMK55hc-38asC8_Y1WcK2LupegpwaLNYFa7ZmQQ9lj-0RVv8vn4LkMkgf8_FGbmK4mv9-ul9aTS7y8tkfGDTcvGjUb4LMyvNnd3iZzKuDB-A5ZbvGzXQOq70kvlPetPHfyzuZyU9ZyU_rr5P3H0"
```
Joe-Edwards-GDS, line 144: (these are the parts of the test RSA encryption key - they tend to be a lot bigger than EC keys!)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -382,3 +383,27 @@ When("I call the healthcheck endpoint", async function (this: World) {
 Then("the healthcheck is successful", async function (this: World) {
   assert.ok(this.healthCheckResult);
 });
+
+When("I call the JWKS endpoint", async function (this: World) {
+  this.jwksResult = await externalClient.jwks();
+});
+
+Then("I get a valid JWKS response", async function (this: World) {
```
Joe-Edwards-GDS, line 391: This is a slightly weaker test than the one in ipv-core-tests (which asserts on the exact contents).

IMO this is probably ok - otherwise we'll just have a load of places that need updating if we ever have to do a key rotation.

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JwksHandler.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public void jwks(Context ctx) throws ParseException {
+        var signingKey =
+                ECKey.parse(configService.getSecret(ConfigurationVariable.SIGNING_KEY_JWK));
+        var encryptionKey =
+                RSAKey.parse(configService.getSecret(ConfigurationVariable.JAR_ENCRYPTION_KEY_JWK));
+
+        var jwks = new JWKSet(List.of(signingKey, encryptionKey));
+
+        ctx.json(jwks.toJSONObject(true));
```
Joe-Edwards-GDS, line 28: (the `true` here is `publicKeysOnly: true` - it shouldn't publish private keys!)

api-tests/src/steps/ipv-steps.ts
Joe-Edwards-GDS: @blakeyp would be good to get a perspective from you here, as I think you were involved in the addition of those tests (after the incident...)

We could still keep the tests in ipv-core-tests, but I'm not sure how useful they are - they don't test production, so if someone was actively tampering then they'd just change prod, not build :shrug:

blakeyp: Agree we don't need to assert the exact contents but could we check that the keys array is non-empty? in the incident the endpoint was still up the array was just empty

Joe-Edwards-GDS: Yeah, that should be tested implicitly by 'must be at least one signing and one encryption key' - but I could add an explicit assertion on length too.

blakeyp: Ah cool that looks sufficient

blakeyp (approved):

PR #2404, opened 2024-09-09
PYIC-7308: return codes api tests

thebauSoftwire:
### What changed
Adds api tests for return codes

### Why did it change
We want to increase coverage of api tests

### Issue tracking

- [PYIC-7308](https://govukverify.atlassian.net/browse/PYIC-7308)

[PYIC-7308]: https://govukverify.atlassian.net/browse/PYIC-7308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/return-codes.feature
+++ b/api-tests/features/return-codes.feature
@@ -1,174 +1,154 @@
 Feature: Return exit codes
-  Rule: no return codes
-    @Build
-    Scenario:  Successful journey with identity - no CIs
-      When I start a new 'medium-confidence' journey
-      Then I get a 'page-ipv-identity-document-start' page response
-      When I submit an 'appTriage' event
-      Then I get a 'dcmaw' CRI response
-      When I submit 'kenneth-passport-valid' details to the CRI stub
-      Then I get a 'page-dcmaw-success' page response
-      When I submit a 'next' event
-      Then I get an 'address' CRI response
-      When I submit 'kenneth-current' details to the CRI stub
-      Then I get a 'fraud' CRI response
-      When I submit 'kenneth-score-2' details to the CRI stub
-      Then I get a 'page-ipv-success' page response
-      When I submit a 'next' event
-      Then I get an OAuth response
-      When I use the OAuth response to get my identity
-      Then I get a 'P2' identity
-      And I don't get any return codes
+  @Build
+  Scenario: no return codes - successful journey with identity - no CIs
+    When I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-passport-valid' details to the CRI stub
+    Then I get a 'page-dcmaw-success' page response
+    When I submit a 'next' event
+    Then I get an 'address' CRI response
+    When I submit 'kenneth-current' details to the CRI stub
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-score-2' details to the CRI stub
+    Then I get a 'page-ipv-success' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+    And I don't get any return codes
 
-  Rule: non-ci-breaching code returned
-    Scenario: Failed identity journey with no CI - user doesn't hold appropriate documents
-      Given I start a new 'medium-confidence' journey
-      Then I get a 'page-ipv-identity-document-start' page response
-      When I submit an 'end' event
-      Then I get a 'page-ipv-identity-postoffice-start' page response
-      When I submit an 'end' event
-      Then I get a 'pyi-escape' page response
-      When I submit a 'end' event
-      Then I get an OAuth response
-      When I use the OAuth response to get my identity
-      Then I get a 'P0' identity
-      And I get 'non-ci-breaching' return code
+  Scenario: non-ci-breaching code returned - failed identity journey with no CI - user doesn't hold appropriate documents
+    Given I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'end' event
+    Then I get a 'page-ipv-identity-postoffice-start' page response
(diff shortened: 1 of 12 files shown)


api-tests/features/return-codes.feature
```diff
@@ -0,0 +1,174 @@
...
+        Then I get a 'f2f' CRI response
+        When I get an 'access_denied' OAuth error from the CRI stub
+        Then I get a 'pyi-another-way' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P0' identity
+        And I get 'non-ci-breaching' return code
+
+      Scenario: Error from DCMAW CRI
```
thebauSoftwire: Not sure whether it's that much more useful to have both of these scenarios? cc: @SreeBandla01

api-tests/features/return-codes.feature
```diff
@@ -0,0 +1,174 @@
...
+        Then I get a 'pyi-post-office' page response
+        When I submit a 'end' event
+        Then I get a 'pyi-another-way' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P0' identity
+        And I get 'non-ci-breaching' return code
+
+    Scenario: CI mitigated in separate session but failed to complete journey
```
thebauSoftwire: @SreeBandla01 Decided to remove
`Scenario: NO CI Failed journey via F2F No Photo with correct exit code` (see ticket)
as it was testing the same thing as the above scenarios but just had an extra bit for testing the return code in a separate journey which this scenario already tests. Let me know if you disagree

SreeBandla01: Thats fine @thebauSoftwire

api-tests/src/clients/cri-stub-client.ts
```diff
@@ -5,6 +5,25 @@ import {
...
+  fraud: "fraud-cri",
+  ukPassport: "passport-cri",
+  drivingLicence: "driving-license-cri",
+  claimedIdentity: "claimed-identity-cri",
+  f2f: "f2f-cri",
+  nino: "nino-cri",
+  dwpKbv: "dwp-kbv-cri",
+  hmrcKbv: "hmrc-kbv-cri",
+  bav: "bav-cri",
+};
```
thebauSoftwire: `generateVc` was failing because the cri id didn't match the subdomain of the CRI e.g.
```
Given the subject already has the following credentials
| CRI             | scenario           |
| drivingLicence  | some-scenario |
```
so I added some mapping here. Not sure if this is the best approach though as it means it would need to be kept up to date

api-tests/features/return-codes.feature
```diff
@@ -0,0 +1,174 @@
+Feature: Return exit codes
+  Rule: no return codes
```
Joe-Edwards-GDS: (Are these `Rule`s adding much? Feels like they could just be merged with the scenario names...)

api-tests/features/return-codes.feature
```diff
@@ -0,0 +1,174 @@
...
+      When I submit an 'end' event
+      Then I get a 'pyi-escape' page response
+      When I submit a 'end' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P0' identity
+      And I get 'non-ci-breaching' return code
+
+    Rule: Failed to complete journey
+      Background: Start alternate doc journey with no CI
```
Joe-Edwards-GDS: I don't think this is an 'alternate doc' journey? Looks like thin file + failure to use dcmaw/f2f

api-tests/features/return-codes.feature
Joe-Edwards-GDS: I'd be happy for these to be combined

api-tests/src/clients/cri-stub-client.ts
```diff
@@ -5,6 +5,25 @@ import {
...
 import config from "../config/config.js";
 
+const CREDENTIAL_ISSUER_SUBDOMAINS: Record<string, string> = {
+  ticf: "ticf",
+  kbv: "experian-kbv-cri",
+  dcmaw: "dcmaw-cri",
+  address: "address-cri",
+  fraud: "fraud-cri",
+  ukPassport: "passport-cri",
+  drivingLicence: "driving-license-cri",
```
Joe-Edwards-GDS, line 15: Is this really `driving-license-cri` 😢   (as opposed to `driving-licence`)

api-tests/src/clients/cri-stub-client.ts
Joe-Edwards-GDS: I guess the alternative is just to use the 'mapped' version in the table, but it's a bit weird that it doesn't match our usual `criId` format. This is probably easier to use :)

(cc @Wynndow - unless you had a different idea for how this would work?)

api-tests/src/clients/cri-stub-client.ts
```diff
@@ -5,6 +5,25 @@ import {
 } from "../types/cri-stub.js";
 import config from "../config/config.js";
 
+const CREDENTIAL_ISSUER_SUBDOMAINS: Record<string, string> = {
```
Joe-Edwards-GDS: Strictly this should be `STUB_CREDENTIAL_ISSUER_SUBDOMAINS` I think - to avoid confusion with 'real' stubs.

Joe-Edwards-GDS (commented):
A couple of comments but looks good!

api-tests/src/clients/cri-stub-client.ts
thebauSoftwire: yeah :')

Joe-Edwards-GDS (approved):

api-tests/src/clients/cri-stub-client.ts
Wynndow: Eww. No, sorry, I think this is probably as reasonable as it gets. We're all used to using the criId's we have, but the subdomains don't align. The other option is to fix that...

PR #2414, opened 2024-09-10
PYIC-7320: additional inherited identity api tests

thebauSoftwire:
### What changed
Adds additional inherited identity api tests

### Why did it change

### Issue tracking

- [PYIC-7320](https://govukverify.atlassian.net/browse/PYIC-7320)

[PYIC-7320]: https://govukverify.atlassian.net/browse/PYIC-7320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/inherited-identity.feature
+++ b/api-tests/features/inherited-identity.feature
@@ -19,34 +19,22 @@ Feature: Inherited Identity
       | medium-confidence-pcl250        | kenneth-vot-pcl250-passport       | PCL250            |
       | medium-confidence-pcl250        | kenneth-vot-pcl250-driving-permit | PCL250            |
 
-  Scenario: Migrate PCL 200 HMRC profile successfully with no evidence and returns with P2
-    Given I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity 'alice-vot-pcl200-no-evidence'
+  Scenario Outline: Migrate <expected-identity> HMRC profile successfully with no evidence and returns with <vtr> - requires identity to be proved for stronger profile
+    Given I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity '<inherited-identity>'
     Then I get an OAuth response
     And an 'IPV_INHERITED_IDENTITY_VC_RECEIVED' audit event was recorded [local only]
     When I use the OAuth response to get my identity
-    Then I get a 'PCL200' identity
-    When I start a new 'medium-confidence' journey
+    Then I get a '<expected-identity>' identity
+    When I start a new '<return-journey-type>' journey
     Then I get a 'page-ipv-identity-document-start' page response
 
-  Scenario: Migrate PCL 200 HMRC profile successfully with no evidence and returns with P2/PCL250
-    Given I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity 'alice-vot-pcl200-no-evidence'
-    Then I get an OAuth response
-    And an 'IPV_INHERITED_IDENTITY_VC_RECEIVED' audit event was recorded [local only]
-    When I use the OAuth response to get my identity
-    Then I get a 'PCL200' identity
-    When I start a new 'medium-confidence-pcl250' journey
-    Then I get a 'page-ipv-identity-document-start' page response
+    Examples:
+    | inherited-identity            | expected-identity | return-journey-type      | vtr       |
+    |  alice-vot-pcl200-no-evidence | PCL200            | medium-confidence        | P2        |
+    |  alice-vot-pcl200-no-evidence | PCL200            | medium-confidence-pcl250 | P2/PCL250 |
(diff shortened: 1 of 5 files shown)


api-tests/features/inherited-identity.feature
```diff
@@ -18,20 +18,110 @@ Feature: Inherited Identity
...
+
+  Scenario: Migrate PCL250 HRMC profile successfully with no evidence and returns with P2/PCL250
+    Given I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity 'alice-vot-pcl250-no-evidence'
     Then I get an OAuth response
     And an 'IPV_INHERITED_IDENTITY_VC_RECEIVED' audit event was recorded [local only]
     When I use the OAuth response to get my identity
-    Then I get a 'PCL200' identity
+    Then I get a 'PCL250' identity
     When I start a new 'medium-confidence-pcl250' journey
+    Then I get an OAuth response
```
Joe-Edwards-GDS, line 45: Should this also get the identity and assert that it's (still) a PCL250?

api-tests/features/inherited-identity.feature
```diff
@@ -18,20 +18,110 @@ Feature: Inherited Identity
       | medium-confidence-pcl250        | kenneth-vot-pcl250-passport       | PCL250            |
       | medium-confidence-pcl250        | kenneth-vot-pcl250-driving-permit | PCL250            |
 
-  Scenario: Migrate PCL 200 HMRC profile successfully with no evidence and returns with P2
-    Given I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity 'alice-vot-pcl200-no-evidence'
+  Scenario Outline: Migrate <expected-identity> HMRC profile successfully with no evidence and returns with <vtr> - requires identity to be proved for stronger profile
```
Joe-Edwards-GDS: Not super important but this reads a bit weirdly to me... What do you think of something like:
`Migrating a <expected-identity> HMRC profile and returning with <vtr> VTR requires a P2 identity to be proved`

e.g. `Migrating a PCL200 HMRC profile and returning with P2 VTR requires a P2 identity to be proved`

api-tests/features/inherited-identity.feature
```diff
@@ -18,20 +18,110 @@ Feature: Inherited Identity
...
+    When I submit an 'next' event
+    Then I get an OAuth response
+    And an 'IPV_INHERITED_IDENTITY_VC_RECEIVED' audit event was recorded [local only]
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+
+    Examples:
+    | inherited-identity                | evidence-state   |
+    | alice-vot-pcl200-no-evidence      | with no evidence |
+    | kenneth-vot-pcl250-driving-permit | with evidence    |
```
Joe-Edwards-GDS: FWIW I think the level of confidence (PCL200/PCL250) is more relevant than the evidence state here

api-tests/src/steps/ipv-steps.ts
```diff
@@ -315,11 +317,24 @@ Then(
...
+          .vc as IdentityCheckCredentialClass)
+      : undefined;
+
+    const namePart = (
+      inheritedIdentity
+        ? (inheritedIdentity.credentialSubject as IdentityCheckSubjectClass)
+        : this.identity[identityCredential]
+    ).name?.[0].nameParts.find((np) => {
+      return field === np.type;
+    });
```
Joe-Edwards-GDS: This change shouldn't really be necessary - suggests there might be an issue in the way we find the name :(

Which scenario break if you don't do this?

Joe-Edwards-GDS: If it is broken, we might need to raise a separate ticket (unless it's a very easy fix)

Joe-Edwards-GDS (commented):
Looks good, thanks! A couple of questions - particularly the last one

We'll want a bunch more scenarios when we come to fine-tune inherited ids later (cc @SreeBandla01 ) but I think it'd be a waste of time to try and define them all now, since they may well change.

api-tests/features/inherited-identity.feature
thebauSoftwire: reads way better! Thanks :) (why are words so hard)

api-tests/src/steps/ipv-steps.ts
thebauSoftwire: note from slack: this was bc in UserIdentityService, the inherited identity with no evidence was being marked as unsuccessful bc it had an empty evidence block so we mark inherited identities as successful

Joe-Edwards-GDS (dismissed):

Wynndow (approved):

PR #2399, opened 2024-09-05
PYIC-7306: add COI fast follow api tests

shivanshuit914:
## Proposed changes

- Create COI fast follow changes in core back api tests

### What changed

- Added API tests feature file

### Issue tracking

- [PYIC-7306](https://govukverify.atlassian.net/browse/PYIC-7306)
 

[PYIC-7306]: https://govukverify.atlassian.net/browse/PYIC-7306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2415, opened 2024-09-10
PYIC-7307: Update details failure test cases 

Wynndow:
### What changed

Update details failure test cases 

### Why did it change

This adds more test cases for failure scenarios when updating details,
both for reuse and for repeat fraud check.

### Issue tracking

- [PYIC-7307](https://govukverify.atlassian.net/browse/PYIC-7307)

[PYIC-7307]: https://govukverify.atlassian.net/browse/PYIC-7307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/data/cri-stub-requests/dcmaw/kenneth-changed-given-name-passport-valid/credentialSubject.json
```diff
@@ -11,7 +11,7 @@
       "nameParts": [
         {
           "type": "GivenName",
-          "value": "Michael"
+          "value": "Ken"
```
Joe-Edwards-GDS, line 14: Is there a reason you changed this? I think it's probably clearer to use a totally different name(?)

I guess it's a bit odd that it's under `kenneth-changed-given-name` if the name is `Michael` 🤔

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
```diff
@@ -0,0 +1,98 @@
+@Build @RfcFailure
```
Joe-Edwards-GDS: What are you using these tags for?

I'm not opposed to them, but I think we need a clear and consistent approach - otherwise we'll end up with a load of random tags all over the place.

If you want to run a particular test suite, you can also do that in a couple of other ways:
- `npm run test:local -- features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature` (by file)
- `npm run test:local -- --name 'Given name change'` (by scenario name)

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
```diff
@@ -0,0 +1,98 @@
...
+  Scenario: Given name change - family name changed
+    When I submit 'kenneth-changed-family-name-driving-permit-valid' details to the CRI stub
+    Then I get a 'page-dcmaw-success' page response
+    When I submit a 'next' event
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-changed-family-name-score-2' details to the CRI stub
+    Then I get a 'sorry-could-not-confirm-details' page response
+
+  Scenario: Given name change - breaching CI received from TICF CRI
+    Given I activate the 'ticfCriBeta' feature set
```
Joe-Edwards-GDS: TICF CRI is now on-by-default for local/dev/build, so you shouldn't need this :)

(same for the reuse version)

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Joe-Edwards-GDS, line 1: I like this - very clear what all the cases are :)

When we write tests for the fast-follow work, we might want to copy these and update appropriately, which will overlap a bit with the work started in https://github.com/govuk-one-login/ipv-core-back/pull/2399

That should mean once we clean up the feature flag we can just delete the old version of the tests and switch to the new one 🤞 

cc @SreeBandla01 / @shivanshuit914 / @thebauSoftwire

api-tests/data/cri-stub-requests/dcmaw/kenneth-changed-given-name-passport-valid/credentialSubject.json
Wynndow: This was just dedupliucation - we had two different "change name" cases - Ken and Michael. Happy to revert to Michael if it's more different, but I'm not sure it should matter too much.

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Wynndow: Just for making it easy to run a particular section of tests. if we're grouping them in feature files already, feels like that might be a useful thing to be able to run on it's own, but yeah consistency would be good.

I think that as we generally use tags across the board maybe we should stick to tags rather than file names or scenario names. Althought I suppose people can do what they prefer - the tags just enable a more familiar option.

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Wynndow: 🤦

api-tests/data/cri-stub-requests/dcmaw/kenneth-changed-given-name-passport-valid/credentialSubject.json
Wynndow: Looking at it again - the vast majority of the change name credentialSubjects are using Ken. And if feels a more likely scenario, so I'm tempted to leave it as is.

Joe-Edwards-GDS: The main thing I wanted to avoid was any implication that the name being changed had to match the previous one (e.g. that Kenneth -> Ken is ok, but Kenneth -> Michael is not) - you should be using this journey to change your legal name, not just use a nickname.

(But also fine to leave)

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Joe-Edwards-GDS: Yeah, I guess I'd prefer to group by using files/directories, rather than arbitrary tags - so it helps us navigate the tests when writing them, as well as running them!

Note you can do directories as well, e.g. `npm run test:local -- features/repeat-fraud-check/`

Wynndow: OK fair - I'll drop them

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
SreeBandla01: Yes @Joe-Edwards-GDS We just need new version of the tests on or after fast follow goes out

Joe-Edwards-GDS (approved):

PR #2417, opened 2024-09-10
BAU: Add birthdate to BAV VCs in pact tests

DanCorderIPV:
### What changed

BAV pact tests

### Why did it change

DoB field has been added to the VCs


MikeCollingwood (approved):
LGTM

PR #2418, opened 2024-09-10
BAU: turn on CI api tests for build

thebauSoftwire:
### What changed
- turns on majority of api tests on for build (with exception of return code ones)
- updates config to allow running of cimit api tests in build

### Why did it change
The obscured CI codes have been added to build which means any api tests that were being skipped in build due to CIs can now be run in build


api-tests/secure-pipeline/run-tests.sh
```diff
@@ -19,6 +19,12 @@ export CRI_STUB_GEN_CRED_API_KEY
 MANAGEMENT_TICF_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/credentialIssuers/ticf/connections/stub/apiKey | jq -r .SecretString)
 export MANAGEMENT_TICF_API_KEY
 
+MANAGEMENT_TICF_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/credentialIssuers/ticf/connections/stub/apiKey | jq -r .SecretString)
+export MANAGEMENT_TICF_API_KEY
+
+MANAGEMENT_CIMIT_STUB_API_KEY=$(aws ssm get-parameter --name /tests/core-back-build/cimit_api_key | jq -r .Parameter.Value)
```
thebauSoftwire, line 22: Do we expect these `/tests/core-back-build/cimit_api_key` (see [here](https://github.com/govuk-one-login/ipv-core-common-infra/blob/7b04ca09bbe46a229aa4c045999b950f77f99690/utils/config-mgmt/app/configs/core.build.params.yaml#L117)) to be removed once the api tests have enough coverage? In which case, it might be better to add this to secret manager instead?

api-tests/.env.build
```diff
@@ -14,7 +14,6 @@ TICF_STUB_BASE_URL="https://ticf.stubs.account.gov.uk"
 
 # These aren't needed for build but are required to have a value for config
 CIMIT_STUB_BASE_URL="https://cimit.stubs.account.gov.uk"
-MANAGEMENT_CIMIT_STUB_API_KEY="some-value" # pragma: allowlist secret
```
Joe-Edwards-GDS, line 17: Should this be added to the list below instead?

api-tests/secure-pipeline/run-tests.sh
```diff
@@ -19,6 +19,12 @@ export CRI_STUB_GEN_CRED_API_KEY
 MANAGEMENT_TICF_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/credentialIssuers/ticf/connections/stub/apiKey | jq -r .SecretString)
 export MANAGEMENT_TICF_API_KEY
 
+MANAGEMENT_TICF_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/credentialIssuers/ticf/connections/stub/apiKey | jq -r .SecretString)
+export MANAGEMENT_TICF_API_KEY
```
Joe-Edwards-GDS: Duplicates the lines above?

api-tests/secure-pipeline/run-tests.sh
Joe-Edwards-GDS: Good question, I'm not sure... it's probably fine to assume these will stay in place for now, and then we can deal with them more broadly later

Joe-Edwards-GDS (commented):
Nice! I assume you've tested they all pass against build :)

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2422, opened 2024-09-11
BAU: Parse CIMIT API errors

Joe-Edwards-GDS:
### What changed

Relax parsing rules on CIMIT responses to ignore unknown properties, and handle the known 'errorMessage' property.

### Why did it change

We want to be lenient when parsing error responses, so we don't fail inappropriately.

### DIFF
--- a/libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/cimit/domain/CimitApiResponse.java
+++ b/libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/cimit/domain/CimitApiResponse.java
@@ -1,7 +1,6 @@
 package uk.gov.di.ipv.core.library.cimit.domain;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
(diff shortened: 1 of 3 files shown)


libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/cimit/domain/CimitApiResponse.java
```diff
@@ -1,7 +1,6 @@
 package uk.gov.di.ipv.core.library.cimit.domain;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
-@JsonInclude(JsonInclude.Include.NON_NULL)
-public record CimitApiResponse(@JsonProperty String result, @JsonProperty String reason) {}
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record CimitApiResponse(String result, String reason, String errorMessage) {}
```
Joe-Edwards-GDS, line 6: Example from staging:
```
{
  "result":"fail",
  "reason":"BAD_VC_ISSUER",
  "errorMessage":"CRI config not found for issuer: https://ticf.account.gov.uk"
}
```

Wynndow (approved):

PR #2425, opened 2024-09-11
PYIC-7333:  API tests for the F2F journeys

shivanshuit914:
## Proposed changes

- Add API tests for the F2F journeys

### What changed

- API tests for the pending, error and successful f2f verification scenarios

### Issue tracking

- [PYIC-7333](https://govukverify.atlassian.net/browse/PYIC-7333)

[PYIC-7333]: https://govukverify.atlassian.net/browse/PYIC-7333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/p2-f2f-journey.feature
+++ b/api-tests/features/p2-f2f-journey.feature
@@ -1,4 +1,4 @@
-@Build
+@f2f
 Feature: P2 F2F journey
 
   Scenario: Pending F2F request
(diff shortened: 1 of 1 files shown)


api-tests/features/p2-f2f-journey.feature
```diff
@@ -22,6 +22,65 @@ Feature: P2 F2F journey
     When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-pending' page response
 
+  Scenario: Pending F2F request delete identity
+    # Initial journey
+    Given I start a new 'medium-confidence' journey
```
MikeCollingwood: Is there a way of summing this up into one step? Similar to `the subject already has the following credentials` but for an entire identity. (It could be an extension of the same method)

Joe-Edwards-GDS: This would be a bit trickier, as it would need to create the pending F2F record which is private to IPV Core (unlike EVCS). TBH given the hairiness around F2F it's probably not a bad thing to run through them properly, within reason.

What could be useful is putting this in a `Background` section for the tests that all require a pending F2F identity. (You can use a `Rule` if you need to group some of the scenarios to do this)

api-tests/features/p2-f2f-journey.feature
```diff
@@ -51,3 +110,80 @@ Feature: P2 F2F journey
...
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-score-2' details to the CRI stub
+    Then I get a 'f2f' CRI response
+    When I get an 'access_denied' OAuth error from the CRI stub
+    Then I get a 'pyi-another-way' page response
+
+    Examples:
+      | doc      | details                      |
+      | passport | kenneth-passport-valid       |
+      | DL       | kenneth-driving-permit-valid |
```
Joe-Edwards-GDS: We don't seem to actually use these? - can this just be a plain old `Scenario`?

api-tests/features/p2-f2f-journey.feature
```diff
@@ -51,3 +110,80 @@ Feature: P2 F2F journey
...
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-score-2' details to the CRI stub
+    Then I get a 'f2f' CRI response
+    When I get an 'temporarily_unavailable' OAuth error from the CRI stub
+    Then I get a 'pyi-technical' page response
+
+    Examples:
+      | doc      | details                      |
+      | passport | kenneth-passport-valid       |
+      | DL       | kenneth-driving-permit-valid |
```
Joe-Edwards-GDS: Ditto

api-tests/features/p2-f2f-journey.feature
shivanshuit914: Thanks, updated it. Moved common steps in background and rule :)

api-tests/features/p2-f2f-journey.feature
shivanshuit914: Aah, removed it thanks

api-tests/features/p2-f2f-journey.feature
shivanshuit914: removed it thanks

Joe-Edwards-GDS (approved):

PR #2426, opened 2024-09-11
PYIC-7321: inherited identity extended scenarios

thebauSoftwire:
### What changed
- Adding api tests for inherited identity extended scenarios
- updates step to allow for sending errors to async queue

### Why did it change

### Issue tracking

- [PYIC-7321](https://govukverify.atlassian.net/browse/PYIC-7321)

[PYIC-7321]: https://govukverify.atlassian.net/browse/PYIC-7321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/inherited-identity/extended-scenarios.feature
```diff
@@ -0,0 +1,290 @@
...
+    When I start a new 'medium-confidence-pcl200-pcl250' journey with inherited identity '<identity-details>'
+    Then I get an OAuth response
+    And an 'IPV_INHERITED_IDENTITY_VC_RECEIVED' audit event was recorded [local only]
+    When I use the OAuth response to get my identity
+    Then I get a '<expected-identity>' identity
+
+    Examples:
+    | inherited-identity | identity-details             | expected-identity |
+    | PCL200             | alice-vot-pcl200-no-evidence | PCL200            |
+    | PCL250             | kenneth-vot-pcl250-passport  | PCL250            |
```
Joe-Edwards-GDS: Do you think there's much value running this with different profiles? Feels like one would probably be sufficient...

Same applies throughout this file tbh - the important part of these tests is how inherited IDs behave with other things (CIs, F2F, etc.), rather than the profile itself

api-tests/features/inherited-identity/extended-scenarios.feature
```diff
@@ -0,0 +1,290 @@
...
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+
+    Examples:
+    | inherited-identity | identity-details             |
+    | PCL200             | alice-vot-pcl200-no-evidence |
+    | PCL250             | kenneth-vot-pcl250-passport  |
+
+  Scenario Outline: Fails to migrate <inherited-identity> HMRC profile when user has failed F2F with CI
```
Joe-Edwards-GDS: Do you know why this is specifically for F2F? Feels like it would be a quicker/easier test to use DCMAW

I think the important part of these tests is checking that you get appropriate behaviour when migrating an inherited ID with a CI, i.e.
- BREACHING -> no match
- ENHANCED_VERIFICATION_REQUIRED -> enhanced verification mitigation journey
- ALTERNATE_DOC_REQUIRED -> alternate doc mitigation journey

api-tests/features/inherited-identity/extended-scenarios.feature
```diff
@@ -0,0 +1,290 @@
...
+    Given I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-passport-valid' details to the CRI stub
+    Then I get a 'page-dcmaw-success' page response
+    When I submit a 'next' event
+    Then I get an 'address' CRI response
+    When I submit 'kenneth-current' details to the CRI stub
+    Then I get a 'fraud' CRI response
```
Joe-Edwards-GDS: I'm not sure this 'partial' journey is adding anything useful?

api-tests/features/inherited-identity/extended-scenarios.feature
```diff
@@ -0,0 +1,290 @@
+@Build
+Feature: Inherited identity extended scenarios
+  Scenario Outline: Successfully migrates <inherited-identity> HRMC profile after partial P2 journey
```
Joe-Edwards-GDS: This is a slightly odd scenario to test - given how our session store works, the partial journey should have no impact on anything!

Joe-Edwards-GDS (commented):
Looks good - I think we might just be able to streamline out a few of the cases, to try and keep our test suite speedy :)

api-tests/features/inherited-identity/extended-scenarios.feature
SreeBandla01: When You inherit Are you not checking which route it failed if not then Its ok, We have these cases in selenium tests as we seen failures in different routes at times so covering all the cases

api-tests/features/inherited-identity/extended-scenarios.feature
SreeBandla01: This is like if the user has partial vc's already in ipv core and then user try to inherit ID scenario and think of merging this scenario already in some other case then we wont need it again

api-tests/features/inherited-identity/extended-scenarios.feature
SreeBandla01: Yes, thats fine

Joe-Edwards-GDS (approved):

api-tests/features/inherited-identity/extended-scenarios.feature
Joe-Edwards-GDS: Hmm, I suppose there are a few different things that could conflict with each other here. Of course it _shouldn't_ matter which journey if you've got a CI.

Maybe let's revisit when we pick up inherited identities again?

PR #2428, opened 2024-09-12
PYIC-7035 nested web doc journey

Joe-Edwards-GDS:
### What changed

Create a new nested journey for web docs (UK Passport/DL), includes new functionality to override the entry event for a nested journey.

Use in new-P2-identity:
image

Nested journey:
image

### Why did it change

The journey map is becoming increasingly complex and repetitive (e.g. across P1 and P2), and judicious use of nested journeys could help address this.

### Issue tracking

- [PYIC-7035](https://govukverify.atlassian.net/browse/PYIC-7035)

[PYIC-7035]: https://govukverify.atlassian.net/browse/PYIC-7035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -23,12 +23,14 @@ states:
   ALTERNATE_DOC_PASSPORT:
     events:
       next:
-        targetState: MITIGATION_04_DL_NO_MATCH_PAGE
+        targetState: WEB_DL_OR_PASSPORT
+        targetEntryEvent: alternate-doc-invalid-dl
```
Joe-Edwards-GDS, line 27: Overriding the entry event like this makes the journey engine a bit clunky, but the alternative seemed worse in the YAML (creating several copies of the nested journey with different entry events specified there).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -712,9 +629,10 @@ states:
             targetJourney: TECHNICAL_ERROR
             targetState: ERROR
       passport:
-        targetState: CRI_UK_PASSPORT_J2
+        targetState: WEB_DL_OR_PASSPORT
+        targetEntryEvent: ukPassport
```
Joe-Edwards-GDS, line 639: Annoyingly there's inconsistency here between `passport` and `ukPassport` 🙄

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -168,3 +168,246 @@ STRATEGIC_APP_TRIAGE:
...
+        lambda: check-gpg45-score
+        lambdaInput:
+          scoreType: fraud
+          scoreThreshold: 2
+      events:
+        met:
+          exitEventToEmit: next
+        unmet:
+          targetJourney: FAILED
+          targetState: FAILED
```
Joe-Edwards-GDS: This is potentially a point of discussion - do we mind nested journeys triggering a journey change, that isn't visible to the caller, or should this emit an exit event that is mapped in the parent?

As written, it's not possible to see this as a possible outcome of the nested journey, even when showing failures. (N.B. this already affects Address & Fraud!)

Wynndow: I feel like this should probably emit an exit event, which is then mapped.
TBH, trying to grok exactly what's happening with nested journeys is difficult. If we can keep as much consistency as possible I think that would be helpful.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
Wynndow: I think this is better - it's much clearer how to read the map. before you needed to understand that the event used to transition to the nested journey state was used again to dive into it. Now you can work it out without needing that knowledge.

Wynndow (commented):
I think this is good - and it certainly declutters the map. But I do think that the change of journey thing should be mapped.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
Joe-Edwards-GDS: Cool, I'll definitely change this one and the other one-offs.

The CRI states are trickier I think - there's 11 different events (not including `next`!) that need to be mapped... At the moment they just inherit from `CRI_STATE` in the parent, which is pretty sketchy but quite convenient!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -326,16 +321,14 @@ WEB_DL_OR_PASSPORT:
         back:
           targetState: MITIGATION_DL_CRI_DRIVING_LICENCE
         returnToRp:
-          targetJourney: FAILED
-          targetState: FAILED_SKIP_MESSAGE
+          exitEventToEmit: return-to-rp
```
Joe-Edwards-GDS, line 324: This one (and the equivalent for passport) now go to slightly different places, however:
- `FAILED_SKIP_MESSAGE` and `INELIGIBLE_SKIP_MESSAGE` are identical (call TICF and return to RP)
- It's arguably more consistent to describe this case as 'ineligible' anyway - you should only reach this page if you go to the DL CRI and then press the button to say you don't have a DVA or DVLA issued licence.

The only tangible difference will be if D&A use this for tracking success rates of failed vs. ineligible (but as above, I think this change is more accurate anyway).

Wynndow (approved):
Nice! One less factor to hold in your head when looking at nested journeys...

PR #2429, opened 2024-09-12
PYIC-7450: update journey map to route to the two variants of the sorry-could-not-confirm-details page

thebauSoftwire:
### What changed
- Updated route to the two variants of the `sorry-could-not-confirm-details` page depending on the journeyContext and when `updateDetailsAccountDeletion` feature flag is enabled

### Why did it change
The `sorry-could-not-confirm-details` has two variants depending on the initial journey (rfc or reuse) which show different content so the routing needed to be updated to allow for this.

Corresponding FE PR: https://github.com/govuk-one-login/ipv-core-front/pull/1573

### Issue tracking

- [PYIC-7450](https://govukverify.atlassian.net/browse/PYIC-7450)

[PYIC-7450]: https://govukverify.atlassian.net/browse/PYIC-7450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
+++ b/api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
@@ -19,17 +19,26 @@ Feature: Repeat fraud check failures
       Then I get a 'dcmaw' CRI response
 
     @FastFollow
-    Scenario: Given name change - DCMAW access denied OAuth error
+    Scenario: DCMAW access denied OAuth error
+      Given I activate the 'updateDetailsAccountDeletion' feature set
       When I get an 'access_denied' OAuth error from the CRI stub
-      Then I get an 'sorry-could-not-confirm-details' page response
-      When I submit a 'end' event
+      Then I get an 'update-details-failed' page response
+      When I submit a 'return-to-service' event
       Then I get an OAuth response
       When I use the OAuth response to get my identity
       Then I get a 'P0' identity
       When I start a new 'medium-confidence' journey
       Then I get a 'confirm-your-details' page response
 
-    Scenario: Given name change - breaching CI received from DCMAW
+    @FastFollow
+    Scenario: User is able to delete details from update-details-failed screen
+      Given I activate the 'updateDetailsAccountDeletion' feature set
+      When I get an 'access_denied' OAuth error from the CRI stub
+      Then I get an 'update-details-failed' page response
+      When I submit a 'delete' event
+      Then I get a 'delete-handover' page response
+
+    Scenario: Breaching CI received from DCMAW
       When I submit 'kenneth-driving-permit-breaching-ci' details to the CRI stub
       Then I get a 'sorry-could-not-confirm-details' page response
       When I submit a 'end' event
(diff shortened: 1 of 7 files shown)


Joe-Edwards-GDS (commented):
I think this will need to be a little bit more complicated, sorry!

If you look at the table of scenarios, we need to make sure that receiving a CI during a reuse journey also results in the new variant.

This probably means updating `CRI_STATE` (in `update-name.yaml` and `update-address.yaml`) to point CI failures (`fail-with-ci`, `enhanced-verification` and `alternate-doc-xx`) to a different place...

Joe-Edwards-GDS:
One other thought - should we include API tests for the new routing as well? Or do you think that's covered by another ticket.

(In future I'm very keen for tickets to include API tests by default! - as you would with unit tests)

api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
```diff
@@ -1,90 +1,141 @@
...
+            When I submit a 'address-only' event
+            Then I get an 'address' CRI response
+            When I get an 'access_denied' OAuth error from the CRI stub
+            Then I get an 'sorry-could-not-confirm-details' page response
+            When I submit a 'end' event
+            Then I get an OAuth response
+            When I use the OAuth response to get my identity
+            Then I get a 'P0' identity
+            When I start a new 'medium-confidence' journey
+            Then I get a 'page-ipv-reuse' page response
```
thebauSoftwire, line 242: I've only added additional failure cases here but these don't test fast-follow changes yet since there isn't a way of testing the variations in content of the page until the reinstate identity work has been merged where we'd want to test against the identity. So can wait until that work is merged or can get this in and add a separate ticket to update these api tests for fast follow?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -131,21 +147,40 @@ states:
...
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
       alternate-doc-invalid-dl:
-        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
       alternate-doc-invalid-passport:
-        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
       fail-with-ci:
-        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
```
Joe-Edwards-GDS: This is a little sneaky, but I think actually the CI events here (`fail-with-ci` etc.) should _also_ go to `COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_INVALID_ID`. And probably equivalently for the `COULD_NOT_UPDATE_DETAILS_PAGE`!

It's a bit of an unusual case, but if TICF produces a CI at this stage, it will also invalidate your existing identity...

api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
Joe-Edwards-GDS: (Answered in slack)

Joe-Edwards-GDS (commented):
Looking good, just a couple of test cases once Chris's change is merged

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
thebauSoftwire: good catch!

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Joe-Edwards-GDS, line 1: You might find it easier to clean up later if the fast-follow versions were in a separate file - but happy to have them all together.

Either way we can tidy up after releasing the feature :)

Joe-Edwards-GDS (approved):
Thanks!

PR #2430, opened 2024-09-12
PYIC-7380: Reinstate identity for failed COI journeys

Wynndow:
### What changed
Reinstate identity for failed COI journeys

### Why did it change

If a user fails a COI journey but still has a valid identity, they should be able to still use it.

This adds some logic to the reset-session-identity lambda with a new reset type to delete the current session VCs and restore the old identity from the long term store. This is plumbed in with some routing in the journey maps.

### Issue tracking

- [PYIC-7380](https://govukverify.atlassian.net/browse/PYIC-7380)

[PYIC-7380]: https://govukverify.atlassian.net/browse/PYIC-7380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
+++ b/api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
@@ -24,7 +24,19 @@ Feature: Identity reuse update details failures
         When I submit a 'continue' event
         Then I get an OAuth response
         When I use the OAuth response to get my identity
-        Then I get a 'P0' identity
+        Then I get a 'P2' identity
+        When I start a new 'medium-confidence' journey
+        Then I get a 'page-ipv-reuse' page response
+
+    @FastFollow
+    Scenario: Given name change - fail-with-no-ci from DCMAW
+        Given I activate the 'updateDetailsAccountDeletion' feature set
+        When I submit 'kenneth-passport-verification-zero' details to the CRI stub
+        Then I get an 'update-details-failed' page response
+        When I submit a 'continue' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P2' identity
         When I start a new 'medium-confidence' journey
         Then I get a 'page-ipv-reuse' page response
 
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
@@ -9,6 +9,11 @@ states:
       next:
         targetState: EVALUATE_GPG45_SCORES
 
+  START_NO_STORE:
+    events:
+      next:
+        targetState: EVALUATE_GPG45_SCORES_NO_STORE
+
   # Parent states
   CRI_TICF_STATE:
     events:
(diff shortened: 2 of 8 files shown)


api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
```diff
@@ -24,7 +24,7 @@ Feature: Identity reuse update details failures
         When I submit a 'continue' event
         Then I get an OAuth response
         When I use the OAuth response to get my identity
-        Then I get a 'P0' identity
+        Then I get a 'P2' identity
```
Joe-Edwards-GDS, line 27: 🎉

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
```diff
@@ -9,6 +9,11 @@ states:
       next:
         targetState: EVALUATE_GPG45_SCORES
 
+  START_NO_STORE:
```
Joe-Edwards-GDS, line 12: Do you think it makes sense to have this in here or `failed.yaml`? I'm not totally sure, but I might slightly lean towards failed, since it's not 'completing' a journey and the states are totally distinct (apart from 'return to RP', which doesn't really count).

Perhaps what I really want is for this to be `new-identity-success.yaml` or something similar 🤔 

For reference, the existing reuse journey does not use this subjourney!

api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
Joe-Edwards-GDS, line 1: Should there also be a scenario for `fail-with-no-ci` as well? (As the other case that hits this)

Joe-Edwards-GDS (commented):
This looks good, I'm just wondering if the new states are in the right place

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
Wynndow: I started adding the states in `failed.yaml` but if felt weird - because they weren't about failing so I shifted them here.

`new-identity-success.yaml` - I think that would confuse things. These new states are for an old identity?

api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature
Wynndow: Potentially yes - although I don't think there was a corresponding selenium tests, so not created here. I'll add one in.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/evaluate-scores.yaml
Joe-Edwards-GDS: Yeah sorry I meant, if this file were called `new-identity-success.yaml` then its purpose would be clear, and the answer would be obvious (that the new states shouldn't be in it!)

I suppose the alternative is something like `return-identity` or something, but then it's odd that normal reuse journeys don't use it.

(at the moment 'evaluate-scores' is a bit unclear to me, as it's not just about evaluating scores, but storing the result and returning to the RP)

Joe-Edwards-GDS (approved):
I'm still not totally convinced about `evaluate-scores.yaml` as a thing, but happy to consider that a separate problem and get this merged.

PR #2433, opened 2024-09-12
PYIC-7336: Add enhanced-verification API tests

Wynndow:
### What changed

Add enhanced-verification API tests

### Why did it change

This adds more scenarios for the mitigation of CIs with enhanced verification.

### Issue tracking

- [PYIC-7336](https://govukverify.atlassian.net/browse/PYIC-7336)

[PYIC-7336]: https://govukverify.atlassian.net/browse/PYIC-7336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/cimit/enhanced-verification-mitigation-dcmaw.feature
```diff
@@ -0,0 +1,118 @@
...
+
+  Background:
+    # Navigate to KBV CRI and apply NEEDS-ENHANCED-VERIFICATION CI
+    Given I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I get an 'access_denied' OAuth error from the CRI stub
+    Then I get a 'page-multiple-doc-check' page response
+    When I submit a 'drivingLicence' event
```
Wynndow, line 12: In the original selenium tests the document used when gathering the initial CI was parameterised - they ran the test with a passport and a driving permit. Unfortunatley you can't parameterise a Background block, so I've settled on just using one document. It doesn't feel like there's much value in doing both - but happy to be corrected if people feel it's important.

Joe-Edwards-GDS: Agree - I'm a big fan of trying to cut cases that don't actually add any real coverage

(Although they are still much quicker, the number of API test cases is already getting pretty big!)

api-tests/features/cimit/enhanced-verification-mitigation-dcmaw.feature
```diff
@@ -0,0 +1,118 @@
...
+      Then I get a 'dcmaw' CRI response
+      When I submit 'kenneth-passport-with-breaching-ci' details to the CRI stub
+      Then I get a 'pyi-no-match' page response
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P0' identity
+
+    Scenario: Separate session DCMAW enhanced verification mitigation - DCMAW is unavailable
+      When I start a new 'medium-confidence' journey with feature set 'dcmawOffTest'
```
Joe-Edwards-GDS, line 111: This is interesting - I wonder if we should have a separate test suite for 'xxx CRI disabled'? (As a separate ticket)

Otherwise this feels like a slightly random test case, for a very rare event!

api-tests/features/cimit/enhanced-verification-mitigation-dcmaw.feature
```diff
@@ -0,0 +1,118 @@
...
+      Then I get a 'page-ipv-identity-document-start' page response
+      When I submit an 'appTriage' event
+      Then I get a 'pyi-technical' page response
```
Joe-Edwards-GDS, line 115: I assume this is deliberate - to avoid unnecessarily pushing people towards F2F... This is the sort of thing that could use a comment tbh, otherwise it looks like it might be a bug.

api-tests/features/cimit/enhanced-verification-mitigation-f2f.feature
```diff
@@ -0,0 +1,138 @@
...
+      When I start a new 'medium-confidence' journey and return to a 'page-ipv-reuse' page response
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P2' identity
+
+      Examples:
+        | document-details             |
+        | kenneth-passport-valid       |
+        | kenneth-driving-permit-valid |
```
Joe-Edwards-GDS: (I'd be quite happy to drop one of the docs here, but up to you)

Joe-Edwards-GDS (dismissed):
Looks sensible, but a couple of comments where we may be able to streamline

api-tests/features/cimit/enhanced-verification-mitigation-dcmaw.feature
Wynndow: It's deliberate - it's specifically there in the journey map if we get that event but dcmaw is disabled. But I think it was probably more of a "we need to handle this case in a simple way". Git blame bring me to this ticket: https://govukverify.atlassian.net/browse/PYIC-3984
I'll add a comment

api-tests/features/cimit/enhanced-verification-mitigation-f2f.feature
Wynndow: Yeah - I think I will. We're already testing the ability of those docs to mitigate elsewhere so not much need to duplicate.

Joe-Edwards-GDS (approved):

PR #2435, opened 2024-09-13
BAU: Update Experian & Nino 400 status codes to 403s

MikeCollingwood:
### What changed

- Update Experian & Nino 400 status codes to 403s
- Update `fetchAccessToken` to reference accurate http status code on `CriApiException`

### Why did it change

- Requested by provider

### DIFF
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/TokenTests.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/TokenTests.java
@@ -199,7 +199,7 @@ class TokenTests {
 
         // Assert
         assertEquals("Invalid token request", exception.getErrorResponse().getMessage());
-        assertEquals(HTTPResponse.SC_FORBIDDEN, exception.getHttpStatusCode());
+        assertEquals(HTTPResponse.SC_BAD_REQUEST, exception.getHttpStatusCode());
     }
 
     @NotNull
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/TokenTests.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/TokenTests.java
@@ -132,7 +132,7 @@ class TokenTests {
     }
 
     @Pact(provider = "ExperianKbvCriTokenProvider", consumer = "IpvCoreBack")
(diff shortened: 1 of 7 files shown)


mikebeeby (dismissed):

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -138,7 +138,7 @@ private BearerAccessToken fetchAccessToken(
                         errorObject.getDescription(),
                         errorObject.getHTTPStatusCode());
                 throw new CriApiException(
-                        HTTPResponse.SC_BAD_REQUEST, ErrorResponse.INVALID_TOKEN_REQUEST);
+                        httpResponse.getStatusCode(), ErrorResponse.INVALID_TOKEN_REQUEST);
```
DanCorderIPV: Do we definitely want to do this? Everywhere else CriApiException is thrown the return code is hard coded. That return code is then returned by core-back so this will change core-back's behaviour in certain circumstances.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/experianKbvCri/TokenTests.java
```diff
@@ -198,7 +198,7 @@ void fetchAccessToken_whenCalledAgainstExperianKbvCriWithInvalidAuthCode_throwsA
 
         // Assert
         assertEquals("Invalid token request", exception.getErrorResponse().getMessage());
-        assertEquals(400, exception.getHttpStatusCode());
+        assertEquals(403, exception.getHttpStatusCode());
```
DanCorderIPV: As you've done in other places this would be nice to change to SC_FORBIDDEN

DanCorderIPV (changes requested):

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
MikeCollingwood: The other noticeable hardcoding is for `fetchVerifiableCredential`, which can be updated to not be hardcoded either (since the logic is duplicated in the catch statement in `ProcessCriCallbackHandler`). Other hardcoded values come from thrown exceptions, which makes sense.

I have walked through all paths from where I am suggesting to update: `fetchAccessToken` & `fetchVerifiableCredential`. There are no functional differences with updating these status codes, it's just that `core-front` gets a more accurate status code - the intention behind the refactor.

That being said, it has a low reward/risk ratio to do this refactor so I will remove it from this PR.

DanCorderIPV (approved):

PR #2440, opened 2024-09-16
PYIC-7339: add f2f volume control api tests

thebauSoftwire:
### What changed
Adds F2F volume control tests

### Why did it change

### Issue tracking

- [PYIC-7339](https://govukverify.atlassian.net/browse/PYIC-7339)

[PYIC-7339]: https://govukverify.atlassian.net/browse/PYIC-7339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/p2-f2f-volume-control.feature
Joe-Edwards-GDS, line 1: Minor thing, but I think 'volume control' is a pretty unclear name for this - now that it's released, it's just 'the web CRI journeys'.

I would probably just move all these cases into `p2-web-journey.feature` - it can use the same background and everything.

api-tests/features/p2-f2f-volume-control.feature
```diff
@@ -0,0 +1,117 @@
+@Build
+Feature: F2F Volume Control
+
+  Background: Start P2 journey via app but DCMAW returns OAuth error
```
Joe-Edwards-GDS: Although it's implemented as an OAuth error, this is really 'Start P2 journey and ineligible for app' (it's what happens if you select that you have no appropriate smartphone, for example)

api-tests/features/p2-f2f-volume-control.feature
```diff
@@ -0,0 +1,117 @@
...
+    Given I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I get an 'access_denied' OAuth error from the CRI stub
+    Then I get a 'page-multiple-doc-check' page response
+
+  Scenario Outline: Allows use of <alternative-doc-cri> when <initial-cri> CRI returns OAuth error
```
Joe-Edwards-GDS: Similar - from a user perspective, this is a drop-out (e.g. they realise they don't have a UK DL after all)

api-tests/features/p2-f2f-volume-control.feature
```diff
@@ -0,0 +1,117 @@
...
+    When I submit a 'ukPassport' event
+    Then I get a 'ukPassport' CRI response
+    When I get an 'access_denied' OAuth error from the CRI stub
+    Then I get a 'prove-identity-another-type-photo-id' page response
+    When I submit a 'returnToRp' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P0' identity
+
+  Scenario: User can use F2F from the prove-identity-another-type-photo-id page to receive an identity
```
Joe-Edwards-GDS: Does this duplicate the F2F case: `Successful P2 identity via F2F using <doc> - DCMAW access_denied`?

api-tests/features/p2-f2f-volume-control.feature
```diff
@@ -0,0 +1,117 @@
...
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+
+    Examples:
+    | initial-cri    | alternative-doc-cri | alternative-doc              |
+    | ukPassport     | drivingLicence      | kenneth-driving-permit-valid |
+    | drivingLicence | ukPassport          | kenneth-passport-valid       |
+
+  Scenario Outline: Mitigation of alternate-doc CI via <mitigating-cri> when <mitigating-cri> initially returns OAuth error
```
Joe-Edwards-GDS: I wonder if this (and the other CI mitigation scenario below) should live in `cimit/alternate-doc.feature` instead?

There's always going to be some overlaps when trying to group scenarios (it's both a CIMIT scenario and a web doc scenario, so where should it be grouped), so I don't have a very strong preference.

Joe-Edwards-GDS (commented):
Scenarios look good, but a few queries about where they live, and one possible duplicate.

api-tests/features/p2-f2f-volume-control.feature
thebauSoftwire: Hmm good point. I guess this test is specifically testing that from the `prove-identity-another-type-photo-id` page (through dropping out of the DL or passport CRIs) that we can still take an F2F journey whereas `Successful P2 identity via F2F using <doc> - DCMAW access_denied` checks we can take F2F from the `page-multiple-doc-check` page

api-tests/features/p2-f2f-volume-control.feature
thebauSoftwire: I agree, I think CIMIT makes sense as the existing tests in there are also web doc routes with cimit scenarios, have moved them :)

Joe-Edwards-GDS (approved):

PR #2441, opened 2024-09-16
PYIC-7263: Send audit events for the EVCS migration

shivanshuit914:
## Proposed changes

- Send audit events for the EVCS migration

### What changed

- Sending success, fail, skipped audit events during user evcs migration. 

### Issue tracking

- [PYIC-7263](https://govukverify.atlassian.net/browse/PYIC-7263)

[PYIC-7263]: https://govukverify.atlassian.net/browse/PYIC-7263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -228,6 +253,12 @@ private void migrateIdentity(
                     e,
                     batchId,
                     pageSummary);
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_EVCS_MIGRATION_FAILURE,
+                            configService.getParameter(ConfigurationVariable.COMPONENT_ID),
+                            new AuditEventUser(reportItem.getHashUserId(), "", "", ""),
```
shivanshuit914: I am not sure, If we need to send actual userId or hash one

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -53,21 +61,33 @@ public class BulkMigrateVcsHandler implements RequestHandler<Request, BatchRepor
...
         this.forkJoinPoolFactory = forkJoinPoolFactory;
+        this.configService = configService;
+        this.auditService = auditService;
     }
 
     @ExcludeFromGeneratedCoverageReport
     public BulkMigrateVcsHandler() {
-        var configService = ConfigService.create();
+        this(ConfigService.create());
+    }
```
Wynndow: Do we need this extra constuctor? What's the use case of having a constructor that allows the ConfigService to be injected on its own?

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -207,13 +227,18 @@ private void migrateIdentity(
...
                     annotateLog(
                             "Skipping migration - not a P2 identity",
                             reportItem,
                             batchId,
                             pageSummary));
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_EVCS_MIGRATION_SKIPPED,
+                            batchId,
+                            new AuditEventUser(reportItem.getHashUserId(), "", "", ""),
```
Wynndow: I think in the audit events we should be using the users actual userID rather than the hashed value. That way audit events can be tied together to form more complete journeys.

Also wondering if it might be better to put a placeholder instead of these empty strings - just to make it clear why we don't have values for them.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -207,13 +227,18 @@ private void migrateIdentity(
...
             LOGGER.info(
                     annotateLog(
                             "Skipping migration - not a P2 identity",
                             reportItem,
                             batchId,
                             pageSummary));
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_EVCS_MIGRATION_SKIPPED,
+                            batchId,
```
Wynndow: Think you need the component ID here intead of batchId

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: Ah, as above, we should be using the userId

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -228,6 +253,12 @@ private void migrateIdentity(
                     e,
                     batchId,
                     pageSummary);
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_EVCS_MIGRATION_FAILURE,
+                            configService.getParameter(ConfigurationVariable.COMPONENT_ID),
+                            new AuditEventUser(reportItem.getHashUserId(), "", "", ""),
+                            new AuditExtensionsEvcsFailedMigration(batchId, 0, e.getMessage())));
```
Wynndow: Instead of using the error's message, I'd be tempted to use something more static - like "Failed to fetch VCs from tactical store". We're logging the error already, so will be able to get those details if needed. As we're catching all exceptions, it's possible that the message recorded like this won't actually be that useful.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -207,13 +227,18 @@ private void migrateIdentity(
...
                             "Skipping migration - not a P2 identity",
                             reportItem,
                             batchId,
                             pageSummary));
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_EVCS_MIGRATION_SKIPPED,
+                            batchId,
+                            new AuditEventUser(reportItem.getHashUserId(), "", "", ""),
+                            new AuditExtensionsEvcsSkippedMigration(batchId, "already_migrated")));
```
Wynndow: I think the reason here is wrong - this is for non P2 identities. So somthing like `not_p2_identity` perhaps

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: Same for all the audit messages added here.

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -282,15 +330,35 @@ private void migrateIdentity(
...
             LOGGER.info(annotateLog("Migrated", reportItem, batchId, pageSummary));
-            // Send migrated audit event
+            List<String> vcsCredentials =
+                    vcs.stream()
+                            .map(
+                                    vc -> {
+                                        String[] parts = vc.getVcString().split("\\.");
+                                        return parts.length > 2 ? parts[2] : "";
+                                    })
+                            .toList();
```
Wynndow: I thinkj this could be streamlined a bit - we should be pretty sure our vc strings are in good shape.

```suggestion
            var vcsCredentials = vcs.stream().map(vc -> vc.getVcString().split("\\.")[2]).toList();
```

You could even inline it into the method below.

Wynndow (commented):
GitHub won't let me comment on the line directly, but I think you need to add a call to `auditService.awaitAuditEvents()` in the finally block on line 190. The audit service sends audit events async, so we need to wait for it to finish before the lambda shuts down to stop "bad things"™️ from happening.

Also, I think you'll need to raise a PR against the [event-catalogue repo](https://github.com/govuk-one-login/event-catalogue), to let TxMA know about the new events. Their docs can be found here: https://govukverify.atlassian.net/wiki/spaces/TMA/pages/4250632243/1.+Follow+the+process+outlined+here#3.-Add-%2F-Update-event-in-event-catalogue

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
shivanshuit914: Just to inject config service, We did similar in other handler example `CheckExistingIdentityHandler`

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
shivanshuit914: Thanks forgot to replace :)

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
shivanshuit914: Thanks

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: I don't think this new constructor is used though - and we have the ability to inject a config service in the all args constructor, which is the one used by the tests.

Looking at that constructor in check existing identity - it's also unused. We should probably remove it...

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
shivanshuit914: I used single line one initially, but one of the test was failing with `java.lang.ArrayIndexOutOfBoundsException` may be I need to update data in the test setup.

lambdas/bulk-migrate-vcs/src/test/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandlerTest.java
```diff
@@ -1,13 +1,11 @@
...
 import com.amazonaws.services.lambda.runtime.Context;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.ArgumentCaptor;
-import org.mockito.Captor;
-import org.mockito.InjectMocks;
-import org.mockito.Mock;
+import org.mockito.*;
```
Wynndow: Wild card import here which we try to avoid - if you're using IntelliJ you can stop it doing this: Settings -> Editor -> Code Style -> Java -> Imports tab -> Set `Class count to use import with '*'` and `Names count to use static import with '*'` to 999

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -334,29 +331,22 @@ private void migrateIdentity(
...
             LOGGER.info(annotateLog("Migrated", reportItem, batchId, pageSummary));
-            List<String> vcsCredentials =
-                    vcs.stream()
-                            .map(
-                                    vc -> {
-                                        String[] parts = vc.getVcString().split("\\.");
-                                        return parts.length > 2 ? parts[2] : "";
-                                    })
-                            .toList();
+            var vcsCredentials = vcs.stream().map(vc -> vc.getVcString().split("\\.")[2]).toList();
```
Wynndow: Maybe rename this var to something like `vcSignatures` to be super clear

lambdas/bulk-migrate-vcs/src/test/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandlerTest.java
```diff
@@ -74,7 +69,7 @@ class BulkMigrateVcsHandlerTest {
     @Mock private PageIterable<ReportUserIdentityItem> mockPageIterable;
     @Mock private Context mockContext;
     @Mock private VerifiableCredential migratedVc;
-    @Mock private VerifiableCredential notMigratedVc;
+    @Mock private VerifiableCredential notMigratedVc = PASSPORT_NON_DCMAW_SUCCESSFUL_VC;
```
Wynndow: I think that this assignment isn't actually doing anything. The `notMigratedVc` is still a mock in the tests.

libs/evcs-migration/src/main/java/uk/gov/di/ipv/core/library/persistence/item/ReportUserIdentityItem.java
```diff
@@ -41,6 +41,11 @@ public String getHashUserId() {
         return hashUserId;
     }
 
+    @JsonIgnore
+    public String getUserId() {
+        return userId;
+    }
+
```
Wynndow: Do we need this? We've already got a getter through the lombok `@Data` annotation, and we have `@JsonIgnore` on the field, which is excluding the value from serialisation/deserialisation.

lambdas/bulk-migrate-vcs/src/test/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandlerTest.java
```diff
@@ -188,6 +202,38 @@ void handlerShouldCorrectlyHandleFailures() throws Exception {
         assertTrue(report.getAllFailedEvcsWriteHashUserIds().get(0).startsWith("04b"));
         assertEquals(1, report.getAllFailedTacticalWriteHashUserIds().size());
         assertTrue(report.getAllFailedTacticalWriteHashUserIds().get(0).startsWith("f34"));
+        verify(auditService, times(3)).sendAuditEvent(auditEventArgumentCaptor.capture());
+        assertEquals(
```
Wynndow, line 213: Might be good to also assert the contents of the audit events.

Wynndow (approved):
Looks good!

PR #2443, opened 2024-09-16
PYIC-7470: Add bulk migration batchIds to tactical store VCs

Wynndow:
**Not to be merged before https://github.com/govuk-one-login/ipv-core-common-infra/pull/1064**

### What changed

Add bulk migration batchIds to tactical store VCs

### Why did it change

This adds the batch ID from the bulk migration to the VCs in the tactical store. This allows for us to more easily deal with having to roll back a batch.

The Bravo team will be responsible for deleting VCs from EVCS. We can then run the script again and remigrate the VCs whos batch IDs match a value in config.

And for on-demand migrations in the check-existing-identity lambda we can follow similar logic - check if the users tactical VCs have a matching batch ID, and remigrate if they do.

### Issue tracking

### DIFF
--- a/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
+++ b/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
@@ -12,11 +12,13 @@ import uk.gov.di.ipv.core.bulkmigratevcs.domain.BatchReport;
 import uk.gov.di.ipv.core.bulkmigratevcs.domain.EvcsMetadata;
 import uk.gov.di.ipv.core.bulkmigratevcs.domain.PageSummary;
 import uk.gov.di.ipv.core.bulkmigratevcs.domain.Request;
+import uk.gov.di.ipv.core.bulkmigratevcs.exceptions.TooManyBatchIdsException;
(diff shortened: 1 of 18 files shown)


lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -368,6 +386,42 @@ private void migrateIdentity(
         }
     }
 
+    private boolean vcsAreFromRollbackBatch(
+            List<VerifiableCredential> vcs,
+            ReportUserIdentityItem reportItem,
+            String batchId,
+            PageSummary pageSummary)
+            throws TooManyBatchIdsException {
+        var vcBatchIds = vcs.stream().map(VerifiableCredential::getBatchId).distinct().toList();
```
Joe-Edwards-GDS, line 395: (I assume this handles `null` ok as well?)

Joe-Edwards-GDS (commented):
This looks ok - hopefully the extra checks for mismatched batches etc. won't be necessary, but shouldn't hurt.

However, I think we'll need to also update the on-demand migration logic to clear the batch ID (in `EvcsMigrationService`) - so that running an on-demand migration will correctly supersede the rolled back migration.

(If this happens before EVCS have reverted on their side we may see errors, but that's probably an acceptable impact)

Joe-Edwards-GDS (approved):

PR #2452, opened 2024-09-17
PYIC-7035 Pull fraud score check out of web doc nested journey

Joe-Edwards-GDS:
### What changed

Included the fraud score check in the nested web doc journey, which meant it inadvertently got added to the P1 journey.

Instead, we can emit different exit events, and allow the main journey to determine how to handle different events. This should be more flexible if (for example) in future we wish to allow users to follow other routes with DL + fraud score 1.

### DIFF
--- a/lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
+++ b/lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
@@ -290,6 +290,13 @@ class JourneyMapTest {
                     }
                 }
             }
+            if (nestedState instanceof NestedJourneyInvokeState nestedNestedState) {
+                for (var event : nestedNestedState.getExitEvents().values()) {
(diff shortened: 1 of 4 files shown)


lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -290,6 +290,13 @@ private static HashSet<String> getActualExitEvents(NestedJourneyInvokeState stat
                     }
                 }
             }
+            if (nestedState instanceof NestedJourneyInvokeState nestedNestedState) {
+                for (var event : nestedNestedState.getExitEvents().values()) {
+                    if (event instanceof ExitNestedJourneyEvent exitNestedJourneyEvent) {
+                        actualExitEvents.add(exitNestedJourneyEvent.getExitEventToEmit());
+                    }
+                }
+            }
```
Joe-Edwards-GDS, line 299: (Pick up exit events mapped from other exit events)

MikeCollingwood (approved):

PR #2455, opened 2024-09-17
PYIC-7482 Add nested journey for KBVs

Joe-Edwards-GDS:
### What changed

Pull all KBV journeys into a nested journey

### Why did it change

The KBVs are highly duplicated, with 6 different copies - making it quite well-suited to pulling out into a nested journey.

One consideration is that this ties our KBV priority together - we'll have to add duplication back in if we want different priorities depending on the route.

### Screenshots

KBV routing with HMRC/DWP disabled:
image

KBV routing with HMRC enabled:
image

KBV routing with DWP enabled:
image

P2 identity:
image

P1 identity:
image

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -776,3 +776,65 @@ states:
           mitigationType: enhanced-verification
       end:
         targetState: RETURN_TO_RP
+
+## Temporary inclusion of KBV states to avoid breaking existing journeys
+
+  PRE_EXPERIAN_KBV_TRANSITION_PAGE:
+    response:
+      type: page
+      pageId: page-pre-experian-kbv-transition
+    events:
+      next:
+        targetState: CRI_EXPERIAN_KBV
+
+  CRI_EXPERIAN_KBV:
+    response:
+      type: cri
+      criId: kbv
+    parent: CRI_STATE
+    events:
+      fail-with-no-ci:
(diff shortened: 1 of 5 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -925,85 +750,91 @@ states:
...
+        targetState: MITIGATION_01_IDENTITY_START_PAGE
+        auditEvents:
+          - IPV_MITIGATION_START
+        auditContext:
+          mitigationType: enhanced-verification
+      end:
+        targetState: RETURN_TO_RP
 
-  MITIGATION_PRE_EXPERIAN_KBV_TRANSITION_PAGE:
+## Temporary inclusion of KBV states to avoid breaking existing journeys
```
Joe-Edwards-GDS, line 780: (ok this has messed up the diff more than I hoped - you might want to ignore the last commit!)

blakeyp (approved):
Nice 🚀

MikeCollingwood:
The KBV CRI states don't have use the `evidence_requested` parameter, instead it is calculated from the `targetVot`. This means P1 & P2 journeys can both use this shared journey and they will both request the correct verification score (1 & 2 respectively).

Using `targetVot` is a bit of a latent method of keeping track of which journey you are on - and actually could be used to deal with the `CHECK_FRAUD_SCORE` issue from `WEB_DL_OR_PASSPORT`? I think it'd be good to choose one method of containing the P1/ P2 choice, so if `targetVot` isn't suitable then maybe it isn't for KBVs either?

PR #2462, opened 2024-09-18
BAU logging fixes

Joe-Edwards-GDS:
### What changed

- Fixed incorrect test evidence
- Added missing `kidJarHeaderEnabled` config for local running, and updated code to handle it
- Removed spurious warning logs
- Tidied up various other warning/error logs that don't use the log helper

### Why did it change

Removing spurious warning logs, and making our logging more consistent

**Probably worth reviewing each commit independently**

### DIFF
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelper.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelper.java
@@ -11,6 +11,7 @@ import com.nimbusds.jose.JWSHeader;
 import com.nimbusds.jose.JWSSigner;
 import com.nimbusds.jose.Payload;
 import com.nimbusds.jose.crypto.RSAEncrypter;
+import com.nimbusds.jose.jwk.ECKey;
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.AuthorizationRequest;
@@ -28,6 +29,7 @@ import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.service.ConfigService;
 
+import java.text.ParseException;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Date;
@@ -36,6 +38,7 @@ import java.util.Objects;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.COMPONENT_ID;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.JWT_TTL_SECONDS;
 import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.SIGNING_KEY_ID;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.SIGNING_KEY_JWK;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.KID_JAR_HEADER;
 
 public class AuthorizationRequestHelper {
@@ -62,8 +65,7 @@ public class AuthorizationRequestHelper {
 
         var headerBuilder = new JWSHeader.Builder(JWSAlgorithm.ES256).type(JOSEObjectType.JWT);
         if (configService.enabled(KID_JAR_HEADER)) {
-            String signingKid = DigestUtils.sha256Hex(configService.getParameter(SIGNING_KEY_ID));
-            headerBuilder.keyID(signingKid);
+            headerBuilder.keyID(getKid(configService));
         }
         JWSHeader header = headerBuilder.build();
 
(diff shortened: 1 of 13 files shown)


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelper.java
```diff
@@ -129,4 +131,16 @@ public static JWEObject createJweObject(RSAEncrypter rsaEncrypter, SignedJWT sig
             throw new HttpResponseExceptionWithErrorBody(500, ErrorResponse.FAILED_TO_ENCRYPT_JWT);
         }
     }
+
+    private static String getKid(ConfigService configService) {
+        if (ConfigService.isLocal()) {
+            try {
+                return ECKey.parse(configService.getSecret(SIGNING_KEY_JWK)).getKeyID();
```
Joe-Edwards-GDS, line 138: (This means if the signing key is defined with a JWK, it'll use the `kid` from that - in local it's just `test-signing-key`)

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -131,12 +123,8 @@ private BearerAccessToken fetchAccessToken(
...
                                 new ErrorObject("unknown", "unknown"));
                 LOGGER.error(
-                        "Failed to exchange token with credential issuer with ID '{}' at '{}'. Code: '{}', Description: {}, HttpStatus code: {}",
-                        criId,
-                        criConfig.getTokenUrl(),
-                        errorObject.getCode(),
-                        errorObject.getDescription(),
-                        errorObject.getHTTPStatusCode());
+                        LogHelper.buildErrorMessage(
+                                "Failed to exchange token with credential issuer", errorObject));
```
Joe-Edwards-GDS, line 127: The CRI ID is automatically added to all log messages by the handler, so there's no need to repeat it - and it turns out this was the only use of `criId` here!

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -603,27 +536,16 @@ private Optional<List<PassportDetails>> generatePassportClaim(List<VerifiableCre
...
             if (isNullOrEmpty(passport)) {
-                StringMapMessage mapMessage =
-                        new StringMapMessage()
-                                .with(
-                                        LOG_MESSAGE_DESCRIPTION.getFieldName(),
-                                        "Passport property is missing from VC or empty passport property.")
-                                .with(
-                                        LOG_CRI_ISSUER.getFieldName(),
-                                        passportVc.get().getCri().getId());
-                LOGGER.warn(mapMessage);
```
Joe-Edwards-GDS, line 614: These should not be warnings - lots of DCMAW VCs don't have a passport...

In the last week there were something like 20,000 of these type of warnings logged!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -69,60 +67,47 @@ public static String getIpvSessionId(JourneyRequest event)
...
+        var language = nullIfBlank(request.getLanguage());
 
         if (language == null) {
-            LOGGER.error(
-                    LogHelper.buildErrorMessage(
-                            "Language choice is missing from this frontend request",
-                            IP_ADDRESS_HEADER));
-            throw new HttpResponseExceptionWithErrorBody(
-                    SC_BAD_REQUEST, ErrorResponse.MISSING_LANGUAGE);
+            LOGGER.warn(LogHelper.buildErrorMessage(ErrorResponse.MISSING_LANGUAGE));
```
Joe-Edwards-GDS, line 93: I'd be willing the change this back, but IMO a missing language header should not be fatal - especially as it's only required for DWP KBVs.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -263,27 +246,19 @@ private static void validateIpvSessionId(
...
     }
 
-    private static void validateIpAddress(String ipAddress, String errorMessage)
-            throws HttpResponseExceptionWithErrorBody {
+    private static void validateIpAddress(String ipAddress) {
         if (ipAddress == null) {
-            LOGGER.error(LogHelper.buildErrorMessage(errorMessage, IP_ADDRESS_HEADER));
-            throw new HttpResponseExceptionWithErrorBody(
-                    SC_BAD_REQUEST, ErrorResponse.MISSING_IP_ADDRESS);
+            LOGGER.warn(LogHelper.buildErrorMessage(ErrorResponse.MISSING_IP_ADDRESS));
```
Joe-Edwards-GDS, line 257: Similarly - I'm not convinced a blank/missing IP Address should be fatal (I think it's only used for audit logging)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/RequestHelper.java
```diff
@@ -263,27 +246,19 @@ private static void validateIpvSessionId(
...
         }
     }
 
-    private static void validateClientOAuthSessionId(
-            String clientSessionId, StringMapMessage message) {
-        if (clientSessionId == null) {
```
Joe-Edwards-GDS, line 284: There was a subtlety here - it would log a warning if the `clientSessionId` was null, but not if it was empty.

Because of the way our VTL works in API Gateway, it would never actually be null - if the header was missing it'd default to an empty string. I've added `nullIfBlank` to get the null we expect.

In practice, it will be missing in 99% of cases anyway, because we only use this for the cross-browser recovery - so we shouldn't bother logging a warning.

Wynndow (approved):
This all looks very sensible.

PR #2463, opened 2024-09-18
BAU: Fix doc type detection

Joe-Edwards-GDS:
### What changed

Only try to deduplicate evidence VCs (i.e. those with a strength + validity)

### Why did it change

Other VCs do not correspond properly to documents - e.g. the HMRC KBV VC contains a NINO in the subject, but should not be treated as NINO evidence.

### DIFF
--- a/libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
+++ b/libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
@@ -1,7 +1,10 @@
 package uk.gov.di.ipv.core.library.gpg45;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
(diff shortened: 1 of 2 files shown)


libs/gpg45-evaluator/src/main/java/uk/gov/di/ipv/core/library/gpg45/Gpg45ProfileEvaluator.java
```diff
@@ -57,13 +61,22 @@ private List<Gpg45Scores.Evidence> deduplicateEvidences(List<VerifiableCredentia
...
                         getEvidences(
                                 idCheckVc.getEvidence() != null
                                         ? idCheckVc.getEvidence()
                                         : Collections.emptyList());
-                if (isEmpty(docType)) {
+
+                // Not an evidence VC - skip
+                if (isNullOrEmpty(evidence)) {
+                    continue;
+                }
```
Joe-Edwards-GDS, line 73: Previously it would blow up at `if (isNullOrEmpty(existing) || evidence.get(0).compareTo(existing.get(0)) > 0) {` if `evidence` was empty.

It was a race condition because it depended whether the NINO VC or the HMRC KBV VC was processed first - if the HMRC KBV VC was processed first, the condition would short-circuit and it would succeed.

Wynndow (approved):

PR #2470, opened 2024-09-20
PYIC-7494 Use vocab classes for CIMIT

Joe-Edwards-GDS:
### What changed

Use the vocab classes for CIs and Mitigations

### Why did it change

No need to use our own copies which may get out of date

### Issue tracking

- [PYIC-7494](https://govukverify.atlassian.net/browse/PYIC-7494)

[PYIC-7494]: https://govukverify.atlassian.net/browse/PYIC-7494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/build-cri-oauth-request/build.gradle
```diff
@@ -7,7 +7,6 @@ plugins {
 
 dependencies {
 	implementation libs.bundles.awsLambda,
-			libs.diVocab,
```
Joe-Edwards-GDS, line 10: (added as an api dependency in `common-services`, so this is now available anywhere that uses the common lib)

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CimitService.java
```diff
@@ -109,23 +107,41 @@ public void submitMitigatingVcList(
...
+    public List<ContraIndicator> getContraIndicators(
             String userId, String govukSigninJourneyId, String ipAddress)
             throws CiRetrievalException {
         var vc = getContraIndicatorsVc(userId, govukSigninJourneyId, ipAddress);
 
         return getContraIndicators(vc);
     }
 
-    public ContraIndicators getContraIndicators(VerifiableCredential vc)
+    public List<ContraIndicator> getContraIndicators(VerifiableCredential vc)
```
Joe-Edwards-GDS, line 118: This is the main change - returning the CIs directly from the VC object

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -22,18 +25,63 @@ public CimitUtilityService(ConfigService configService) {
         this.configService = configService;
     }
 
+    public int getContraIndicatorScore(List<ContraIndicator> contraIndicators)
```
Joe-Edwards-GDS, line 28: This was only used in `CimitUtilityService` anyway, so it makes sense to move here and keep the CIs as pure data classes.

Wynndow (approved):
Nice!

PR #2475, opened 2024-09-23
PYIC-7470: Recreate EvcsClient in bulk migration

Wynndow:
### What changed

Recreate EvcsClient in bulk migration

### Why did it change

Testing of the bulk migration script has shown that we're getting predicatable GOAWAY responses from the API gateway in front of the EVCS stub.

These are happening in clusters at a cadence of every 11500 - 11750 identities migrated. I can't find anything in AWS docs to directly suggest the reason, however I did note that an ALB will do this if it receives over 10'000 requests from any given connection. So I suspect there is a similar mechanism happening here.

To resolve this, we can generate a new EVCS client for each page of results we're processing. This should keep us well under the figure noted above. There doesn't seem to be any way to force the client to drop any open connections easily, so recreating might be the easiest.

I did consider adding a GOAWAY response as a valid reason for a retry, in the retry logic in in the client, but didn't want to impact the core system code.

### Issue tracking

### DIFF
--- a/lambdas/bulk-migrate-vcs/build.gradle
+++ b/lambdas/bulk-migrate-vcs/build.gradle
@@ -20,6 +20,7 @@ dependencies {
 	annotationProcessor libs.lombok
 
 	aspect libs.powertoolsLogging,
+			libs.powertoolsTracing,
 			libs.aspectj
 
 	testImplementation libs.junitJupiter,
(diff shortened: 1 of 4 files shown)


lambdas/bulk-migrate-vcs/src/test/java/uk/gov/di/ipv/core/bulkmigratevcs/factories/EvcsClientFactoryTest.java
Joe-Edwards-GDS, line 1: TBH I'd kinda prefer not to have this test and just skip the coverage check for this review, but not important.

The one thing worse than having no tests is having useless tests!

Joe-Edwards-GDS (dismissed):
It sounds like the 'best' solution would be to retry - and I'm not too worried about touching the prod code.

However, I'm not exactly sure how it behaves with other threads etc., so perhaps it's cleaner just to recreate the client as you've done 🤷 

I think the http clients are lazy, so you might not see the new connection overhead until you send the first request. Anecdotally it can be an extra few hundred ms - so potentially several minutes of additional time, but much better than having to dig back through random errors.

lambdas/bulk-migrate-vcs/src/test/java/uk/gov/di/ipv/core/bulkmigratevcs/factories/EvcsClientFactoryTest.java
Wynndow: It's also brought in a code smell, so I'm just going to drop it 🫠

Joe-Edwards-GDS (approved):

PR #2476, opened 2024-09-23
PYIC-7304: add journeyContext visualisation

thebauSoftwire:
### What changed
- allows journey contexts to be visualised
- fixes expanding nested journeys in place

| Description | Screenshot  |
|-|-|
| Possible targets (with or without journeyCtx) always shown | Screenshot 2024-09-23 at 08 56 51 |
| List of journeyContexts that are checked for in description | Screenshot 2024-09-23 at 08 56 16 |
| Fixes nested journeys not expanding | Screenshot 2024-09-23 at 08 56 04 |

### Why did it change
We've added journeyContexts which means the journey map visualisation tool needs to be updated to show these

### Issue tracking

- [PYIC-7034](https://govukverify.atlassian.net/browse/PYIC-7034)

[PYIC-7034]: https://govukverify.atlassian.net/browse/PYIC-7034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/public/helpers.mjs
+++ b/journey-map/public/helpers.mjs
@@ -44,7 +44,7 @@ export const getOptions = (journeyMaps, nestedJourneys) => {
     return { disabledOptions, featureFlagOptions };
 };
 
-export const resolveEventTargets = (definition, resolvedEventTargets, formData) => {
+export const resolveEventTargets = (definition, formData, resolvedEventTargets) => {
     const resolvedTargets = resolvedEventTargets || [];
 
     // Look for an override for disabled CRIs
@@ -53,20 +53,16 @@ export const resolveEventTargets = (definition, resolvedEventTargets, formData) 
     if (disabledResolution) {
         return resolveEventTargets(
             {...definition.checkIfDisabled[disabledResolution], journeyContext: definition.journeyContext},
-            resolvedTargets,
-            formData);
+            formData,
+            resolvedTargets);
     }
 
-    const journeyContexts = Object.keys(definition.checkJourneyContext || {});
-    if (journeyContexts.length > 0) {
-        for (let n = 0; n < journeyContexts.length; n++) {
-            const targets = resolveEventTargets(
-                    {...definition.checkJourneyContext[journeyContexts[n]], journeyContext: journeyContexts[n]},
-                    undefined,
-                    formData)
-            resolvedTargets.push(...targets);
-        }
-    }
+    Object.keys(definition.checkJourneyContext || {}).forEach(journeyContext => {
+        const targets = resolveEventTargets(
+            {...definition.checkJourneyContext[journeyContext], journeyContext: journeyContext},
+            formData)
+        resolvedTargets.push(...targets);
+    })
 
     // Look for an override for feature flags
     const featureFlags = formData.getAll('featureFlag');
@@ -74,8 +70,8 @@ export const resolveEventTargets = (definition, resolvedEventTargets, formData) 
     if (featureFlagResolution) {
         return resolveEventTargets(
             {...definition.checkFeatureFlag[featureFlagResolution], journeyContext: definition.journeyContext},
-            resolvedTargets,
-            formData);
+            formData,
+            resolvedTargets);
     }
 
     return [...resolvedTargets, definition];
--- a/journey-map/public/helpers.mjs
+++ b/journey-map/public/helpers.mjs
@@ -14,16 +14,27 @@ const addDefinitionOptions = (definition, disabledOptions, featureFlagOptions) =
 };
 
 // Traverse the journey map to collect the available 'disabled' and 'featureFlag' options
-export const getOptions = (journeyMaps) => {
+export const getOptions = (journeyMaps, nestedJourneys) => {
     const disabledOptions = ['ticf'];
     const featureFlagOptions = [];
 
-    Object.values(journeyMaps).forEach((journeyMap) => {
-        Object.values(journeyMap.states).forEach((definition) => {
-            const events = definition.events || definition.exitEvents || {};
-            Object.values(events).forEach((def) => {
-                addDefinitionOptions(def, disabledOptions, featureFlagOptions);
-            });
+    const states = [
+        ...Object.values(journeyMaps)
+            .flatMap(journeyMap => Object.values(journeyMap.states)),
+        ...Object.values(nestedJourneys)
+            .flatMap(nestedJourney => Object.values(nestedJourney.nestedJourneyStates)),
+    ];
+
+    states.forEach((definition) => {
+        const events = definition.events || definition.exitEvents || {};
+        Object.values(events).forEach((def) => {
+            addDefinitionOptions(def, disabledOptions, featureFlagOptions);
+        });
+    });
+
+    Object.values(nestedJourneys).forEach(nestedJourney => {
+        Object.values(nestedJourney.entryEvents).forEach((def) => {
+            addDefinitionOptions(def, disabledOptions, featureFlagOptions);
         });
     });
 
--- a/journey-map/public/helpers.mjs
+++ b/journey-map/public/helpers.mjs
@@ -1,5 +1,41 @@
+const addDefinitionOptions = (definition, disabledOptions, featureFlagOptions) => {
+    Object.entries(definition.checkIfDisabled || {}).forEach(([opt, def]) => {
+        if (!disabledOptions.includes(opt)) {
+            disabledOptions.push(opt);
+        }
+        addDefinitionOptions(def, disabledOptions, featureFlagOptions);
+    });
+    Object.entries(definition.checkFeatureFlag || {}).forEach(([opt, def]) => {
+        if (!featureFlagOptions.includes(opt)) {
+            featureFlagOptions.push(opt);
+        }
+        addDefinitionOptions(def, disabledOptions, featureFlagOptions);
+    });
+};
+
+// Traverse the journey map to collect the available 'disabled' and 'featureFlag' options
+export const getOptions = (journeyMaps) => {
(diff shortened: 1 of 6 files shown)


journey-map/public/render.mjs
```diff
@@ -70,73 +26,69 @@ const expandParents = (journeyStates, otherStates) => {
...
-            Object.entries(subjourney.entryEvents).forEach(([entryEvent, def]) => {
+            Object.entries(entryEvents).forEach(([entryEvent, entryEventDef]) => {
+                mapTargetStateToExpandedState(entryEventDef, subJourneyState);
+
                 Object.values(journeyMap).forEach((journeyDef) => {
                     if (journeyDef.events?.[entryEvent]) {
-                        const target = resolveEventTarget(journeyDef.events[entryEvent], formData);
-                        if (target.targetState === state && !target.targetEntryEvent) {
-                            target.targetState = `${def.targetState}_${state}`;
+                        const target = resolveEventTargets(journeyDef.events[entryEvent], undefined, formData).find(t => !t.journeyContext)
```
thebauSoftwire: `resolveEventTargets` atm only returns multiple targets if there is a journeyContext in the definition (since we only replace targets when checking feature flags or when CRIs are disabled and only want to show the different routes we can take depending on journeyCtx) so we take the first target definition without a journeyContext

journey-map/public/render.mjs
```diff
@@ -70,73 +26,69 @@ const expandParents = (journeyStates, otherStates) => {
...
+                        if (target.targetState === subJourneyState && !target.targetEntryEvent) {// if (target.targetState === subJourneyState && !target.targetEntryEvent) {
+                            journeyDef.events[entryEvent] = entryEventDef;
```
thebauSoftwire, line 83: We replace the event with the whole target definition as we also want to include the `{..., checkJourneyContext}` bits which is handled in `renderTransitions`

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -225,6 +225,7 @@ states:
         targetState: CHECK_FRAUD_AFTER_DL
       next-passport:
         targetState: KBV_PHOTO_ID
+        targetEntryEvent: next
```
Joe-Edwards-GDS, line 228: ack - do you think this was broken in prod as well?

journey-map/public/helpers.mjs
```diff
@@ -0,0 +1,125 @@
...
+    if (disabledResolution) {
+        return resolveEventTargets(
+            {...definition.checkIfDisabled[disabledResolution], journeyContext: definition.journeyContext},
+            resolvedTargets,
+            formData);
+    }
+
+    const journeyContexts = Object.keys(definition.checkJourneyContext || {});
+    if (journeyContexts.length > 0) {
+        for (let n = 0; n < journeyContexts.length; n++) {
```
Joe-Edwards-GDS: I think this is equivalent to
```
Object.keys(definition.checkJourneyContext || {}).forEach((journeyContext) => {
  const targets = resolveEventTargets(
    { ...definition.checkJourneyContext[journeyContext], journeyContext },
    undefined,
    formData);
  resolvedTargets.push(...targets);
});
```

journey-map/public/helpers.mjs
```diff
@@ -0,0 +1,125 @@
...
+            {...definition.checkFeatureFlag[featureFlagResolution], journeyContext: definition.journeyContext},
+            resolvedTargets,
+            formData);
+    }
+
+    return [...resolvedTargets, definition];
+}
+
+const addJourneyContextFromDefinition = (definition, journeyContexts) => {
+    Object.entries(definition.checkIfDisabled || {}).forEach(([_, def]) => {
```
Joe-Edwards-GDS, line 81: (If you don't need the key, you can just use `Object.values()`, but this is fine)

journey-map/public/index.mjs
```diff
@@ -1,7 +1,8 @@
 import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
 import svgPanZoom from 'https://cdn.jsdelivr.net/npm/svg-pan-zoom@3.6.1/+esm';
 import yaml from 'https://cdn.jsdelivr.net/npm/yaml@2.3.2/+esm';
-import { getOptions, render } from './render.mjs';
+import { render } from './render.mjs';
+import {getJourneyContexts, getNestedJourneyStates, getOptions} from "./helpers.mjs";
```
Joe-Edwards-GDS: (tiny thing but can we space this like the others?)

journey-map/public/index.mjs
```diff
@@ -249,6 +276,20 @@ const renderSvg = async (selectedJourney, selectedNestedJourney, formData) => {
     });
     // Pan to correct header offset
     svgPanZoomInstance.panBy({ x: 0, y: headerContent.offsetHeight / 2 });
+
+    // Style journeyContext transitions
```
Joe-Edwards-GDS: It's possible to do this in the render too, which might be neater:

You can add a `linkStyle` declaration, but this requires you to know the index of the link you want to style (which might require a bit of refactoring, idk) - https://mermaid.js.org/syntax/flowchart.html?id=flowcharts-basic-syntax#styling-links:
e.g.: `linkStyle 1,2,7 color:blue;`

You can include HTML in the transition definition, e.g.:
`SOURCE_STATE-->|next - <span class="journeyContext">context: foo</span>|TARGET_STATE`

The second option feels best to me!

journey-map/public/render.mjs
```diff
@@ -163,54 +115,62 @@ const renderTransitions = (journeyStates, formData) => {
...
-
-            // Special case for disabling TICF, to match the special case in the journey engine
-            if (journeyStates[resolvedTarget.targetState]?.response?.lambda === 'call-ticf-cri' &&
-                formData.getAll('disabledCri').includes('ticf')) {
-                resolvedTarget = resolveEventTarget(
-                    journeyStates[resolvedTarget.targetState].events.next, formData);
-            }
+            let resolvedEventTargets = resolveEventTargets(def, undefined, formData);
+
+            for (let t = 0; t < resolvedEventTargets.length; t++)  {
```
Joe-Edwards-GDS: `resolvedEventTargets.forEach`?

journey-map/public/helpers.mjs
```diff
@@ -0,0 +1,125 @@
...
+        });
+    });
+
+    disabledOptions.sort();
+    featureFlagOptions.sort();
+
+    return { disabledOptions, featureFlagOptions };
+};
+
+export const resolveEventTargets = (definition, resolvedEventTargets, formData) => {
```
Joe-Edwards-GDS: If `resolvedEventTargets` is typically `undefined` to start with, you might want to leave it last in the args list - that way you can just omit it when calling in the other files.

If you don't like that, you could have a private one with all three args, and an exported one with just two.

journey-map/public/render.mjs
Joe-Edwards-GDS: I'm not totally sure I follow this - don't we (potentially) need to replace targets with a journeyContext as well?

Not too worried if you want to make this a bit of a simplification though - I pretty much never use this option anyway!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
thebauSoftwire: PR for fix here: https://github.com/govuk-one-login/ipv-core-back/pull/2477

journey-map/public/render.mjs
thebauSoftwire: From what I understand from before, this bit is just updating the targetStates of the entry events to a mapped subjourney target value (e.g. `targetState: TARGET` => `targetState: TARGET_SUBJOURNEY` to display which states are part of a subjourney, in which case we don't need to worry about replacing with a journeyContext bc mapping of all nestedJourneyStates is done above (from line 50) and we handle the transitions between these (inc when there is a journeyContext) in `renderTransitions`

journey-map/public/index.mjs
thebauSoftwire: should we add prettier (maybe even eslint) to the journey map subfolder? Could do as a separate ticket to keep these changes separate?

journey-map/public/render.mjs
thebauSoftwire: I don't think this should be a `forEach` bc it won't let us iterate through the added targets (on line 126) so I've used a for loop instead

Joe-Edwards-GDS: Ah I see... Yeah I think in general it's better to try and avoid mutating an array you're iterating over, but it's not too bad if you're just appending.

I think you could still use a `for (const resolvedTarget of resolvedEventTargets) { ... }` but not too fussed (and again, it's not totally clear what the semantics will actually be!)

journey-map/public/index.mjs
Joe-Edwards-GDS: Yeah, it's definitely big enough to warrant it now...

Definitely a separate PR - if you think it's quick then I'd be happy to review as a 'BAU' change, but otherwise feel free to create a ticket :)

Joe-Edwards-GDS (dismissed):

journey-map/public/render.mjs
thebauSoftwire: ooo didn't know there was a difference between `for...in...` and `for...of...` 😮

Joe-Edwards-GDS (approved):

PR #2478, opened 2024-09-23
PYIC-7470: Only recreate EVCS client when needed

Wynndow:
### What changed

Only recreate EVCS client when needed

### Why did it change

Recreating the EVCS client solved the GOAWAY issue, but it introduced a new one. The audit service was unable to open a socket due to too many files being open.

This updates the logic to only recreate the client when the number of migrated identities is approaching 10_000. This should dramatically reduce the number of client created.

There doesn't seem to be an easy way to close an HttpClient in Java 17 (in 21 it's been made AutoCloseable), which would have been nicer.

This also handles errors thrown by the AuditService to ensure that we get the report back from the lambda. The errors will be logged in the application logs.

### Issue tracking

### DIFF
--- a/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
+++ b/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
@@ -29,6 +29,7 @@ import uk.gov.di.ipv.core.library.config.EnvironmentVariable;
 import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
 import uk.gov.di.ipv.core.library.dto.EvcsCreateUserVCsDto;
 import uk.gov.di.ipv.core.library.enums.Vot;
+import uk.gov.di.ipv.core.library.exception.AuditException;
 import uk.gov.di.ipv.core.library.exception.EvcsServiceException;
 import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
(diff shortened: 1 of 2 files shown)


lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -139,6 +144,15 @@ public BatchReport handleRequest(Request request, Context context) {
                             USER_ID,
                             IDENTITY)) {
 
+                if ((report.getTotalMigrated() + page.count()) > EVCS_CLIENT_GOAWAY_LIMIT) {
```
Joe-Edwards-GDS: Will we ever hit this twice in one request? (i.e. when we hit 20k records)

I wonder if we should try going down the retry-in-EvcsClient approach 🤔

Wynndow: Yeah, we actually might. And I've just noticed that as soon as we breach 10_000 this logic becomes silly.

Wynndow: I've just pushed a change which should make it work properly. I think this'll be ok for now, without going down the retry route. I think adding retries to this script might actually make it a bit more unpredictable - if something fundamentally wrong happens we could end up backing ourselves off until the lambda times out.

Joe-Edwards-GDS (approved):
Feels a bit janky but can't see an obviously cleaner approach :shrug:

PR #2480, opened 2024-09-23
PYIC-7470: Output report in logs

Wynndow:
### What changed

Output report in logs

### Why did it change

If for some reason we fail to catch the output of the lambda invocation we still want to be able to retrieve the report.

### Issue tracking

### DIFF
--- a/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
+++ b/lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
@@ -2,6 +2,8 @@ package uk.gov.di.ipv.core.bulkmigratevcs;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
```diff
@@ -220,6 +226,7 @@ public BatchReport handleRequest(Request request, Context context) {
             forkJoinPool.shutdown();
         }
 
+        logReport(report);
```
Joe-Edwards-GDS: If you put this in the `finally` block, then you don't need to repeat it - and it's clearer what the intent is IMO

Joe-Edwards-GDS (dismissed):
The report only includes hashed user IDs right (i.e. nothing potentially sensitive)? Seems like a sensible thing to do.

One small suggestion...

lambdas/bulk-migrate-vcs/src/main/java/uk/gov/di/ipv/core/bulkmigratevcs/BulkMigrateVcsHandler.java
Wynndow: I think that's a good shout - I'll shift it

Wynndow: And yeah - just hashes of user IDs

Joe-Edwards-GDS (approved):

PR #2481, opened 2024-09-23
BAU: add prettier and eslint to journey map

thebauSoftwire:
### What changed
- adds prettier and eslint to journey-map folder
- adds lint check for journey map to pre-commit config


Joe-Edwards-GDS (dismissed):
Thanks!

I'm not always a fan of what prettier does, but it's a lot better than no checking, and we want consistency across repos.

Joe-Edwards-GDS (approved):

PR #2487, opened 2024-09-24
PYIC-7347: misc unhappy paths api tests

thebauSoftwire:
### What changed
- add tests for misc unhappy paths
- adds a way to check against thrown errors

### Why did it change

### Issue tracking

- [PYIC-7347](https://govukverify.atlassian.net/browse/PYIC-7347)

[PYIC-7347]: https://govukverify.atlassian.net/browse/PYIC-7347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/inherited-identity/inherited-identity.feature
```diff
@@ -128,3 +128,7 @@ Feature: Inherited Identity
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
     And my identity 'GivenName' is 'Kenneth'
+
+  Scenario: Invalid inherited identity JWT from orch
+    When I start a new 'medium-confidence-pcl200-pcl250' inherited identity journey with an invalid inherited identity JWT
+    Then I get a 'pyi-technical' page response
```
Joe-Edwards-GDS, line 134: Can we test that you get the correct error code after proceeding from here? Something like:
```
When I start a new 'medium-confidence-pcl200-pcl250' inherited identity journey with an invalid inherited identity JWT
Then I get a 'pyi-technical' page response
When I submit a 'next' event
Then I get an OAuth response
Then I get an 'invalid_inherited_identity' error code
```
(I don't know if any other tests check for error codes yet - you might need to set that up)

api-tests/features/orchestration-errors.feature
```diff
@@ -0,0 +1,10 @@
+@Build
+Feature: Errors from Orchestration
+
+  Scenario: Unrecoverable error from Orchestration
+    When I start a new 'medium-confidence' journey with invalid redirect url 'https://orch.stubs.account.gov.uk/invalid-callback'
+    Then I get an error from 'InitialiseIpvSession' with message 'Failed to parse the session start request' and with status code '400'
+
+  Scenario: Recoverable error from Orchestration
+    When I start a new 'medium-confidence-with-invalid-audience' journey
+    Then I get a 'pyi-technical' page response
```
Joe-Edwards-GDS, line 10: As above, it might be neat to prove that this is recoverable by proceeding and getting an OAuth response (with an appropriate error code)

api-tests/features/p2-web-journey.feature
```diff
@@ -93,3 +93,18 @@ Feature: P2 Web document journey
     Then I get an OAuth response
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
+
+  Scenario Outline: Failed P2 journey via Web using <cri>
```
Joe-Edwards-GDS, line 192: This isn't quite what we wanted to test - although it might be worth keeping this in as a valid scenario anyway!

The specific case is for using a Driving Licence and only achieving a fraud score of 1 - that fraud score is sufficient with passport (M1A), but not DL (M1B) - https://www.gov.uk/government/publications/identity-proofing-and-verification-of-an-individual/identity-profiles#medium-confidence

api-tests/src/utils/jar-generator.ts
```diff
@@ -37,27 +37,36 @@ export const generateJarPayload = async (
...
     "https://vocab.account.gov.uk/v1/storageAccessToken"
   ].values = [await createEvcsAccessToken(session.subject)];
 
-  if (session.inheritedIdentityId) {
-    const inheritedIdentity = JSON.parse(
-      await fs.readFile(
-        path.join(
-          __dirname,
-          `../../data/inherited-identities/${session.inheritedIdentityId}.json`,
+  if (session.inheritedIdentity) {
```
Joe-Edwards-GDS, line 48: I'd also be happy to use a magic value here if it makes things neater (i.e. `if (inheritedIdentity === 'invalid') ... `), but this is more explicit :)

api-tests/src/utils/jar-generator.ts
```diff
@@ -37,27 +37,36 @@ export const generateJarPayload = async (
       sub: session.subject,
       govuk_signin_journey_id: session.journeyId,
       state: getRandomString(16),
-      redirect_uri: config.orch.redirectUrl,
+      redirect_uri: session.redirectUrl || config.orch.redirectUrl,
```
Joe-Edwards-GDS: I wonder if we should make the overrides here cleverer - e.g. if you had:
```
redirect_uri: payloadData.redirect_uri || config.orch.redirectUrl
```
So if you have a payload JSON with `"redirect_uri": ""` then it'll get populated automatically, but if you have `"redirect_uri": "invalid"` then it won't.

(That would save you passing the url around everywhere, and it'd work more like the invalid audience)

thebauSoftwire: good idea!

api-tests/src/steps/ipv-steps.ts
```diff
@@ -131,27 +129,15 @@ When(
...
 );
 
 When(
-  "I start a new {string} journey with invalid redirect url {string}",
-  async function (
-    this: World,
-    journeyType: string,
-    redirectUrl: string,
-  ): Promise<void> {
+  "I start a new {string} journey with invalid redirect uri",
```
Joe-Edwards-GDS, line 137: Do we need this any more - could we just use the 'normal' start new journey step?

Joe-Edwards-GDS (dismissed):
One small comment, but looks cleaner now I think :)

api-tests/src/steps/ipv-steps.ts
thebauSoftwire: I think I kept this separate because for this case, we expect an error to get thrown from `initialiseIpvSession` that we want to catch and keep track of so we can assert against it in a future step, whereas in the 'normal' start new journey step, if an error gets thrown from say `initialiseIpvSession` we'd want it to prevent the subsequent steps from running

Wynndow (approved):

PR #2520, opened 2024-10-01
IPS-1109: Canary deployments

louiseli1:
### What changed

Enabled canary deployments for JourneyEngineStepFunction and the following lambdas:

1. IssueClientAccessTokenFunction - `/token` (external)
2. InitialiseIpvSessionFunction - `/session/initialise` (internal)
3. ProcessCriCallbackFunction - `/cri/callback` (internal)
4. BuildUserIdentityFunction - `/user-identity` (external)
5. BuildProvenUserIdentityDetailsFunction - `/user/proven-identity-details`(internal)
6. ProcessAsyncCriCredentialFunction

Lambdas each have alarms for:
- A lambda error
- 5xx error (1-6 only)

These alert #di-ipv-core-non-prod-alarms and #di-ipv-core-prod-alarms

Add DeploymentStrategies:
- JourneyEngineStepFunction to use Versions of lambda, rather than the alias
- Update the API spec to point to the step function alias

### Why did it change

- Canary deployments across identity

### Issue tracking

- [IPS-1109](https://govukverify.atlassian.net/browse/IPS-1109)

#### Canaries in progress
1-CanariesInProgress

#### Starting a failed execution
2-StartFailedExecution

#### Alarm triggered
3-AlarmTriggered

#### Canary reverted to V1
4-CanaryRevertToV1

#### Stack rollback
5-StackRollback

#### Next deployment - canaries in progress
6-CanariesInProgress

#### Canaries complete
7-CanariesComplete

:journey:next

### Lambda 
LambdaDeployment

### Alarms
image

### Other considerations

[IPS-1109]: https://govukverify.atlassian.net/browse/IPS-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -71,6 +71,10 @@ Globals:
         - !Sub
           - '{{resolve:secretsmanager:${SecretArn}:SecretString:JAVA_LAYER}}' # pragma: allowlist secret or NODEJS_LAYER or PYTHON_LAYER
           - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, dynatraceSecretArn ]
+    DeploymentPreference:
+      Type: !Ref LambdaDeploymentPreference
+      Role: !GetAtt CodeDeployServiceRole.Arn
+    AutoPublishAlias: live
 
 Parameters:
   Environment:
@@ -89,6 +93,28 @@ Parameters:
     Description: "The ARN of the permissions boundary to apply when creating IAM roles"
     Type: String
     Default: "none"
+  LambdaDeploymentPreference:
+    Description: "Specifies the configuration to enable gradual Lambda deployments."
+    Type: String
+    Default: "AllAtOnce"
+    AllowedValues:
+      - AllAtOnce
+      - Canary10Percent5Minutes
+      - Canary10Percent10Minutes
+      - Canary10Percent15Minutes
+      - Canary10Percent30Minutes
+      - Linear10PercentEvery10Minutes
+      - Linear10PercentEvery1Minute
+      - Linear10PercentEvery2Minutes
+      - Linear10PercentEvery3Minutes
+  StepFunctionDeploymentPreference:
+    Description: "Specifies the configuration to enable gradual StepFunction deployment."
+    Type: String
+    Default: "ALL_AT_ONCE"
+    AllowedValues:
+      - ALL_AT_ONCE
+      - CANARY
+      - LINEAR
 
 Conditions:
   IsDevelopment: !Or
@@ -114,6 +140,9 @@ Conditions:
   IsTestApiEnv: !Or
     - !Condition IsDevelopment
     - !Equals [ !Ref Environment, build ]
+  UseCanaryDeploymentAlarms: !Or
+    - !Not [ !Equals [ !Ref StepFunctionDeploymentPreference, ALL_AT_ONCE ]]
+    - !Not [ !Equals [ !Ref LambdaDeploymentPreference, AllAtOnce ]]
 
 # The AWS Account Id is used in the following mapping section because we have
 # multiple developer environments and it is undesirable to have to keep this
@@ -626,6 +655,11 @@ Resources:
       PackageType: Zip
       CodeUri: ../lambdas/issue-client-access-token
       Tracing: Active
+      DeploymentPreference:
+        Alarms: !If
+          - UseCanaryDeploymentAlarms
+          - [!Ref IssueClientAccessTokenFunctionErrorCanaryAlarm, !Ref IssueClientAccessTokenFunction5xxErrorCanaryAlarm]
+          - [!Ref AWS::NoValue]
       Environment:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
@@ -781,6 +815,11 @@ Resources:
       PackageType: Zip
       CodeUri: ../lambdas/initialise-ipv-session
       Tracing: Active
+      DeploymentPreference:
+        Alarms: !If
+          - UseCanaryDeploymentAlarms
+          - [!Ref InitialiseIpvSessionFunctionErrorCanaryAlarm, !Ref InitialiseIpvSessionFunction5xxErrorCanaryAlarm]
+          - [!Ref AWS::NoValue]
       Environment:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
@@ -981,6 +1020,11 @@ Resources:
       CodeUri: ../lambdas/process-cri-callback
       MemorySize: 2048
       Tracing: Active
+      DeploymentPreference:
+        Alarms: !If
+          - UseCanaryDeploymentAlarms
+          - [!Ref ProcessCriCallbackFunctionErrorCanaryAlarm, !Ref ProcessCriCallbackFunction5xxErrorCanaryAlarm]
+          - [!Ref AWS::NoValue]
       Environment:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
@@ -1099,6 +1143,11 @@ Resources:
       PackageType: Zip
       CodeUri: ../lambdas/build-user-identity
       Tracing: Active
+      DeploymentPreference:
+        Alarms: !If
+          - UseCanaryDeploymentAlarms
+          - [!Ref BuildUserIdentityFunctionErrorCanaryAlarm, !Ref BuildUserIdentityFunction5xxErrorCanaryAlarm]
+          - [!Ref AWS::NoValue]
       Environment:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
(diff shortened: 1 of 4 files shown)


deploy/template.yaml
```diff
@@ -89,6 +93,24 @@ Parameters:
...
+    AllowedValues:
+      - AllAtOnce
+      - Canary10Percent5Minutes
+      - Canary10Percent10Minutes
+      - Canary10Percent15Minutes
+      - Canary10Percent30Minutes
+  StepFunctionsDeploymentPreference:
+    Description: "Specifies the configuration to enable gradual StepFunction deployment."
+    Type: String
+    Default: "CANARY"
```
sandragds: Shall we have default as ALL_AT_ONCE? I think that's the configuration for other services

deploy/template.yaml
```diff
@@ -1247,19 +1297,31 @@ Resources:
...
+        CallTicfCriLambdaArn: !Ref CallTicfCriFunction.Version
+        CallDcmawAsyncCriLambdaArn: !Ref CallDcmawAsyncCriFunction.Version
+        StoreIdentityLambdaArn: !Ref StoreIdentityFunction.Version
+        ResetSessionIdentityFunctionArn: !Ref ResetSessionIdentityFunction.Version
+        CheckCoiFunctionArn: !Ref CheckCoiFunction.Version
+      AutoPublishAlias: live
+      DeploymentPreference:
+        Type: !Ref StepFunctionsDeploymentPreference
+        Interval: 10
+        Percentage: 20
```
sandragds: has the canary strategy been agreed with core?

deploy/template.yaml
```diff
@@ -3003,6 +3101,509 @@ Resources:
...
+  ####################################################################
+  #                                                                  #
+  # Canary Alarms                                                    #
+  #                                                                  #
+  ####################################################################
+
+  JourneyEngineStepFunctionExecutionFailedCanaryAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      ActionsEnabled: true
```
sandragds: start with false so they don't fire to slack for production where it is set to not all at once?

deploy/template.yaml
```diff
@@ -1247,19 +1297,31 @@ Resources:
...
+        CheckExistingIdentityFunctionArn: !Ref CheckExistingIdentityFunction.Version
+        BuildCriOauthRequestFunctionArn: !Ref BuildCriOauthRequestFunction.Version
+        BuildClientOauthResponseFunctionArn: !Ref BuildClientOauthResponseFunction.Version
+        CheckGpg45ScoreFunctionArn: !Ref CheckGpg45ScoreFunction.Version
+        EvaluateGpg45ScoresFunctionArn: !Ref EvaluateGpg45ScoresFunction.Version
+        CallTicfCriLambdaArn: !Ref CallTicfCriFunction.Version
+        CallDcmawAsyncCriLambdaArn: !Ref CallDcmawAsyncCriFunction.Version
+        StoreIdentityLambdaArn: !Ref StoreIdentityFunction.Version
+        ResetSessionIdentityFunctionArn: !Ref ResetSessionIdentityFunction.Version
+        CheckCoiFunctionArn: !Ref CheckCoiFunction.Version
```
sandragds, line 1326: alarms for all the above functions should be added and configured in the state machine as they are associated with a state machine e.g. in bravo - https://github.com/govuk-one-login/di-id-reuse-storage/blob/3d040e9f94162790a1ad57d0ee862e2a44207901/src/infra/main/template.yaml#L587

deploy/template.yaml
louiseli1: Not yet tbc

louiseli1: Yep, strategy updated, we will deploy canaries to build first

sandragds:
These lambdas are without canaries set up and not associated to state machines:
UserReverificationFunction
ReportUserIdentityFunction
BulkMigrateVcsFunction
Should they be set up with canaries too or they dont need them either?

louiseli1:
> These lambdas are without canaries set up and not associated to state machines: UserReverificationFunction ReportUserIdentityFunction BulkMigrateVcsFunction Should they be set up with canaries too or they dont need them either?

They were identified as non critical lambdas, they receive very low traffic and it doesn't seem like canaries would add much benefit

sandragds (dismissed):
Looks good!

deploy/template.yaml
```diff
@@ -3004,6 +3101,728 @@ Resources:
...
+          Value: POST
+        - Name: Stage
+          Value: !Ref Environment
+        - Name: Resource
+          Value: /token
+      Statistic: Sum
+      Unit: Count
+      Period: 60
+      EvaluationPeriods: 1
+      Threshold: 1
```
Joe-Edwards-GDS, line 3181: Are these deliberately stricter than the error alarms?

openAPI/core-back-internal.yaml
```diff
@@ -75,7 +75,7 @@ paths:
               \"clientOAuthSessionId\":\"$util.escapeJavaScript($input.params('client-session-id'))\",
               \"language\":\"$util.escapeJavaScript($input.params('language'))\"
               }",
-              "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}"
+              "stateMachineArn": "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${JourneyEngineStepFunction.Name}:live"
```
Joe-Edwards-GDS, line 78: Do we need to worry about any permissions issues with this change? E.g. I know when we switched to invoking CIMIT lambdas using an alias we had to update our permissions to specify the alias as well.

Joe-Edwards-GDS (commented):
Looks reasonable to me, but a couple of queries.

To elaborate on the 'non-critical' lambdas:

*UserReverificationFunction* - This is only used for MFA reset journeys, which is not yet live - but once this journey is live we probably should have canaries for it... (May be several months!)
*ReportUserIdentityFunction* and *BulkMigrateVcsFunction* - These are manually invoked lambdas, used for the bulk EVCS migration, so it doesn't make sense to canary these.

Goes without saying, but please don't merge until next week!

openAPI/core-back-internal.yaml
louiseli1: I’ve looked at 3 types of permissions:

1. APIGW to step function with `JourneyEngineStepFunctionApiGateWayIamRole`: No changes needed to this, APIGW can invoke any version of the live alias of the state machine with a response 200

2. Step function to lambdas with LambdaInvokePolicy: Also no changes needed, this is an inbuilt policy and gives permission to invoke a function, alias, or version (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-policy-template-list.html#lambda-invoke-policy)

3. Test APIGW to lambdas: It looks like the APIGW already has access to `Function.Alias` https://github.com/govuk-one-login/ipv-core-back/blob/eaf5cd912f91798148f77427ef6b5b5c75c7c38e/deploy/template.yaml#L858 - but do you know how I could double check this works for both versions during a canary deployment? Ie what query strings/headers do I need to test permissions from apigw to these lambdas for build-user-identity, initialise-ipv-session, issue-client-access-token etc

Is there anything else I should check? I’ve used https://orch-dev-sre.01.core.dev.stubs.account.gov.uk/ to run through journeys (not sure if I’m doing this bit right?), and it looks like the canaries are splitting traffic, see screenshot from when I upped the ratio from 10:90 to 50:50):

step-function-core-canaries

deploy/template.yaml
louiseli1: Not intentional, there is a baseline error rate so I've modified these to also be in 3x1 minute, same for the state machine `ExecutionsFailed` alarm E.g. this is the sum over 5 min periods in build over the last 4 weeks:

executions-failed-core-build

louiseli1:
> Looks reasonable to me, but a couple of queries.
> 
> To elaborate on the 'non-critical' lambdas:
> 
> _UserReverificationFunction_ - This is only used for MFA reset journeys, which is not yet live - but once this journey is live we probably should have canaries for it... (May be several months!) _ReportUserIdentityFunction_ and _BulkMigrateVcsFunction_ - These are manually invoked lambdas, used for the bulk EVCS migration, so it doesn't make sense to canary these.
> 
> Goes without saying, but please don't merge until next week!

I’ve updated the PR description to remove the manual lambdas, and have taken note to add UserReverificationFunction

deploy/template.yaml
Joe-Edwards-GDS: Thanks - that's a higher baseline than I'd have expected tbh, so maybe something for me to look into!

openAPI/core-back-internal.yaml
Joe-Edwards-GDS: TBH if you've run through journeys then that's the best test :)

Joe-Edwards-GDS (approved):

PR #2524, opened 2024-10-02
PYIC-7476: CRI disabled API tests

Wynndow:
### Why did it change

[PYIC-7476: Allow nested journey exit events behind conditionals](https://github.com/govuk-one-login/ipv-core-back/commit/ba8e1abe885e68dd52bcc6f5589aa96933b77be6)

The state manchine initializer wasn't correctly initializing exit nested
journey events that were defined behind journey map conditionals
(checkIfDisabled etc). This change makes sure that they are.

It also fixes a couple of places in the journey map where we were trying
to exit to states defined outside of the nested journey definition. It
now uses an exit event to route to the expected state.

[PYIC-7476: Add cimit-stub client](https://github.com/govuk-one-login/ipv-core-back/commit/87095fa06900ec4bf9b99fe7b360a3f539b7757d) 

We have the ability to send VCs directly to the EVCS stub to allow quick
creation of existing identities. This however skips sending the VCs to
CIMIT.

This adds a new client for doing that. It doesn't use the management API
of CIMIT and instead posts directly into the internal API.

This allows for quick creation of separate session mitigation journeys.

[PYIC-7476: Add disabled CRI API tests](https://github.com/govuk-one-login/ipv-core-back/commit/75ee60f138aef98c9c4ecebdca631300ef3bb608)

This adds a suite of API tests for journeys when we've disabled a CRI.

The tests for currently disabled by default CRIs (BAV, HMRC and DWP
KBVs) are slightly future proofing - when those CRIs are enabled these
tests should still be valid and pass.

This commit adds a couple of `txn` properties to some of the evidence
JSON blobs. Without these our CIMIT stub gets very upset - and evidence
blocks should have them anyway.

Also adds some logic to prevent exiting feature sets from being wiped
out if they exist.

### Issue tracking

- [PYIC-7476](https://govukverify.atlassian.net/browse/PYIC-7476)

[PYIC-7476]: https://govukverify.atlassian.net/browse/PYIC-7476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/cimit/alternate-doc.feature
+++ b/api-tests/features/cimit/alternate-doc.feature
@@ -36,7 +36,8 @@ Feature: CIMIT - Alternate doc
       | 'ukPassport'        | 'kenneth-passport-needs-alternate-doc'       | 'pyi-passport-no-match-another-way'        | 'drivingLicence'| 'kenneth-driving-permit-valid' |
 
   Scenario Outline: Alternate doc mitigation via passport or DL - DWP KBV
-    Given I start a new 'medium-confidence' journey with feature set 'dwpKbvTest'
+    Given I activate the 'dwpKbvTest' feature set
+    When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-identity-document-start' page response
     When I submit an 'appTriage' event
     Then I get a 'dcmaw' CRI response
@@ -71,7 +72,8 @@ Feature: CIMIT - Alternate doc
       | 'ukPassport'        | 'kenneth-passport-needs-alternate-doc'       | 'pyi-passport-no-match-another-way'        | 'drivingLicence'| 'kenneth-driving-permit-valid' |
 
   Scenario Outline: Alternate doc mitigation via passport or DL - DWP KBV PIP page dropout
-    Given I start a new 'medium-confidence' journey with feature set 'dwpKbvTest'
+    Given I activate the 'dwpKbvTest' feature set
+    When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-identity-document-start' page response
     When I submit an 'appTriage' event
     Then I get a 'dcmaw' CRI response
@@ -106,7 +108,8 @@ Feature: CIMIT - Alternate doc
       | 'ukPassport'        | 'kenneth-passport-needs-alternate-doc'       | 'pyi-passport-no-match-another-way'        | 'drivingLicence'| 'kenneth-driving-permit-valid' |
 
   Scenario Outline: Alternate doc mitigation via passport or DL - DWP KBV transition page dropout
-    Given I start a new 'medium-confidence' journey with feature set 'dwpKbvTest'
+    Given I activate the 'dwpKbvTest' feature set
+    When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-identity-document-start' page response
     When I submit an 'appTriage' event
     Then I get a 'dcmaw' CRI response
--- a/api-tests/features/disabled-cri-journeys.feature
+++ b/api-tests/features/disabled-cri-journeys.feature
@@ -101,7 +101,7 @@ Feature: Disabled CRI journeys
       When I submit an 'end' event
       Then I get a 'pyi-escape' page response
 
-    Scenario Outline: Access denied from passport or driving licence CRI returns user to multidoc page
+    Scenario Outline: Choosing another way after access-denied from passport or DL CRIs leads to escape page
       Given I activate the 'f2fDisabled' feature set
       When I start a new 'medium-confidence' journey
       Then I get a 'page-ipv-identity-document-start' page response
@@ -112,12 +112,14 @@ Feature: Disabled CRI journeys
       When I submit a '<cri>' event
       Then I get a '<cri>' CRI response
       When I submit an 'access-denied' event
-      Then I get a 'page-multiple-doc-check' page response
+      Then I get a 'prove-identity-another-type-photo-id' page response with context '<context>'
+      When I submit an 'f2f' event
+      Then I get a 'pyi-escape' page response
 
       Examples:
-        | cri            |
-        | ukPassport     |
-        | drivingLicence |
+        | cri            | context        |
+        | ukPassport     | passport       |
+        | drivingLicence | drivingLicence |
 
     Scenario: Separate session mitigation with enhanced verification and no photo ID leads to ineligible
       Given the subject already has the following credentials
--- a/api-tests/features/disabled-cri-journeys.feature
+++ b/api-tests/features/disabled-cri-journeys.feature
@@ -1,0 +1,233 @@
+@Build
+Feature: Disabled CRI journeys
+
+  Rule: DCMAW is disabled
+
+    Scenario: A P1 journey takes the user on a no photo ID journey
+      Given I activate the 'dcmawOffTest,p1Journeys' feature sets
+      When I start a new 'low-confidence' journey
+      Then I get a 'page-ipv-identity-document-start' page response
+      When I submit an 'appTriage' event
+      Then I get a 'prove-identity-no-photo-id' page response with context 'nino'
+
+    Scenario: A P2 journey takes the user to the document select page
+      Given I activate the 'dcmawOffTest' feature set
+      When I start a new 'medium-confidence' journey
+      Then I get a 'page-ipv-identity-document-start' page response
+      When I submit an 'appTriage' event
+      Then I get a 'page-multiple-doc-check' page response
+
+    Scenario: Choosing DCMAW after escaping from KBV CRIs leads to technical failure
+      Given I activate the 'dcmawOffTest' feature set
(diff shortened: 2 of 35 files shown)


api-tests/features/disabled-cri-journeys.feature
```diff
@@ -0,0 +1,233 @@
...
+      Then I get a 'page-ipv-identity-document-start' page response
+      When I submit an 'appTriage' event
+      Then I get a 'prove-identity-no-photo-id' page response with context 'nino'
+
+    Scenario: A P2 journey takes the user to the document select page
+      Given I activate the 'dcmawOffTest' feature set
+      When I start a new 'medium-confidence' journey
+      Then I get a 'page-ipv-identity-document-start' page response
+      When I submit an 'appTriage' event
+      Then I get a 'page-multiple-doc-check' page response
```
Wynndow, line 18: This scenario is repeated in the below scenarios. If we keep the below (they might be too edge-casey), it might be worth removing this one.

api-tests/features/disabled-cri-journeys.feature
```diff
@@ -0,0 +1,233 @@
...
+      When I submit an 'appTriage' event
+      Then I get a 'dcmaw' CRI response
+      When I submit an 'access-denied' event
+      Then I get a 'page-multiple-doc-check' page response
+      When I submit an 'end' event
+      Then I get a 'pyi-post-office' page response
+      When I submit an 'end' event
+      Then I get a 'pyi-escape' page response
+
+  Rule: HMRC and DWP KBVs are disabled
```
Joe-Edwards-GDS, line 197: Should there be another case for 'DWP KBVs are disabled' (but HMRC KBVs are enabled)?

Feels like this might overlap with some of the other work that Hilux are doing, if so then we could potentially put them in a separate suite. (cc @blakeyp)

api-tests/features/disabled-cri-journeys.feature
Joe-Edwards-GDS: They are pretty edge-case-y, but I guess it's covering all the cases.

For now the tests seem to still be fairly speedy, but could be a good candidate for trimming...

api-tests/src/steps/cri-steps.ts
```diff
@@ -289,5 +290,9 @@ When(
       this.userId,
       generatePostVcsBody(credentials),
     );
+    for (const credential of credentials) {
+      await cimitStubClient.postDetectCi({ signed_jwt: credential });
+    }
+    await cimitStubClient.postMitigateCi({ signed_jwts: credentials });
```
Joe-Edwards-GDS: Would this actually work? The CIMIT stub doesn't have any mitigation rules, it has to be primed by the CRI stub...

I suppose it doesn't hurt, but potentially misleading!

api-tests/src/steps/ipv-steps.ts
```diff
@@ -89,7 +89,7 @@ const startNewJourney = async (
 ): Promise<void> => {
   world.userId = world.userId ?? getRandomString(16);
   world.journeyId = getRandomString(16);
-  world.featureSet = featureSet;
+  world.featureSet = featureSet || world.featureSet;
```
Joe-Edwards-GDS: (this inconsistency is a bit annoying - how much work would it be to make everything use the `I activate the feature set` step?)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
```diff
@@ -195,7 +195,7 @@ WEB_DL_OR_PASSPORT:
           targetState: PROVE_ANOTHER_WAY_AFTER_PASSPORT
           checkIfDisabled:
             f2f:
-              targetState: MULTIPLE_DOC_CHECK_PAGE
+              exitEventToEmit: access-denied-passport-or-dl
```
Joe-Edwards-GDS: There's an interesting inconsistency here, depending on when exactly F2F is disabled 🤔 

Most of the time it'll be caught here and you'll go -> MULTIPLE_DOC_CHECK_PAGE
But if it's slightly later you'll go -> PROVE_ANOTHER_WAY_AFTER_PASSPORT -> PYI_ESCAPE

I wonder if we can unify them (and potentially just drop one of them), i.e.:
- Make this also emit `end`
- Make the `end` event go to `MULTIPLE_DOC_CHECK_PAGE` if F2F is disabled

What do you think? We could potentially check with Joe U when he's back, but it seems pretty low-stakes!

api-tests/src/steps/cri-steps.ts
```diff
@@ -289,5 +290,9 @@ When(
       this.userId,
       generatePostVcsBody(credentials),
     );
+    for (const credential of credentials) {
+      await cimitStubClient.postDetectCi({ signed_jwt: credential });
```
Joe-Edwards-GDS, line 339: I assume this is fairly speedy - we could try and skip if we know there's no CI, but this is a better reflection of core behaviour :shrug:

Joe-Edwards-GDS (commented):
Assume you've updated the github secrets (inc dependabot)!

api-tests/src/steps/cri-steps.ts
Wynndow: That's a good point. I was just following what core does, but it's not neccessary and is just introducing an extra API call to the tests for no reason. In future, if we wanted to arrange some state to have already mitigated CIs we could just use the management endpoint. I'll yank this.

api-tests/src/steps/ipv-steps.ts
Wynndow: Let's find out...

api-tests/src/steps/cri-steps.ts
Wynndow: Yeah, I did think about that but then decided not to. For the amount we're actually using this step I don't think it's a going to be a massive timesaver.

Wynndow:
> Assume you've updated the github secrets (inc dependabot)!

Not needed which is nice - the new stuff is just using the already exsting API key for CIMIT, already in place for running core locally in GHA

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journey-definitions.yaml
Wynndow: Yeah that's true. I think the best thing for the user, without changing too much stuff. is to drop the f2f disabled check on the access-denied event and allow the user to progress to `PROVE_ANOTHER_WAY_AFTER_PASSPORT`. The working might be a little confusing but it will give them two options -> DL CRI or "try using another type of photo ID". If they choose the other way (f2f) then catch that and send them to the escape page. 
Sending them back to the multidoc page will also have slightly confusing wording, but gives them two options that won't work instead of one...

Wynndow: Actually, I just remembered how the user gets `access-denied` from passport, so it won't be confusing at all if they get the `PROVE_ANOTHER_WAY_AFTER_PASSPOR` page.

api-tests/features/cimit/alternate-doc.feature
```diff
@@ -36,7 +36,8 @@ Feature: CIMIT - Alternate doc
       | 'ukPassport'        | 'kenneth-passport-needs-alternate-doc'       | 'pyi-passport-no-match-another-way'        | 'drivingLicence'| 'kenneth-driving-permit-valid' |
 
   Scenario Outline: Alternate doc mitigation via passport or DL - DWP KBV
-    Given I start a new 'medium-confidence' journey with feature set 'dwpKbvTest'
+    Given I activate the 'dwpKbvTest' feature set
+    Given I start a new 'medium-confidence' journey
```
Joe-Edwards-GDS: (should probably be `When`)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/ExitNestedJourneyEvent.java
```diff
@@ -8,6 +8,7 @@
 
 import java.util.Map;
 
+@SuppressWarnings({"java:S7027"}) // Circular dependency with Event interface
```
Joe-Edwards-GDS: :eyes: just confirming my thinking that we'd be better off implementing this in TS!

Joe-Edwards-GDS (dismissed):
Thanks!

api-tests/features/cimit/alternate-doc.feature
Wynndow: Good spot

Joe-Edwards-GDS (approved):

PR #2525, opened 2024-10-02
PYIC-4111: add consumer side contract tests for the EVCS client

shivanshuit914:
## Proposed changes

- Add consumer side contract tests for the EVCS

### What changed

- Added Pact contract tests for EVCS client

### Issue tracking

- [PYIC-4111](https://govukverify.atlassian.net/browse/PYIC-4111)

[PYIC-4111]: https://govukverify.atlassian.net/browse/PYIC-4111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
+++ b/libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
@@ -62,6 +62,11 @@ class ContractTest {
                                     "timestampMs", "1714478033959"),
                             ONLINE));
 
+    private static final List<EvcsCreateUserVCsDto> INVALID_CREATE_USER_VCS_DTO =
+            List.of(
+                    new EvcsCreateUserVCsDto(
+                            VcFixtures.VC_ADDRESS.getVcString(), EvcsVCState.CURRENT, null, null));
+
     private static final List<EvcsUpdateUserVCsDto> EVCS_UPDATE_USER_VCS_DTO =
             List.of(
                     new EvcsUpdateUserVCsDto(
@@ -118,18 +123,7 @@ class ContractTest {
                                                             VcFixtures.DCMAW_PASSPORT_VC
                                                                     .getVcString());
                                                     vc.stringType("state", "PENDING_RETURN");
-                                                    vc.object(
-                                                            "metadata",
-                                                            metadata -> {
-                                                                metadata.stringType(
-                                                                        "reason", "testing");
-                                                                metadata.numberType(
-                                                                        "timestampMs",
-                                                                        1711721297123L);
-                                                                metadata.stringType(
-                                                                        "txmaEventId",
-                                                                        "1a116fe7-2ff9-4f7c-940d-d55fa7d03d66");
-                                                            });
+                                                    vc.object("metadata", metadata -> {});
                                                 });
                                     });
                         })
@@ -323,13 +317,25 @@ class ContractTest {
         return builder.given("EVCS client exist")
                 .given("User has a valid VC")
                 .uponReceiving("A request to create EVCS user VCs")
-                .path("/vcs/" + INVALID_USER_ID)
+                .path("/vcs/" + TEST_USER_ID)
                 .method("POST")
                 .headers(Map.of(CONTENT_TYPE, ContentType.APPLICATION_JSON.toString()))
-                .body(getRequestBodyForUserVC())
+                .body(invalidRequestBodyForUserVC())
                 .willRespondWith()
                 .status(400)
                 .toPact();
+    }
+
+    private DslPart invalidRequestBodyForUserVC() {
+        return newJsonArray(
+                        array -> {
+                            array.object(
+                                    vcObject -> {
+                                        vcObject.stringType("vc", "invalid-vc-string");
+                                        vcObject.stringType("state", "CURRENT");
+                                    });
+                        })
+                .build();
     }
 
     @Test
@@ -340,7 +346,7 @@ class ContractTest {
         assertThrows(
                 EvcsServiceException.class,
                 () -> {
-                    evcsClient.storeUserVCs(INVALID_USER_ID, EVCS_CREATE_USER_VCS_DTO);
+                    evcsClient.storeUserVCs(TEST_USER_ID, INVALID_CREATE_USER_VCS_DTO);
                 });
     }
 
--- a/libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
+++ b/libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
@@ -1,0 +1,430 @@
+package uk.gov.di.ipv.core.library.pact;
+
+import au.com.dius.pact.consumer.MockServer;
+import au.com.dius.pact.consumer.dsl.DslPart;
+import au.com.dius.pact.consumer.dsl.PactDslWithProvider;
+import au.com.dius.pact.consumer.junit.MockServerConfig;
+import au.com.dius.pact.consumer.junit5.PactConsumerTestExt;
+import au.com.dius.pact.consumer.junit5.PactTestFor;
+import au.com.dius.pact.core.model.RequestResponsePact;
+import au.com.dius.pact.core.model.annotations.Pact;
+import org.apache.http.entity.ContentType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.di.ipv.core.library.client.EvcsClient;
+import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
+import uk.gov.di.ipv.core.library.dto.EvcsCreateUserVCsDto;
+import uk.gov.di.ipv.core.library.dto.EvcsUpdateUserVCsDto;
+import uk.gov.di.ipv.core.library.enums.EvcsVCState;
+import uk.gov.di.ipv.core.library.exception.EvcsServiceException;
+import uk.gov.di.ipv.core.library.fixtures.VcFixtures;
+import uk.gov.di.ipv.core.library.service.ConfigService;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static au.com.dius.pact.consumer.dsl.LambdaDsl.newJsonArray;
+import static au.com.dius.pact.consumer.dsl.LambdaDsl.newJsonBody;
+import static org.apache.http.HttpHeaders.CONTENT_TYPE;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.when;
+import static uk.gov.di.ipv.core.library.enums.EvcsVCState.PENDING_RETURN;
+import static uk.gov.di.ipv.core.library.enums.EvcsVcProvenance.ONLINE;
+
+@ExtendWith(PactConsumerTestExt.class)
+@ExtendWith(MockitoExtension.class)
+@PactTestFor(providerName = "EvcsProvider")
+@MockServerConfig(hostInterface = "localhost")
+class ContractTest {
+    private static final String EVCS_API_KEY =
+            "L2BGccX59Ea9PMJ3ipu9t7r99ykD2Tlh1KYpdjdg"; // pragma: allowlist secret
+    private static final String TEST_EVCS_ACCESS_TOKEN = "TEST_EVCS_ACCESS_TOKEN";
+    private static final String TEST_USER_ID = "9bd7f130-4238-4532-83cd-01cb29584834";
+    private static final String INVALID_USER_ID = "invalid-user-id";
+    private static final List<EvcsVCState> VC_STATES_FOR_QUERY = List.of(PENDING_RETURN);
+
+    private static final List<EvcsCreateUserVCsDto> EVCS_CREATE_USER_VCS_DTO =
+            List.of(
+                    new EvcsCreateUserVCsDto(
+                            VcFixtures.VC_ADDRESS.getVcString(),
+                            EvcsVCState.CURRENT,
+                            Map.of(
+                                    "reason", "testing",
+                                    "txmaEventId", "txma-event-id-2",
+                                    "timestampMs", "1714478033959"),
+                            ONLINE));
+
+    private static final List<EvcsUpdateUserVCsDto> EVCS_UPDATE_USER_VCS_DTO =
+            List.of(
+                    new EvcsUpdateUserVCsDto(
+                            "VC_Signature1",
+                            EvcsVCState.HISTORIC,
+                            Map.of(
+                                    "reason", "testing",
+                                    "txmaEventId", "txma-event-id-2",
+                                    "timestampMs", "1714478033959")));
+
+    @Mock ConfigService mockConfigService;
+
+    @BeforeEach
+    void setUp(MockServer mockServer) {
+        when(mockConfigService.getParameter(ConfigurationVariable.EVCS_APPLICATION_URL))
+                .thenReturn("http://localhost:" + mockServer.getPort());
+        lenient()
+                .when(mockConfigService.getSecret(ConfigurationVariable.EVCS_API_KEY))
+                .thenReturn(EVCS_API_KEY);
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "EvcsConsumer")
+    public RequestResponsePact validRetrieveVcRequestReturnsUsersVcWith200(
+            PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
+                .uponReceiving("A request to get VCS")
+                .path("/vcs/" + TEST_USER_ID)
+                .query("state=" + PENDING_RETURN)
+                .method("GET")
+                .headers(
+                        Map.of(
+                                "x-api-key",
+                                EVCS_API_KEY,
+                                "Authorization",
+                                "Bearer " + TEST_EVCS_ACCESS_TOKEN))
+                .willRespondWith()
+                .status(200)
+                .body(getResponseBodyForUserVC())
+                .toPact();
+    }
+
+    private DslPart getResponseBodyForUserVC() {
+        return newJsonBody(
+                        body -> {
+                            body.array(
+                                    "vcs",
+                                    vcs -> {
+                                        vcs.object(
+                                                vc -> {
+                                                    vc.stringType("vc", "vc");
+                                                    vc.stringType("state", "ABANDONED");
+                                                    vc.object(
+                                                            "metadata",
+                                                            metadata -> {
+                                                                metadata.stringType(
+                                                                        "reason",
+                                                                        "abandoned due inactivity");
+                                                                metadata.numberType(
+                                                                        "timestampMs",
+                                                                        1711721297123L);
+                                                                metadata.stringType(
+                                                                        "txmaEventId",
+                                                                        "1a116fe7-2ff9-4f7c-940d-d55fa7d03d66");
+                                                            });
+                                                });
+                                    });
+                        })
+                .build();
+    }
+
+    @Test
+    @PactTestFor(pactMethod = "validRetrieveVcRequestReturnsUsersVcWith200")
+    void testRetrieveVcRequestReturnsUsersVcWith200(MockServer mockServer) throws Exception {
+        // Under Test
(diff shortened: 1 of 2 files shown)


libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+import static uk.gov.di.ipv.core.library.enums.EvcsVCState.PENDING_RETURN;
+import static uk.gov.di.ipv.core.library.enums.EvcsVcProvenance.ONLINE;
+
+@ExtendWith(PactConsumerTestExt.class)
+@ExtendWith(MockitoExtension.class)
+@PactTestFor(providerName = "EvcsProvider")
+@MockServerConfig(hostInterface = "localhost")
+class ContractTest {
+    private static final String EVCS_API_KEY =
+            "L2BGccX59Ea9PMJ3ipu9t7r99ykD2Tlh1KYpdjdg"; // pragma: allowlist secret
```
Joe-Edwards-GDS: Where has this come from? If it's just a test value, can we make it something like `test-evcs-api-key` to make it obvious

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+    private static final String TEST_EVCS_ACCESS_TOKEN = "TEST_EVCS_ACCESS_TOKEN";
+    private static final String TEST_USER_ID = "9bd7f130-4238-4532-83cd-01cb29584834";
```
Joe-Edwards-GDS: Where has this come from? Our real user ids have a `urn:uuid:` prefix...

If it doesn't need that, then can we make it clear that it's a test value - `test-user-id` or similar.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+    @BeforeEach
+    void setUp(MockServer mockServer) {
+        when(mockConfigService.getParameter(ConfigurationVariable.EVCS_APPLICATION_URL))
+                .thenReturn("http://localhost:" + mockServer.getPort());
+        lenient()
+                .when(mockConfigService.getSecret(ConfigurationVariable.EVCS_API_KEY))
+                .thenReturn(EVCS_API_KEY);
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "EvcsConsumer")
```
Joe-Edwards-GDS: All our other contract tests have `consumer = IpvCoreBack`, should we be consistent?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+    public RequestResponsePact validRetrieveVcRequestReturnsUsersVcWith200(
+            PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
```
Joe-Edwards-GDS: I'm not quite sure what all these givens are supposed to mean - I think they should be clear about what state is expected to be set up - i.e. all the test access tokens, user ids etc.

How about something like:
```
.given("test-api-key is a valid API key")
.given("test-acess-token is a valid access token")
.given("test-user-id has one PENDING_RETURN VC")
...
```

This gives you a pattern for setting up other tests, where one of those conditions may not be true!

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+    private DslPart getResponseBodyForUserVC() {
+        return newJsonBody(
+                        body -> {
+                            body.array(
+                                    "vcs",
+                                    vcs -> {
+                                        vcs.object(
+                                                vc -> {
+                                                    vc.stringType("vc", "vc");
+                                                    vc.stringType("state", "ABANDONED");
```
Joe-Edwards-GDS: The request is for `PENDING_RETURN` but the state here is `ABANDONED`?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+        assertEquals(1, evcsGetUserVCsDto.vcs().size());
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "EvcsConsumer")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith400(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
+                .uponReceiving("A request to get VCS")
+                .path("/vcs/" + INVALID_USER_ID)
```
Joe-Edwards-GDS: Would you expect an invalid user id to give a 400? I'd expect that to be a 404...

You might get a 400 with an invalid state, for example.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+                    evcsClient.getUserVcs(
+                            INVALID_USER_ID, TEST_EVCS_ACCESS_TOKEN, VC_STATES_FOR_QUERY);
+                });
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "EvcsConsumer")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith404(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
```
Joe-Edwards-GDS: What's the difference between this case and the one above?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+    void testRetrieveVcRequestReturns404(MockServer mockServer) throws Exception {
+        // Under Test
+        EvcsClient evcsClient = new EvcsClient(mockConfigService);
+        var userVcs =
+                evcsClient.getUserVcs(INVALID_USER_ID, TEST_EVCS_ACCESS_TOKEN, VC_STATES_FOR_QUERY);
+        assertEquals(Collections.emptyList(), userVcs.vcs());
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "EvcsConsumer")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith401(PactDslWithProvider builder) {
```
Joe-Edwards-GDS: perhaps better to rephrase as `invalidApiKeyReturns401` to be really clear about what you're doing

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,430 @@
...
+        return newJsonArray(
+                        array -> {
+                            array.object(
+                                    vcObject -> {
+                                        vcObject.stringType("vc", "vc");
+                                        vcObject.stringType("state", "CURRENT");
+                                        vcObject.object(
+                                                "metadata",
+                                                metadata -> {
+                                                    metadata.stringType("reason", "ABANDON");
```
Joe-Edwards-GDS: What is this metadata?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
Joe-Edwards-GDS, line 1: There are a few other cases we've seen that might be worth including:
- GET returns a 404 if a user has no VCs (even if they are a valid user!)
- GET returns a 403 if the access token is for the wrong user
- POST returns a 409 if the VC already exists

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: Not sure, I Copied from swagger doc.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: copy paste mistake, let me update with correct value. Thanks

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: hm yes makes sense thanks

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: Its just diff status code we receive when its not finding user, also 404 not error on our end. I think I need to update the given example.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
Joe-Edwards-GDS: Do we still need `EVCS client exist/VCS exist/EVC has users VC` steps? What additional state are you expecting the provider-side tests to configure here (what would you expect to happen if we didn't include them?)

(Similarly with all tests)

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+                                                            "metadata",
+                                                            metadata -> {
+                                                                metadata.stringType(
+                                                                        "reason", "testing");
+                                                                metadata.numberType(
+                                                                        "timestampMs",
+                                                                        1711721297123L);
+                                                                metadata.stringType(
+                                                                        "txmaEventId",
+                                                                        "1a116fe7-2ff9-4f7c-940d-d55fa7d03d66");
```
Joe-Edwards-GDS: I think the metadata here is made up - I wonder if we should either:
- Not care about its exact shape, and just allow any object
- Define a specific testing metadata blob

We could ask @TaherKapasi if he has an opinion

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith400(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
+                .given("test-evcs-api-key is not provided")
+                .given("test-acess-token is a valid access token")
+                .given("test-user-id has one PENDING_RETURN VC")
+                .uponReceiving("An invalid request to get VCS")
```
Joe-Edwards-GDS: What way is this meant to be invalid? Omitting the API key is probably a 401, not a 400

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+        assertThrows(
+                EvcsServiceException.class,
+                () -> {
+                    evcsClient.getUserVcs(
+                            TEST_USER_ID, TEST_EVCS_ACCESS_TOKEN, VC_STATES_FOR_QUERY);
+                });
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith404(PactDslWithProvider builder) {
```
Joe-Edwards-GDS: This is not an invalid request - it's a valid request for a user with no VCs.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+        // Assertions
+        assertEquals(1, evcsGetUserVCsDto.vcs().size());
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact invalidRetrieveVcRequestReturnsWith400(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("VCS exist")
+                .given("EVC has users VC")
+                .given("test-evcs-api-key is not provided")
```
Joe-Edwards-GDS: This isn't the right place for this - we are the ones providing the API key, not EVCS.

If you want to omit it, then omit it in the headers below (as you have done). If anything, we should leave in the `test-evcs-api-key is a valid API key`, to prove that the issue is coming from omitting the header.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+        EvcsClient evcsClient = new EvcsClient(mockConfigService);
+        assertThrows(
+                EvcsServiceException.class,
+                () -> {
+                    evcsClient.storeUserVCs(INVALID_USER_ID, EVCS_CREATE_USER_VCS_DTO);
+                });
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact validCreateUserVcReturns409(PactDslWithProvider builder) {
```
Joe-Edwards-GDS: This could be a bit more specific - perhaps something like `createUserVcForExistingVcReturns409`?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+
+    private DslPart getResponseBodyForUserVC() {
+        return newJsonBody(
+                        body -> {
+                            body.array(
+                                    "vcs",
+                                    vcs -> {
+                                        vcs.object(
+                                                vc -> {
+                                                    vc.stringType("vc", "vc");
```
Joe-Edwards-GDS: You might need to make this at least a _plausible_ looking VC - EVCS is going to break this apart to extract the signature.

I suggest using a simple test VC value - no need to worry about the contents, but it'll need to be a valid JWT.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,499 @@
...
+                .status(204)
+                .toPact();
+    }
+
+    private DslPart getRequestBodyUpdateVC() {
+        return newJsonArray(
+                        array -> {
+                            array.object(
+                                    vcObject -> {
+                                        vcObject.stringType("signature", "signature");
```
Joe-Edwards-GDS: This signature should probably correspond to the signature of the test VC

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: hm yes, I have removed the 400 status code test for now, I am not sure if we need it for now.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: Yes I checked with him, Metadata is schema less object. There is not validation on their end.

Joe-Edwards-GDS: There is no validation at their end, but this will still make it part of the contract - i.e. they will need to generate identical metadata as part of their test.

IMO it's a bit misleading to have this (given that it is not reflective of any real metadata we're likely to use), so I'd prefer to either:
- Remove it
- Make it obvious that it's a test-specific example (e.g. remove the timestamp/txma id, which sound like real things)

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,457 @@
...
+            fail("EvcsServiceException was thrown");
+        }
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact invalidCreateUserVcReturns400(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
+                .given("User has a valid VC")
+                .uponReceiving("A request to create EVCS user VCs")
+                .path("/vcs/" + INVALID_USER_ID)
```
Joe-Edwards-GDS: I'm not really convinced an invalid user id is a 400 (how would EVCS know whether a user id is real?) - much more likely to get a 400 is an invalid request body...

If you're confident that it's a valid scenario then we can keep it.

Joe-Edwards-GDS (commented):
A couple more comments, but we should almost be at the stage where we'll need to publish the Pact and get feedback from EVCS to do much more.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,463 @@
...
+        try {
+            evcsClient.storeUserVCs(TEST_USER_ID, EVCS_CREATE_USER_VCS_DTO);
+        } catch (EvcsServiceException e) {
+            fail("EvcsServiceException was thrown");
+        }
+    }
+
+    @Pact(provider = "EvcsProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact invalidCreateUserVcReturns400(PactDslWithProvider builder) {
+        return builder.given("EVCS client exist")
```
Joe-Edwards-GDS: You've still got two of these steps, can we remove them?

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,463 @@
...
+                                        vcObject.stringType("state", "CURRENT");
+                                        vcObject.object(
+                                                "metadata",
+                                                metadata -> {
+                                                    metadata.stringType("reason", "testing");
+                                                    metadata.stringType(
+                                                            "timestampMs", "1711721297123");
+                                                    metadata.stringType(
+                                                            "txmaEventId",
+                                                            "1a116fe7-2ff9-4f7c-940d-d55fa7d03d66");
```
Joe-Edwards-GDS: I think it's more reasonable for us to provide some metadata here - but I'd still like it to be clear that this is arbitrary test data.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
```diff
@@ -0,0 +1,463 @@
...
+                                        vcObject.stringType("provenance", "ONLINE");
+                                    });
+                        })
+                .build();
+    }
+
+    private DslPart getResponseBodyCreateUserVC() {
+        return newJsonBody(
+                        body -> {
+                            body.stringType("messageId", "bd8359d9-d559-47dd-9467-2a31e88a9e2d");
```
Joe-Edwards-GDS: Do you think we need to include this in the setup, so their test knows what `messageId` to return? I'm not sure tbh

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: yes, removed it thanks

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/pact/ContractTest.java
shivanshuit914: hm, yes we not using messageID on our end, removed this setup. Thanks

Joe-Edwards-GDS (approved):
Probably worth getting this in now, and we can get feedback from Bravo if the pacts don't fit on their side

PR #2529, opened 2024-10-02
PYIC-7489 openapi types

Joe-Edwards-GDS:
### What changed

- Added response schemas to the external OpenAPI schema
- Load and validate those schemas against responses received in the API tests

### Why did it change

- Our external API needs to be accurately documented, and other teams may rely on this
- Validating in the API tests helps ensure these definitions do not get out of date

### Issue tracking

- [PYIC-7489](https://govukverify.atlassian.net/browse/PYIC-7489)

[PYIC-7489]: https://govukverify.atlassian.net/browse/PYIC-7489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Screenshots

Endpoint configuration after pasting into swagger editor:

image

Token exchange request

image

User identity response (with Vocab schema types)

image

### DIFF
--- a/api-tests/src/clients/core-back-external-client.ts
+++ b/api-tests/src/clients/core-back-external-client.ts
@@ -1,10 +1,43 @@
+import * as assert from "assert";
+import path from "path";
+import { BeforeAll } from "@cucumber/cucumber";
 import { JSONWebKeySet } from "jose";
+import type { Validator } from "jsonschema";
 import config from "../config/config.js";
 import {
   MfaResetResult,
   TokenResponse,
   UserIdentity,
 } from "../types/external-api.js";
+import { createValidator } from "../utils/schema-validator.js";
+
+let schemaValidator: Validator;
+
+BeforeAll(async () => {
+  schemaValidator = await createValidator(
+    path.resolve(
+      import.meta.dirname,
+      "../../../openAPI/core-back-external.yaml",
+    ),
+  );
+});
+
+const validateResponseSchema = async (
+  body: unknown,
+  schemaName: string,
+): Promise<void> => {
+  const result = schemaValidator.validate(
+    body,
+    schemaValidator.schemas[`/${schemaName}`],
+  );
+
+  if (result.errors.length) {
+    const { property, message } = result.errors[0];
+    assert.fail(
+      `External API response did not match schema: ${property} ${message}`,
+    );
+  }
+};
 
 export const exchangeCodeForToken = async (
   tokenExchangeBody: string,
@@ -23,7 +56,10 @@ export const exchangeCodeForToken = async (
     );
   }
 
-  return await response.json();
+  const body = await response.json();
+  await validateResponseSchema(body, "tokenResponse");
+
+  return body;
 };
 
 export const getIdentity = async (
@@ -41,7 +77,10 @@ export const getIdentity = async (
     throw new Error(`getIdentity request failed: ${response.statusText}`);
   }
 
-  return (await response.json()) as UserIdentity;
+  const body = await response.json();
+  await validateResponseSchema(body, "userIdentityResponse");
+
+  return body;
 };
 
 export const getMfaResetResult = async (
@@ -59,15 +98,25 @@ export const getMfaResetResult = async (
     throw new Error(`getMfaResetResult request failed: ${response.statusText}`);
   }
 
-  return (await response.json()) as MfaResetResult;
+  const body = await response.json();
+  await validateResponseSchema(body, "reverificationResponse");
+
+  return body;
 };
 
-export const healthcheck = async (): Promise<boolean> => {
+export const healthcheck = async (): Promise<string> => {
   const response = await fetch(`${config.core.externalApiUrl}/healthcheck`, {
     method: "GET",
   });
 
-  return response.ok;
+  if (!response.ok) {
+    throw new Error(`healthcheck request failed: ${response.statusText}`);
+  }
+
+  const body = await response.json();
+  await validateResponseSchema(body, "healthcheckResponse");
+
+  return body.healthcheck;
 };
 
 export const jwks = async (): Promise<JSONWebKeySet> => {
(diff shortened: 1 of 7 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -28,6 +29,7 @@
 @Setter
 @ExcludeFromGeneratedCoverageReport
 @Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
Joe-Edwards-GDS, line 32: This is a change to the output, but I'm hoping it won't affect any consumers and should help streamline our responses.

(If we don't do this, we have to allow `null` as a valid value for the property in the schemas, which is a bit of a pain!)

cc @CarlyG55

openAPI/core-back-external.yaml
```diff
@@ -117,3 +115,123 @@ paths:
...
+                # $ref: https://vocab.account.gov.uk/v1/json-schemas/BirthDate.json
+                type: object
+          required:
+            - name
+            - birthDate
+          additionalProperties: false
+        "https://vocab.account.gov.uk/v1/address":
+          type: "array"
+          items:
+            $ref: https://vocab.account.gov.uk/v1/json-schemas/PostalAddress.json
```
Joe-Edwards-GDS, line 269: Somewhat surprisingly this seems to Just Work in the Swagger editor!

api-tests/src/clients/core-back-external-client.ts
```diff
@@ -1,10 +1,43 @@
...
+let schemaValidator: Validator;
+
+BeforeAll(async () => {
+  schemaValidator = await createValidator(
+    path.resolve(
+      import.meta.dirname,
+      "../../../openAPI/core-back-external.yaml",
+    ),
+  );
+});
```
Wynndow, line 23: Why do you need to use this constuct? Could you not assign schemaValidator to the result of createValidator directly?

api-tests/src/types/world.ts
```diff
@@ -26,7 +26,7 @@ export interface World extends CucumberWorld {
   mfaResetResult?: MfaResetResult;
 
   // Healthcheck results
-  healthCheckResult?: boolean;
+  healthCheckResult?: string;
```
Wynndow, line 29: Why's this change needed?

api-tests/src/utils/schema-validator.ts
```diff
@@ -0,0 +1,56 @@
...
+    await readFile(openApiPath, "utf-8"),
+  ) as OpenApiSchema;
+
+  Object.entries(openApiSchema.components.schemas).forEach(([key, schema]) =>
+    validator.addSchema(schema, `/${key}`),
+  );
+
+  while (validator.unresolvedRefs.length) {
+    const ref = validator.unresolvedRefs[0];
+    const schema = await fetchVocabSchema(ref);
```
Wynndow, line 46: neat

openAPI/core-back-external.yaml
```diff
@@ -23,19 +47,34 @@ paths:
 
   /user-identity:
     get:
-      description: "Returns a list of Verifiable Credentials representig the users identity"
+      description: "List of Verifiable Credentials and claims representing the user's identity"
+      security:
+        - bearerAuth: ["user-identity"]
```
Wynndow: What does `"user-identity"` here signify?

openAPI/core-back-external.yaml
```diff
@@ -45,17 +84,34 @@ paths:
 
   /reverification:
     get:
-      description: "Returns a success or failure response based on the results of the reverification journey"
+      description: "Result of a reverification journey"
+      security:
+        - bearerAuth: ["reverification"]
```
Wynndow: Same question as above

Wynndow (dismissed):

api-tests/src/clients/core-back-external-client.ts
Joe-Edwards-GDS: as in, just have `const schemaValidator = createValidator(...)`?

The trouble is it's async, so we can't just do it when initialising the module.

api-tests/src/types/world.ts
Joe-Edwards-GDS: It's not really needed, but means that we're checking the `healthcheck` property in the response rather than the status code :shrug:

Would also be happy to revert and just `return true`

openAPI/core-back-external.yaml
Joe-Edwards-GDS: These are the OAuth scopes required...

This is a bit of a bodge tbh, since this only does anything when we use the 'OAuth' security scheme... If we do that, then it wants to know about the authorization/token endpoint etc. required, which is confusing when we're defining those explicitly ourselves. See: https://swagger.io/docs/specification/v3_0/authentication/oauth2/#scopes-extra

Could just remove it, since it doesn't really do anything.

Wynndow: I'm not sure the anchor in your link works properly, and I couldn't see what you were referring to?
This was confusing to me because these endpoints are protected by bearer auth, but it's the access token. So having the scope in here doesn't seem to make sense.
I think if we don't need it, maybe remove it?

Joe-Edwards-GDS: The access token is linked to a scope, but I was mistaken about how this was used - we actually use the same scope when generating all tokens, and don't persist it anyway...

We probably 'should' use a different scope depending on the endpoint, persist that, and check it - but it would just end up duplicating the scope from the JAR (which we check instead). It'd be more relevant if we had separate authorization vs. protected resource services.

...I'll remove it

Wynndow (approved):

PR #2530, opened 2024-10-02
PYIC-7493: Emit dl-auth-source-check event

Wynndow:
### What changed

Emit dl-auth-source-check event

### Why did it change

This adds a special case to the process-cri-callback lamdba for when a user is returning from DCMAW with a successful VC, issued against a driving licence, _and_ the user doesn't already have a successful driving licence CRI VC.

In this scenario we emit an event that will be
used to take them to the driving licence CRI so we can do an authoritative source check.

### Issue tracking

- [PYIC-7493](https://govukverify.atlassian.net/browse/PYIC-7493)

[PYIC-7493]: https://govukverify.atlassian.net/browse/PYIC-7493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS (approved):
👍

PR #2537, opened 2024-10-03
Bau improve api test cri calls

DanCorderIPV:
### What changed

Extended CRI calling steps in API tests to check that all extra properties sent in the JAR to the CRI are specified in the test.

### Why did it change

To ensure that we don't send anything unexpected to the CRIs and to explicitly document the values that we do send.


api-tests/src/steps/cri-steps.ts
```diff
@@ -336,3 +418,38 @@ When(
     );
   },
 );
+
+function assertNoUnexpectedJarProperties(
```
Joe-Edwards-GDS: We've now got a mixture of `function` and arrow function definitions - this is definitely a matter of taste, but it would be nice to be consistent. At the moment pretty much everything else is arrow functions...

api-tests/src/steps/cri-steps.ts
```diff
@@ -336,3 +418,38 @@ When(
...
+      v !== "iss" &&
+      v !== "response_type" &&
+      v !== "client_id" &&
+      v !== "govuk_signin_journey_id" &&
+      v !== "aud" &&
+      v !== "nbf" &&
+      v !== "redirect_uri" &&
+      v !== "state" &&
+      v !== "exp" &&
+      v !== "iat",
```
Joe-Edwards-GDS: Should this just be an array of allowed values - `jarValues.filter((v) => ALLOWED_JAR_ATTRIBUTES.includes(v))`

api-tests/src/steps/cri-steps.ts
```diff
@@ -336,3 +418,38 @@ When(
...
+function assertNoUnexpectedJarProperties(
+  jarPayload: CriStubResponseJarPayload,
+  expectedNonStandardValues: string[] | undefined = undefined,
+) {
+  for (const key of removeStandardJarValues(Object.keys(jarPayload))) {
+    const expectedValueExists =
+      expectedNonStandardValues === undefined
+        ? false
+        : expectedNonStandardValues.includes(key);
+    assert.equal(
```
Joe-Edwards-GDS: You can use `assert.ok` if you just want truthiness

Joe-Edwards-GDS:
I think on the whole this is a net positive, but worth noting that this does make our tests a bit more verbose and bit more brittle (in the cases where you aren't too fussed about the 'extra' properties).

E.g. at some point in the future we might start including `evidence_requested` on _all_ CRI requests, which would add a lot of noise.

Was there a bug you've seen that this would have prevented?

api-tests/src/steps/cri-steps.ts
DanCorderIPV: Thanks, I was surprised `assert.true` didn't exist

api-tests/src/steps/cri-steps.ts
DanCorderIPV: Good point

DanCorderIPV:
As discussed, this wasn't in response to a particular bug, just to make our interface with CRI's more explicit.
It did highlight the strength = 0 requests to F2F which while not wrong were surprising to me.

api-tests/src/steps/cri-steps.ts
```diff
@@ -340,3 +422,34 @@ When(
...
+      expectedNonStandardValues?.includes(key);
+
+    assert.ok(
+      propertyIsExpected,
+      `Non-standard JAR property ${key} with value ${JSON.stringify(jarPayload[key as keyof typeof jarPayload])} sent to CRI but not specified in test`,
+    );
+  }
+};
+
+const STANDARD_JAR_VALUES = [
```
Joe-Edwards-GDS, line 442: (conventionally I'd put constants at the top of the module, but this is fine...)

api-tests/src/types/cri-stub.ts
```diff
@@ -34,10 +34,12 @@ interface CriStubUserInfoEndpointErrorRequest {
...
 export interface CriStubResponse {
   redirectUri: string;
-  jarPayload: {
-    context?: string;
-    evidence_requested?: { strength?: number; validity?: number };
-  };
+  jarPayload: CriStubResponseJarPayload;
+}
+
+export interface CriStubResponseJarPayload {
```
Joe-Edwards-GDS, line 40: (I wonder if it's worth adding a note that this is actually only a small subset of this interface - it's missing all the 'standard' properties!)

Joe-Edwards-GDS (approved):
Couple of tiny notes but looks good

PR #2542, opened 2024-10-03
PYIC-7507: Journey routing for DL auth source checks

Wynndow:
### What changed
Journey routing for DL auth source checks

### Why did it change

This adds journey routing to all the instances of a DCMAW state we have, and handles their unique subtleties... sort of.

### Issue tracking

- [PYIC-7507](https://govukverify.atlassian.net/browse/PYIC-7507)

[PYIC-7507]: https://govukverify.atlassian.net/browse/PYIC-7507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/dl-auth-source-checks.feature
+++ b/api-tests/features/dl-auth-source-checks.feature
@@ -1,0 +1,183 @@
+@Build
+Feature: Authoritative source checks with driving licence CRI
+
+  Scenario Outline: Journey through DCMAW with driving licence requires authoritative source check
+    Given I activate the 'drivingLicenceAuthCheck,p1Journeys' feature sets
+    When I start a new '<journey-type>' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
+    Then I get a 'drivingLicence' CRI response
+    When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'page-dcmaw-success' page response
+
+    Examples:
+      | journey-type       |
+      | low-confidence    |
+      | medium-confidence |
+
+  Scenario Outline: Journey with auth source check that attracts a CI leads to a mitigation journey
+    Given I activate the 'drivingLicenceAuthCheck,p1Journeys' feature sets
+    When I start a new '<journey-type>' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
+    Then I get a 'drivingLicence' CRI response
+    When I submit 'kenneth-driving-permit-needs-alternate-doc' details with attributes to the CRI stub
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'pyi-driving-licence-no-match-another-way' page response
+
+    Examples:
+      | journey-type       |
+      | low-confidence    |
+      | medium-confidence |
+
+  Scenario Outline: Journey where user backs out of driving licence CRI leads to other doc type page
+    Given I activate the 'drivingLicenceAuthCheck,p1Journeys' feature sets
+    When I start a new '<journey-type>' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
+    Then I get a 'drivingLicence' CRI response
+    When I call the CRI stub with attributes and get an 'access_denied' OAuth error
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'prove-identity-another-type-photo-id' page response with context 'drivingLicence'
+
+    Examples:
+      | journey-type      |
+      | low-confidence    |
+      | medium-confidence |
+
+  Scenario: Separate session enhanced verification mitigation with DCMAW and driving licence requires auth source check
+    Given I activate the 'drivingLicenceAuthCheck' feature set
+    And the subject already has the following credentials
+      | CRI        | scenario                            |
+      | ukPassport | kenneth-passport-valid              |
+      | address    | kenneth-current                     |
+      | fraud      | kenneth-score-2                     |
+      | kbv        | kenneth-needs-enhanced-verification |
+
+    # Return journey
+    When I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub that mitigate the 'NEEDS-ENHANCED-VERIFICATION' CI
+    Then I get a 'drivingLicence' CRI response
+    When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'page-dcmaw-success' page response
+
+  Scenario: Same session enhanced verification mitigation with DCMAW and driving licence requires auth source check
+    Given I activate the 'drivingLicenceAuthCheck' feature set
+    When I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I call the CRI stub and get an 'access_denied' OAuth error
+    Then I get a 'page-multiple-doc-check' page response
+    When I submit a 'ukPassport' event
+    Then I get a 'ukPassport' CRI response
+    When I submit 'kenneth-passport-valid' details to the CRI stub
+    Then I get an 'address' CRI response
+    When I submit 'kenneth-current' details to the CRI stub
+    Then I get a 'fraud' CRI response
+    When I submit 'kenneth-score-2' details to the CRI stub
+    Then I get a 'page-pre-experian-kbv-transition' page response
+    When I submit a 'next' event
+    Then I get a 'kbv' CRI response
+    When I submit 'kenneth-needs-enhanced-verification' details with attributes to the CRI stub
+      | Attribute          | Values                                          |
+      | evidence_requested | {"scoringPolicy":"gpg45","verificationScore":2} |
+    Then I get a 'pyi-suggest-other-options' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub that mitigate the 'NEEDS-ENHANCED-VERIFICATION' CI
+    Then I get a 'drivingLicence' CRI response
+    When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'page-ipv-success' page response
+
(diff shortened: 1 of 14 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -730,6 +801,16 @@ states:
...
+  CRI_DRIVING_LICENCE_AUTH_SOURCE_CHECK_AFTER_PYI_ESCAPE:
+    response:
+      type: cri
+      criId: drivingLicence
+    parent: CRI_STATE
+    events:
+      next:
+        targetJourney: EVALUATE_SCORES
+        targetState: START
+
```
Wynndow: This is for same session enhanced verification mitigation. If the user backs out of the drivingLicence CRI  we'll get an access-denied event. At written this will fail the users journey. But we could instead make changes to the reset-session-identity lambda to clear out the DCMAW VC and offer the user F2F as another way to achieve a verification score.
We could potentially do this for any non CI attracting failure from driving licence.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
```diff
@@ -45,6 +45,9 @@ states:
       alternate-doc-invalid-passport:
         targetJourney: FAILED
         targetState: FAILED
+      dl-auth-source-check:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS: It's a bit annoying that we now have to add this as a state for _all_ CRIs - even ones that will never emit it :/

I don't think it's _that_ bad - and does ensure that we never get an unhandled state :shrug:. Otherwise we could finesse the journey map test so that it's aware of certain CRI-specific cases.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -79,6 +79,9 @@ states:
       alternate-doc-invalid-passport:
         targetJourney: FAILED
         targetState: FAILED
+      dl-auth-source-check:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS: Do you think `FAILED` or `TECHNICAL_ERROR` is a better default? (Since this event is still a 'success')

Of course it's a largely academic question!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -198,6 +203,41 @@ states:
       fail-with-no-ci:
         targetState: MULTIPLE_DOC_CHECK_PAGE
 
+  CRI_DRIVING_LICENCE_AUTH_SOURCE_CHECK:
+    response:
+      type: cri
+      criId: drivingLicence
```
Joe-Edwards-GDS, line 209: As per https://gds.slack.com/archives/C068GQY76EL/p1727363487409729 I think we'll want to add `context: check_details` here as well

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -634,6 +676,33 @@ states:
             targetJourney: INELIGIBLE
             targetState: INELIGIBLE
 
+  MITIGATION_01_CRI_DRIVING_LICENCE_AUTH_SOURCE_CHECK:
```
Joe-Edwards-GDS, line 681: All these wordy states are a bit awkward in the journey map but I don't have a better suggestion 🤷

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -634,6 +676,33 @@ states:
...
+      type: cri
+      criId: drivingLicence
+    parent: CRI_STATE
+    events:
+      next:
+        targetState: POST_DCMAW_SUCCESS_PAGE
+      access-denied:
+        targetState: MITIGATION_01_RESET_SESSION_BEFORE_PYI_POST_OFFICE
+      temporarily-unavailable:
+        targetState: MITIGATION_01_RESET_SESSION_BEFORE_PYI_POST_OFFICE
```
Joe-Edwards-GDS: Why do we map `temporarily-unavailable` here, but not for any of the others?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Joe-Edwards-GDS: Ah, so this is the specific case where the user has used a passport in the web route, but for some reason chose to use a DL for the DCMAW mitigation?

We can take it to Joe/Simon to confirm, but I think it's probably fine to let them fail at that point.

Joe-Edwards-GDS (commented):
A couple of thoughts:
- We haven't properly defined the escape routes yet - I think you've defined some sensible options, but they may change.
- Is there any value in extracting DCMAW into a nested journey, now that it's potentially more than just a single state?
    - Could also be a separate task - it needs to be a two-step change to avoid breaking existing states

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Wynndow: Yep fair - anything that isn't a `next` event here will fallback to the defailt parent states, which are all fails.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
Wynndow: Yeah it is a little annoying, but it does give us some protection for future changes. I wonder if we could define parent states in one place, and then have other journey maps reference them...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
Wynndow: I suppose technical error may be better, as the user will have gotten themselves into an unexpected situation. I think I was following the pattern for the events above which all come into the same category. I'll this one to technical error, but might be worth applying that logic across the board.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Wynndow: Do you mean the other events that are defined on `MITIGATION_01_CRI_DCMAW`? If so the other events are `not-found`, `fail-with-no-ci` and `enhanced-verification` - none of which should be possible to get from the DL CRI I think?

Wynndow:
> * Is there any value in extracting DCMAW into a nested journey,

Yeah that's probably a good shout. Might be worth doing after the dust has settled on whatever is going to happen with mitigations though?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
Joe-Edwards-GDS: I mean we don't handle `temporarily-unavailable` on any other instances of the driving licence CRI

(afaict we only handle it from DCMAW - so I assume other CRIs never actually emit it, but happy to be proved wrong!)

Wynndow: I figured we're handling it on `MITIGATION_01_CRI_DCMAW`, so by extension to preserve the functionality of the system (offer them a f2f mitigation route if they can't do it with the app), we'd need to handle it here.
Thinking about it though, this is part of the wider conversation about mitigations. I'll route this to a failure state.

api-tests/features/dl-auth-source-checks.feature
```diff
@@ -0,0 +1,183 @@
...
+    When I start a new '<journey-type>' journey
+    Then I get a 'page-ipv-identity-document-start' page response
+    When I submit an 'appTriage' event
+    Then I get a 'dcmaw' CRI response
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
+    Then I get a 'drivingLicence' CRI response
+    When I submit 'kenneth-driving-permit-needs-alternate-doc' details with attributes to the CRI stub
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'pyi-driving-licence-no-match-another-way' page response
```
Wynndow, line 33: It would be nice if we could test that the reset-session lambda is run at this point, but we don't have a mechanism, and the lambda doesn't emit an audit event...

Joe-Edwards-GDS: I guess the 'test' would need to complete the mitigation journey and either:
- check that the identity bundle at the end doesn't contain a DCMAW VC
- complete the mitigation route with a different name (on the basis that it would fail correlation if the DCMAW VC wasn't wiped) - but this is a pretty indirect way of doing it!

api-tests/features/dl-auth-source-checks.feature
```diff
@@ -0,0 +1,183 @@
...
+      | medium-confidence |
+
+  Scenario: Separate session enhanced verification mitigation with DCMAW and driving licence requires auth source check
+    Given I activate the 'drivingLicenceAuthCheck' feature set
+    And the subject already has the following credentials
+      | CRI        | scenario                            |
+      | ukPassport | kenneth-passport-valid              |
+      | address    | kenneth-current                     |
+      | fraud      | kenneth-score-2                     |
+      | kbv        | kenneth-needs-enhanced-verification |
```
Joe-Edwards-GDS, line 69: One observation about this - with the session store, you actually won't have any of these credentials - you'll just have the CI.

It's not really a problem (and historic users might well have the credentials, at least until we perform the EVCS migration).

Joe-Edwards-GDS (dismissed):
A couple of notes about the tests but otherwise looks good!

We can think about extracting a nested journey for DCMAW after this, but could be easier once we have the full scope.

One last thought - do you think there's any value in putting the journey routes behind a feature flag? I can't really see one, since generation of the event is feature-flagged instead...

api-tests/features/dl-auth-source-checks.feature
Wynndow: Ah yeah, option 1 is the simplest. I'll add it in.

api-tests/features/dl-auth-source-checks.feature
Wynndow: Yeah, that's a good point - sort of means the changes I recently made to post to cimit in this step are a bit off. I'll look at fixing it up and adding something to just send CIs to cimit, but probably not in this PR

Wynndow:
> do you think there's any value in putting the journey routes behind a feature flag?

No, not particularly I don't think. The event is flagged, which I think is enough...

api-tests/features/dl-auth-source-checks.feature
```diff
@@ -31,6 +31,10 @@ Feature: Authoritative source checks with driving licence CRI
       | Attribute | Values          |
       | context   | "check_details" |
     Then I get a 'pyi-driving-licence-no-match-another-way' page response
+    When I submit an 'end' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P0' identity without a drivingLicence VC
```
Joe-Edwards-GDS: Do you mean `drivingLicence` or `dcmaw` here? I think dcmaw is the problematic one, since it'll be valid!

Wynndow: oh flip, good spot!

Joe-Edwards-GDS (approved):
Thanks!

PR #2553, opened 2024-10-07
PYIC-6236: Make ProcessAsyncCriCredentialHandler inclusive for DCMAW

MikeCollingwood:
### What changed

- Use state to index userId's cri response instead of criId.
- Make sending audit events conditional only for F2F cri.
- Remove criId from expected queue response dto.

### Why did it change

- We need to find out the criId instead of hardcoding F2F, as we are trying to be inclusive to DCMAW async VCs.
- Cri Id isn't sent via queue message, but state is.
- Audit events are only expected for F2F at the moment.

### Issue tracking

- [PYIC-5039](https://govukverify.atlassian.net/browse/PYIC-5039)

### Notes

- Tested it and it works: DCMAW async VC gets stored & no audit events
- Still want to check assumptions (see comments)
- CRI oauth configs to have signing key

[PYIC-5039]: https://govukverify.atlassian.net/browse/PYIC-5039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
+++ b/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
@@ -19,6 +19,7 @@ import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionErrorParams;
 import uk.gov.di.ipv.core.library.cimit.exception.CiPostMitigationsException;
 import uk.gov.di.ipv.core.library.cimit.exception.CiPutException;
 import uk.gov.di.ipv.core.library.config.ConfigurationVariable;
+import uk.gov.di.ipv.core.library.domain.Cri;
 import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
 import uk.gov.di.ipv.core.library.exception.EvcsServiceException;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
@@ -44,13 +45,14 @@ import java.text.ParseException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 
 import static uk.gov.di.ipv.core.library.auditing.helpers.AuditExtensionsHelper.getExtensionsForAudit;
 import static uk.gov.di.ipv.core.library.auditing.helpers.AuditExtensionsHelper.getRestrictedAuditDataForF2F;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EVCS_ASYNC_WRITE_ENABLED;
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.EVCS_READ_ENABLED;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.UNEXPECTED_ASYNC_VERIFIABLE_CREDENTIAL;
-import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_CRI_ISSUER;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_ERROR_CODE;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_ERROR_DESCRIPTION;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
@@ -149,14 +151,13 @@ public class ProcessAsyncCriCredentialHandler
     }
 
     private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriResponse) {
-        CriResponseItem responseItem =
-                criResponseService.getCriResponseItem(
-                        errorAsyncCriResponse.getUserId(),
-                        errorAsyncCriResponse.getCredentialIssuer());
+        var userId = errorAsyncCriResponse.getUserId();
+        var state = errorAsyncCriResponse.getOauthState();
+        Optional<CriResponseItem> criResponseItem = getCriResponseItem(userId, state);
 
-        if (responseItem != null) {
-            responseItem.setStatus(CriResponseService.STATUS_ERROR);
-            criResponseService.updateCriResponseItem(responseItem);
+        if (criResponseItem.isPresent()) {
+            criResponseItem.get().setStatus(CriResponseService.STATUS_ERROR);
+            criResponseService.updateCriResponseItem(criResponseItem.get());
         }
 
         LOGGER.error(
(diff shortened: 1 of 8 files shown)


lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -167,44 +168,49 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
...
                                 LOG_ERROR_DESCRIPTION.getFieldName(),
                                 errorAsyncCriResponse.getErrorDescription())
-                        .with(LOG_ERROR_CODE.getFieldName(), errorAsyncCriResponse.getError())
-                        .with(
-                                LOG_CRI_ISSUER.getFieldName(),
-                                errorAsyncCriResponse.getCredentialIssuer()));
+                        .with(LOG_ERROR_CODE.getFieldName(), errorAsyncCriResponse.getError()));
 
-        sendIpvVcErrorAuditEvent(errorAsyncCriResponse);
+        if (criResponseItem.isPresent()
```
MikeCollingwood: This will only be called when there is a cri response, as we need it to check for F2F cri Id.

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -167,44 +168,49 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
                         .with(
                                 LOG_ERROR_DESCRIPTION.getFieldName(),
                                 errorAsyncCriResponse.getErrorDescription())
-                        .with(LOG_ERROR_CODE.getFieldName(), errorAsyncCriResponse.getError())
-                        .with(
```
MikeCollingwood, line 171: Assuming we can't replace this with the oauth state.

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -224,51 +228,47 @@ private void processSuccessAsyncCriResponse(SuccessAsyncCriResponse successAsync
...
             boolean isSuccessful)
             throws UnrecognisedVotException {
-        AuditEvent auditEvent =
-                AuditEvent.createWithoutDeviceInformation(
-                        AuditEventTypes.IPV_F2F_CRI_VC_RECEIVED,
-                        configService.getParameter(ConfigurationVariable.COMPONENT_ID),
-                        auditEventUser,
-                        getExtensionsForAudit(verifiableCredential, isSuccessful));
-        auditService.sendAuditEvent(auditEvent);
+        if (Cri.F2F.equals(cri)) {
```
MikeCollingwood, line 252: It might be nicer to put this logic in the parent method but it makes it too complex according the SonarCloud

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -167,44 +168,47 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
...
-                criResponseService.getCriResponseItem(
-                        successAsyncCriResponse.getUserId(),
-                        successAsyncCriResponse.getCredentialIssuer());
-        validateOAuthState(successAsyncCriResponse, criResponseItem);
+        var userId = successAsyncCriResponse.getUserId();
+        var state = successAsyncCriResponse.getOauthState();
+        Optional<CriResponseItem> criResponseItem = getCriResponseItem(userId, state);
+
+        if (criResponseItem.isEmpty()) {
+            LOGGER.error(LogHelper.buildLogMessage("No response item found"));
```
blakeyp: maybe 'No response item found given user id and state'? as this is effectively replacing the state validation

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -167,44 +168,47 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
...
+            throw new AsyncVerifiableCredentialException(UNEXPECTED_ASYNC_VERIFIABLE_CREDENTIAL);
+        }
 
-        configService.setFeatureSet(criResponseItem.getFeatureSet());
+        var cri = Cri.fromId(criResponseItem.get().getCredentialIssuer());
```
blakeyp, line 196: Where do we actually set the CRI id on the response item? I seem to remember we just assumed it's F2F, but this could become a problem if we need to retrieve and distinguish pending records in check existing identity 🤔

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
blakeyp: I would agree but I think this is ok for now as we do have a follow-up to handle audit events for both CRIs (possibly with new generic events)

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -224,51 +228,47 @@ private void processSuccessAsyncCriResponse(SuccessAsyncCriResponse successAsync
...
+                            AuditEventTypes.IPV_F2F_CRI_VC_CONSUMED,
+                            configService.getParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser,
+                            null,
+                            getRestrictedAuditDataForF2F(vc));
+            auditService.sendAuditEvent(auditEvent);
+        }
     }
 
     private void sendIpvVcErrorAuditEvent(ErrorAsyncCriResponse errorAsyncCriResponse) {
```
blakeyp: to be consistent it might be clearer to do the same here (ie. move the conditional inside the method), for now?

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
MikeCollingwood: It's set by (the entire cri response created by) the `call-dcmaw-async-cri` which produces a "pending" CRI response before you fill in details to get the VC which then appears on the queue. In this handler there is a section of code:

```java
criStoringService.recordCriResponse(
    request, DCMAW_ASYNC, oauthState, clientOAuthSessionItem, featureSets);
```

Which saves the cri response with `dcmawAsync` as the cri id. So it should be ok as we have also seen these cri responses in DynamoDb during testing.

We _did_ use to assume criId=F2F when creating the `(success/error)AsyncCriResponse` object but you can see we have altogether removed that attribute now as it is really just a dto for the queue response which doesn't include a cri id.

blakeyp (approved):

PR #2561, opened 2024-10-08
BAU DWP KBV credential request headers and debug logs

blakeyp:
### What changed

For DWP KBV debugging the credential request, set Content-Type header with valid value and Content-Length 0. Also add some more debug logs

### Why did it change

For DWP KBV debugging


MikeCollingwood (approved):
Looks sensible for debugging

PR #2577, opened 2024-10-10
PYIC-7500: throw error when client auth jwt id has already been used

thebauSoftwire:
### What changed
When we detect a replay of client auth JWTs or if the id is missing or empty, we now fail the request by throwing an error and no longer storing the ID instead of just logging a warning message.

### Why did it change
The request should fail anyway as the auth code has already been consumed so we should be throwing an error instead of just logging a warning.

### Issue tracking

- [PYIC-7500](https://govukverify.atlassian.net/browse/PYIC-7500)

[PYIC-7500]: https://govukverify.atlassian.net/browse/PYIC-7500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
+++ b/lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
@@ -72,7 +72,7 @@ public class TokenRequestValidator {
         }
     }
 
-    private void validateJwtId(JWTAuthenticationClaimsSet claimsSet) {
+    private void validateJwtId(JWTAuthenticationClaimsSet claimsSet) throws InvalidClientException {
         JWTID jwtId = claimsSet.getJWTID();
         if (jwtId == null || StringUtils.isBlank(jwtId.getValue())) {
             LOGGER.warn(LogHelper.buildLogMessage("The client auth JWT id (jti) is missing"));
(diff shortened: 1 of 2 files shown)


lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
```diff
@@ -72,7 +72,7 @@ private void validateMaxAllowedAuthClientTtl(JWTAuthenticationClaimsSet claimsSe
         }
     }
 
-    private void validateJwtId(JWTAuthenticationClaimsSet claimsSet) {
+    private void validateJwtId(JWTAuthenticationClaimsSet claimsSet) throws InvalidClientException {
         JWTID jwtId = claimsSet.getJWTID();
         if (jwtId == null || StringUtils.isBlank(jwtId.getValue())) {
             LOGGER.warn(LogHelper.buildLogMessage("The client auth JWT id (jti) is missing"));
```
Joe-Edwards-GDS: Do you think we can turn this into an error now as well? (Otherwise you can bypass the replay detection by omitting a JTI!)

I can't see any instances of this log message in prod, but we should be a little careful that it doesn't break anything!

Joe-Edwards-GDS (dismissed):
Thanks - one comment about behaviour if the JTI is missing though

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
thebauSoftwire: ack good point - thanks!

Joe-Edwards-GDS: No worries - it wasn't on the ticket :)

Joe-Edwards-GDS (approved):

PR #2580, opened 2024-10-10
BAU Further DWP KBV debug logging

blakeyp:
Logging staging (only) token and complete requests so can be reproduced locally


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -284,8 +285,9 @@ private JourneyResponse getJourneyResponse(CriCallbackRequest callbackRequest)
 
         // Retrieve, store and check cri credentials
         var accessToken = criApiService.fetchAccessToken(callbackRequest, criOAuthSessionItem);
-        // Temporarily logging DWP KBV token for integration debugging
-        if (callbackRequest.getCredentialIssuer().equals(DWP_KBV)
+        // Temporarily logging staging token for debugging purposes
+        if ((callbackRequest.getCredentialIssuer().equals(DWP_KBV)
+                        || callbackRequest.getCredentialIssuer().equals(NINO))
```
blakeyp, line 290: I want to be able to reproduce NINO CRI VC requests locally too, as NINO CRI weirdly seems to require a Content-Type set to empty string which doesn't seem right

blakeyp:
Sonarcloud is complaining because the new log method is not covered - but this is just temporary code to be removed soon

shivanshuit914 (approved):

PR #2583, opened 2024-10-11
BAU: Allow nested objects in received EVCS metadata

Wynndow:
### What changed

Allow nested objects in received EVCS metadata

### Why did it change

We need to be more liberal in what we accept in the metadata on VCs received from EVCS. For the migration we're sending them a nested object, so we need to be able to receive it.


Joe-Edwards-GDS (approved):
I feel like I saw this ages ago and twigged a 'what if the metadata is an unexpected shape' but obviously I totally forgot about it!

Well done for spotting!

PR #2587, opened 2024-10-14
PYIC-7319: Add no photo id api tests

MikeCollingwood:
### What changed

- Add no photo id api tests from selenium

### Why did it change

- Move towards api tests for reliability and speed

### Issue tracking

- [PYIC-7319](https://govukverify.atlassian.net/browse/PYIC-7319)

[PYIC-7319]: https://govukverify.atlassian.net/browse/PYIC-7319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/p2-no-photo-id.feature
```diff
@@ -0,0 +1,435 @@
...
+      When I submit 'kenneth-score-2' details with attributes to the CRI stub
+        | Attribute          | Values                                          |
+        | evidence_requested | {"scoringPolicy":"gpg45","verificationScore":2} |
+      Then I get a 'page-ipv-success' page response
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P2' identity
+
+    Scenario: P2 no photo id journey - Experian - BAV dropout:
```
DanCorderIPV, line 41: We don't get to the KBV so it probably doesn't matter, but the end-to-end tests used the HmrcKbv flag for BAV and NINO dropout.
Given that we don't get to the KBV stage in these tests, maybe move them to the `dropout` section below?

DanCorderIPV (approved):

api-tests/features/p2-no-photo-id.feature
MikeCollingwood: Thank you for the feedback. I understand enabling a flag when not needed isn't ideal but I'm struggling to piece together how the re-categorisation would work. I am keen to keep as much as possible in the two main rule groupings Experian and KBV just to make it easier reading. The other rule groups are there because they need to delve into a specific state really. With that, I will merge what's here, and if anything comes up, I can take another look.

PR #2595, opened 2024-10-15
BAU DWP KBV Try making barebones token request

blakeyp:
### What changed

For DWP KBV debugging, try making a barebones token POST request to the endpoint without any headers or body

### Why did it change

DWP's API gateway is still blocking our requests with a 403 Forbidden error despite them lifting their firewall. They believe this may be due to something we're setting in our request, so we want to try just hitting the endpoint to rule that theory in/out


blakeyp:
This is temporary debugging code to be reverted soon, so ignoring sonar issues

shivanshuit914 (approved):

PR #2596, opened 2024-10-15
PYIC-7578: Retry on `SnapStartNotReadyException`

Wynndow:
### What changed

Retry on `SnapStartNotReadyException`

### Why did it change

We've seen some errors in perf tests where SnapStart not being ready has caused the lambda invocation to fail. This happens when a function hasn't been used for more than 14 days and the snapstart snapshot has been binned by AWS. We can catch that error and retry to try to prevent failures.

This doesn't go as far as retrying for other error types. Mostly because we're not sure how idempotent our lambdas are, and we haven't seen a problem so far so no need to fix it.

It was suggested we could try to retry the issues with are seeing when a lambda segfaults. Unfortunatley the error returned to the step function (`Runtime.ExitError`) isn't specific enought to reliably only retry that error.

This also only fixes this issue for lambdas orchstrated by a stepfunction. Those that are invoked directly by the API gateway will suffer from this issue. APIGateway doesn't appear to offer anything, so we'd need to handle and retry on the client side (core-front) if wanted to try and fix them.

### Issue tracking

- [PYIC-7578](https://govukverify.atlassian.net/browse/PYIC-7578)

[PYIC-7578]: https://govukverify.atlassian.net/browse/PYIC-7578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/journeyEngineStepFunction.asl.json
+++ b/deploy/journeyEngineStepFunction.asl.json
@@ -11,7 +11,14 @@
         "featureSet.$": "$$.Execution.Input.featureSet",
         "deviceInformation.$": "$$.Execution.Input.deviceInformation"
       },
-      "Next": "ProcessJourneyStepResult"
+      "Next": "ProcessJourneyStepResult",
+      "Retry": [{
+        "ErrorEquals": ["Lambda.SnapStartNotReadyException"],
+        "IntervalSeconds": 1,
+        "BackoffRate": 2,
+        "MaxAttempts": 5,
+        "MaxDelaySeconds": 4
+      }]
     },
     "ProcessJourneyStepResult": {
       "Type": "Choice",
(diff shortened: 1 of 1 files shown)


deploy/journeyEngineStepFunction.asl.json
Joe-Edwards-GDS, line 1: I assume we have to define this block per lambda, and can't set a default.

I guess that means we _could_ change which errors get included depending on which lambdas we view as idempotent :shrug: - but for now I don't think it's worth looking at, unless we see evidence of lambdas failing in prod.

Joe-Edwards-GDS (approved):

PR #2600, opened 2024-10-16
BAU DWP KBV debugging - build up token request

blakeyp:
### What changed

Try a barebones http token request including the body - this should return 200

### Why did it change

Temporary DWP KBV debugging code


blakeyp:
Temporary debugging code so ignoring sonar

shivanshuit914 (approved):

PR #2602, opened 2024-10-16
BAU DWP KBV clean up debug logging and explicitly set headers

blakeyp:
### What changed

Clean up DWP KBV debugging logging to date
Explicitly set Accept and Content-Type headers for DWP KBV API requests, for now

### Why did it change

DWP KBV API's gateway is rejecting our requests because nimbus sets header values in the underlying http requests that include `charset` parameters eg. `Accept: application/json; charset=UTF-8`. While these are valid header values, DWP doesn't like them so while discussions are ongoing we'll explicitly set the headers without the charset for now.


blakeyp:
Temporary debugging code so ignoring sonar issues

DanCorderIPV (approved):

PR #2616, opened 2024-10-22
PYIC-7581: Migration reconciliation script

Wynndow:
### What changed

Migration reconciliation script

### Why did it change

This adds a script, packaged in a lambda, to reconcile the VCs migrated into EVCS with those in the tactical store.

It will always compare the VCs directly. It can optionally be invoked to check the VCs signatures, or if the VCs can form a P2 identity.

### MemorySize vs duration/cost (red - just checking VC equality, blue - also checking signatures)
Screenshot 2024-10-25 at 09 11 00

### ~AWS Lambda Power Tuning tool [results](https://lambda-power-tuning.show/#AAQACAAMABAAFAAYABwAIAAkACg=;VUZfR6uqv0UAbKZFANi1RavyskVVOadFq8abRau6oUVVIZZFq6WzRg==;6MdLOg/zLjk922M5G/elOTctzDn/9+Q5+s/4OUegEzrfLBo6ufNMOw==) - run with signature validation on 5000 identities~ THIS IS A LOAD OF RUBBISH, IGNORE
Screenshot 2024-10-24 at 14 04 23

### Issue tracking

- [PYIC-7581](https://govukverify.atlassian.net/browse/PYIC-7581)

[PYIC-7581]: https://govukverify.atlassian.net/browse/PYIC-7581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -0,0 +1,412 @@
...
+                    LogHelper.buildErrorMessage("Parallel execution failed", e)
+                            .with(LOG_BATCH_ID.getFieldName(), reconciliationReport.getBatchId())
+                            .with(LOG_ERROR_STACK.getFieldName(), e.getStackTrace()));
+            reconciliationReport.setExitReason("Parallel execution failed");
+        } finally {
+            forkJoinPool.shutdownNow();
+            var unprocessedIdentities = new ArrayList<>(request.userIds());
+            unprocessedIdentities.removeAll(processedIdentities);
+            reconciliationReport.setUnprocessedHashedUserIds(
+                    unprocessedIdentities.stream().map(DigestUtils::sha256Hex).toList());
```
Wynndow, line 149: calculating the hash here to avoid returning a list of user identities. This does mean though that we need to do some processing on the result to be able to complete the batch. Do we need to do this, or would returning the unhashed IDs be acceptable - I'm guessing not as we also log the report.

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    private void checkSignatures(List<VerifiableCredential> evcsVcs, String hashedUserId) {
+        var identityFailedSigValidation = false;
+        for (var vc : evcsVcs) {
+            try {
+                if (!validSignature(vc, hashedUserId)) {
+                    LOGGER.info(annotateLog("Failed to validate signature", hashedUserId));
+                    reconciliationReport.incrementInvalidVcSignatureCount(hashedUserId, vc);
+                    identityFailedSigValidation = true;
```
Joe-Edwards-GDS, line 245: (You could simplify this a little by inverting the condition and returning here if the validation passes - but not very important)

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/domain/ReconciliationReport.java
```diff
@@ -0,0 +1,108 @@
...
+        identityWithInvalidSignaturesCount++;
+        identityWithAnyInvalidSignaturesHashedUserIds.add(hashedUserId);
+    }
+
+    public synchronized void incrementInvalidVcSignatureCount(
+            String hashedUserId, VerifiableCredential vc) {
+        invalidVcSignatureCount++;
+        invalidVcSignatureHashedUserIdAndSigs.add(
+                String.format(
+                        "%s - %s", hashedUserId, vc.getSignedJwt().getSignature().toString()));
```
Joe-Edwards-GDS: I wonder if it'd be more useful to include the CRI than the signature (or both?)

...I'm not sure we'll actually do this in prod anyway, but still

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -0,0 +1,413 @@
...
+                var shortestLength = Math.min(evcsVcLength, tacticalVcLength);
+                int diffCount = 0;
+                for (int j = 0; j < shortestLength; j++) {
+                    if (evcsVc.charAt(j) != tacticalVc.charAt(j)) {
+                        diffCount++;
+                    }
+                }
+                logMap.with(String.format("pair%dDiffCount", i), diffCount)
+                        .with(String.format("evcsVc%dLength", i), evcsVcLength)
+                        .with(String.format("tacticalVc%dLength", i), evcsVcLength);
```
Joe-Edwards-GDS: This is neat, but is it actually useful?

If there are any differences then I think we'll want to inspect the VCs ourselves anyway... Feels like it'd be better to list the CRIs and/or signatures for VCs present in each store but not the other.

(Similar to the comparison logic in `CheckExistingIdentity`)

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Joe-Edwards-GDS: Yeah, I think we should stick to our principle of not logging plaintext user ids.

I suppose you could log the hashed IDs, but return them unhashed for convenience?

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/domain/Request.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.di.ipv.core.reconcilemigratedvcs.domain;
+
+import java.util.List;
+
+public record Request(
+        String batchId,
+        List<String> userIds,
```
Joe-Edwards-GDS, line 7: Have you discussed a plan for constructing this list with Taher?

If we end up deciding to do 100k users at a time (the size of a batch), it'll be ~5MB of JSON which is getting quite close to the lambda limit (6MB).

(If it's only a small %, then it may not be a concern)

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Wynndow: Problem is we're logging the report - in case we lose the connection somehow from the local machine. I've got a PR for a script which can do a reverse lookup using the reports table pretty quick, so I'm actually not too worried about it

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Wynndow: Not sure I follow...

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/domain/ReconciliationReport.java
Wynndow: Yeah - I suppose it might be slightly more useful, allowing us to target just the VC we want to inspect from the tactical store. Also more concise.

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Wynndow: This is probably not really very useful - I was aiming at being able to see in the logs if there were large differences, or just a byte had change. But we'll probably just dive in and look at the VCs.
I'd rather avoid having to parse the VCs if possible, so logging the signatures for differing VCs might be better.

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/domain/Request.java
Wynndow: I believe the plan is to export a CSV of IDs from the success audit events. As yet not discussed if we're going to reconcile everything or a sample.
i've been running tests with 100000 identities provided in the json without issue - although I should probably ensure that the format of the ids is the same...

deploy/template.yaml
```diff
@@ -2564,6 +2564,76 @@ Resources:
       FilterPattern: ""
       LogGroupName: !Ref BulkMigrateVcsFunctionLogGroup
 
+  ReconcileMigratedVcsFunction:
```
TaherKapasi, line 2567: Will this lambda automatically incorporate the DynaTrace layer,  if so is the memory allocation sufficient?

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -0,0 +1,413 @@
...
+        reconciliationReport.incrementIdentitiesFullyProcessed();
+
+        checkTimeout(context, hashedUserId);
+    }
+
+    private void checkSignatures(List<VerifiableCredential> evcsVcs, String hashedUserId) {
+        var identityFailedSigValidation = false;
+        for (var vc : evcsVcs) {
+            try {
+                if (!validSignature(vc, hashedUserId)) {
```
TaherKapasi, line 242: Will the `validSignature` method deem `exp`ired VCs as still valid and not have a maxAge restriction

deploy/template.yaml
Wynndow: Yeah it will - and the memory as configured should be fine - it's how all our other lambdas run, including the bulk migration one.
Having said that, I've been playing around with memory size and how that affects the performance/cost of the lambda. And I'm going to add a commit to max it out. The speed up is significat, and the extra cost is approx 1 cent per invocation. Seems worth it.

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Wynndow: Yeah, we're not inspecting anything other than the signature. We have other validators for inspecting the claims but we're not using them here.

TaherKapasi (dismissed):

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Joe-Edwards-GDS: ok sure

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Joe-Edwards-GDS: Hmm, now I can't work it out - just ignore!

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
Joe-Edwards-GDS: Makes sense - I think the chance of a single byte flipping is essentially zero.

What might be worth ensuring is that you have something to handle a truncated VC adequately (e.g. if the signature has been dropped). We'll investigate any issue manually anyway, but just so long as the script doesn't fail. (something like `<missing signature>` would be fine!)

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/domain/Request.java
Wynndow: Looks like if we're using the same format of user ID we breach the limit. Simpliest thing is to just chop it in two and run smaller lists in to it.

Joe-Edwards-GDS: I hope you like running this script a lot 😰

Wynndow: We could actually just run it in parallel - it doesn't mutate anything so shouldn't be an issue.

lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -103,7 +103,7 @@ public ReconciliationReport handleRequest(Request request, Context context) {
         var forkJoinPool =
                 forkJoinPoolFactory.getForkJoinPool(
                         request.parallelism() == null
-                                ? DEFAULT_PARALLELISM
+                                ? Runtime.getRuntime().availableProcessors()
```
Joe-Edwards-GDS, line 106: Do you know if this will do what you want in the lambda runtime? I didn't get useful results with it when I tried in the migration script.

It's also not necessarily a good default - I would expect most of the time in the script is waiting for requests to EVCS/DynamoDB, so won't be utilizing a processor.

Anyway, you've played with the parallelism so you can tell me what's most effective!

Wynndow: I wanted to work out how many CPUs were available when upping the memory to max. 6 it turns out. There is only a small increase in perf (and probably not statistically significant with a sample size of 1) between 4 threads and 6. But seeing as we have those CPUs available, and we also have the ability to set the parallelism through the request to whatever we want, I figured may as well leave it in.

Joe-Edwards-GDS (approved):
Happy for you to merge this so you can start testing in a higher environment

PR #2622, opened 2024-10-25
PYIC-6402: Add backLink to specific mobile triage flow pages

AmritSidhu:
### What changed

Add backLink to specific mobile triage flow pages

### Issue tracking

- [PYIC-6402](https://govukverify.atlassian.net/browse/PYIC-6402)

[PYIC-6402]: https://govukverify.atlassian.net/browse/PYIC-6402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-triage.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-triage.yaml
@@ -28,6 +28,8 @@ nestedJourneyStates:
         targetState: DAD_SELECT_SMARTPHONE
       smartphone:
         targetState: MAM_SELECT_SMARTPHONE
+      back:
+        targetState: IDENTIFY_DEVICE
 
   DAD_SELECT_SMARTPHONE:
     response:
@@ -41,6 +43,8 @@ nestedJourneyStates:
         targetState: DAD_ANDROID_START_SESSION
       neither:
         targetState: DAD_SELECT_SMARTPHONE_EXIT_BUFFER
+      back:
+        targetState: SELECT_DEVICE_PAGE
 
   DAD_SELECT_SMARTPHONE_EXIT_BUFFER:
     response:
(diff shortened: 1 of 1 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-triage.yaml
```diff
@@ -28,6 +28,8 @@ nestedJourneyStates:
         targetState: DAD_SELECT_SMARTPHONE
       smartphone:
         targetState: MAM_SELECT_SMARTPHONE
+      back:
```
blakeyp, line 31: Do the back events definitely need to be defined here? My understanding of #2015 is the journey engine can find the previous state by itself, at least for page responses

AmritSidhu: I did talk to Dan about this. I dont believe it does need to be there but we do have it defined elsewhere we use the back links. So i put them here to show these pages do have them enabled otherwise we have no clear way of knowing which pages are using them and which ones are without manually checking the tempaltes. But happy to remove them if you think we shouldn't be defining them?

blakeyp: @Joe-Edwards-GDS do you have any thoughts on this? from what I can see I think we only define back events where one of the radio options is actually a 'go back' option

Joe-Edwards-GDS: I would prefer not to define back events explicitly, unless they need to route somewhere unexpected (i.e. not the previous screen!)

This means that it'll always go to the right place, and most usefully, you can route multiple pages into the same one, and the back button will follow the correct path back (otherwise you need to define multiple copies, just so they know where 'back' goes).

It is a bit annoying that it doesn't show up in the journey map... My hope is that as we start including `back` on more and more pages, the default assumption shifts to 'all pages allow you to go back' (if the previous state is a page as well).

Happy to take dissenting views/ideas, if you think it's too confusing.

PR #2623, opened 2024-10-25
PYIC-7571: Enabled parsing of RSA signatures

Wynndow:
### What changed

Enabled parsing of RSA signatures

### Why did it change

Looks like we have some vcs signed with RSA rather than EC, so lets enable it.

### Issue tracking

### DIFF
--- a/lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
+++ b/lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
@@ -7,8 +7,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JWSVerifier;
 import com.nimbusds.jose.crypto.ECDSAVerifier;
+import com.nimbusds.jose.crypto.RSASSAVerifier;
 import com.nimbusds.jose.crypto.impl.ECDSA;
-import com.nimbusds.jose.jwk.ECKey;
+import com.nimbusds.jose.jwk.JWK;
 import com.nimbusds.jose.util.Base64URL;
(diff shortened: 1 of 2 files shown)


lambdas/reconcile-migrated-vcs/src/main/java/uk/gov/di/ipv/core/reconcilemigratedvcs/ReconcileMigratedVcsHandler.java
```diff
@@ -136,10 +138,14 @@ public ReconciliationReport handleRequest(Request request, Context context) {
...
                 Thread.currentThread().interrupt();
             }
-            LOGGER.error(
+            var logMessage =
                     LogHelper.buildErrorMessage("Parallel execution failed", e)
                             .with(LOG_BATCH_ID.getFieldName(), reconciliationReport.getBatchId())
-                            .with(LOG_ERROR_STACK.getFieldName(), e.getStackTrace()));
+                            .with(LOG_ERROR_STACK.getFieldName(), e.getStackTrace());
+            if (e.getCause() != null && e.getCause().getStackTrace() != null) {
+                logMessage.with("causeErrorStack", e.getCause().getStackTrace());
```
Wynndow, line 146: This is just adding logging a stack trace for the cause - otherwise the stacktrace for an ExecutionException just takes you to the `get()` call on the task (line 130)

Joe-Edwards-GDS: This feels like a bit of a smell tbh, but I'm willing to leave it in just for this script.

Ideally we'd leverage the error logging capability within Log4J (which must be able to log stack traces sensibly), and if that's not possible then use something like `printStackTrace`.

For example this won't handle more than one nested `cause`...

Joe-Edwards-GDS (approved):
I think we could also just confirm that we're _not_ going to verify signatures and rip all this out...

Wynndow:
> I think we could also just confirm that we're _not_ going to verify signatures and rip all this out...

I'd be delighted for this to happen!

PR #2624, opened 2024-10-25
PYIC-7571: Fetch all CRI by issuer once

Wynndow:
### What changed

Fetch all CRI by issuer once

### Why did it change

When running the reconciliation lambda in parallel (more than one lambda execution at once), we were seeing SSM rate limiting errors from powertools. This is weird because it would appear that it's not always respecting the cache.

To avoid this issue this create a new method on the config service which will create a map of all configured issuers for the env to allow a lookup of the associated CRI. We then use this to populate a map for the script which it can refer to throughout the duration of its running.

I've run four parallel exections of this script with this change and experienced no issues. This allowed the checking of the signatures for 100000 identities in less than two minutes.

### Issue tracking


Joe-Edwards-GDS (approved):
This is a bit strange - I've also seen some issues confirming powertools caching behaviour when messing about with the TS lambda, but :shrug:

Wynndow:
> This is a bit strange - I've also seen some issues confirming powertools caching behaviour when messing about with the TS lambda, but 🤷

I sort of assumed this was because we were pulling using a path prefix - but I've just seen it happen again but for dead normal param. And about 20 seconds after a lambda started. Which sort of makes me thing it's not caching properly at all...

PR #2626, opened 2024-10-25
PYIC-7571: Fetch historic signing keys up front

Wynndow:
### What changed

Fetch historic signing keys up front

### Why did it change

This updates the reconciliation script to ask the config service for the historic signing keys just once, rather than relying on the values existing in its cache. The cache is being weird and I don't trust it.

This also processes the keys into verifiers which simplifies the validation logic later on.

### Issue tracking


Joe-Edwards-GDS (approved):
This is probably more efficient anyway, since we don't need to construct a new verifier each time.

PR #2635, opened 2024-10-28
PYIC-7601: extract dcmaw-driving-licence into it's own nested journey

thebauSoftwire:
### What changed
Extract dcmaw + driving licence into its own nested journey

### Why did it change
With the addition of the DL auth source checks, we can commonise the dcmaw-driving licence checks into its own nested journey

### Issue tracking

- [PYIC-7601](https://govukverify.atlassian.net/browse/PYIC-7601)

[PYIC-7601]: https://govukverify.atlassian.net/browse/PYIC-7601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/NestedJourneyTypes.java
```diff
@@ -13,7 +13,8 @@ public enum NestedJourneyTypes {
     ADDRESS_AND_FRAUD("address-and-fraud"),
     KBVS("kbvs"),
     STRATEGIC_APP_TRIAGE("strategic-app-triage"),
-    WEB_DL_OR_PASSPORT("web-dl-or-passport");
+    WEB_DL_OR_PASSPORT("web-dl-or-passport"),
+    DCMAW_DRIVING_LICENCE("dcmaw-driving-licence");
```
Joe-Edwards-GDS: Looks like the comment above has failed 😱 - we need to update the variable in the journey map visualisation as well!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/dcmaw-driving-licence.yaml
```diff
@@ -0,0 +1,112 @@
+name: DCMAW and Driving Licence
```
Joe-Edwards-GDS: I'm not sure I like this name (sorry!) - it supports a range of docs, it just so happens that DL will perform an extra check.

Perhaps something like `APP_DOC_CHECK`, and the description can add a comment that it 'may' perform authoritative source checks on certain document types.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/dcmaw-driving-licence.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+  POST_DCMAW_SUCCESS_PAGE:
+    response:
+      type: page
+      pageId: page-dcmaw-success
+    events:
+      next:
+        exitEventToEmit: next-address
+
+  MITIGATION_01_CRI_DCMAW:
```
Joe-Edwards-GDS: This duplication seems a bit suspicious - and indicates that we probably haven't got the abstraction quite right!

What's forcing us to create two copies, and can we avoid it?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/dcmaw-driving-licence.yaml
```diff
@@ -0,0 +1,112 @@
...
+      type: cri
+      criId: dcmaw
+    parent: CRI_STATE
+    events:
+      next:
+        targetState: POST_DCMAW_SUCCESS_PAGE
+      dl-auth-source-check:
+        targetState: CRI_DRIVING_LICENCE_AUTH_SOURCE_CHECK
+      not-found:
+        exitEventToEmit: next-multiple-doc
```
Joe-Edwards-GDS: I'm not sure about these event names - IMO `next` should be reserved for successful completion of the check.

Thinking about all the possibilities, I could imagine the exit states being something like:
- `next` - Completed the DCMAW doc check (ready to move onto address and fraud)
- `incomplete` (or something equivalent) - did not complete the doc check, but did not acquire any CIs (dropout, a problem with the device, temporary unavailability etc.)
    - If useful, this could be refined further to distinguish the sub-cases
- `incomplete-invalid-dl` - Dropped out at the DL auth source check (e.g. because of OCR misread)
- `alternate-doc-invalid-dl` - Got a DL doc CI after the DL auth source check

In general it should describe what happened in this nested journey, not a prediction about what may/may not come next. They might not need to go to address, if it's part of a name change or mitigation route!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
Joe-Edwards-GDS, line 1: We're not using the new subjourney for the dropout (PYI_ESCAPE) case, or the update name journeys?

If you're leaving that for future tickets, then no worries :) - it might help illustrate my point about the event names though!

Joe-Edwards-GDS (commented):
Thanks for looking at this - suggested a few tweaks to make sure the abstraction is right

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/NestedJourneyTypes.java
thebauSoftwire: and I wrote this too 🤦‍♀️ I think it may have just been me not reading the code properly

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
thebauSoftwire: I think I missed these! For the PYI_ESCAPE cases, there isn't a `POST_DCMAW_SUCCESS_PAGE` after successful completion. Is there a reason why this is? Wondering if making it consistent would simplify extracting this into a nested journey.

If not, the ways I could think of are to maybe use a `journeyContext` which controls whether we use `POST_DCMAW_SUCCESS_PAGE` but this would override a context set earlier on in the journey and we'd need to know how to reset it to the previous one (maybe find a way to support multiple contexts?); or, we could duplicate the existing nested journey states and use the entry events to control which variation to use.

Wanted to get your thoughts on the above

Joe-Edwards-GDS: Ah - good question!

I think the best approach here is probably to remove the success page from the nested journey entirely... IMO it's poorly labelled - it's as much 'pre-address-and-fraud' as it is 'post-dcmaw', and really it's an intermediate page between two stages in the journey. (Also notice that the COI journeys have wildly different versions of this page).

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -0,0 +1,69 @@
...
+        targetState: RESET_SESSION_BEFORE_PROVE_ANOTHER_WAY_AFTER_DL
+      alternate-doc-invalid-dl:
+        targetState: RESET_SESSION_BEFORE_DL_ALT_DOC_MITIGATION
+
+  RESET_SESSION_BEFORE_PROVE_ANOTHER_WAY_AFTER_DL:
+    response:
+      type: process
+      lambda: reset-session-identity
+      lambdaInput:
+        resetType: ALL
```
Joe-Edwards-GDS, line 48: (Don't worry about this now - but when we add the escape route we might need to change this so it only resets DCMAW, not everything)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -0,0 +1,69 @@
...
+    parent: CRI_STATE
+    events:
+      next:
+        targetState: POST_DCMAW_SUCCESS_PAGE
+      access-denied:
+        targetState: RESET_SESSION_BEFORE_PROVE_ANOTHER_WAY_AFTER_DL
+      alternate-doc-invalid-dl:
+        targetState: RESET_SESSION_BEFORE_DL_ALT_DOC_MITIGATION
+
+  RESET_SESSION_BEFORE_PROVE_ANOTHER_WAY_AFTER_DL:
```
Joe-Edwards-GDS: (We could align these state names with the exit event - what do you think of `RESET_SESSION_INCOMPLETE_DL` and `RESET_SESSION_INVALID_DL`?)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -0,0 +1,69 @@
...
+      type: cri
+      criId: dcmaw
+    parent: CRI_STATE
+    events:
+      next:
+        targetState: POST_DCMAW_SUCCESS_PAGE
+      dl-auth-source-check:
+        targetState: CRI_DRIVING_LICENCE_AUTH_SOURCE_CHECK
+      not-found:
+        exitEventToEmit: incomplete-not-found
```
Joe-Edwards-GDS: This smells suspicious to me - why do we treat it differently in a mitigation journey 🤔 

I've asked in slack...

thebauSoftwire: note from slack: this should be treated the same as the other `incomplete` failures in the mitigation journey ie as a soft failure

Joe-Edwards-GDS (dismissed):
Thanks, this looks like a nice simplification!

We're still missing the `CRI_DCMAW_PYI_ESCAPE` case - but happy to do that in https://govukverify.atlassian.net/browse/PYIC-7641 if you prefer.

Joe-Edwards-GDS (approved):

PR #2636, opened 2024-10-28
PYIC-7580: Handle access-denied error response

AmritSidhu:
### What changed

Rather than a VC we may receive an asynchronous error response off the queue. 
In the case of a user abandon/abort (where the user has opted to abandon the app) the error response will contain an access-denied 
At present when we process an asychronous response error, we simply update the CRI response item ('pending record') with status error regardless of the specific error code. 
We should use a new status type abandon for this scenario and update the CRI response accordingly (with error used for all other error codes).

### Issue tracking

- [PYIC-7580](https://govukverify.atlassian.net/browse/PYIC-7580)

[PYIC-7580]: https://govukverify.atlassian.net/browse/PYIC-7580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
+++ b/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
@@ -157,7 +157,11 @@ public class ProcessAsyncCriCredentialHandler
 
         criResponseItem.ifPresent(
                 responseItem -> {
-                    responseItem.setStatus(CriResponseService.STATUS_ERROR);
+                    if ("access_denied".equals(errorAsyncCriResponse.getError())) {
+                        responseItem.setStatus(CriResponseService.STATUS_ABANDON);
+                    } else {
+                        responseItem.setStatus(CriResponseService.STATUS_ERROR);
+                    }
                     criResponseService.updateCriResponseItem(responseItem);
                 });
 
--- a/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
+++ b/lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
@@ -157,7 +157,8 @@ public class ProcessAsyncCriCredentialHandler
 
         criResponseItem.ifPresent(
                 responseItem -> {
-                    if ("access_denied".equals(errorAsyncCriResponse.getError())) {
+                    if (CriResponseService.ERROR_ACCESS_DENIED.equals(
+                            errorAsyncCriResponse.getError())) {
                         responseItem.setStatus(CriResponseService.STATUS_ABANDON);
                     } else {
                         responseItem.setStatus(CriResponseService.STATUS_ERROR);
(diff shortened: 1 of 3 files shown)


lambdas/process-async-cri-credential/src/test/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandlerTest.java
```diff
@@ -176,7 +178,27 @@ void shouldProcessSuccessfullyIfEvcsFailsAndEvcsReadsIsNotEnabled() throws Excep
...
+
+        verify(criResponseService, times(1)).updateCriResponseItem(TEST_CRI_RESPONSE_ITEM);
+
+        assertEquals(0, batchResponse.getBatchItemFailures().size());
+
+        ArgumentCaptor<AuditEvent> auditEventCaptor = ArgumentCaptor.forClass(AuditEvent.class);
+        verify(auditService, times(1)).sendAuditEvent(auditEventCaptor.capture());
+        List<AuditEvent> auditEvents = auditEventCaptor.getAllValues();
+        assertEquals(1, auditEvents.size());
+        assertEquals(AuditEventTypes.IPV_F2F_CRI_VC_ERROR, auditEvents.get(0).getEventName());
```
DanCorderIPV, line 181: Maybe check the response item status in this test too?

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -157,7 +157,11 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
 
         criResponseItem.ifPresent(
                 responseItem -> {
-                    responseItem.setStatus(CriResponseService.STATUS_ERROR);
+                    if ("access_denied".equals(errorAsyncCriResponse.getError())) {
```
DanCorderIPV: `access_denied` should probably be a constant here

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
```diff
@@ -157,7 +157,11 @@ private void processErrorAsyncCriResponse(ErrorAsyncCriResponse errorAsyncCriRes
...
+                        responseItem.setStatus(CriResponseService.STATUS_ABANDON);
+                    } else {
```
DanCorderIPV, line 163: It would be good to have a way of noticing any new response error codes.
Either by logging an error here, or even throwing an exception if the code isn't one we recognise.

DanCorderIPV (changes requested):

lambdas/process-async-cri-credential/src/main/java/uk/gov/di/ipv/core/processasynccricredential/ProcessAsyncCriCredentialHandler.java
AmritSidhu: As part of this ticket it says we should use a new status type abandon specifically access_denied. 'error' to be used for all other error codes. So I dont think as part of this ticket we want to start differentiating or treating other error codes any differently yet

DanCorderIPV (approved):

PR #2641, opened 2024-10-29
PYIC-7571: Track verifier use

Wynndow:
### What changed

Track verifier use and number of VCs matched

### Why did it change

Lots of VCs were failing signature verification in staging. Turns out they were signed with an unexpected key. It would be useful to understand how many VCs are signed with this key (which has been added to config).
This would also potentially be interesting when we recondile in production to understand the proportion of passport VCs signed with the old and new keys.

Also, It would be good to be able to easily say that we've matched the same
number of VCs as expected, rather than just the number of identities.

### Issue tracking


Joe-Edwards-GDS (approved):
This is a bit weird, but sure...

I wouldn't spend too long on this though, since we're about to rip it all out :)

PR #2664, opened 2024-11-01
PYIC-7625: Update core local running with new keys

Wynndow:
### What changed

Update core local running with new keys

### Why did it change

Orch has changed how it consumes the keys, so we need to update here.

### Issue tracking

- [PYIC-7625](https://govukverify.atlassian.net/browse/PYIC-7625)

[PYIC-7625]: https://govukverify.atlassian.net/browse/PYIC-7625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/local-running/compose.yaml
+++ b/local-running/compose.yaml
@@ -12,7 +12,7 @@ services:
       context: ../../ipv-stubs/di-ipv-orchestrator-stub
       dockerfile: core-dev-deploy/Dockerfile
     environment:
-      INHERITED_IDENTITY_JWT_SIGNING_KEY: MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgOXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthWhRANCAAQT1nO46ipxVTilUH2umZPN7OPI49GU6Y8YkcqLxFKUgypUzGbYR2VJGM+QJXk0PI339EyYkt6tjgfS+RcOMQNO # pragma: allowlist secret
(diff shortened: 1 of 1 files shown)


local-running/compose.yaml
```diff
@@ -12,21 +12,22 @@ services:
...
       ORCHESTRATOR_PORT: 4500
       ORCHESTRATOR_REDIRECT_URL: http://localhost:4500/callback
       EVCS_ACCESS_TOKEN_ENDPOINT: https://mock.credential-store.build.account.gov.uk/generate
       EVCS_ACCESS_TOKEN_TTL: 60
-      EVCS_ACCESS_TOKEN_SIGNING_KEY_JWK: '{"kty":"EC","d":"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU","crv":"P-256","x":"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM","y":"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04"}' # pragma: allowlist secret
+      EVCS_ACCESS_TOKEN_SIGNING_JWK: '{"kty":"EC","kid":"evcs-token-local-running-FI4xysvMVdRtkt6xmO5gqcaTF4Tf9NKD1zdg3T8y69M","use":"sig","d":"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU","crv":"P-256","x":"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM","y":"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04"}' # pragma: allowlist secret
```
Joe-Edwards-GDS, line 30: I don't think this is actually specific to local-running - perhaps it should be, but at the moment I think EVCS stub just accepts a single signing key?

Joe-Edwards-GDS (approved):

local-running/compose.yaml
Wynndow: This is another place that we use the same test key - so the `evcs-token-local-running` part is just to differentiate it. But yeah, the stub will be set up to just use that key

PR #2666, opened 2024-11-01
PYIC-7632: Include kid in private key jwt for token exchange

Wynndow:
### What changed

Include kid in private key jwt for token exchange

### Why did it change

When we exchange an auth code for a token with a CRI we use a signed JWT to prove the request is coming from a legitimate source. This change updates the header of the JWT to include the kid of the signing key.

We already had logic to include the key when signing - it was being used when generating a JAR. It was implemented for DWP and is behind a flag to allow it to be tested in staging before hitting prod.

This moves the logic to a shared place and utilises it for both signing operations.

### Issue tracking

- [PYIC-7632](https://govukverify.atlassian.net/browse/PYIC-7632)

[PYIC-7632]: https://govukverify.atlassian.net/browse/PYIC-7632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
```diff
@@ -8,38 +8,42 @@
...
+public class JwtHelper {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final Logger LOGGER = LogManager.getLogger();
 
     private JwtHelper() {}
 
-    public static <T> SignedJWT createSignedJwtFromObject(T claimInput, JWSSigner signer)
-            throws JOSEException {
-        JWSHeader jwsHeader = generateHeader();
+    public static <T> SignedJWT createSignedJwtFromObject(
```
Joe-Edwards-GDS: I wonder if we should change this signature to take a `JWTClaimsSet` instead, and get rid of `generateClaims`...

Looking at the other usage (for the private key JWT), we'd be better served getting rid of our `ClientAuthClaims` object and using the built-in nimbus classes. E.g.:
```
            var clientAuthClaims = new JWTClaimsSet.Builder()
                    .issuer(criConfig.getClientId())
                    .subject(criConfig.getClientId())
                    .audience(criConfig.getComponentId())
                    .expirationTime(...)
                    .jwtID(secureTokenHelper.generate())
                    .build();
```
IMO this is clearer and safer than using our custom class and serialising ourselves! And it means you can use a shared method for the signing.

Appreciate it's nothing really to do with your change, so happy to put into a different PR if you prefer.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
```diff
@@ -8,38 +8,42 @@
...
+            T claimInput, JWSSigner signer, ConfigService configService) throws JOSEException {
```
Joe-Edwards-GDS: I don't _love_ that this takes a `configService` - it feels like a bad abstraction for a static helper class, but can live with it if there's no neater option. (I might prefer to have `includeKeyId` as a parameter, perhaps)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
```diff
@@ -58,4 +62,26 @@ public static boolean signatureIsDerFormat(SignedJWT signedJWT) throws JOSEExcep
         return signedJWT.getSignature().decode().length
                 != ECDSA.getSignatureByteArrayLength(JWSAlgorithm.ES256);
     }
+
+    private static String getKid(ConfigService configService) {
```
Joe-Edwards-GDS: It's annoying that you can't get this from the `JWSSigner` - it feels like an omission :(

Means that this must be kept in sync with the logic in `SignerFactory`, to make sure the key id actually matches the key being used in the signer!

Probably not the end of the world, but a potential improvement might be to either wrap in our own class (that does expose the key id) or combine this helper class with the signer factory, to have a single 'signing service' that handles the key selection in one place.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
```diff
@@ -8,38 +8,42 @@
...
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.ipv.core.library.service.ConfigService;
 
+import java.text.ParseException;
 import java.util.Map;
 
-public class JwtHelper {
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.SIGNING_KEY_ID;
+import static uk.gov.di.ipv.core.library.config.ConfigurationVariable.SIGNING_KEY_JWK;
+import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.KID_JAR_HEADER;
```
Joe-Edwards-GDS: I hadn't realised this was already behind a feature flag! Convenient - means we can test both headers at the same time :D

Joe-Edwards-GDS (commented):
Looks good but left a few comments about possible improvements :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
Wynndow: Yep, that makes sense. I'll fix it up here as I'm in that headspace.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
Wynndow: Yeah, we could pass the keyId in, and then include it if it's there.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
```diff
@@ -1,48 +1,30 @@
...
+    public static SignedJWT createSignedJwt(
+            JWTClaimsSet claimsSet, CoreSigner signer, boolean includeKid) throws JOSEException {
+        JWSHeader jwsHeader = generateHeader(includeKid ? signer.getKid() : null);
         SignedJWT signedJWT = new SignedJWT(jwsHeader, claimsSet);
         signedJWT.sign(signer);
         return signedJWT;
     }
 
-    public static JWSHeader generateHeader(ConfigService configService) {
+    public static JWSHeader generateHeader(String kid) {
```
Joe-Edwards-GDS: (could now be private)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/fixtures/TestFixtures.java
```diff
@@ -18,12 +18,12 @@ public interface TestFixtures {
     String EC_PRIVATE_KEY =
             "MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgOXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthWhRANCAAQT1nO46ipxVTilUH2umZPN7OPI49GU6Y8YkcqLxFKUgypUzGbYR2VJGM+QJXk0PI339EyYkt6tjgfS+RcOMQNO"; // pragma: allowlist secret
     String EC_PRIVATE_KEY_JWK =
-            "{\"kty\":\"EC\",\"d\":\"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"; // pragma: allowlist secret
+            "{\"kty\":\"EC\",\"kid\":\"test-fixtures-ec-key\",\"use\":\"sig\",\"d\":\"OXt0P05ZsQcK7eYusgIPsqZdaBCIJiW4imwUtnaAthU\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"; // pragma: allowlist secret
```
Joe-Edwards-GDS, line 21: (FYI @thebauSoftwire this will probably clash)

Joe-Edwards-GDS (dismissed):
Thanks - I think that's improved things!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/JwtHelper.java
Wynndow: yep, will update

Joe-Edwards-GDS (dismissed):

Joe-Edwards-GDS (approved):

PR #2671, opened 2024-11-04
PYIC-7630: consume cri oauth enc key from well-known endpoint

thebauSoftwire:
### What changed
- if `jwksUrl` is present in OAuth cri config, get enc key from this endpoint
- if get request fails, default to enc key specified in config
- enc key cached for configured amount of time

### Why did it change
When the enc keys are rotated for the CRIs, we can just call the well-known endpoint and get the first valid key to avoid having to update the config each time

### Issue tracking

- [PYIC-7630](https://govukverify.atlassian.net/browse/PYIC-7630)

[PYIC-7630]: https://govukverify.atlassian.net/browse/PYIC-7630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -56,6 +56,10 @@ Globals:
           - IsDevelopment
           - 0
           - 3
+        OAUTH_CRI_ENC_KEY_CACHE_DURATION_MINUTES: !If
```
Joe-Edwards-GDS: This could probably just be 'standard' config - i.e. using an SSM parameter, unless there's a good reason to make it static?

I'd also suggest just having one value for `OAUTH_KEY_CACHE_DURATION`, I don't think we're likely to need different values for the oauth key types.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +351,45 @@ private JWEObject signEncryptJar(
...
 
-        RSAEncrypter rsaEncrypter = new RSAEncrypter(oauthCriConfig.getParsedEncryptionKey());
-        return AuthorizationRequestHelper.createJweObject(rsaEncrypter, signedJWT);
+        var encKey = getCriEncryptionKey(oauthCriConfig);
+        RSAEncrypter rsaEncrypter = new RSAEncrypter(encKey);
+        return AuthorizationRequestHelper.createJweObject(
+                rsaEncrypter, signedJWT, encKey.getKeyID());
+    }
+
+    private RSAKey getCriEncryptionKey(OauthCriConfig criConfig) throws ParseException {
```
Joe-Edwards-GDS: I think it's worth pulling this out into a separate `OauthKeyService` or similar - this class is already quite big (and does several things!)

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -96,6 +102,10 @@ public class BuildCriOauthRequestHandler
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
     private final SessionCredentialsService sessionCredentialsService;
+    private final CriApiService criApiService;
+
+    private static final Integer DEFAULT_OAUTH_CRI_ENC_KEY_CACHE_DURATION_MINUTES = 5;
+    private static CachedOAuthCriEncryptionKey cachedOAuthCriEncryptionKey;
```
Joe-Edwards-GDS: Careful - although in dev/build all our CRIs use the same keys, in production they won't!

We'll probably need to store a cache indexed by CRI + connection id

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +351,45 @@ private JWEObject signEncryptJar(
...
+        var encKey = getCriEncryptionKey(oauthCriConfig);
+        RSAEncrypter rsaEncrypter = new RSAEncrypter(encKey);
+        return AuthorizationRequestHelper.createJweObject(
+                rsaEncrypter, signedJWT, encKey.getKeyID());
+    }
+
+    private RSAKey getCriEncryptionKey(OauthCriConfig criConfig) throws ParseException {
+        var jwksUrl = criConfig.getJwksUrl();
+
+        if (jwksUrl != null) {
```
Joe-Edwards-GDS: (I assume this being null basically 'just works')

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CachedOAuthCriEncryptionKey.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.di.ipv.core.buildcrioauthrequest.domain;
+
+import com.nimbusds.jose.jwk.RSAKey;
+
+import java.util.Date;
+
+public record CachedOAuthCriEncryptionKey(RSAKey key, Date created) {
+    public Boolean isExpired(Integer cacheDurationMinutes) {
```
Joe-Edwards-GDS: Should be `boolean` if it doesn't need to be nullable

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/domain/CachedOAuthCriEncryptionKey.java
```diff
@@ -0,0 +1,12 @@
...
+        var expiryDate = new Date(created.getTime() + cacheDurationMinutes * 60 * 1000);
+        return expiryDate.before(new Date());
```
Joe-Edwards-GDS: You could simplify this by writing the `expiryTime` into the record, rather than the `createdTime` - it's a bit weird that this method takes an argument.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +351,45 @@ private JWEObject signEncryptJar(
...
+                var parsedKey = key.get().toRSAKey();
+                cacheEncryptionKey(parsedKey);
+                return parsedKey;
+            }
+        }
+        return criConfig.getParsedEncryptionKey();
+    }
+
+    private static void cacheEncryptionKey(RSAKey key) {
+        cachedOAuthCriEncryptionKey = new CachedOAuthCriEncryptionKey(key, new Date());
```
Joe-Edwards-GDS: (this could just be inlined)

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -68,6 +78,7 @@ public class CriApiService {
     private final SignerFactory signerFactory;
     private final SecureTokenHelper secureTokenHelper;
     private final Clock clock;
+    private final HttpClient httpClient;
```
Joe-Edwards-GDS: There'll be a conflict here with https://github.com/govuk-one-login/ipv-core-back/pull/2672, but should be easy enough to resolve - we can share the HTTP client between the two.

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -358,4 +384,37 @@ private VerifiableCredentialResponse getVerifiableCredentialResponseForApplicati
         }
         return vcResponseBuilder.build();
     }
+
+    @Tracing
+    public Optional<JWK> getKeyFromJwksEndpoint(URI jwksEndpoint) {
```
Joe-Edwards-GDS: One thing to consider - if you pull the other logic out into an `OAuthKeyService`, should this move there as well?

In the next ticket we're going to call the JWKS endpoint from orchestration, rather than a CRI - which will have an almost identical implementation.

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -358,4 +384,37 @@ private VerifiableCredentialResponse getVerifiableCredentialResponseForApplicati
...
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "%d: Error retrieving public encryption key.",
+                                        httpResponse.statusCode())));
+                return Optional.empty();
+            }
+
+            var keys = JWKSet.parse(httpResponse.body()).getKeys();
+
+            return keys.stream().filter(key -> key.getKeyUse() == KeyUse.ENCRYPTION).findFirst();
```
Joe-Edwards-GDS: At the moment the method says 'getKeyFromJwksEndpoint' but this is specifically returning the first encryption key (not an arbitrary key).

I'd suggest making this method return the entire JWKS object, and then consumers can decide which key they want. That's probably also what we want to store in our cache too.

Joe-Edwards-GDS (commented):
Thanks, I think this is on the right lines but I think we can refactor a bit to make the next ticket much easier!

deploy/template.yaml
thebauSoftwire: PR for this is here: https://github.com/govuk-one-login/ipv-core-common-infra/pull/1109/files

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
thebauSoftwire: oops good point!

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
thebauSoftwire: it was initially inlined but sonar complained that you shouldn't be mutating a static variable inside a non-static method so I've extracted it here to shush it

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
thebauSoftwire: I've moved the method which uses this into its own service now so hopefully shouldn't be any conflicts here

thebauSoftwire: oh I just had a proper look at the PR, is it for all requests to the CRIs? I'll update this if the other one gets merged first

Joe-Edwards-GDS: I think this is probably irrelevant now you've pulled it into a separate service - I don't mind that, the JWKS endpoint isn't so much of a 'CRI API' as the token/credential endpoints.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,86 @@
...
+    public OAuthKeyService() {
+        this.httpClient = HttpClient.newHttpClient();
+    }
+
+    @ExcludeFromGeneratedCoverageReport
+    public OAuthKeyService(HttpClient httpClient) {
+        this.httpClient = httpClient;
+    }
+
+    public RSAKey getValidEncryptionKey(OauthCriConfig criConfig) throws ParseException {
```
Joe-Edwards-GDS: What does `valid` mean here?

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Tracing
+    private List<JWK> getKeysFromJwksEndpoint(URI jwksEndpoint) {
+        try {
+            LOGGER.info(LogHelper.buildLogMessage("Retrieving key from well-known endpoint."));
+            var request = HttpRequest.newBuilder().uri(jwksEndpoint).GET().build();
+            var httpResponse = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+
+            if (httpResponse.statusCode() != SC_OK) {
+                LOGGER.warn(
```
Joe-Edwards-GDS: I know we'll recover for now, but I think these should probably just be `error`

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,86 @@
...
+                        LogHelper.buildLogMessage(
+                                String.format(
+                                        "%d: Error retrieving public encryption key.",
+                                        httpResponse.statusCode())));
```
Joe-Edwards-GDS: Can we attach the statusCode as JSON log property, i.e.:
```
LogHelper.buildLogMessage("Error retrieving JWKS")
  .with(LOG_STATUS_CODE, httpResponse.statusCode())
```
That way it's easier to aggregate/query on

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -96,6 +101,10 @@ public class BuildCriOauthRequestHandler
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
     private final SessionCredentialsService sessionCredentialsService;
+    private final OAuthKeyService oAuthKeyService;
+
+    private static final Integer DEFAULT_OAUTH_CRI_ENC_KEY_CACHE_DURATION_MINUTES = 5;
+    private static Map<String, CachedOAuthCriEncryptionKey> cachedOAuthCriEncryptionKeys;
```
Joe-Edwards-GDS: I think we should move the caching logic into the `OauthKeyService` as well - this class shouldn't really have to worry about cache implementation details.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +345,41 @@ private JWEObject signEncryptJar(
...
+
+        RSAEncrypter rsaEncrypter = new RSAEncrypter(encKey);
+        return AuthorizationRequestHelper.createJweObject(
+                rsaEncrypter, signedJWT, encKey.getKeyID());
+    }
+
+    private static void cacheEncryptionKey(RSAKey key, Cri cri, Integer cacheDuration) {
+        // Initialise cachedOAuthCriEncryptionKeys if still null
+        if (cachedOAuthCriEncryptionKeys == null) {
+            cachedOAuthCriEncryptionKeys = new HashMap<>();
```
Joe-Edwards-GDS: It's worth noting that this is _not_ thread-safe - if you have multiple threads then the map might be corrupted.

That should be fine in production because each lambda is effectively single-threaded, but could cause issues with the API tests... It might be worth changing this to be a `ConcurrentHashMap` instead.

You could also initialise this variable directly, unless there's a particular reason to do it lazily? It would simplify a bunch of your logic and make it `final`.

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +345,41 @@ private JWEObject signEncryptJar(
...
+        }
+        // Cache the key in a variable outside of the handler so it exists in
+        // the Lambda's memory and is still accessible between invocations that
+        // occur at short intervals from one another
+        var expiryDate = LocalDateTime.now().plusMinutes(cacheDuration);
+        cachedOAuthCriEncryptionKeys.put(
+                cri.getId(), new CachedOAuthCriEncryptionKey(key, expiryDate));
```
Joe-Edwards-GDS: I think you need to key this on CRI _and_ connection ID (the `connection` variable in the handle method) - e.g. in staging we might be using either the stub or real CRI, depending on a feature flag (and these will have different endpoints).

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -333,8 +345,41 @@ private JWEObject signEncryptJar(
...
+                && !cachedOAuthCriEncryptionKeys.isEmpty()
+                && cachedOAuthCriEncryptionKeys.get(cri.getId()) != null
+                && !cachedOAuthCriEncryptionKeys.get(cri.getId()).isExpired()) {
+            encKey = cachedOAuthCriEncryptionKeys.get(cri.getId()).key();
+        } else {
+            encKey = oAuthKeyService.getValidEncryptionKey(oauthCriConfig);
+
+            var cacheDuration =
+                    configService.getParameter(ConfigurationVariable.OAUTH_KEY_CACHE_DURATION_MINS)
+                                    == null
```
Joe-Edwards-GDS: Can/should this ever be null? I think the config service will throw if the config is not found...

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
Joe-Edwards-GDS: Alternatively, if you move the logic into `OAuthKeyService`, you could cache on the JWKS URL instead

Joe-Edwards-GDS (commented):
Looking better, but I'm quite keen to keep the caching logic out of the request handlers :)

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
thebauSoftwire: hmm I think I was trying to refer to the fact we have a bit of filtering ie the first key that has `use:enc` but I can see how this doesn't exactly imply `valid` either. How about just `getEncryptionKey?`

Joe-Edwards-GDS: Sounds good to me

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/AuthorizationRequestHelper.java
```diff
@@ -115,14 +115,16 @@ ResponseType.CODE, new ClientID(oauthCriConfig.getClientId()))
...
+    public static JWEObject createJweObject(
+            RSAEncrypter rsaEncrypter, SignedJWT signedJWT, String keyId)
             throws HttpResponseExceptionWithErrorBody {
         try {
             JWEObject jweObject =
                     new JWEObject(
                             new JWEHeader.Builder(
                                             JWEAlgorithm.RSA_OAEP_256, EncryptionMethod.A256GCM)
                                     .contentType("JWT")
+                                    .keyID(keyId)
```
Joe-Edwards-GDS: It shouldn't matter, but can we put this behind the kid header feature flag - `configService.enabled(KID_JAR_HEADER)`, just in case it breaks something.

Compare to Chris's change in `JwtHelper`: https://github.com/govuk-one-login/ipv-core-back/pull/2666/files#diff-ae88c342eed7926d0ba81bb36d87eda6b92d047f611cfc5462dd8e2502264c32

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,128 @@
...
+        this.httpClient = httpClient;
+    }
+
+    public RSAKey getEncryptionKey(OauthCriConfig criConfig) throws ParseException {
+        var jwksUrl = criConfig.getJwksUrl();
+        var keyFromConfig = criConfig.getParsedEncryptionKey();
+
+        if (jwksUrl != null) {
+            if (!cachedOAuthEncryptionKeys.isEmpty()
+                    && cachedOAuthEncryptionKeys.get(jwksUrl.toString()) != null
```
Joe-Edwards-GDS: (is this possible?)

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+import static com.nimbusds.oauth2.sdk.http.HTTPResponse.SC_OK;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_STATUS_CODE;
+
+public class OAuthKeyService {
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final ConcurrentHashMap<String, CachedOAuthCriEncryptionKey>
+            cachedOAuthEncryptionKeys = new ConcurrentHashMap<>();
```
Joe-Edwards-GDS: If it helps, this might not actually need to be static - I think AWS should construct a single instance of the handler class (and therefore this class) per lambda instance.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
Joe-Edwards-GDS, line 1: This looks good, but once you pick up https://govukverify.atlassian.net/browse/PYIC-7631 you might find that it's actually better to cache the `JWKSet` rather than specific keys. Happy for you to work out the details there though.

libs/oauth-key-service/src/test/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyServiceTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Captor private ArgumentCaptor<HttpRequest> httpRequestCaptor;
+    @Mock private HttpClient mockHttpClient;
+    @Mock private HttpResponse<String> mockHttpResponse;
+    @Mock private ConfigService mockConfigService;
+    @InjectMocks private OAuthKeyService oAuthKeyService;
+
+    @BeforeEach
+    void setUp() throws Exception {
+        oAuthKeyService = new OAuthKeyService(mockConfigService, mockHttpClient);
```
Joe-Edwards-GDS: Do you need to reconstruct this each time, if it's got the `@InjectMocks` it should be constructed by Mockito...

libs/oauth-key-service/src/test/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyServiceTest.java
```diff
@@ -0,0 +1,210 @@
...
+        oauthCriConfig =
+                OauthCriConfig.builder()
+                        .jwksUrl(new URI(TEST_JWKS_ENDPOINT))
+                        .encryptionKey(TEST_KEY)
+                        .build();
```
Joe-Edwards-GDS, line 55: (Technically this is probably just `BeforeAll`, or even a static initialiser)

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -0,0 +1,128 @@
...
+                    LogHelper.buildLogMessage(
+                            "Interrupted while trying to fetch public key for encryption."));
+            return List.of();
+        }
+    }
+
+    // This method is only used during testing to reset the cachedOAuthCriEncryptionKeys
+    // between tests
+    public void clearCache() {
+        cachedOAuthEncryptionKeys.clear();
```
Joe-Edwards-GDS: If the cached keys are static, this method probably should be too

libs/oauth-key-service/src/test/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyServiceTest.java
```diff
@@ -0,0 +1,210 @@
...
+        var oauthConfigNoJwksUrl = OauthCriConfig.builder().encryptionKey(TEST_KEY).build();
+        var key = oAuthKeyService.getEncryptionKey(oauthConfigNoJwksUrl);
+
+        assertEquals(RSAKey.parse(TEST_KEY), key);
+    }
+
+    @Test
+    void getEncryptionKeyShouldReturnCachedKeyIfItExistsAndIsNotExpired() throws Exception {
+        // Set up
+        // Create a valid key already in the cache
```
Joe-Edwards-GDS: Instead of doing this, I wonder if you could just call it twice and assert that the mockHttpClient is only called once? That's the public behaviour that we're trying to test - and the implementation doesn't really matter.

For the test below, you could do the same but set the config cache duration to zero...

Joe-Edwards-GDS (commented):
I think this looks much cleaner thanks :)

One slightly paranoid comment about the feature flag, and a handful of other optional suggestions.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
thebauSoftwire: ohh interesting! Thanks didn't know this

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
thebauSoftwire: removed the static as suggested :)

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
thebauSoftwire: This bit checks if the key exists in the map but it isn't as readable - have changed to check if key exists instead which has simplified it, thanks :)

Joe-Edwards-GDS: Ah of course 🤦 - yeah using `containsKey` is definitely the better option!

Joe-Edwards-GDS (approved):
Thanks for all the updates :)

PR #2672, opened 2024-11-04
BAU: Traced CRI HTTP requests

Joe-Edwards-GDS:
### What changed

Use a custom `HTTPRequestSender` for CRI requests, which can use our existing `TracingHttpClient` based on `java.net.HttpClient`.
- Token exchange endpoint
- Credential endpoint

Also updated the tests and removed wiremock as a dependency.

### Why did it change

The existing implementation uses an internal Nimbus implementation, based on `java.net.HttpURLConnection`.

This makes it difficult to instrument for X-Ray tracing, and means we don't get other benefits from using a proper HTTP client, such as connection reuse.

The additional instrumentation means we can monitor response times per CRI and endpoint, if we wish.

### DIFF
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingServiceTest.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingServiceTest.java
@@ -2,7 +2,6 @@ package uk.gov.di.ipv.core.processcricallback.service;
 
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.http.HTTPResponse;
-import org.apache.commons.lang3.StringUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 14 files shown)


lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingServiceTest.java
```diff
@@ -2,7 +2,6 @@
 
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.http.HTTPResponse;
-import org.apache.commons.lang3.StringUtils;
```
Joe-Edwards-GDS, line 5: (This was a transitive dependency of wiremock)

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/JavaHttpRequestSender.java
Joe-Edwards-GDS, line 1: This is the the main addition - an implementation of `com.nimbusds.oauth2.sdk.http.HTTPRequestSender` that uses an underlying `java.net.HttpClient`

thebauSoftwire (approved):
Looks good!

PR #2686, opened 2024-11-07
PYIC-7461 Fix up tests and routing for COI FF go-live

Joe-Edwards-GDS:
### What changed

- Switch the updateDetailsAccountDeletion flag on for local-running
- Update the tests to handle this
- Update a couple of routes to combine with the flag appropriately

### Why did it change

We're turning it on in prod soon


api-tests/features/mfa-reset-journey.feature
```diff
@@ -68,8 +68,8 @@ Feature: MFA reset journey
     When I submit 'alice-passport-valid' details to the CRI stub
     Then I get a 'page-dcmaw-success' page response
     When I submit a 'next' event
-    Then I get a 'sorry-could-not-confirm-details' page response
-    When I submit a 'end' event
+    Then I get a 'pyi-no-match' page response
+    When I submit a 'next' event
```
Joe-Edwards-GDS, line 72: This will change anyway, so not thinking too hard about this one!

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature
Joe-Edwards-GDS, line 1: In general this is: removing the 'default' version of each test, and making the `@FastFollow` version the default.

We probably want to think about how to make this easy to do when introducing new features in future 🤔

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
```diff
@@ -209,9 +209,9 @@ states:
         targetState: FAILED_UPDATE_DETAILS
       alternate-doc-invalid-dl:
         targetJourney: FAILED
-        targetState: FAILED_UPDATE_DETAILS
+        targetState: FAILED_CONFIRM_DETAILS_INVALID_ID
```
Joe-Edwards-GDS, line 212: @Wynndow this is the crossover with DL auth source checks - we missed it because without the feature flag, `FAILED_UPDATE_DETAILS`/`FAILED_CONFIRM_DETAILS` both do the same thing!

Wynndow (approved):

PR #2688, opened 2024-11-08
PYIC-7631: Verify client signatures with key from JWKS endpoint

Wynndow:
### What changed

Verify client signatures with key from JWKS endpoint

### Why did it change

[PYIC-7631: Extend OAuthKeyService for client signing keys](https://github.com/govuk-one-login/ipv-core-back/commit/651133170d3ed19294bd5526cadf9eee079ec8f9)

This adds a new method to the OAuthKeyService to allow us to fetch
client signing keys by key id. It also refactors the service to cache
the entire jwks set rather than the individual keys.

[PYIC-7631: Consume signing key from key-service](https://github.com/govuk-one-login/ipv-core-back/commit/6dd13790341d40701ae8057c876bf8ababaecce6)

This updates the initialise-ipv-session lambda and the
issue-client-access-token lambda to consume a key from the key-service
when verifying signatures.

### Issue tracking

- [PYIC-7631](https://govukverify.atlassian.net/browse/PYIC-7631)

[PYIC-7631]: https://govukverify.atlassian.net/browse/PYIC-7631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
```diff
@@ -49,32 +39,19 @@ public List<? extends PublicKey> selectPublicKeys(
...
+        var algorithm = jwsHeader.getAlgorithm();
+        if (algorithm != ES256) {
+            throw new InvalidClientException(
+                    String.format(
+                            "%s algorithm is not supported. Received from client ID '%s'",
+                            algorithm, clientId));
+        }
 
         try {
-            switch (algorithm.toString()) {
```
Wynndow, line 58: This strips out support for signatures with RS256. Our clients don't use it, and it would add more complexity to the OAuth key selector. 
If we want to start supporting RS256 we could definitely add it in, but it's not needed right now.

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
```diff
@@ -30,25 +31,29 @@ public class TokenRequestValidator {
...
         this.configService = configService;
         this.clientAuthJwtIdService = clientAuthJwtIdService;
+        this.oAuthKeyService = oAuthKeyService;
     }
 
     public void authenticateClient(String requestBody) throws ClientAuthenticationException {
-        if (verifier == null) {
-            this.verifier = getClientAuthVerifier(configService);
+        if (clientAuthVerifier == null) {
+            this.clientAuthVerifier = getClientAuthVerifier(configService);
```
Wynndow: This is a bit weird, but I think it's like this so that the audience config value doesn't get baked in to the snapstart image.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -51,64 +65,114 @@ public RSAKey getEncryptionKey(OauthCriConfig criConfig) throws ParseException {
...
             var httpResponse = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
 
             if (httpResponse.statusCode() != SC_OK) {
                 LOGGER.error(
                         LogHelper.buildLogMessage("Error retrieving JWKS")
-                                .with(LOG_STATUS_CODE.getFieldName(), httpResponse.statusCode()));
-                return List.of();
+                                .with(LOG_STATUS_CODE.getFieldName(), httpResponse.statusCode())
+                                .with(LOG_JWKS_URL.getFieldName(), jwksEndpoint));
+                return new JWKSet();
```
Joe-Edwards-GDS, line 167: It's worth thinking briefly about this - this will mean that a transient error will cache an empty (invalid) value, and it'll need to expire before re-fetching.

On the other hand, if we _don't_ cache errors, then a persistent error will cause the endpoint to get hit every time until it comes back up.

The former is possibly better while we still have the fallback, but I wonder if the latter is the better long-term? An ideal solution might include a circuit-breaker so we don't hammer the endpoint if it's persistently down, but definitely not worth implementing now.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -51,64 +65,114 @@ public RSAKey getEncryptionKey(OauthCriConfig criConfig) throws ParseException {
...
+        try {
+            return URI.create(configService.getParameter(CLIENT_JWKS_URL, clientId));
+        } catch (ConfigParameterNotFoundException e) {
+            return null;
+        }
+    }
+
+    private JWKSet getCachedJWKSet(URI jwksUrl) {
+        if (cachedJwkSets.computeIfAbsent(jwksUrl, this::createCachedJWKSet).isExpired()) {
+            cachedJwkSets.put(jwksUrl, createCachedJWKSet(jwksUrl));
```
Joe-Edwards-GDS: There is potentially a race condition here - although it won't matter very much since our lambdas are single-threaded.

I think the 'correct' solution is probably to use `compute` and do everything atomically:
```
        return cachedJwkSets.compute(jwksUrl, (key, value) -> {
            if (value == null || value.isExpired()) {
                return createCachedJWKSet(key);
            }
            return value;
        }).jwkSet();
```

This will block other update operations (i.e. other concurrent requests trying to refresh the cache), but that's probably the behaviour you want.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -51,64 +65,114 @@ public RSAKey getEncryptionKey(OauthCriConfig criConfig) throws ParseException {
...
+                                                    .with(
+                                                            LOG_CRI_ISSUER.getFieldName(),
+                                                            criConfig.getComponentId())
+                                                    .with(LOG_JWKS_URL.getFieldName(), jwksUrl));
+                                    return keyFromConfig;
+                                });
+    }
 
-            var keys = getKeysFromJwksEndpoint(jwksUrl);
+    public ECKey getClientSigningKey(String clientId, JWSHeader jwsHeader) throws ParseException {
```
Joe-Edwards-GDS, line 86: (Very minor - do you think it's more appropriate to pass the keyId here instead of the entire header?)

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
Joe-Edwards-GDS: Makes some sense, but is still a bit dubious - it'll still keep the audience value for the duration of the lambda instance, which may be several hours.

I wonder if we should just construct a new one each time - I don't imagine it's excessively heavyweight.

Not really related to your change, so happy for you to worry about separately if you prefer.

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
Joe-Edwards-GDS: Are we 100% sure about this? I know there was a bit of a kerfuffle with DWP KBV key types, but this is for clients so must be unrelated.

If you're sure then happy to remove - it's pretty suspicious that we (used to) load it from an X.509 rather than a JWK.

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
```diff
@@ -49,32 +39,19 @@ public List<? extends PublicKey> selectPublicKeys(
             Context context)
             throws InvalidClientException {
```
Joe-Edwards-GDS: Perhaps worth observing that we're not respecting the `forceRefresh` parameter here, but looking at the usages of this, we'll never actually see it :)

To get it, we should pass the `CLIENT_HAS_REMOTE_JWK_SET` hint to the verify method - which will call this again with forceRefresh = true if the first attempt fails.

I'm not sure we really need it right now though - especially as we'd have to do something else to make it work for the JAR validation and it's a bit pointless if they're not in sync.

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
Joe-Edwards-GDS, line 1: The name of this class is a bit less meaningful now - it will return keys from a JWKS endpoint instead if available!

Joe-Edwards-GDS (commented):
A few slightly fiddly comments, but on the whole this looks good, thanks!

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
Wynndow: Yeah, I think back in the day, before we were live, we were using RS256 from our stub. But then the real auth/orch (before the split) were going to use ES256 so we added support from it. Now we just use ES256.
I think we were using certs because that's what we were used to handling from Verify days. But try getting a signed X509 cert using KMS...

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
Wynndow: Yeah, everything involved looks pretty lightweight. And it'd make things more sensible to test. I'll update it.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
Wynndow: Hmm, that's a good point. I suppose that once we've removed the config key, we'd probably want to just maintain the value in the cache in the case of an HTTP failure. If  this happend to be on a cold start we'll definitely see a bunch of failures, but warm lambdas might be able to chug on for a bit.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
Wynndow: Ah yeah - the expired check is happening outside the atomic operation. Will update.

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
Wynndow: I think not - we're extracting the `kid` to find the right key, but we could go further. You can create a matching object directly from a header, which could use things like the algorithm and the use to refine the JWKSet. I think it'd be better to leave the service to decide what info it wants to use to find the key - will avoid having to update anything that consumes this method if we change the logic in future.

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/domain/ConfigurationServicePublicKeySelector.java
Wynndow: Very true - will update

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
Joe-Edwards-GDS: Yeah - if there's a failure and we still have a cached value, then continuing to use the cached value is probably better than failing outright.

Joe-Edwards-GDS (approved):
I think there might still be room to optimise what happens on failure, but that could wait until we remove support for the fallback.

If not doing it now, perhaps worth leaving a note on https://govukverify.atlassian.net/browse/PYIC-7697 ?

PR #2689, opened 2024-11-08
BAU: Retry on HTTP/2 GOAWAY

Joe-Edwards-GDS:
### What changed

Add a single retry to our base HTTP Client if it receives an HTTP/2 GOAWAY signal.

### Why did it change

This is similar to an HTTP/1.1 connection close - the server no longer wishes to process requests on that connection. This seems to happen after the connection has been open for ~1 hour with the fraud and address CRIs.

The simplest solution is to just retry those requests.

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -4,7 +4,10 @@ import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.AWSXRayRecorder;
 import com.amazonaws.xray.entities.Namespace;
 import com.amazonaws.xray.entities.Subsegment;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+import uk.gov.di.ipv.core.library.helpers.LogHelper;
 
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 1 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -100,6 +103,25 @@ public <T> HttpResponse<T> send(
...
+    @Override
+    public <T> HttpResponse<T> send(
+            HttpRequest request, HttpResponse.BodyHandler<T> responseBodyHandler)
+            throws IOException, InterruptedException {
+        try {
+            return sendWithTracing(request, responseBodyHandler);
+        } catch (IOException e) {
+            // We see occasional HTTP/2 GOAWAY messages from AWS when connections last
+            // for a long time (>1 hour). Retry them once, to recreate the connection.
+            if (e.getMessage().contains("GOAWAY received")) {
```
Joe-Edwards-GDS, line 115: This is a bit unsatisfactory, but I couldn't see a cleaner solution... The messages look like `/169.254.76.1:51122: GOAWAY received` (where the port varies per request)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Joe-Edwards-GDS, line 1: It's not _really_ the responsibility of this file to do the retry, but it's a convenient place to do it. We want to apply across multiple requests, and creating another wrapper class feels a bit heavyweight.

Wynndow (approved):

PR #2690, opened 2024-11-08
PYIC-7641: journey routing when user indicates DL details are incorrect and drops out of DL CRI during auth source check

thebauSoftwire:
### What changed
- routing the user to new screens for DL auth source check
- clearing DCMAW VC after dropping out of DL CRI during DL auth source check
- update unit tests + api tests

### Why did it change
If a user indicates that their DL details are incorrect (e.g. bc of bad scan) then they should be routed to new DL auth source check pages which gives them options to try proving their identity in other ways or return to the RP.
The DCMAW VC gained during their session is also deleted to avoid issues with data integrity further down the journey/

### Issue tracking

- [PYIC-7641](https://govukverify.atlassian.net/browse/PYIC-7641)

[PYIC-7641]: https://govukverify.atlassian.net/browse/PYIC-7641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/dl-auth-source-checks/dl-cri-dropout.feature
+++ b/api-tests/features/dl-auth-source-checks/dl-cri-dropout.feature
@@ -135,12 +135,12 @@ Feature: Dropping out of authoritative source checks with DL CRI (e.g. due to in
       When I submit a 'update-details' event
       Then I get a 'update-details' page response
 
-    Scenario Outline: Change of name only journey - User backs out of DL CRI - Returns to DCMAW to use passport
-      When I submit a '<update-type>' event
+    Scenario: Change of name only journey - User backs out of DL CRI - Returns to DCMAW to use passport
+      When I submit a 'given-names-only' event
       Then I get a 'page-update-name' page response
       When I submit a 'update-name' event
       Then I get a 'dcmaw' CRI response
-      When I submit '<dl-details>' details to the CRI stub
+      When I submit 'kenneth-changed-given-name-driving-permit-valid' details to the CRI stub
       Then I get a 'drivingLicence' CRI response
       When I call the CRI stub with attributes and get an 'access_denied' OAuth error
         | Attribute | Values          |
@@ -148,11 +148,11 @@ Feature: Dropping out of authoritative source checks with DL CRI (e.g. due to in
       Then I get a 'uk-driving-licence-details-not-correct' page response
       When I submit a 'next' event
       Then I get a 'dcmaw' CRI response
-      When I submit '<passport-details>' details to the CRI stub
+      When I submit 'kenneth-changed-given-name-passport-valid' details to the CRI stub
       Then I get a 'page-dcmaw-success' page response with context 'coiNoAddress'
       When I submit a 'next' event
       Then I get a 'fraud' CRI response
-      When I submit '<fraud-details>' details to the CRI stub
+      When I submit 'kenneth-changed-given-name-score-2' details to the CRI stub
       Then I get a 'page-ipv-success' page response
       When I submit a 'next' event
       Then I get an OAuth response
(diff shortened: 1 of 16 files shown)


api-tests/features/dl-auth-source-checks/dl-cri-dropout.feature
```diff
@@ -0,0 +1,309 @@
...
+    When I call the CRI stub with attributes and get an 'access_denied' OAuth error
+      | Attribute | Values          |
+      | context   | "check_details" |
+    Then I get a 'uk-driving-licence-details-not-correct' page response
+    When I submit an 'end' event
+    Then I get a 'prove-identity-another-way' page response
+    When I submit a 'returnToRp' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P0' identity without a 'dcmaw' VC
```
Joe-Edwards-GDS: (We don't actually care about the DCMAW VC if they return with a P0 - they don't have an identity either way, but it doesn't hurt to check!)

api-tests/features/dl-auth-source-checks/dl-cri-dropout.feature
```diff
@@ -0,0 +1,309 @@
...
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P2' identity
+      And I have a dcmaw VC without 'drivingPermit' details
+
+      Examples:
+        | update-type             | dl-details                                       | passport-details                           | fraud-details                                   |
+        | given-names-and-address | kenneth-changed-given-name-driving-permit-valid  | kenneth-changed-given-name-passport-valid  | kenneth-changed-given-name-and-address-score-2  |
+        | family-name-and-address | kenneth-changed-family-name-driving-permit-valid | kenneth-changed-family-name-passport-valid | kenneth-changed-family-name-and-address-score-2 |
```
Joe-Edwards-GDS: I've mentioned to Chris, but I'm campaigning to combine these routes together (by performing the same COI check), which should halve a lot of these test cases 🤞

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -36,19 +36,19 @@ nestedJourneyStates:
...
         targetState: RESET_SESSION_INVALID_DL
 
-  RESET_SESSION_INCOMPLETE_DL:
+  REMOVE_DCMAW_FROM_SESSION:
     response:
       type: process
       lambda: reset-session-identity
       lambdaInput:
-        resetType: ALL
+        resetType: DCMAW
```
Joe-Edwards-GDS, line 48: Do you think the other reset should be DCMAW as well? I don't think there are currently any scenarios where it matters, but it feels more consistent.

(The D02 only really invalidates the DCMAW VC)

Joe-Edwards-GDS (commented):
Looks good to me - just a couple of minor comments in the routing that might be worth tweaking.

api-tests/features/dl-auth-source-checks/dl-cri-dropout.feature
thebauSoftwire: oooo nice! Should I get rid of these tests then or leave in for the meantime?

Joe-Edwards-GDS: Whichever is easiest tbh - if you keep it in then you could make a note on https://govukverify.atlassian.net/browse/PYIC-7701, or just squash them pre-emptively.

Joe-Edwards-GDS (approved):
Thanks! Maybe wait until after lunch to merge, so we avoid the no-photo-id run-through

PR #2698, opened 2024-11-12
PYIC-7651: Update journeyMap with new exit page

AmritSidhu:
### What changed

We should display this page if we receive an access-denied error code from DCMAW and the user is an international address user

### Issue tracking

- [PYIC-7651](https://govukverify.atlassian.net/browse/PYIC-7651)

[PYIC-7651]: https://govukverify.atlassian.net/browse/PYIC-7651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -806,7 +806,19 @@ states:
       end:
         targetState: RETURN_TO_RP
 
-  # Live in uk journey
+  # International Address journey
+  APP_DOC_CHECK_INTERNATIONAL_ADDRESS:
+    nestedJourney: APP_DOC_CHECK
+    exitEvents:
+      next:
+        targetState: POST_APP_DOC_CHECK_SUCCESS_PAGE
+      incomplete:
+        targetState: NON_UK_NO_APP_PAGE
+      incomplete-invalid-dl:
+        targetState: NON_UK_NO_APP_PAGE
+      alternate-doc-invalid-dl:
+        targetState: NON_UK_NO_APP_PAGE
+
   LIVE_IN_UK_PAGE:
     response:
       type: page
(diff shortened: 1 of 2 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -806,14 +806,59 @@ states:
...
+    exitEvents:
+      next:
+        targetState: POST_APP_DOC_CHECK_SUCCESS_PAGE
+      incomplete:
+        targetState: NON_UK_NO_APP_PAGE
+      incomplete-invalid-dl:
+        targetState: PROVE_IDENTITY_ANOTHER_WAY_AFTER_APP_DOC_CHECK
+      alternate-doc-invalid-dl:
+        targetState: WEB_DL_OR_PASSPORT
+        targetEntryEvent: alternate-doc-invalid-dl-another-way
```
blakeyp: We'll want these to go to the non-uk-no-app page too, as app with passport is the only viable route for an international address user. (They'll be told they can only use a passport anyway but there's a chance they still try to use a driving licence in the app itself and end up in these states)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -806,14 +806,59 @@ states:
       end:
         targetState: RETURN_TO_RP
 
-  # Live in uk journey
+  # International Address journey
+  IDENTITY_START_PAGE_IA:
```
blakeyp: Prefer to avoid using abbreviations in naming states (makes it not immediately obvious to follow eg. in the journey map visualiser), even if it makes the names a bit long `IDENTITY_START_PAGE_INTERNATIONAL_ADDRESS`?

blakeyp (approved):

PR #2704, opened 2024-11-14
BAU: Retry on Connection reset errors

Wynndow:
### What changed

Retry on Connection reset errors

### Why did it change

We're seeing exceptions due to connections being reset when attempting to fetch an access token from the CRI stubs in the build env.

This seems to be happening due to idle connections being picked up from the connection pool of the http client. It seems the connection is being reset by something upstream. I've seen it happen when a connection is over 8 minutes old, but it could be happening earlier.

Setting the keepalive timeout system property to a low value should resolve this, but it doesn't. I've tried it at 30 seconds and we're still getting the issue.

Retrying seems to be the simplest option, although it comes with a risk that the retry will occur with a connection in the pool which is has also been reset.

The inherently spikey nature of traffic in the build env - nothing, then a bunch of parallel tests hitting it at once - might explain why we see this. We create a bunch of connections to satisfy all the requests, and then they become stale.

There is some chat online about it being a NAT gateway sending a RST response, rather than FIN that is causing the issue. We seem to have the same issue as described there, including using keepalive timeout not working: https://stackoverflow.com/questions/63605720/intermittent-java-io-ioexception-connection-reset-not-connection-reset-by-peer

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -21,6 +21,7 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


Joe-Edwards-GDS (dismissed):
Happy to merge this, but in light of https://govukverify.atlassian.net/browse/PYIC-7705 it might not work either :/

Will be interesting to see...

Wynndow:
> Happy to merge this, but in light of https://govukverify.atlassian.net/browse/PYIC-7705 it might not work either :/
> 
> Will be interesting to see...

I've been able to recreate and then resolve the connections timeouts in my dev env, so I think it should work

Joe-Edwards-GDS (approved):

PR #2706, opened 2024-11-15
PYIC-7508: Check doc ID in auth source check

Wynndow:
### What changed
Check doc ID in auth source check

### Why did it change

We want to be more sure that when using a previously attained driving licence VC to dodge the auth source check, that the document used is the same as that used at DCMAW.

This checks that the document identifiers in both VCs match, and if not route to an auth source check.

### Issue tracking

- [PYIC-7508](https://govukverify.atlassian.net/browse/PYIC-7508)

[PYIC-7508]: https://govukverify.atlassian.net/browse/PYIC-7508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
+++ b/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
@@ -36,8 +36,7 @@ import uk.gov.di.ipv.core.library.verifiablecredential.domain.VerifiableCredenti
 import uk.gov.di.ipv.core.library.verifiablecredential.helpers.VcHelper;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredentialsService;
 import uk.gov.di.ipv.core.processcricallback.exception.InvalidCriCallbackRequestException;
-import uk.gov.di.model.DrivingPermitDetails;
-import uk.gov.di.model.PersonWithDocuments;
+import uk.gov.di.model.IdentityCheckSubject;
 
 import java.util.Arrays;
 import java.util.List;
@@ -45,7 +44,10 @@ import java.util.Objects;
 import java.util.Optional;
 
 import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.DL_AUTH_SOURCE_CHECK;
+import static uk.gov.di.ipv.core.library.domain.Cri.DCMAW;
+import static uk.gov.di.ipv.core.library.domain.Cri.DRIVING_LICENCE;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_VALIDATE_VERIFIABLE_CREDENTIAL_RESPONSE;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_CRI_ID;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_ACCESS_DENIED_PATH;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_DL_AUTH_SOURCE_CHECK_PATH;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_ERROR_PATH;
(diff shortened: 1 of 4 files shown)


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,29 +260,40 @@ public JourneyResponse checkVcResponse(
...
-    private Optional<VerifiableCredential> getSuccessfulDlVc(List<VerifiableCredential> vcs) {
-        return vcs.stream()
-                .filter(vc -> Cri.DRIVING_LICENCE.equals(vc.getCri()))
-                .filter(VcHelper::isSuccessfulVc)
+                                        ? personWithDocuments.getDrivingPermit()
+                                        : null)
+                .filter(Objects::nonNull)
+                .filter(drivingPermitDetailsList -> !drivingPermitDetailsList.isEmpty())
+                .map(drivingPermitDetailsList -> drivingPermitDetailsList.get(0))
+                .map(DrivingPermitDetails::getPersonalNumber)
```
Joe-Edwards-GDS: I wonder if we want to use the SPOT/CIMIT logic for this - https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3889922273/RFC+0042+Unique+identifiers+for+evidence+documents+for+de-duplicating+contra-indicators+in+Identity+Proofing+and+Verification+IPV#Driving-permits

That'll include the issuer etc., not just the number

(Up to you whether you want to implement the logic for _other_ document types)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,29 +260,40 @@ public JourneyResponse checkVcResponse(
...
+                .map(
+                        identifier ->
+                                !identifier.equals(
+                                        extractDrivingPermitIdentifier(
+                                                        sessionVcs, Cri.DRIVING_LICENCE, true)
+                                                .orElse(null)))
+                .orElse(false);
+    }
+
+    private Optional<String> extractDrivingPermitIdentifier(
```
Joe-Edwards-GDS: This is a bit of a strange compound method - I wonder if it would be clearer to split into
- `Optional<VC> findVcFromCri(List<VC> vcs, Cri cri)`
- `String getDrivingPermitIdentifier(VC vc)`

(not massively important...)

I do like splitting out `requiresAuthoritativeSourceCheck` though :)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,29 +260,40 @@ public JourneyResponse checkVcResponse(
...
+            List<VerifiableCredential> vcs, Cri cri, boolean successfulCheck) {
         return vcs.stream()
-                .filter(vc -> Cri.DCMAW.equals(vc.getCri()))
-                .filter(
+                .filter(vc -> cri.equals(vc.getCri()))
+                .filter(successfulCheck ? VcHelper::isSuccessfulVc : x -> true)
```
Joe-Edwards-GDS: Do we need this condition? I don't think we ever care about unsuccessful VCs...

I suppose you could argue that if you already had a _failed_ DL VC for the same doc id then we should treat that as a failure here, but that seems too complicated - might as well just send them on the DL auth source check.

Wynndow: Is the problem here not that if they had a failed DL VC from earlier in the journey that matched identifiers, by not checking that the succuess state for that VC, the default is to _not_ send them for an auth source check?

Joe-Edwards-GDS: Ah, I really meant the ternary as opposed to the success check itself - not that it matters now you've refactored :)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,29 +262,53 @@ public JourneyResponse checkVcResponse(
...
+                        instanceof IdentityCheckSubject identityCheckSubject
+                && identityCheckSubject.getDrivingPermit() != null
+                && !identityCheckSubject.getDrivingPermit().isEmpty()) {
+            var permit = identityCheckSubject.getDrivingPermit().get(0);
+            return String.format(
+                    "drivingPermit/%s/%s/%s/%s",
+                    permit.getIssuingCountry(),
+                    permit.getIssuedBy(),
+                    permit.getPersonalNumber(),
+                    permit.getIssueDate());
```
Joe-Edwards-GDS, line 307: Do you need to default these to 'unknown' (or anything else) if they are missing?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,29 +262,53 @@ public JourneyResponse checkVcResponse(
...
+        }
+        LOGGER.warn(
+                LogHelper.buildLogMessage("Unable to get driving permit identifier from VC")
+                        .with(LOG_CRI_ID.getFieldName(), vc.getCri()));
+        return null;
```
Joe-Edwards-GDS, line 312: This will mean that we'll skip the auth source check if this happens - but it's not going to work if the DL claim was missing anyway!

Happy to leave it as a warning - we could consider elevating to an error later if we want.

Joe-Edwards-GDS (dismissed):

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
Wynndow: They'll default to a liternal null if they're not present, which I think is fine for our purposes here

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
Wynndow: Yep - I went back on fourth on this quite a bit about if we should be throwing in this scenario - and decided to just log. If we end up in there there's something pretty broken with the DL VC which will rear its head elsewhere.

Joe-Edwards-GDS (approved):

PR #2707, opened 2024-11-15
PYIC-7590: add session vcs to submit cimit api

shivanshuit914:
## Proposed changes

- Add session vcs to submit cimit api

### What changed

- Sending session VCs to cimit api 

### Issue tracking

- [PYIC-7590](https://govukverify.atlassian.net/browse/PYIC-7590)

[PYIC-7590]: https://govukverify.atlassian.net/browse/PYIC-7590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
+++ b/libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
@@ -33,6 +33,7 @@ import uk.gov.di.ipv.core.library.verifiablecredential.dto.VerifiableCredentialR
 import uk.gov.di.ipv.core.library.verifiablecredential.helpers.VcHelper;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredentialsService;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import static uk.gov.di.ipv.core.library.auditing.helpers.AuditExtensionsHelper.getExtensionsForAudit;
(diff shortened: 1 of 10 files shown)


libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
```diff
@@ -149,6 +150,14 @@ public void storeVcs(
                 new AuditEventUser(
                         userId, ipvSessionItem.getIpvSessionId(), govukSigninJourneyId, ipAddress);
 
+        List<VerifiableCredential> mitigationVcList = new ArrayList<>();
+        if (!cri.equals(TICF)) {
+            mitigationVcList.addAll(
+                    sessionCredentialsService.getCredentials(
+                            ipvSessionItem.getIpvSessionId(), userId, true));
```
Joe-Edwards-GDS: We're already fetching the session credentials once in the `CriCheckingService`, so it's a shame to fetch them again here...

It might be a bit fiddly, but is it possible to pull that out so we only fetch the session VCs once per callback?

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
```diff
@@ -149,6 +150,14 @@ public void storeVcs(
                 new AuditEventUser(
                         userId, ipvSessionItem.getIpvSessionId(), govukSigninJourneyId, ipAddress);
 
+        List<VerifiableCredential> mitigationVcList = new ArrayList<>();
+        if (!cri.equals(TICF)) {
```
Joe-Edwards-GDS: Why does TICF need special logic here?

If we're doing this, do you think we should just skip the mitigation call entirely for the TICF CRI? It doesn't seem totally unreasonable, since TICF is unlikely to ever mitigate a CI.

Joe-Edwards-GDS (commented):
Thanks, I think this will work, but keen to cut out extra dynamodb calls if we can!

libs/cri-storing-service/src/main/java/uk/gov/di/ipv/core/library/cristoringservice/CriStoringService.java
shivanshuit914: Hm, it was failing most of the API tests for the TCIF. Let me retry again. Thanks

Joe-Edwards-GDS (commented):
This looks good to me, but we should make sure not to merge until Chris is ready.

He's also suggested we might want a feature flag, in case something goes wrong - but he sounds reasonably confident.

Joe-Edwards-GDS (approved):
Definitely want to check this is working in staging with real CIMIT :)

PR #2708, opened 2024-11-15
PYIC-7600: COI check for users with existing identity

Wynndow:
### What changed

COI check for users with existing identity

### Why did it change

[PYIC-7600: Add check-coi lambda to all IPV journeys](https://github.com/govuk-one-login/ipv-core-back/pull/2708/commits/6e52be8edfdb99b7f373ac9d9721f19be2f96b04)
This adds the COI check to the evaluate subjourney so that we run it for
all successful IPV journeys.

It also removes the fallback to the tactical store if EVCS doesn't have
any VCs. All identities have now been migrated, and EVCS is the source
of truth. This simplifies things here as we don't have to worry about
partial identies being retrieved from tactical.

[PYIC-7600: Skip COI check if no idenity](https://github.com/govuk-one-login/ipv-core-back/pull/2708/commits/86414247d5549949700907536c00b9e6ac5064ed)
If a user has no existing identity in EVCS then performing a COI check
shouldn't be neccessary. This adds a special case to the
process-journey-event lambda to skip over the check-coi lambda if the
user has no VCs in EVCS.

### Issue tracking

- [PYIC-7600](https://govukverify.atlassian.net/browse/PYIC-7600)

[PYIC-7600]: https://govukverify.atlassian.net/browse/PYIC-7600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/data/cri-stub-requests/drivingLicence/kenneth-driving-permit-dva-valid/credentialSubject.json
```diff
@@ -0,0 +1,35 @@
+{
```
Wynndow, line 1: I accidentally addded this scenario - but I'm leaving it here as it may be useful in future.

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -152,6 +152,27 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
...
+                            }
+                            if (userIdentityService.areFamilyNameAndDobCorrelatedForCoiCheck(
+                                    combinedCredentials)) {
+                                anyValidPassed = FAMILY_NAME_AND_DOB.toString();
+                                yield true;
+                            }
+                            if (userIdentityService.areVcsCorrelated(combinedCredentials)) {
+                                anyValidPassed = FULL_NAME_AND_DOB.toString();
+                                yield true;
+                            }
```
Joe-Edwards-GDS: Is it possible for this to pass, if the two previous checks have failed? If it's an interesting distinction, we should probably run the checks from strictest to weakest.

api-tests/features/cimit/alternate-doc.feature
```diff
@@ -1,241 +1,275 @@
...
+    Scenario: Mitigating when a user already has an identity should be subject to a COI check
+      Given the subject already has the following credentials
+        | CRI        | scenario               |
+        | ukPassport | kenneth-passport-valid |
+        | address    | kenneth-current        |
+        | fraud      | kenneth-score-2        |
+        | kbv        | kenneth-score-2        |
+      And the subject has the following CIs
+        | code                | document                                           |
+        | NEEDS-ALTERNATE-DOC | drivingPermit/GB/DVLA/DECER607085K98FGA/2005-02-02 |
```
Joe-Edwards-GDS: Do you think it's worth running through the entire journey here (i.e. the change of details -> D02 -> return), to record the real-life scenario that might encounter this?

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
Wynndow: That's a good point - and I think that for what we're doing the full name check isn't necessary. I'm going to wait for @thebauSoftwire's PR to get merged, as I think that will mean there's no need to add the new check type. We can just use the new one created in that PR.

api-tests/features/cimit/alternate-doc.feature
Wynndow: Yeah, I suppose it might be - will fix it up

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -356,6 +394,16 @@ private State executeStateTransition(
         return result.state();
     }
 
+    private boolean hasExistingIdentity(ClientOAuthSessionItem clientOAuthSessionItem)
```
Joe-Edwards-GDS, line 397: Hmm - I'm not a massive fan of `ProcessJourneyEventHandler` having anything to do with VCs or business logic like this, but it feels hard to do anything else :/ (even worse than the TICF case, because at in that case we do already inspect CRI enablement etc.)

Maybe let's leave it in as a compromise for now, and can think about it more carefully later. It means we'll fetch the existing identity twice when it's used, but that should be a very small % compared with the value gained by skipping...

Joe-Edwards-GDS: Other options I can think of
- Store a session variable for 'has existing identity' and use that instead - but that has other awkward problems
- Squash down all the end-journey processing into a single lambda

I quite like the second option tbh - similar to the squashing we did for ProcessCriCallback and CheckExistingIdentity, but will be a bigger piece of work than this ticket!

api-tests/src/steps/cimit-steps.ts
```diff
@@ -0,0 +1,21 @@
+import { DataTable, When } from "@cucumber/cucumber";
+import { World } from "../types/world.js";
+import { getRandomString } from "../utils/random-string-generator.js";
+import * as cimitStubClient from "../clients/cimit-stub-client.js";
+
+When(
+  "the subject has the following CIs",
```
Joe-Edwards-GDS, line 7: Is this actually used? I guess it might be useful in future scenarios, so happy to leave it in if you want

Joe-Edwards-GDS (approved):
Interested in your thoughts about the comments...

We should perhaps also keep an eye on this when it goes out, to see if it's ever getting triggered (until we introduce DL auth source checks it should never be run in prod!)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Wynndow: Yeah, i did think about setting the session var, but it felt pretty scarily prone to errors. This way felt closer to the source of truth, but yes, it's not ideal. Due to the way that different routes feed into evaluate journey, this felt like the most centralised way to do it.

And yeah, option 2 does sounds attractive. Especially if it helps to speed up that whole end journey part.

api-tests/src/steps/cimit-steps.ts
Wynndow: No, it's a leftover from running the full journey in the api-test to attract the CI, rather than setting it direct. But I thought I'd leave it in case it is useful at some point.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: I'll create a ticket - it's been a year since consolidating CRI callback so about time for another big refactor :D

PR #2713, opened 2024-11-18
PYIC-7705: add http2 client recreation logic to avoid goaway message

shivanshuit914:
## Proposed changes

- Fix occational goaway 403 error message.

### What changed

- Added time based recreation for http tracing client to avoid goaway error message

### Issue tracking

- [PYIC-7705](https://govukverify.atlassian.net/browse/PYIC-7705)

[PYIC-7705]: https://govukverify.atlassian.net/browse/PYIC-7705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -35,9 +35,9 @@ public class TracingHttpClient extends HttpClient {
     private static final List<String> HTTP_ERROR_MESSAGES_TO_RETRY =
             List.of("GOAWAY received", "Connection reset");
     private static final Duration MAX_CLIENT_AGE = Duration.ofHours(1);
-    private final HttpClient baseClient;
+    private HttpClient baseClient;
     private final AWSXRayRecorder recorder;
-    private final Instant creationTime;
+    private Instant creationTime;
 
     private TracingHttpClient(HttpClient baseClient) {
         this.baseClient = baseClient;
@@ -49,8 +49,13 @@ public class TracingHttpClient extends HttpClient {
         return new TracingHttpClient(HttpClient.newHttpClient());
     }
 
-    private boolean isClientTooOld() {
-        return Duration.between(creationTime, Instant.now()).compareTo(MAX_CLIENT_AGE) > 0;
+    // Synchronize to prevent race conditions when multiple threads attempt to recreate client
+    private synchronized void recreateBaseClientIfNecessary() {
+        if (creationTime.plus(MAX_CLIENT_AGE).isBefore(Instant.now())) {
+            LOGGER.info("Recreating base HTTP client due to age");
+            this.baseClient = HttpClient.newHttpClient();
+            this.creationTime = Instant.now(); // Reset creation time
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -33,18 +34,25 @@ public class TracingHttpClient extends HttpClient {
...
         this.recorder = AWSXRay.getGlobalRecorder();
+        this.creationTime = Instant.now();
     }
 
     public static HttpClient newHttpClient() {
         return new TracingHttpClient(HttpClient.newHttpClient());
     }
 
+    private boolean isClientTooOld() {
+        return Duration.between(creationTime, Instant.now()).compareTo(MAX_CLIENT_AGE) > 0;
```
Joe-Edwards-GDS: It's a very trivial thing, but I think `return creationTime.plus(MAX_CLIENT_AGE).isAfter(Instant.now())` might convey the intent better. The `between` threw me for a second!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -33,18 +34,25 @@ public class TracingHttpClient extends HttpClient {
...
         this.baseClient = baseClient;
         this.recorder = AWSXRay.getGlobalRecorder();
+        this.creationTime = Instant.now();
     }
 
     public static HttpClient newHttpClient() {
         return new TracingHttpClient(HttpClient.newHttpClient());
     }
 
+    private boolean isClientTooOld() {
```
Joe-Edwards-GDS: I think this needs an explanatory comment - e.g. linking to the Jira ticket, otherwise it's a bit unclear what's going on.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -110,6 +118,10 @@ private <T> HttpResponse<T> sendWithTracing(
     public <T> HttpResponse<T> send(
             HttpRequest request, HttpResponse.BodyHandler<T> responseBodyHandler)
             throws IOException, InterruptedException {
+        if (isClientTooOld()) {
+            LOGGER.info("Recreating HTTP client due to age");
+            return newHttpClient().send(request, responseBodyHandler);
```
Joe-Edwards-GDS: This isn't quite right - it's creating a new `TracingHttpClient` and then only using it for this single request.

Instead, I think we probably want to recreate the `baseClient` (and reassign it). It may be worth pulling the logic out into a `synchronized` method, to avoid the risk of race conditions, but it's not critical.

Joe-Edwards-GDS (approved):
Thanks Shiv, did you manage to reproduce the behaviour yourself?

If not, we might just need to try this and see...

shivanshuit914:
> Thanks Shiv, did you manage to reproduce the behaviour yourself?
> 
> If not, we might just need to try this and see...

No, I tried few ways but didn't work. I tried on my dev env.

PR #2714, opened 2024-11-18
PYIC-7701: Use single coi check for update name journeys

thebauSoftwire:
### What changed
- use single COI check for update family + given name journeys
- combines above routing
- update api tests as now user can still successfully update their e.g. given name despite selecting to update their family name at the start of the journey

Update to audit events is here: https://github.com/govuk-one-login/event-catalogue/pull/308

### Why did it change
Using a singe coi check for update-given and update-family name journeys means we can reduce duplication and simplify the journey map. It also lets us cut down on some of the api tests.

Screenshot 2024-11-18 at 15 38 44

### Issue tracking

- [PYIC-7701](https://govukverify.atlassian.net/browse/PYIC-7701)

[PYIC-7701]: https://govukverify.atlassian.net/browse/PYIC-7701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/repeat-fraud-check/p2-repeat-fraud-check.feature
+++ b/api-tests/features/repeat-fraud-check/p2-repeat-fraud-check.feature
@@ -50,16 +50,6 @@ Feature: Repeat fraud check journeys
     Then I get a 'dcmaw' CRI response
     When I submit 'kenneth-changed-family-name-driving-permit-valid' details to the CRI stub
     Then I get a 'page-dcmaw-success' page response with context 'coiNoAddress'
-    When I submit a 'next' event
-    Then I get a 'fraud' CRI response
-    When I submit 'kenneth-changed-family-name-score-2' details to the CRI stub
-    Then I get a 'page-ipv-success' page response
-    When I submit a 'next' event
-    Then I get an OAuth response
-    When I use the OAuth response to get my identity
-    Then I get a 'P2' identity
-    And my identity 'GivenName' is 'Kenneth'
-    And my identity 'FamilyName' is 'Smith'
 
   Scenario: Fraud 6 Months Expiry + Address Update
     # Repeat fraud check with update address
@@ -105,18 +95,6 @@ Feature: Repeat fraud check journeys
     Then I get a 'dcmaw' CRI response
     When I submit 'kenneth-changed-given-name-driving-permit-valid' details to the CRI stub
     Then I get a 'page-dcmaw-success' page response with context 'coiAddress'
-    When I submit a 'next' event
-    Then I get a 'address' CRI response
-    When I submit 'kenneth-changed' details to the CRI stub
-    Then I get a 'fraud' CRI response
-    When I submit 'kenneth-changed-given-name-score-2' details to the CRI stub
-    Then I get a 'page-ipv-success' page response
-    When I submit a 'next' event
-    Then I get an OAuth response
-    When I use the OAuth response to get my identity
-    Then I get a 'P2' identity
-    And my identity 'GivenName' is 'Ken'
-    And my address 'streetName' is 'King Road'
 
   Scenario: Unsupported Changes
     # Repeat fraud check with various unsupported events and back navigation
--- a/api-tests/features/reuse-update-details/p2-reuse-update-details.feature
+++ b/api-tests/features/reuse-update-details/p2-reuse-update-details.feature
@@ -12,52 +12,29 @@ Feature: Identity reuse update details
         When I submit a 'update-details' event
         Then I get a 'update-details' page response
 
-    Scenario: Given Name Change Failure
-        When I submit a 'given-names-only' event
+    Scenario Outline: Successful name change - <selected-name-change> name change but user updates <actual-name-change> name instead
+        When I submit a '<selected-name-change>' event
         Then I get a 'page-update-name' page response
         When I submit a 'update-name' event
         Then I get a 'dcmaw' CRI response
-        When I submit 'kenneth-changed-family-name-driving-permit-valid' details to the CRI stub
+        When I submit '<details>' details to the CRI stub
         Then I get a 'page-dcmaw-success' page response with context 'coiNoAddress'
         When I submit a 'next' event
         Then I get a 'fraud' CRI response
-        When I submit 'kenneth-changed-family-name-score-2' details to the CRI stub
-        Then I get a 'sorry-could-not-confirm-details' page response with context 'existingIdentityValid'
+        When I submit '<fraud-details>' details to the CRI stub
+        Then I get a 'page-ipv-success' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P2' identity
+        And my identity 'GivenName' is '<expected-given-name>'
+        And my identity 'FamilyName' is '<expected-family-name>'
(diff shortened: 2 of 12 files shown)


api-tests/features/repeat-fraud-check/p2-repeat-fraud-check.feature
```diff
@@ -50,16 +50,6 @@ Feature: Repeat fraud check journeys
     Then I get a 'dcmaw' CRI response
     When I submit 'kenneth-changed-family-name-driving-permit-valid' details to the CRI stub
     Then I get a 'page-dcmaw-success' page response with context 'coiNoAddress'
-    When I submit a 'next' event
```
thebauSoftwire, line 53: I've cut this down (and the one on line 90) instead of just removing it to test the routing in `repeat-fraud-check.yaml` and the bits after after tested in the test above it since we've now merged the journeys together

api-tests/features/reuse-update-details/p2-reuse-update-details.feature
```diff
@@ -99,18 +76,6 @@ Feature: Identity reuse update details
         Then I get a 'dcmaw' CRI response
         When I submit 'kenneth-changed-given-name-driving-permit-valid' details to the CRI stub
         Then I get a 'page-dcmaw-success' page response with context 'coiAddress'
-        When I submit a 'next' event
```
thebauSoftwire, line 102: similar reasoning as [here](https://github.com/govuk-one-login/ipv-core-back/pull/2714/files#r1846900747)

api-tests/features/reuse-update-details/p2-reuse-update-details.feature
```diff
@@ -12,52 +12,29 @@ Feature: Identity reuse update details
         When I submit a 'update-details' event
         Then I get a 'update-details' page response
 
-    Scenario: Given Name Change Failure
-        When I submit a 'given-names-only' event
+    Scenario Outline: Successful name change - <selected-name-change> name change but user updates <actual-name-change> name instead
```
Joe-Edwards-GDS, line 15: This is fine to include (although not super important), but does that mean we need a new scenario for a failure in its place?

api-tests/features/repeat-fraud-check/p2-repeat-fraud-check.feature
Joe-Edwards-GDS: Makes sense :)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -179,40 +179,30 @@ public boolean areVcsCorrelated(List<VerifiableCredential> vcs)
...
 
-        if (!checkBirthDateCorrelationInCredentials(successfulVcs)) {
-            LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_BIRTHDATE_CORRELATION));
-            return false;
-        }
-        return true;
+        var isBirthDateCorrelated = checkBirthDateCorrelationInCredentials(successfulVcs);
+
+        return (areGivenNamesCorrelated && isBirthDateCorrelated)
+                || (isFamilyNameCorrelated && isBirthDateCorrelated);
```
Joe-Edwards-GDS, line 205: (Note that pulling the variables out will mean the logic won't short-circuit - not really an issue as it should all be very fast anyway)

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -146,10 +146,8 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
             var combinedCredentials = Stream.concat(oldVcs.stream(), sessionVcs.stream()).toList();
             var successfulCheck =
                     switch (checkType) {
-                        case GIVEN_NAMES_AND_DOB -> userIdentityService
-                                .areGivenNamesAndDobCorrelated(combinedCredentials);
-                        case FAMILY_NAME_AND_DOB -> userIdentityService
-                                .areFamilyNameAndDobCorrelatedForCoiCheck(combinedCredentials);
+                        case GIVEN_OR_FAMILY_NAME_AND_DOB -> userIdentityService
+                                .areNamesAndDobCorrelated(combinedCredentials);
```
Joe-Edwards-GDS, line 150: I wonder if it'd be useful to indicate _which_ check passed - e.g. in a log? It's not a requirement though.

cc @Wynndow

Joe-Edwards-GDS (commented):
Really just the question about whether we're missing a COI check fail test

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
Wynndow: I've got something similar in my PR if you wanted to defer?

api-tests/features/reuse-update-details/p2-reuse-update-details.feature
thebauSoftwire: Would the tests in [p2-reuse-update-details-failure.feature](https://github.com/govuk-one-login/ipv-core-back/blob/a25de84d0ecdc1a81e01a399ccd06445c3239251/api-tests/features/reuse-update-details/p2-reuse-update-details-failure.feature#L108) and [p2-repeat-fraud-check-failure.feature](https://github.com/govuk-one-login/ipv-core-back/blob/a25de84d0ecdc1a81e01a399ccd06445c3239251/api-tests/features/repeat-fraud-check/p2-repeat-fraud-check-failure.feature#L81) cover the failure scenario here?

Joe-Edwards-GDS: Ah - sorry there is one for a failed COI check. I lost it among all the other failures..!

Joe-Edwards-GDS (approved):

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
thebauSoftwire: could you expand a bit more on this? Interested in what you mean by the logic not short-circuiting

Joe-Edwards-GDS: If you write something like:
```
return isBirthDateCorrelated(vcs) && (isGivenNamesCorrelated(vcs) || isFamilyNameCorrelated(vcs);
```
then if `isBirthDateCorrelated` returns false, it will 'short-circuit' the logic and won't execute any of the other methods (since it's guaranteed to return false regardless) - https://en.wikipedia.org/wiki/Short-circuit_evaluation

If you pull them out into variables first, then they'll all be computed in advance.

thebauSoftwire: ohh I see, thanks that makes sense :) I saw the approval before and merged but I can update this as part of another PR

PR #2717, opened 2024-11-19
BAU: Fix local running client jwks urls

Wynndow:
### What changed

Fix local running client jwks urls

### Why did it change

When running as separate containers on the same host, we need to use a special hostname to allow inter container communication.

### DIFF
--- a/local-running/core.local.params.yaml
+++ b/local-running/core.local.params.yaml
@@ -42,14 +42,14 @@ core:
       publicKeyMaterialForCoreToVerify: "{\"kty\":\"EC\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"
       validRedirectUrls: "http://localhost:4500/callback"
       validScopes: "openid"
-      jwksUrl: "http://localhost:4500/.well-known/jwks.json"
+      jwksUrl: "http://host.docker.internal:4500/.well-known/jwks.json"
     authStub:
       id: authStub
       issuer: authStub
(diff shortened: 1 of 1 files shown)


local-running/core.local.params.yaml
```diff
@@ -42,14 +42,14 @@ core:
...
-      jwksUrl: "http://localhost:4500/.well-known/jwks.json"
+      jwksUrl: "http://host.docker.internal:4500/.well-known/jwks.json"
     authStub:
       id: authStub
       issuer: authStub
       publicKeyMaterialForCoreToVerify: "{\"kty\":\"EC\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"
       validRedirectUrls: "http://localhost:4500/callback"
       validScopes: "reverification"
-      jwksUrl: "http://localhost:4500/.well-known/jwks.json"
+      jwksUrl: "http://host.docker.internal:4500/.well-known/jwks.json"
```
Joe-Edwards-GDS, line 52: Presumably this breaks things if you run orch-stub as a separate container? Is there another way to get the docker container to route through localhost to make it agnostic?

I'm not super-fussed, as you won't need to do it very often, but could be a bit of a gotcha.

Wynndow: I think this should be ok? `host.docker.internal` inside the container resolves to the host machine - so should work either way. As long as something is listening on 4500

Joe-Edwards-GDS: Ah I see - I guess it means that it'll fail if you run the local running app instead? If anything that's more annoying - I do that quite a lot for the API tests...

This has also made me realise that we're not passing the `kid` header for the API tests - though this might not be trivial to fix, since there won't be an orch-stub running.

I wonder if we should remove this JWKS URL for now (so it uses the hardcoded config) and raise a proper ticket to sort this out.

Wynndow: If you just spin up core-back on its own it'll attempt to fetch the config and fail, and so fall back to the hardcoded option.
We could do a similar thing to what we do for the AWS service and have a local running version of the oauth key service that just grabs stuff from config, rather than reaching out to an endpoint?

PR #2721, opened 2024-11-20
Pyic 7453 add nested journey audit events

DanCorderIPV:
### What changed

Add ability to raise audit events on journey events going to sub journeys

### Why did it change

To support adding the new IPV_INTERNATIONAL_ADDRESS_START event

### Issue tracking

- [PYIC-7453](https://govukverify.atlassian.net/browse/PYIC-7453)

[PYIC-7453]: https://govukverify.atlassian.net/browse/PYIC-7453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/TransitionResult.java
```diff
@@ -12,6 +14,40 @@ public record TransitionResult(
         Map<String, String> auditContext,
         String targetEntryEvent) {
     public TransitionResult(State state) {
-        this(state, null, null, null);
+        this(state, new ArrayList<>(), new HashMap<>(), null);
+    }
+
+    public static List<AuditEventTypes> mergeAuditEvents(
```
Joe-Edwards-GDS: Do you think it's worth making these generic collection utils - i.e. to merge nullable collections together? I don't think there's anything specific about audit events in them.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
```diff
@@ -30,8 +31,8 @@ public class BasicEvent implements Event {
     private LinkedHashMap<String, Event> checkIfDisabled;
     private LinkedHashMap<String, Event> checkFeatureFlag;
     private LinkedHashMap<String, Event> checkJourneyContext;
-    private List<AuditEventTypes> auditEvents;
-    private LinkedHashMap<String, String> auditContext;
+    private List<AuditEventTypes> auditEvents = new ArrayList<>();
+    private LinkedHashMap<String, String> auditContext = new LinkedHashMap<>();
```
Joe-Edwards-GDS: You're adding a default here, but also null-checking - do we need both?

If we stick with the defaults, do these need to be mutable?

Joe-Edwards-GDS (commented):
Looks reasonable to me - a couple of style comments but seems legit from a journey engine POV

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/TransitionResult.java
DanCorderIPV: I'll do that - I was initially concerned that the behaviour of returning an empty collection when both inputs are null was maybe not generic enough. But on reflection it seems like a good idea.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
DanCorderIPV: I'll revert this and update the test instead - if we were to go down the path of adding defaults I think it would need more time than we really have to do it properly.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/TransitionResult.java
```diff
@@ -12,6 +14,6 @@ public record TransitionResult(
         Map<String, String> auditContext,
         String targetEntryEvent) {
     public TransitionResult(State state) {
-        this(state, null, null, null);
+        this(state, new ArrayList<>(), new HashMap<>(), null);
```
Joe-Edwards-GDS: `List.of()` (or `Collections.emptyList()`) and the equivalent map, are fractionally more performant (saves an allocation). Doesn't really matter for our performance, but perhaps a good habit!

(Fine to leave if you prefer)

Joe-Edwards-GDS (dismissed):

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/TransitionResult.java
DanCorderIPV: Nice, thanks for the tip

Joe-Edwards-GDS (approved):

PR #2723, opened 2024-11-20
PYIC-7461 Clean up COI Fast Follow feature flag

Joe-Edwards-GDS:
### What changed

Remove usages of the feature flag, and the routing that was using it

### Why did it change

Tidy up old behaviour

### Issue tracking

- [PYIC-7461](https://govukverify.atlassian.net/browse/PYIC-7461)

[PYIC-7461]: https://govukverify.atlassian.net/browse/PYIC-7461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
@@ -106,10 +106,7 @@ states:
           updateFields: family-name,address
           updateSupported: true
       dob:
-        targetState: UPDATE_NAME_DOB
-        checkFeatureFlag:
-          updateDetailsAccountDeletion:
-            targetState: UPDATE_NAME_DOB_ALLOW_ACCOUNT_DELETION
+        targetState: UPDATE_NAME_DOB_ALLOW_ACCOUNT_DELETION
         auditEvents:
           - IPV_USER_DETAILS_UPDATE_START
           - IPV_USER_DETAILS_UPDATE_SELECTED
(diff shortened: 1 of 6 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
```diff
@@ -106,109 +106,79 @@ states:
           updateFields: family-name,address
           updateSupported: true
       dob:
-        targetState: UPDATE_NAME_DOB
-        checkFeatureFlag:
-          updateDetailsAccountDeletion:
-            targetState: UPDATE_NAME_DOB_ALLOW_ACCOUNT_DELETION
+        targetState: UPDATE_NAME_DOB_ALLOW_ACCOUNT_DELETION
```
Wynndow, line 109: Might be nice to rename the now default states - so this might become just `UPDATE_NAME_DOB`

Wynndow (approved):
Just a thought about the names of the states

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
Joe-Edwards-GDS: Yeah, it'd be nice - the only slightly annoying thing is that we'd need to do it in a two-step deployment to avoid breaking sessions (though for these screens the chance of actually hitting one is very low...)

We almost need a ticket every month to go through and revise all the names that have drifted like this(!)

PR #2737, opened 2024-11-25
PYIC-7757: update local-running set-up to use jwks endpoints

thebauSoftwire:
### What changed
- update jar signing keys with kid and send in header
- separate client configs for api tests when running core-back locally
  - when running the api tests against a containerised core-back, should use `npm run test:local-container` instead
  - adds some unit tests for new YamlConfigService logic
  - updates README for api tests
  - when running core-back outside of a container, we don't use the jwks endpoints bc for some reason, we get a `Connection refused error` :/ so falling back to just using the hard-coded key instead and relying on the selenium tests to cover this functionality

### Why did it change
We want the api-tests be as close to production as possible which means using the client jwks endpoint when validating client signature. 

### Issue tracking

- [PYIC-7757](https://govukverify.atlassian.net/browse/PYIC-7757)

[PYIC-7757]: https://govukverify.atlassian.net/browse/PYIC-7757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
```diff
@@ -56,6 +73,22 @@ public YamlConfigService(File parametersFile, File secretsFile) {
...
 
+    private JsonNode getParamsWithCorrectClientConfig(File parametersFile, File apiClientConfigFile)
+            throws IOException {
+        try {
+            var params = YAML_OBJECT_MAPPER.readTree(parametersFile).get(CORE);
+
+            if (apiClientConfigFile != null) {
+                var apiClientConfig = YAML_OBJECT_MAPPER.readTree(apiClientConfigFile).get(CLIENTS);
+                ((ObjectNode) params).replace(CLIENTS, apiClientConfig);
+            }
```
thebauSoftwire: Left it like this for simplicity as I know we're only overriding the client config for the api tests but maybe this could override explicit config defined in the yaml (similar to how dev-deploy works with its user configs) instead of just the client? Couldn't find a handy package to do this in java though so might need to make a util from scratch.

local-running/core.api-tests.params.yaml
```diff
@@ -0,0 +1,14 @@
...
+    issuer: orchStub
+    publicKeyMaterialForCoreToVerify: "{\"kty\":\"EC\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"
+    validRedirectUrls: "http://localhost:4500/callback"
+    validScopes: "openid"
+  authStub:
+    id: authStub
+    issuer: authStub
+    publicKeyMaterialForCoreToVerify: "{\"kty\":\"EC\",\"crv\":\"P-256\",\"x\":\"E9ZzuOoqcVU4pVB9rpmTzezjyOPRlOmPGJHKi8RSlIM\",\"y\":\"KlTMZthHZUkYz5AleTQ8jff0TJiS3q2OB9L5Fw4xA04\"}"
+    validRedirectUrls: "http://localhost:4500/callback"
+    validScopes: "reverification"
```
thebauSoftwire: if we were to make a more flexible override function, we can make this look more like the config in `core.local.params`

api-tests/src/utils/jwt-signer.ts
Joe-Edwards-GDS, line 1: Do you need to make the same change in `jar-generator.ts`, for the JWE?

api-tests/.env.local
```diff
@@ -6,7 +6,7 @@ CORE_BACK_EXTERNAL_API_URL="http://localhost:4502"
 CORE_BACK_PUBLIC_ENCRYPTION_KEY='{"kty": "RSA","n": "0465qJwo8nCkC2tvV4niuWF6IM6pNjmeYszhTwHPY609-HVAtO8PoRLUyA86rzQ-QzbT7XxbzCjfyRXoRFOGleZqTuwlc25ezDxV58bhecPiWFMaFYOS1W7zIDsVFo37gjjvtkcD6OqK8PKAv6n5tUphjDCcnnmpTMIyGAnzmQCbSkJWu6V_gc3tirAugXoZukMCohxw3_-c6prhMN0smDNv0qWmva3oqokabePwe1OS72DXyXR-TPd_Dtz4-tRr9jvZwHulX4Zcs1BBbjBpIim3WNY8asv9yjlBxkdt-nckhCMZekPuT7xWSTrvccB_fnnSUgEQW_5irLNdnr5MWQ","e": "AQAB","alg": "RS256"}' # pragma: allowlist secret
```
Joe-Edwards-GDS, line 6: Do the encryption keys need a `kid` as well?

Joe-Edwards-GDS: I suppose the alternative is to read the encryption key from the Core JWKS endpoint instead 🤔

What do you think? Should be a separate PR, but we have the method in `core-back-external-client` already! (Should probably only fetch once per test run though)

local-running/core.api-tests.params.yaml
Joe-Edwards-GDS: Sorry I meant something slightly different here!

Each of these blocks (i.e. `orchStub` and `authStub`) are 'clients' for IPV Core, and are identified by the `id` - which corresponds to the `client_id` parameter in the JAR (i.e. in `/jar-requests/*.json`).

That should mean you can add these blocks into the `core.local.params.yaml`, but with different client ids - e.g. `orchApiTest` or something. Then you don't need to 'override' the config like this.

(You'll also need to add the config to the build configuration as well 🙂 )

thebauSoftwire: ohh I see, that's much simpler thanks :)

api-tests/.env.local
thebauSoftwire: that's a good idea, I'll do that as part of a separate PR

local-running/core.api-tests.params.yaml
thebauSoftwire: PR for build config update: https://github.com/govuk-one-login/ipv-core-common-infra/pull/1147

api-tests/package.json
```diff
@@ -8,6 +8,7 @@
     "test:dev": "CORE_ENV=dev cucumber-js",
     "test:build": "CORE_ENV=build cucumber-js --tags '@Build'",
     "test:local": "CORE_ENV=local cucumber-js",
+    "test:local-container": "AUTH_CLIENT=authStub ORCH_CLIENT=orchStub CORE_ENV=local cucumber-js",
```
Joe-Edwards-GDS: Is this necessary - wouldn't we still use the same client configuration as plain old `local`?

Joe-Edwards-GDS (commented):
One question, but this looks much cleaner thanks :)

api-tests/package.json
thebauSoftwire: by default, local-running has the `jwksUrl` set to host.docker.internal for `orchStub` and `authStub` which we want for running in a container. The api test specific config doesn't have the jwksUrl bc of that connection refused error I was getting :/

Joe-Edwards-GDS: Could we just use the hardcoded key whenever we run the API tests - that way it's always consistent (and, if desirable, we could generate our own key specifically for the API tests)?

thebauSoftwire: yeah, that's a fair point - will remove!

api-tests/src/utils/jar-generator.ts
```diff
@@ -30,6 +31,14 @@ export const generateJarPayload = async (
     ),
   ) as JarRequest;
 
+  payloadData.iss = ORCH_CLIENT || "orchStub";
+  payloadData.client_id = ORCH_CLIENT || "orchStub";
+
+  if (session.journeyType === "reverification") {
+    payloadData.iss = AUTH_CLIENT || "authStub";
+    payloadData.client_id = AUTH_CLIENT || "authStub";
+  }
```
Joe-Edwards-GDS: Do we still need all this stuff with ORCH_CLIENT and AUTH_CLIENT?

thebauSoftwire: ah 🤦🏻‍♀️ thanks for the spot

Joe-Edwards-GDS (dismissed):
Thanks! I think this is a neat compromise :)

Joe-Edwards-GDS (approved):

PR #2742, opened 2024-11-26
PYIC-6068: Remove reads/writes from/to tactical

Wynndow:
**Not to be merged until we've been given the green light**

### What changed

Remove reads and writes from tactical store

### Why did it change

This removes all the reads and writes from the tactical store, instead purely using EVCS

### Issue tracking

- [PYIC-6068](https://govukverify.atlassian.net/browse/PYIC-6068)

[PYIC-6068]: https://govukverify.atlassian.net/browse/PYIC-6068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-mobile-app-vc-receipt/src/main/java/uk/gov/di/ipv/core/checkmobileappvcreceipt/CheckMobileAppVcReceiptHandler.java
```diff
@@ -165,13 +169,20 @@ private JourneyResponse getJourneyResponse(CheckMobileAppVcReceiptRequest reques
...
         // Retrieve and validate cri response and vc
-        var criResponse = criResponseService.getCriResponseItem(userId, Cri.DCMAW_ASYNC);
+        var criResponse = criResponseService.getCriResponseItem(userId, DCMAW_ASYNC);
         if (criResponse == null) {
             throw new InvalidCriResponseException(ErrorResponse.CRI_RESPONSE_ITEM_NOT_FOUND);
         }
 
-        var vc = verifiableCredentialService.getVc(userId, Cri.DCMAW_ASYNC.getId());
-        if (CriResponseService.STATUS_PENDING.equals(criResponse.getStatus()) && vc == null) {
+        var dcmawAsyncVc =
```
Wynndow, line 186: @MikeCollingwood Would you mind reviewing this change? I've made the assumption that we can just return a 404 if the VC isn't found, regardless of the status of the CRI response, but I'm missing all of the context so am quite probably wrong.
I think that the previous version would have blown up if the status was anytthing other than pending and there was no VC found, as the use of `List.of()` on line 198 doesn't allow nulls as input.

MikeCollingwood: We return a not found if the VC isn't found and the CriResponseItem status is pending.

The other status checks below should ensure that everything past them will have a non-null vc. I.e. a check on each status directly leads to a return except STATUS_PENDING where the vc is also checked, so the only case that could occur after ln184/194 is STATUS_PENDING and non-null vc. 

This change you've made would work if you put the checks:
```java
        if (CriResponseService.STATUS_ERROR.equals(criResponse.getStatus())) {
            return JOURNEY_ERROR;
        }
        if (CriResponseService.STATUS_ABANDON.equals(criResponse.getStatus())) {
            return JOURNEY_ABANDON;
        }
```
above the check for dcmawAsyncVc.isEmpty()

Wynndow: Ah ok, so the only status possible is PENDING, ABANDON or ERROR. Will update, thanks Mike

Joe-Edwards-GDS (dismissed):
LGTM, if Mike is happy

(Subject to the caveat about waiting for approval from SMT!)

lambdas/check-mobile-app-vc-receipt/src/main/java/uk/gov/di/ipv/core/checkmobileappvcreceipt/CheckMobileAppVcReceiptHandler.java
```diff
@@ -183,8 +182,23 @@ private JourneyResponse getJourneyResponse(CheckMobileAppVcReceiptRequest reques
             return JOURNEY_ABANDON;
         }
 
+        var dcmawAsyncVc =
+                evcsService
+                        .getVerifiableCredentialsByState(
+                                userId, clientOAuthSessionItem.getEvcsAccessToken(), CURRENT)
+                        .get(CURRENT)
```
MikeCollingwood: After learning more about EVCS, I realise instead we'll want to fetch from `PENDING_RETURN` and load these VCs into the session credentials, if they exist.

```java
        ...

        if (CriResponseService.STATUS_ERROR.equals(criResponse.getStatus())) {
            return JOURNEY_ERROR;
        }
        if (CriResponseService.STATUS_ABANDON.equals(criResponse.getStatus())) {
            return JOURNEY_ABANDON;
        }

        var dcmawAsyncVc =
                evcsService
                        .getVerifiableCredentialsByState(
                                userId, clientOAuthSessionItem.getEvcsAccessToken(), PENDING_RETURN)
                        .get(PENDING_RETURN)
                        .stream()
                        .filter(vc -> DCMAW_ASYNC.equals(vc.getCri()))
                        .findFirst();

        if (dcmawAsyncVc.isEmpty()) {
            return null;
        }

        sessionCredentialsService.persistCredentials(
                List.of(dcmawAsyncVc.get()), auditEventUser.getSessionId(), false);

        ...
```

MikeCollingwood (dismissed):
Looks good to me

Joe-Edwards-GDS (approved):

PR #2746, opened 2024-11-27
PYIC-7637: MFA failureCode and failureDescription

thebauSoftwire:
### What changed
- returns identity_check_incomplete by default if reverification is unsuccessful
- returns identity_check_failed in the CRI callback when user fails CRI
- returns identity_did_not_match in the COI check when check fails
- updates api tests
- updates
Screenshot 2024-11-27 at 10 27 55
 openAPI schema

note: contract tests do not need to be updated as the consumer tests only have cases for a successful reverification response and an invalid access token.

### Why did it change
We were originally using the `errorCode` from the ipvSession to set these values on the reverification response but these are used for OAuth errors and will never be set as the `reverification` endpoint won't ever be reached once we get one. Instead, we should use a separate `failureCode` + desc to be returned on the reverification response

### Issue tracking

- [PYIC-7637](https://govukverify.atlassian.net/browse/PYIC-7637)

[PYIC-7637]: https://govukverify.atlassian.net/browse/PYIC-7637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/mfa-reset-journey.feature
+++ b/api-tests/features/mfa-reset-journey.feature
@@ -29,7 +29,7 @@ Feature: MFA reset journey
     When I submit a 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my MFA reset result
-    Then I get an unsuccessful MFA reset result with failureCode 'identity_check_failed'
+    Then I get an unsuccessful MFA reset result with failure code 'identity_check_failed'
 
     # New journey with same user id
     When I start a new 'medium-confidence' journey
@@ -42,7 +42,7 @@ Feature: MFA reset journey
     When I submit a 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my MFA reset result
-    Then I get an unsuccessful MFA reset result with failureCode 'identity_check_incomplete'
+    Then I get an unsuccessful MFA reset result with failure code 'identity_check_incomplete'
 
   Scenario: Failed MFA reset journey - no photo id
     When I submit an 'end' event
@@ -50,7 +50,7 @@ Feature: MFA reset journey
     When I submit an 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my MFA reset result
-    Then I get an unsuccessful MFA reset result with failureCode 'identity_check_incomplete'
+    Then I get an unsuccessful MFA reset result with failure code 'identity_check_incomplete'
 
   Scenario: Failed MFA reset journey - failed verification score
     When I submit an 'appTriage' event
@@ -60,7 +60,7 @@ Feature: MFA reset journey
     When I submit a 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my MFA reset result
-    Then I get an unsuccessful MFA reset result with failureCode 'identity_check_failed'
+    Then I get an unsuccessful MFA reset result with failure code 'identity_check_failed'
 
   Scenario: Failed MFA reset journey - non-matching identity
     When I submit an 'appTriage' event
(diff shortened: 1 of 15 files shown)


api-tests/features/mfa-reset-journey.feature
```diff
@@ -29,7 +29,7 @@ Feature: MFA reset journey
     When I submit a 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my MFA reset result
-    Then I get an unsuccessful MFA reset result
+    Then I get an unsuccessful MFA reset result with failureCode 'identity_check_failed'
```
Joe-Edwards-GDS: Nice!

api-tests/src/steps/ipv-steps.ts
```diff
@@ -503,8 +503,12 @@ Then(
 );
 
 Then(
-  /^I get a(?:n)? (successful|unsuccessful) MFA reset result$/,
-  async function (this: World, expectedMfaResetResult: string): Promise<void> {
+  /^I get a(?:n)? (successful|unsuccessful) MFA reset result( with failureCode '([\w_]+)')?$/,
```
Joe-Edwards-GDS: (I think it would be ok to have 'failure code' as it is meaningful, but this is fine)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -269,6 +272,11 @@ && requiresAuthoritativeSourceCheck(newVcs, sessionVcs)) {
         return JOURNEY_NEXT;
     }
 
+    private void setFailedIdentityCheckOnIpvSessionItem(IpvSessionItem ipvSessionItem) {
+        ipvSessionItem.setFailureCode(ReverificationFailureCode.IDENTITY_CHECK_FAILED);
+        ipvSessionItem.setFailureDescription("Identity check failed.");
+    }
```
Joe-Edwards-GDS, line 288: An important gotcha here: updating the session object will not automatically persist the change* - you need to call `ipvSessionService.updateIpvSession(ipvSessionItem)` to update it in DynamoDB.

Ideally we want to minimise the number of times we do this, but it's better to make sure it happens - even if some later step throws an exception! (We already update the session earlier in `ProcessCriCallbackHandler`, for example).

*It does when running locally, but this is a flaw in our local setup - see https://github.com/govuk-one-login/ipv-core-back/pull/2139/files/f936f6e3fd0edce5a5d092059612907688e53378#r1668974060 for some discussion

Joe-Edwards-GDS, line 288: Do you think we should skip this if the scope isn't reverification? I guess it doesn't hurt too much, but could be an extra round trip to DynamoDB.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -37,6 +38,10 @@ public class IpvSessionItem implements PersistenceItem {
     private ReverificationStatus reverificationStatus;
     private List<String> stateStack = new ArrayList<>();
 
+    // These are used as part of an unsuccessful reverification response
+    private ReverificationFailureCode failureCode;
+    private String failureDescription;
```
Joe-Edwards-GDS: Is the description ever variable - could we just store the code and put the description in the enum as well?

openAPI/core-back-external.yaml
```diff
@@ -324,9 +324,10 @@ components:
           type: boolean
         sub:
           type: string
-        error_code:
+        failure_code:
           type: string
-        error_description:
+          enum: ["no_identity_available", "identity_check_incomplete", "identity_check_failed", "identity_did_not_match"]
+        failure_description:
```
Joe-Edwards-GDS, line 330: Nice!

Joe-Edwards-GDS (approved):
Nice!

PR #2747, opened 2024-11-27
PYIC-7757: use public enc key from core-back jwks endpoint for api tests

thebauSoftwire:
### What changed
- uses public enc key from core-back jwks endpoint for api tests
- removes `CORE_BACK_PUBLIC_ENCRYPTION_KEY` config from endpoint

### Issue tracking

- [PYIC-7757](https://govukverify.atlassian.net/browse/PYIC-7757)

[PYIC-7757]: https://govukverify.atlassian.net/browse/PYIC-7757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/src/utils/jar-generator.ts
+++ b/api-tests/src/utils/jar-generator.ts
@@ -1,4 +1,5 @@
 import * as jose from "jose";
+import { BeforeAll } from "@cucumber/cucumber";
 import config from "../config/config.js";
 import fs from "node:fs/promises";
 import * as path from "path";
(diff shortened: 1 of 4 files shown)


api-tests/src/utils/jar-generator.ts
```diff
@@ -70,8 +69,23 @@ export const generateJarPayload = async (
...
+    throw Error("No public encryption key found for core-back");
+  }
+  return key;
+};
+
+export const encryptJarRequest = async (
+  payload: JarRequest,
+): Promise<string> => {
+  const encryptionKeyJwk = await getCoreBackEncKey();
+  const encKey = await jose.importJWK(encryptionKeyJwk, encAlg);
```
Joe-Edwards-GDS: Do you think it's still worth caching these at the module level, so we don't have to refetch this for every test case? (We wouldn't need to worry about expiry or anything, just store the result once)

The endpoint will be pretty quick, so not critical.

Joe-Edwards-GDS (commented):
Thanks!

api-tests/src/utils/jar-generator.ts
thebauSoftwire: good idea, thanks :) (I assume we don't need anything more complex like an expiry)

Joe-Edwards-GDS: Yeah, if the keys change while the tests are running then we must be very unlucky!

Joe-Edwards-GDS (approved):
🎉

PR #2751, opened 2024-11-28
PYIC-7076: Add check-reverification-identity lambda

Wynndow:
### What changed

Add check-reverification-identity lambda

### Why did it change

We need to ensure that if a user starts a reverification journey that they have some form of existing identity to re-verify. This adds a new lamdba that will check if they've got a P2, P1 or operational identity. It's not concerned about CIs, expired VCs, account interventions or pending identities.

### Issue tracking

- [PYIC-7076](https://govukverify.atlassian.net/browse/PYIC-7076)

[PYIC-7076]: https://govukverify.atlassian.net/browse/PYIC-7076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
```diff
@@ -0,0 +1,197 @@
...
+            return new JourneyErrorResponse(
+                            JOURNEY_ERROR_PATH, SC_SERVER_ERROR, FAILED_TO_PARSE_ISSUED_CREDENTIALS)
+                    .toObjectMap();
+        } catch (Exception e) {
+            LOGGER.error(LogHelper.buildErrorMessage("Unhandled lambda exception", e));
+            throw e;
+        }
+    }
+
+    private boolean vcsContainIdentity(List<VerifiableCredential> vcs)
```
Joe-Edwards-GDS: This seems very similar to `CheckExistingIdentityHandler#getStrongestAttainedVotForVtr` (with all valid VOTs and no CIs) - do you think there's any merit in trying to commonise the two? Either by merging the lambdas or pulling out a shared service...

There might be a question about whether we _need_ to check for a profile here at all - in principle we should only have an identity in EVCS if they match(ed) a valid profile. It probably doesn't hurt - and should be fairly speedy (if we don't call CIMIT), so I'm not wildly fussed.

Wynndow: They seem similar because I used it to inform this code! I figured that they were different enough though to stand on their own. Can look at pulling out a common service but I suspect it may be more trouble than it's worth.

I'd rather not combine the two lambdas - check-existing-identity is already pretty complicated and sprawling, and shoehorning another code path through it gives me the fear.

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
```diff
@@ -0,0 +1,186 @@
...
+
+            var reverificationIdentityVot = getReverificationIdentityVot(vcs);
+            if (reverificationIdentityVot.isEmpty()) {
+                ipvSession.setFailureCode(NO_IDENTITY_AVAILABLE);
+                ipvSessionService.updateIpvSession(ipvSession);
+                return NOT_FOUND_RESPONSE;
+            }
+
+            if (GPG45.equals(reverificationIdentityVot.get().vot().getProfileType())) {
+                ipvSession.setTargetVot(reverificationIdentityVot.get().vot());
```
Wynndow: Do we actually care about this? If the previous identity was a P2, do we _need_ the user to prove their identity to a P2 level? Would P1 give us enough confidence to let them use their MFA again?

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
```diff
@@ -0,0 +1,186 @@
...
+            throw e;
+        }
+    }
+
+    private Optional<VotAndProfile> getReverificationIdentityVot(List<VerifiableCredential> vcs)
+            throws ParseException, HttpResponseExceptionWithErrorBody {
+        var gpg45Vcs = VcHelper.filterVCBasedOnProfileType(vcs, GPG45);
+        var gpg45Scores = gpg45ProfileEvaluator.buildScore(gpg45Vcs);
+        var gpg45VcsCorrelated = userIdentityService.areVcsCorrelated(gpg45Vcs);
+        var operationalVcs = VcHelper.filterVCBasedOnProfileType(vcs, OPERATIONAL_HMRC);
```
Joe-Edwards-GDS: Should we do the filtering in the `VotMatcher` - looks quite similar for both usages?

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Joe-Edwards-GDS: IMO this is kinda meaningless - the user doesn't prove to any GPG45 level (e.g. there's no fraud score!)

The only places it's used is to determine the CI threshold, and if we include F2F or KBV, then the scores being requested.

I wonder if we should approach this in a different way - either setting it to P2 automatically, or deliberately leaving it null and handling explicitly in all of those places.

(cc @DanCorderIPV who I think spotted this issue originally)

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
```diff
@@ -0,0 +1,186 @@
...
+                            CURRENT);
+
+            var reverificationIdentityVot = getReverificationIdentityVot(vcs);
+            if (reverificationIdentityVot.isEmpty()) {
+                ipvSession.setFailureCode(NO_IDENTITY_AVAILABLE);
+                ipvSessionService.updateIpvSession(ipvSession);
+                return NOT_FOUND_RESPONSE;
+            }
+
+            if (GPG45.equals(reverificationIdentityVot.get().vot().getProfileType())) {
```
Joe-Edwards-GDS: (regardless of what we do below, I think we still need to do something with the operation profiles!)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/VotMatcher.java
```diff
@@ -0,0 +1,111 @@
...
+
+        for (Vot vot : vots) {
+            if (vot.getProfileType().equals(GPG45) && areGpg45VcsCorrelated) {
+                var matchedGpg45Profile =
+                        achievedWithGpg45Profile(vot, gpg45Vcs, gpg45Scores, contraIndicators);
+                if (matchedGpg45Profile.isPresent()) {
+                    return Optional.of(new VotAndProfile(vot, matchedGpg45Profile.get()));
+                }
+            }
+            if (hasOperationalProfileVc(vot, operationalVcs, contraIndicators)) {
```
Joe-Edwards-GDS: Should this be an else, so we don't try and check this for GPG45 VOTs?

Joe-Edwards-GDS (commented):
I think this feels better, and probably more reusable - what do you think?

(And takes some of the logic out of CheckExistingIdentity, which only feels like a good thing)

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Wynndow: I intentionally had them being passed in as we use the filtered VCs and the gpg45 scores elsewhere in the check-existing-identity lambda (for sending an audit event). I didn't want to have to do the filtering again, or send an audit event with details that weren't acutally used when checking the VOT.
Another option is we send that audit event (`IPV_GPG45_PROFILE_MATCHED`) in the new lambda but I figured it's not required and may be confusing.

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Wynndow: Looking at the reverification journey, we'll never touch any of those things (checking if breaching CI threshold, or determining evidence required for F2F/KBV). So my inclination is to just leave it null and then handle.

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Wynndow: if we do nothing (leave taget VOT as null), what else is there to do?

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Joe-Edwards-GDS: If you're confident that we'll never hit it, then that might be cleaner.

It does feel like we'd be leaving a little bit of a foot-gun if we change it later (e.g. to introduce KBVs), but I suppose we would hit it pretty quickly.

The remaining use-case that I'm not sure about is for the TICF CRI (i.e. `IpvSessionItem#getThresholdVot`)

DanCorderIPV: That sounds reasonable to me - I suspect we just set it to P2 originally because things broke if we didn't. If you're going to check those areas and make them not break then that should be OK.

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Joe-Edwards-GDS: Ok sure - I think the overhead of filtering the list twice is pretty negligible, so I'd go for whichever looks neatest. The GPG45 scores are a bit more awkward (still pretty cheap to calculate, but worse than the filtering) :(

In terms of the audit event, I'm not sure! I could probably justify either way... If you want to retain the control in the calling code (which makes sense), I suppose you could have the method return a triple of Vot/GPG45Profile/GPG45Scores or something. Perhaps call it something like `ProfileMatchingResult` if it makes you feel better 😄

Joe-Edwards-GDS: Overall I quite like `ProfileMatchingResult matchProfile(List<Vot> vots, List<VerifiableCredential> vcs, List<ContraIndicator> cis)` as a good clean signature!

Wynndow: It wasn't so much about the overhead of filtering, more that I wanted to ensure that the _same_ list used to match the VOT was used in the audit event. If we recalculate it for the audit event if feels like there's an opportunity in future for those lists to get differentiated somehow. I'm probably being over cautious. I'll do the triple thing then.

lambdas/check-reverification-identity/src/main/java/uk/gov/di/ipv/core/checkreverificationidentity/CheckReverificationIdentityHandler.java
Wynndow: I'm gonna row back on my idea of leaving it as null. I failed to realise that we'll be checking for CIs as part of the return from a CRI, as well as in TICF, which requires a VOT.
We'll either need a mechanism to skip the CI check (which doesn't feel right), or decide on a threshold.

I don't want to put too much info about scores here, so I think lets talk about this offline.

Joe-Edwards-GDS (approved):
Thanks, I think this is a bit more consistent now

PR #2753, opened 2024-11-28
PYIC-7753: Update reverification coi check

thebauSoftwire:
### What changed
Coi check rules for reverification:
- DOB must match exactly
- 1 char of given name(s)
- 3 chars of family name(s)
Small bit of refactoring + updating/adding unit tests for new logic

### Why did it change
Match coi check rules for reverification to latest requirements

### Issue tracking

- [PYIC-7753](https://govukverify.atlassian.net/browse/PYIC-7753)

[PYIC-7753]: https://govukverify.atlassian.net/browse/PYIC-7753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
+++ b/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
@@ -50,6 +50,7 @@ import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_PARSE_ISSUED_CREDENTIALS;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.IPV_SESSION_NOT_FOUND;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.UNKNOWN_CHECK_TYPE;
+import static uk.gov.di.ipv.core.library.enums.CoiCheckType.FULL_NAME_AND_DOB;
 import static uk.gov.di.ipv.core.library.enums.CoiCheckType.FULL_NAME_WITH_ALLOWANCE_AND_DOB;
 import static uk.gov.di.ipv.core.library.enums.EvcsVCState.CURRENT;
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_CHECK_TYPE;
@@ -149,6 +150,8 @@ public class CheckCoiHandler implements RequestHandler<ProcessRequest, Map<Strin
                                 .areNamesAndDobCorrelated(combinedCredentials);
                         case FULL_NAME_WITH_ALLOWANCE_AND_DOB -> userIdentityService
                                 .areNamesAndDobCorrelatedForReverification(combinedCredentials);
+                        case FULL_NAME_AND_DOB -> userIdentityService.areVcsCorrelated(
+                                combinedCredentials);
                     };
 
             var scopeClaims = clientOAuthSession.getScopeClaims();
@@ -271,7 +274,7 @@ public class CheckCoiHandler implements RequestHandler<ProcessRequest, Map<Strin
             LOGGER.info(
                     LogHelper.buildLogMessage(
                             "Reprove identity flag set - checking full name and DOB"));
-            return FULL_NAME_WITH_ALLOWANCE_AND_DOB;
+            return FULL_NAME_AND_DOB;
         }
         return RequestHelper.getCoiCheckType(request);
     }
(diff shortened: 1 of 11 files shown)


libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/UserIdentityServiceTest.java
```diff
@@ -1694,6 +1695,143 @@ void findIdentityReturnsIdentityClaimForOperationalVcWithNoEvidence() throws Exc
...
+                                    createCredentialWithNameAndBirthDate(
+                                            "Jimbo", "Jones", "1000-01-01")),
+                            vcHmrcMigrationPCL200());
+
+            // Act & Assert
+            assertTrue(userIdentityService.areNamesAndDobCorrelatedForReverification(vcs));
+        }
+
+        @Test
+        void shouldReturnTrueWhenFamilyNamesAreDifferentButMatchWithinCharAllowance()
```
thebauSoftwire, line 1733: I did think about combining these tests into parameterized tests but we might end up losing the readability of the test names which I think is more useful for these cases e.g. minor differences for checking char allowance

local-running/core.local.params.yaml
```diff
@@ -32,7 +32,8 @@ core:
         UserTriggeredIdentityResetConfirmationF2f: "dummy-value"
     coi:
       # Test COI config
-      familyNameChars: 5
+      familyNameChars: 3
```
thebauSoftwire: matching this to common infra

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
```diff
@@ -147,8 +147,8 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
                     switch (checkType) {
                         case GIVEN_OR_FAMILY_NAME_AND_DOB -> userIdentityService
                                 .areNamesAndDobCorrelated(combinedCredentials);
-                        case FULL_NAME_AND_DOB -> userIdentityService.areVcsCorrelated(
-                                combinedCredentials);
+                        case FULL_NAME_AND_DOB -> userIdentityService
```
Joe-Edwards-GDS: Do you think `FULL_NAME_AND_DOB` is still right? I don't think it's necessarily wrong, but it'd be easy to assume that it means the entire name (rather than just a number of chars). On the other hand, it does kinda match the one above in spirit!

Happy to leave, or change, as you prefer.

local-running/core.local.params.yaml
Joe-Edwards-GDS: see slack message

lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java
thebauSoftwire: I've gone for `FULL_NAME_WITH_ALLOWANCE_AND_DOB` to match the above and make it more clear it's not an exact match. I'm pretty rubbish at naming things so let me know if you disagree!

thebauSoftwire:
I've spotted [this](https://github.com/govuk-one-login/ipv-core-back/blob/3ae770a0c55a8a49d0d1b4efa464da12ff5c2d3d/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java#L274) full name check in the CoiHandler when the reprove identity flag is set. This was using the `FULL_NAME_DOB` check too but should they have the same check rules? It doesn't look like it's being used for reverification?

(cc @Wynndow)

Wynndow:
> I've spotted [this](https://github.com/govuk-one-login/ipv-core-back/blob/3ae770a0c55a8a49d0d1b4efa464da12ff5c2d3d/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java#L274) full name check in the CoiHandler when the reprove identity flag is set. This was using the `FULL_NAME_DOB` check too but should they have the same check rules? It doesn't look like it's being used for reverification?
> 
> (cc @Wynndow)

This is used when the user comes to IPV with the `reprove_identity` intervention flag set. I _would_ say that just having the same full check that we use elsewhere would be find. But I also know this change is under some scrutiny from the fraud team, so it may be that we need exact name correlation. Not a very helpful answer, I know.

thebauSoftwire:
> > I've spotted [this](https://github.com/govuk-one-login/ipv-core-back/blob/3ae770a0c55a8a49d0d1b4efa464da12ff5c2d3d/lambdas/check-coi/src/main/java/uk/gov/di/ipv/core/checkcoi/CheckCoiHandler.java#L274) full name check in the CoiHandler when the reprove identity flag is set. This was using the `FULL_NAME_DOB` check too but should they have the same check rules? It doesn't look like it's being used for reverification?
> > (cc @Wynndow)
> 
> This is used when the user comes to IPV with the `reprove_identity` intervention flag set. I _would_ say that just having the same full check that we use elsewhere would be find. But I also know this change is under some scrutiny from the fraud team, so it may be that we need exact name correlation. Not a very helpful answer, I know.

Thanks Chris :) in that case I'll reinstate the previous `FULL_NAME_AND_DOB` for reprove identity journeys just to be safe

Joe-Edwards-GDS (approved):
I think this feels cleaner, what do you reckon?

cc @Wynndow this will clash with the F2F COI PR

PR #2767, opened 2024-12-04
PYIC-7547: Add audit events IPV_APP_MISSING_CONTEXT and IPV_APP_SESSION_RECOVERED

MikeCollingwood:
### What changed

- Add audit events IPV_APP_MISSING_CONTEXT and IPV_APP_SESSION_RECOVERED
- Add logic to get previous ipv session id

### Why did it change

- Need to know previous ipv session id for IPV_APP_SESSION_RECOVERED extension 

### Follow-up PR

- I will follow this work up with a PR to remove `criOAuthSessionId` as a global secondary index of IPV Session as it is unused (even, it wasn't possible to programatically use in its current state).

### Note

- Due to having Chris' commits here, I will have to rebase this like I did for PYIC-6246, before merging.

### Issue tracking

- [PYIC-7547](https://govukverify.atlassian.net/browse/PYIC-7547)

[PYIC-7547]: https://govukverify.atlassian.net/browse/PYIC-7547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MikeCollingwood:
This might cause performance issues

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -476,6 +482,34 @@ private JourneyResponse buildF2FNoMatchResponse(
...
+            Vot lowestGpg45ConfidenceRequested,
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            AuditEventUser auditEventUser)
+            throws ClientOauthSessionNotFoundException, IpvSessionNotFoundException,
+                    VerifiableCredentialException {
+        var criResponseItem =
+                criResponseService.getCriResponseItem(
+                        clientOAuthSessionItem.getUserId(), DCMAW_ASYNC);
+        var previousIpvSessionItem =
+                ipvSessionService.getIpvSessionByCriOAuthSessionId(criResponseItem.getOauthState());
```
MikeCollingwood: This `criOAuthSessionId` is already a Global Secondary Index of the IPV session table. This is `O(logN)`.

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
```diff
@@ -144,64 +120,57 @@ private MobileAppCallbackRequest parseCallbackRequest(APIGatewayProxyRequestEven
...
+        if (criResponse == null || !Objects.equals(criResponse.getOauthState(), criState)) {
             throw new InvalidMobileAppCallbackRequestException(
                     ErrorResponse.CRI_RESPONSE_ITEM_NOT_FOUND);
         }
 
         if (CriResponseService.STATUS_ERROR.equals(criResponse.getStatus())) {
             throw new InvalidCriResponseException(ErrorResponse.ERROR_MOBILE_APP_RESPONSE_STATUS);
         }
+
+        // does this need a STATUS_ABANDON check?
```
DanCorderIPV: This needs addressing?

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
```diff
@@ -144,64 +120,57 @@ private MobileAppCallbackRequest parseCallbackRequest(APIGatewayProxyRequestEven
...
                         ipvSessionItem.getClientOAuthSessionId());
         var userId = clientOAuthSessionItem.getUserId();
-        configService.setFeatureSet(callbackRequest.getFeatureSet());
 
-        // Attach variables to logs
-        LogHelper.attachGovukSigninJourneyIdToLogs(
-                clientOAuthSessionItem.getGovukSigninJourneyId());
-        LogHelper.attachIpvSessionIdToLogs(callbackRequest.getIpvSessionId());
-        LogHelper.attachFeatureSetToLogs(callbackRequest.getFeatureSet());
-        LogHelper.attachComponentId(configService);
```
DanCorderIPV, line 168: Is it OK to drop this logging section?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/IpvSessionServiceTest.java
```diff
@@ -167,6 +167,26 @@ void shouldRetryGettingSessionItemByAccessToken() throws Exception {
...
+    @Test
+    void shouldReturnSessionItemByCriOAuthSessionId() throws IpvSessionNotFoundException {
+        // Arrange
+        var ipvSessionId = SecureTokenHelper.getInstance().generate();
+        var criOAuthSessionId = "56789";
+
+        IpvSessionItem ipvSessionItem = new IpvSessionItem();
+        ipvSessionItem.setIpvSessionId(ipvSessionId);
+
+        when(mockDataStore.getItemByIndex(eq("criOAuthSessionId"), anyString()))
```
DanCorderIPV: Should use `criOAuthSessionId` rather than `anyString()`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -109,6 +109,20 @@ public IpvSessionItem getIpvSessionByAccessToken(String accessToken)
                 });
     }
 
+    public IpvSessionItem getIpvSessionByCriOAuthSessionId(String criOAuthSessionId)
+            throws IpvSessionNotFoundException {
+        return callRunTaskWithBackoff(
+                () -> {
+                    var item = dataStore.getItemByIndex("criOAuthSessionId", criOAuthSessionId);
```
DanCorderIPV: Interesting that we have this index but we don't appear to use it anywhere at the moment

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -586,6 +613,18 @@ private void sendAuditEvent(
                         new AuditRestrictedDeviceInformation(deviceInformation)));
     }
 
+    private void sendAuditEventWithPreviousIpvSessionId(
+            AuditEventTypes auditEventTypes,
+            AuditEventUser auditEventUser,
+            String previousIpvSessionId) {
+        auditService.sendAuditEvent(
+                AuditEvent.createWithoutDeviceInformation(
```
DanCorderIPV: @blakeyp would we want to add device information to this event?

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
```diff
@@ -144,64 +120,57 @@ private MobileAppCallbackRequest parseCallbackRequest(APIGatewayProxyRequestEven
...
+        var criOAuthSessionItem = criOAuthSessionService.getCriOauthSessionItem(criState);
+        if (criOAuthSessionItem == null) {
+            throw new InvalidMobileAppCallbackRequestException(ErrorResponse.INVALID_OAUTH_STATE);
+        }
+
+        // Check IPV session
+        var ipvSessionId = callbackRequest.getIpvSessionId();
+        if (StringUtils.isBlank(ipvSessionId)) {
+            auditService.sendAuditEvent(
+                    AuditEvent.createWithDeviceInformation(
```
DanCorderIPV, line 152: with/without device information should probably be consistent between the two events.

DanCorderIPV (changes requested):

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
blakeyp: Yep we should have device information if it's a user-initiated rather than purely backend process

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -476,6 +476,33 @@ private JourneyResponse buildF2FNoMatchResponse(
...
+            VerifiableCredentialBundle credentialBundle,
+            Vot lowestGpg45ConfidenceRequested,
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            AuditEventUser auditEventUser)
+            throws IpvSessionNotFoundException, VerifiableCredentialException {
+        var criResponseItem =
+                criResponseService.getCriResponseItem(
+                        clientOAuthSessionItem.getUserId(), DCMAW_ASYNC);
+        var previousIpvSessionItem =
+                ipvSessionService.getIpvSessionByCriOAuthSessionId(criResponseItem.getOauthState());
```
blakeyp: Is there a chance this might be missing/expired in which case we need some safety around it?

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
blakeyp: I don't think so, abandon will come as an 'access-denied' async response on the queue for which the onward journey is handled by CheckMobileAppVcReceipt?

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
MikeCollingwood: This is copying what already existed for the existing DCMAW audit events

lambdas/process-mobile-app-callback/src/main/java/uk/gov/di/ipv/core/processmobileappcallback/ProcessMobileAppCallbackHandler.java
MikeCollingwood: No, good spot thank you.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
MikeCollingwood: This is copying the existing DCMAW audit events, but I will add in the device information and make the change to the event catalogue PR

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
MikeCollingwood: It shouldn't be possible to get there without a CriResponseItem, but I will put a null check in for certainty

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
MikeCollingwood: As it turns out the IpvSession table _doesn't have a column for `criOAuthSessionId`._ 

With this, I will replace `criOAuthSessionId` with `clientOAuthSessionId` as GSI to the IPV session table, since `criOAuthSessionId` is unused.

MikeCollingwood:
Tested in dev 🟢

blakeyp (dismissed):

api-tests/features/p2-strategic-app.feature
```diff
@@ -84,6 +84,7 @@ Feature: M2B Strategic App Journeys
     Then I get an OAuth response
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
+    And audit events for 'strategic-app-cross-browser-journey' are recorded [local only]
```
blakeyp: As I understand it, we should pull this out into its own scenario in audit-events.feature per the new approach (is that right @thebauSoftwire?)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -456,6 +456,11 @@ Then(
       const expectedEvents = await getAuditEventsForJourneyType(journeyName);
       const actualEvents = await auditClient.getAuditEvents(this.userId);
 
+      if (journeyName === "strategic-app-cross-browser-journey") {
+        // Find events with no userId (only IPV_APP_MISSING_CONTEXT)
+        actualEvents.push(...(await auditClient.getAuditEvents(undefined)));
+      }
```
blakeyp, line 462: Hmm is there a better way we could handle this (or, is user id even necessary at all?) [not too fussed as it's tests code but if there's an easy improvement!]

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -487,6 +498,44 @@ private JourneyResponse buildF2FNoMatchResponse(
...
+                        clientOAuthSessionItem.getUserId(), DCMAW_ASYNC);
+        if (criResponseItem == null) {
+            return null;
+        }
+        var criOAuthSessionItem =
+                criOAuthSessionService.getCriOauthSessionItem(criResponseItem.getOauthState());
+        if (criOAuthSessionItem == null) {
+            return null;
+        }
+        var previousIpvSessionItem =
```
blakeyp, line 519: Is it possible this might be null and then we blow up?

MikeCollingwood: It shouldn't be possible for it the be null as there needs to have been an IPV session which (is associated to clientOAuthSessionId and) produced the DCMAW VC. If it is null, then the code has failed to behave as we should expect so appropriate to not define a custom exception

api-tests/src/steps/ipv-steps.ts
MikeCollingwood: As far as I can tell, the method to fetch the audit events needs to use userId to differentiate the events from all of the others produced during local running. This exception is annoying but I think it is the simplest way to find the only case where no identifier isn't provided.

The other way, which I am unconvinced is better, is to make device information a uuid, letting us correlate the userId and this to be able to find all of those associated with the same API test. This seems overkill for a single test case.

Here's the code for the MISSING_CONTEXT audit event showing how bare it is:

```java
            auditService.sendAuditEvent(
                    AuditEvent.createWithDeviceInformation(
                            AuditEventTypes.IPV_APP_MISSING_CONTEXT,
                            configService.getParameter(ConfigurationVariable.COMPONENT_ID),
                            null,
                            new AuditRestrictedDeviceInformation(
                                    callbackRequest.getDeviceInformation())));
```

blakeyp (approved):

MikeCollingwood:
See https://github.com/govuk-one-login/ipv-core-back/pull/2902

PR #2770, opened 2024-12-04
PYIC-7797: process candidate identity lambda

thebauSoftwire:
### What changed
consolidates the following lambdas into one:
- check-coi
- evaluate-gpg45-scores
- call-ticf-cri
- store-identity

The inputs of the lambda:
- processIdentityType: ALL, COI, TICF_ONLY, STORE_IDENTITY
- checkType: CoiCheckType (when using check coi)
- identityType: IdentityType (for when storing identity)

The outputs of the lambda:
- Next (successful processing)
- Coi-check-failed
- Error
- vcs-not-correlated
- gpg45-unmet 
- and any of the possible journey events when calling the original `call-ticf-cri` lambda (enhanced-verification, alternate-doc-invalid-dl, alternate-doc-invalid-passport, fail-with-ci)

### Why did it change
To reduce overhead as these lambdas are often called together or some variation.

### Issue tracking


lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,297 @@
...
+                                request,
+                                deviceInformation,
+                                ipAddress)
+                        .toObjectMap();
+                case TICF_ONLY -> processIdentityTicfOnly(
+                                ipvSessionItem,
+                                clientOAuthSessionItem,
+                                deviceInformation,
+                                ipAddress)
+                        .toObjectMap();
```
thebauSoftwire: I'm wasn't sure how to do this better. I wanted to abstract away as much of the processing logic as possible from the handler (which is why I added the `IdentityProcessingService` as an extra layer) so that the only bits here are worried about which processes to call and so keeping it more lightweight. However, passing the same args down multiple layers of functions feels like a code smell. I thought either to combine this inputs into an object (but it didn't feel right bc they were only related by the fact that they are shared inputs) or take away the `IdentityProcessingService` and move the functions there into the handler.

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/service/TicfCriServiceTest.java
```diff
@@ -0,0 +1,288 @@
+package uk.gov.di.ipv.core.processcandidateidentity.service;
```
thebauSoftwire: we'll need a consumer test for this lambda and ticf but can probably be done as a separate pr?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
Joe-Edwards-GDS: Interesting - this is a very functional sort of approach! I'm not quite sure it works in Java though :( - the main challenge is the error handling (as I imagine you might have found).

An alternative might be something more like:
```
if (COI_CHECK_TYPES.contains(processIdentityType)) {
  var coiCheckSuccessful = checkCoiService.isCoiCheckSuccessful(...);
  if (!coiCheckSuccessful) {
    return JOURNEY_COI_CHECK_FAILED;
  }
}

if (GPG45_CHECK_TYPES.contains()) {
  // check gpg45
}

// etc.
```
You can still pull these blocks out into separate methods (if they get complicated) or abstract into another service if they get too complicated.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,297 @@
...
+
+            ipvSessionItem = ipvSessionService.getIpvSession(ipvSessionId);
+            ClientOAuthSessionItem clientOAuthSessionItem =
+                    clientOAuthSessionDetailsService.getClientOAuthSession(
+                            ipvSessionItem.getClientOAuthSessionId());
+
+            String govukSigninJourneyId = clientOAuthSessionItem.getGovukSigninJourneyId();
+            LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
+
+            return switch (processIdentityType) {
```
Joe-Edwards-GDS: I think basically all these steps require access to the session credentials?

I'd suggest loading the credentials at the top level (i.e. here) and then pass in, to avoid fetching them repeatedly (this is one of the benefits of joining these steps together!)

Not sure if the same applies to anything else - but you should watch out for dynamodb records that are being fetched more than once.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/IdentityProcessingService.java
```diff
@@ -0,0 +1,310 @@
...
+            return new JourneyErrorResponse(
+                    JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse());
+        } catch (CiRetrievalException e) {
+            LOGGER.error(LogHelper.buildErrorMessage(FAILED_TO_GET_STORED_CIS.getMessage(), e));
+            return new JourneyErrorResponse(
+                    JOURNEY_ERROR_PATH,
+                    HttpStatus.SC_INTERNAL_SERVER_ERROR,
+                    FAILED_TO_GET_STORED_CIS);
+        } finally {
+            auditService.awaitAuditEvents();
```
Joe-Edwards-GDS: Can this be lifted into the handler class, so it's only called once right at the end?

If you put it here, then we won't be able to proceed with the next step until it's finished (which somewhat nullifies making the audit events async!)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/IdentityProcessingService.java
Joe-Edwards-GDS, line 1: Not a fan of this class, sorry! It doesn't really satisfy the single responsibility principle :(

Instead, I'd suggest that most of the logic in here should either be pulled up into handler, or pushed down into the corresponding service.

As a principle I'd suggest that:
- 'Business logic' - e.g. anything that performs matching logic, should go down into the service layer
- 'Routing logic' - e.g. mapping results to journey events, should go up into the handler

If you think that makes one of the other classes too big, then I'd probably look for a different way of slicing off some functionality... Happy to talk about specific cases!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/ProcessIdentityType.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.di.ipv.core.library.enums;
+
+public enum ProcessIdentityType {
+    ALL,
+    TICF_ONLY,
+    COI,
+    STORE_IDENTITY
```
Joe-Edwards-GDS: So just to get my head around this, we have:
- `ALL`
  - COI, GPG45, TICF, Store Identity
  - Used in most successful proving/update journeys
- `TICF_ONLY`
  - TICF
  - Used in unsuccessful journeys, reuse/migration, and before F2F handoff
- `COI`
  - COI, TICF
  - Used in reverification
- `STORE_IDENTITY`
  - TICF, Store Identity
  - Used in repeat fraud check

We will very shortly need to use the COI/TICF version in the F2F handoff, and we could even do COI/TICF/Store Identity instead (as per my slack comment), which might change things a bit.

Joe-Edwards-GDS: What do you think about relabelling these to talk about the type of 'candidate' identity you're trying to check?

E.g. something like:
```
PROCESS_CANDIDATE_IDENTITY:
    response:
      type: process
      lambda: process-candidate-identity
      lambdaInput:
        identityType: EXISTING / NEW / PENDING / REVERIFICATION / INCOMPLETE
```

Where:

| identityType | Invokes | Examples |
| -- | -- | -- |
| `EXISTING` | TICF | Reuse existing identity |
| `NEW` | COI, GPG45, TICF, Store new identity | New P1/P2, 'reuse with store' |
| `PENDING` | COI, TICF, Store pending identity | before F2F |
| `REVERIFICATION` | COI, TICF | Reverification |
| `INCOMPLETE` | TICF | unsuccessful journeys |

I'm least sure about the last case, since it's not really a 'candidate identity' at that point... 🤔

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/IdentityProcessingService.java
thebauSoftwire: yep that makes sense! I also wasn't really sure about this class but it didn't feel exactly right to put some of the logic into the services since they were returning journey responses

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: thanks for this! This simplified quite a lot of the journey response mapping and helped us get rid of that unneeded IdentityProcessingService layer

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+            ProcessRequest request,
+            CandidateIdentityType processIdentityType,
+            IpvSessionItem ipvSessionItem,
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            String deviceInformation,
+            String ipAddress,
+            List<VerifiableCredential> sessionVcs)
+            throws EvcsServiceException, HttpResponseExceptionWithErrorBody, CiRetrievalException,
+                    UnknownCoiCheckTypeException, CredentialParseException {
+        if (COI_CHECK_TYPES.contains(processIdentityType)) {
```
thebauSoftwire, line 294: see slack for q on handling update journeys

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+import uk.gov.di.ipv.core.library.cristoringservice.CriStoringService;
+import uk.gov.di.ipv.core.library.domain.Cri;
+import uk.gov.di.ipv.core.library.domain.JourneyErrorResponse;
+import uk.gov.di.ipv.core.library.domain.JourneyResponse;
+import uk.gov.di.ipv.core.library.domain.ProcessRequest;
+import uk.gov.di.ipv.core.library.domain.VerifiableCredential;
+import uk.gov.di.ipv.core.library.enums.CandidateIdentityType;
+import uk.gov.di.ipv.core.library.enums.Vot;
+import uk.gov.di.ipv.core.library.exception.EvcsServiceException;
+import uk.gov.di.ipv.core.library.exceptions.*;
```
Joe-Edwards-GDS: Some wildcard imports have snuck in here 👀

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
Joe-Edwards-GDS: I would be happy to add a new `UPDATE` type as well - I think it's only really used in the audit event right?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+            CandidateIdentityType processIdentityType,
+            IpvSessionItem ipvSessionItem,
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            String deviceInformation,
+            String ipAddress,
+            List<VerifiableCredential> sessionVcs)
+            throws EvcsServiceException, HttpResponseExceptionWithErrorBody, CiRetrievalException,
+                    UnknownCoiCheckTypeException, CredentialParseException {
+        if (COI_CHECK_TYPES.contains(processIdentityType)) {
+            var coiCheckType = RequestHelper.getCoiCheckType(request);
```
Joe-Edwards-GDS: Do we need to supply this in the request - can we deduce from the `CandidateIdentityType` (for reverification) and the session (for account intervention)?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+
+        if (matchingGpg45Profile.isEmpty()) {
+            logLambdaResponse("No GPG45 profiles have been met", JOURNEY_GPG45_UNMET);
+            return JOURNEY_GPG45_UNMET;
+        }
+
+        ipvSessionItem.setVot(Vot.fromGpg45Profile(matchingGpg45Profile.get()));
+        logLambdaResponse("A GPG45 profile has been met", JOURNEY_NEXT);
+
+        return JOURNEY_NEXT;
```
Joe-Edwards-GDS: This is fine, but I wonder if it should instead return either `null` or `Optional.empty()` - to make it clear that this is a special 'check successful, carry on' result, and not a real journey event?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                "CI score is breaching threshold - setting VOT to P0"));
+                ipvSessionItem.setVot(Vot.P0);
+
+                return journeyResponse.get();
+            }
+
+            LOGGER.info(LogHelper.buildLogMessage("CI score not breaching threshold"));
+            return JOURNEY_NEXT;
```
Joe-Edwards-GDS: (same as above)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,402 @@
...
+                    ERROR_PROCESSING_TICF_CRI_RESPONSE);
+        }
+    }
+
+    private void logLambdaResponse(String lambdaResult, JourneyResponse journeyResponse) {
+        var message =
+                new StringMapMessage()
+                        .with(LOG_LAMBDA_RESULT.getFieldName(), lambdaResult)
+                        .with(LOG_JOURNEY_RESPONSE.getFieldName(), journeyResponse);
+        LOGGER.info(message);
```
Joe-Edwards-GDS: This `lambdaResponse` log is now a bit misleading, since it's not necessarily the response to the lambda?

If you think there's something interesting to log here, perhaps we should do that as a 'normal' log message...

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/dto/TicfCriDto.java
Joe-Edwards-GDS, line 1: Perhaps worth considering whether this should go into a library - e.g. `libs/ticf-service`, to make it easier to reuse in future?

(And right now, deduplicate between the CallTicfCri lambda - even if that will be deleted after this change)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/exceptions/UnknownProcessIdentityType.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.di.ipv.core.library.exceptions;
+
+import lombok.Getter;
+
+@Getter
+public class UnknownProcessIdentityType extends Exception {
```
Joe-Edwards-GDS: Generally exception classes should end with `Exception`

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/service/TicfCriServiceTest.java
Joe-Edwards-GDS: Can we just move the existing tests? I'm not super familiar with Pact, but if we keep the provider/consumer and test names, will it actually change the pact?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/EvaluateGpg45ScoresService.java
Joe-Edwards-GDS, line 1: @Wynndow has beaten you to the punch a little on this one - have a look at `VotMatcher`. You might need to tweak it slightly, but I think it basically does exactly what you want...

You'll still need to figure out the audit events - Chris decided to leave the audit events out of it, so the calling code could decide. There's some discussion at https://github.com/govuk-one-login/ipv-core-back/pull/2751#discussion_r1869989620 - I'm open to moving the audit event back inside, but have some other mechanism for suppressing it in reverification (or just asking whether we can include it in that journey!)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
```diff
@@ -0,0 +1,204 @@
...
+            String deviceInformation,
+            String ipAddress,
+            List<VerifiableCredential> sessionVcs)
+            throws HttpResponseExceptionWithErrorBody, CredentialParseException,
+                    EvcsServiceException {
+
+        var ipvSessionId = ipvSessionItem.getIpvSessionId();
+        var userId = clientOAuthSession.getUserId();
+        var govukSigninJourneyId = clientOAuthSession.getGovukSigninJourneyId();
+        LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
```
Joe-Edwards-GDS: Don't we do this in `ProcessCandidateIdentityHandler` now?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
```diff
@@ -0,0 +1,204 @@
...
+
+        if (TRUE.equals(clientOAuthSession.getReproveIdentity())) {
+            LOGGER.info(
+                    LogHelper.buildLogMessage(
+                            "Reprove identity flag set - checking full name and DOB"));
+            checkType = ACCOUNT_INTERVENTION;
+        }
+
+        var auditEventUser =
+                new AuditEventUser(userId, ipvSessionId, govukSigninJourneyId, ipAddress);
```
Joe-Edwards-GDS: Would it be any simpler if passed an auditEventUser around to these services? It's cheap to construct so not really fussed, but would save passing around the ipaddress etc. separately.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
```diff
@@ -0,0 +1,204 @@
...
+                checkType,
+                null,
+                auditEventUser,
+                null,
+                null,
+                deviceInformation);
+
+        var oldVcs =
+                evcsService.getVerifiableCredentials(
+                        userId, clientOAuthSession.getEvcsAccessToken(), CURRENT);
```
Joe-Edwards-GDS, line 106: Could we save a round trip here, if we pull this up into `ProcessCandidateIdentity`? I think we also read the user's VCs from EVCS when storing the user's identity.

Not a very big deal if it's awkward, we could always do this as a later refinement :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
```diff
@@ -0,0 +1,204 @@
...
+                sessionVcs,
+                deviceInformation);
+
+        if (isReverification) {
+            setIpvSessionReverificationStatus(
+                    ipvSessionItem,
+                    successfulCheck ? ReverificationStatus.SUCCESS : ReverificationStatus.FAILED);
+            if (!successfulCheck) {
+                ipvSessionItem.setFailureCode(ReverificationFailureCode.IDENTITY_DID_NOT_MATCH);
+                ipvSessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS: I should have spotted this before (or maybe it was a slightly dodgy merge), but we're doubling up the DynamoDB calls here...

It could probably just be:
```
if (isReverification) {
    ipvSession.setReverificationStatus(successfulCheck
            ? ReverificationStatus.SUCCESS
            : ReverificationStatus.FAILED);
    if (!successfulCheck) {
        ipvSession.setFailureCode(ReverificationFailureCode.IDENTITY_DID_NOT_MATCH);
    }
    ipvSessionService.updateIpvSession(ipvSession);
}
```

Joe-Edwards-GDS: (The `updateIpvSession` is this bit that actually updates it in DynamoDB)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/dto/TicfCriDto.java
Joe-Edwards-GDS: Same sort of applies to the others, although they feel slightly less separate. It's fine to leave for now if we're definitely clearing up the existing copies.

Joe-Edwards-GDS (commented):
Not really gone through the tests, but left another pass of comments :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: good spot! thanks :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
thebauSoftwire: ahh thanks for the spot! I've removed the `ipvSessionService.updateIpvSession(ipvSession);` from here though bc there's one in `finally` block in the [handler](https://github.com/govuk-one-login/ipv-core-back/blob/d82922f6a8dacae0307ebe3906f21c54977babda/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java#L227)

thebauSoftwire: actually scratch that, we probably don't want to do a save for every call to this lambda if no updates were made (though it's just extra safety incase we miss an update). I think only store identity and coi are updating the ipv session item (and ticf if user has breaching CIs). TICF only for example is called quite a bit. Let me know what you think though!

Joe-Edwards-GDS: Yeah - it's not the end of the world to update more than strictly necessary (and we definitely don't want to risk dropping an update), but this case seems particularly easy to optimise!

This is the sort of thing where frameworks can help - they'll track when the record is 'dirty' (i.e. modified) and save conditionally, but I don't think we really need anything so complicated.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
thebauSoftwire: Tried time-boxing this and it seems a little more fiddly than I thought so I think this could be done as a separate bit of refinement (this branch is quite long-lived now and worried if I take more time I'm gonna miss new things added to the four lambdas). I'm happy to make a ticket for it myself and pick it up later :)

Joe-Edwards-GDS: Yep, sounds like a good plan :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/dto/TicfCriDto.java
thebauSoftwire: yeah I initially started with trying to make the others into their own library but they felt more tied to this new lambda and so I decided to put them into their own services only within this lambda's directory (part of the reason why this ended up being a big PR instead of a couple smaller ones)

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/service/TicfCriServiceTest.java
thebauSoftwire: ahh good point, thanks!

thebauSoftwire:
More a note for myself when refactoring is more or less done:
Unit tests to add:
- ProcessIdentityHandler:
  - all non-next journey paths
  - all possible errors
  - CandidateIdentityType.UPDATE (for name + address individually)
  - reprove identity case

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/CheckCoiService.java
thebauSoftwire: Here's the ticket for the above: https://govukverify.atlassian.net/browse/PYIC-7909

libs/ticf-cri-service/src/main/java/uk/gov/di/ipv/core/library/ticfcriservice/TicfCriService.java
```diff
@@ -1,13 +1,10 @@
-package uk.gov.di.ipv.core.callticfcri.service;
+package uk.gov.di.ipv.core.library.ticfcriservice;
```
Joe-Edwards-GDS: Not very important, but the `service` feels a bit awkward here (since it's a bit meaningless and duplicated with the class name) - I might go for `uk.gov.di.ipv.core.library.ticf` perhaps putting the service itself in `uk.gov.di.ipv.core.library.ticf.service` if you like?

settings.gradle
```diff
@@ -58,3 +60,5 @@ dependencyResolutionManagement {
 		}
 	}
 }
+include 'libs:ticf-cri-service'
+findProject(':libs:ticf-cri-service')?.name = 'ticf-cri-service'
```
Joe-Edwards-GDS: What's this for?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -0,0 +1,461 @@
...
+        this.evaluateGpg45ScoresService = evaluateGpg45ScoresService;
+        this.storeIdentityService = storeIdentityService;
+        this.ticfCriService = ticfCriService;
+        this.cimitUtilityService = cimitUtilityService;
+    }
+
+    @Override
+    @Tracing
+    @Logging
+    public Map<String, Object> handleRequest(ProcessRequest request, Context context) {
```
Joe-Edwards-GDS, line 182: Chris has added `@Metrics(captureColdStart = true)` to all the lambdas as well now, so we should match (compare with other classes if you rebase)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/EvaluateGpg45ScoresService.java
```diff
@@ -0,0 +1,138 @@
...
+            AuditEventUser auditEventUser)
+            throws HttpResponseExceptionWithErrorBody, ParseException {
+        if (!userIdentityService.checkRequiresAdditionalEvidence(vcs)) {
+            var requestedVotsByStrength =
+                    clientOAuthSessionItem.getParsedVtr().getRequestedVotsByStrengthDescending();
+
+            var gpg45Vots =
+                    requestedVotsByStrength.stream()
+                            .filter(vot -> vot.getProfileType() == ProfileType.GPG45)
+                            .toList();
```
Joe-Edwards-GDS: This is an interesting question actually - is there any value in filtering to GPG45 scores here 🤔

Fine to leave in for now, for parity with the old behaviour, but I think we should perhaps just be asking 'does the user match any valid profile' for their journey.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/EvaluateGpg45ScoresService.java
```diff
@@ -0,0 +1,138 @@
...
+
+            var matchedVot =
+                    votMatcher.matchFirstVot(
+                            gpg45Vots,
+                            vcs,
+                            contraIndicators,
+                            userIdentityService.areVcsCorrelated(
+                                    VcHelper.filterVCBasedOnProfileType(vcs, GPG45)));
```
Joe-Edwards-GDS: Note that we've already checked for VC correlation (in `ProcessCandidateIdentityHandler`) so this should always be true!

I wonder if you should consider either moving the correlation check in here as well, or moving all this logic into `ProcessCandidateIdentityHandler` (once you've removed this and the breaching check it's not doing very much beyond the audit event!)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/EvaluateGpg45ScoresService.java
```diff
@@ -0,0 +1,138 @@
...
+
+            if (matchedVot.isEmpty()) {
+                return Optional.empty();
+            }
+
+            var isBreaching =
+                    contraIndicators != null
+                            && cimitUtilityService.isBreachingCiThreshold(
+                                    contraIndicators, matchedVot.get().vot());
```
Joe-Edwards-GDS: The `VotMatcher` will only match GPG45 profiles which do not breach the threshold, so this check shouldn't be necessary(?)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
Joe-Edwards-GDS: (still there)

Joe-Edwards-GDS (commented):
I haven't been over everything from scratch (trusting my past self...) but I think this is nearly there - and tbh I'm keen to get it merged sooner rather than later, especially if it's not being used in any deployed environments yet.

I've tried to resolve anything I think is fixed, so just the open comments to check 🤞

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -268,20 +281,6 @@ private CoiCheckType getCoiCheckType(
         return CoiCheckType.STANDARD;
     }
 
-    private boolean isUpdateAddressJourney(
```
Joe-Edwards-GDS: FYI @blakeyp this will mean that we do perform a COI check on update address, but I think that's a generally more consistent approach.

LMK if you have any concerns - this won't be used in any deployed environments yet :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -349,10 +337,20 @@ private Map<String, Object> processCandidateThroughJourney(
             ipvSessionService.updateIpvSession(ipvSessionItem);
         }
 
+        if (STORE_IDENTITY_TYPES.contains(processIdentityType)) {
```
Joe-Edwards-GDS, line 340: (cc @thebauSoftwire I think these were in the wrong order - we were storing before doing profile matching or TICF!)

Joe-Edwards-GDS (approved):
Approval for all of Thea/Shiv's commits

lambdas/process-candidate-identity/build.gradle
```diff
@@ -18,10 +17,6 @@ dependencies {
 			project(":libs:evcs-service"),
 			project(":libs:ticf-cri-service")
 
-	aspect libs.powertoolsLogging,
-			libs.powertoolsTracing,
-			libs.aspectj
-
```
thebauSoftwire: out of curiosity, why do we get rid of these? As I think we're using them or are they bundled up with any of the above (libs.bundles.awsLambda maybe?)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: 🤦‍♀️

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -369,38 +376,48 @@ private JourneyResponse getJourneyResponseFromGpg45ScoreEvaluation(
...
+                        areVcsCorrelated);
 
-        if (matchingGpg45Profile.isEmpty()) {
+        if (votResult.isEmpty()) {
             LOGGER.info(LogHelper.buildLogMessage("No GPG45 profiles have been met"));
             return JOURNEY_GPG45_UNMET;
         }
 
-        ipvSessionItem.setVot(Vot.fromGpg45Profile(matchingGpg45Profile.get()));
+        ipvSessionItem.setVot(votResult.get().vot());
```
thebauSoftwire: nice!

lambdas/process-candidate-identity/build.gradle
Joe-Edwards-GDS: I think Chris added something to add it automatically - have a look at `lambdas/build.gradle`

shivanshuit914 (approved):

PR #2776, opened 2024-12-06
PYIC-7761: Add OTEL tracing to AWS SDK and Java Http

Wynndow:
### What changed

Add OTEL tracing to AWS SDK and Java Http

### Why did it change

This adds instrumentation to the AWS SDK and the Java HTTP client, so that we can see more details traces in dynatrace.

This will be more useful on lambdas that are invoked directly by the API gateway, as the trace ID from core-front will be propogated correctly, allowing for a completely joined up view in dynatrace. For step-function invoked lambdas the WC3 trace headers are not propegated correctly, and so a new root span is geneated in the lambda. The AWS and http calls be be traced with that new root as their parent. This may still be useful.

### Link to dynatrace 

https://khw46367.apps.dynatrace.com/ui/apps/dynatrace.distributedtracing/explorer?filter=dt.entity.service.entity.name+%3D+di-ipv-core-front+AND+host.name+%3D+CORE-FRONT-WYNNE+AND+endpoint.name+%3D+%22%2Fcredential-issuer%2Fcallback%22*&cv=u%2Cfalse&sidebar=a%2Cfalse&tf=now-30m%3Bnow&traceId=6641f937895839b5a76522146864ddb1&tt=2024-12-06T13%3A53%3A38.370217000Z&spanId=ae7938ec57b9495c

### Screenshot

Screenshot 2024-12-06 at 14 00 41

### Issue tracking

- [PYIC-7761](https://govukverify.atlassian.net/browse/PYIC-7761)

[PYIC-7761]: https://govukverify.atlassian.net/browse/PYIC-7761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build.gradle
+++ b/lambdas/build.gradle
@@ -9,3 +9,14 @@ allprojects {
 		}
 	}
 }
+
+subprojects {
+	afterEvaluate { subproject ->
+		if (subproject.plugins.hasPlugin('java')) {
+			dependencies {
+				runtimeOnly platform(libs.openTelemetryBom),
+						libs.openTelemetryAwsSdkAutoConfigure
+			}
+		}
+	}
+}
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -4,6 +4,8 @@ import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.AWSXRayRecorder;
 import com.amazonaws.xray.entities.Namespace;
 import com.amazonaws.xray.entities.Subsegment;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.instrumentation.httpclient.JavaHttpClientTelemetry;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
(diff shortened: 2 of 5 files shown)


lambdas/build.gradle
```diff
@@ -9,3 +9,14 @@ allprojects {
 		}
 	}
 }
+
+subprojects {
+	afterEvaluate { subproject ->
+		if (subproject.plugins.hasPlugin('java')) {
```
Wynndow, line 15: This check is neccessary as we have one TypeScript lambda which doesn't have a runtimeOnly configuration

Joe-Edwards-GDS: (Weird that gradle even picks it up as a project, but sure!)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -46,7 +48,10 @@ private TracingHttpClient(HttpClient baseClient) {
     }
 
     public static HttpClient newHttpClient() {
-        return new TracingHttpClient(HttpClient.newHttpClient());
+        return new TracingHttpClient(
+                JavaHttpClientTelemetry.builder(GlobalOpenTelemetry.get())
+                        .build()
+                        .newHttpClient(HttpClient.newHttpClient()));
```
Joe-Edwards-GDS, line 54: Does this cause any issues in environments without Dynatrace - i.e. local/dev. The API tests failed, but not sure if that's really related :/

Joe-Edwards-GDS (approved):
We might want to keep an eye on the performance in case there's any adverse effect, but should be unlikely 🤞

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Wynndow: Doesn't seem to - API tests are green, and it seems happy in my dev env deployed without dynatrace

PR #2777, opened 2024-12-06
PYIC-7811: Reprove identity with F2F

Wynndow:
### What changed

Reprove identity with F2F

### Why did it change

This change allows a user to use the F2F route to reprove their identity after receiving an intervention.

We store the fact that a F2F journey was started after an intervention on the CriResponse item, and then clear that flag once they've completed the journey.

### Issue tracking

- [PYIC-7811](https://govukverify.atlassian.net/browse/PYIC-7811)

[PYIC-7811]: https://govukverify.atlassian.net/browse/PYIC-7811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/p2-reprove-identity.feature
+++ b/api-tests/features/p2-reprove-identity.feature
@@ -1,7 +1,7 @@
 @Build
 Feature: Reprove Identity Journey
 
-    Scenario: User needs to reprove their identity
+    Scenario: User reproves with F2F
         Given the subject already has the following credentials
             | CRI     | scenario                     |
             | dcmaw   | kenneth-driving-permit-valid |
@@ -11,6 +11,28 @@ Feature: Reprove Identity Journey
         Then I get a 'reprove-identity-start' page response
         When I submit a 'next' event
         Then I get a 'page-ipv-identity-document-start' page response
+        When I submit an 'end' event
+        Then I get a 'page-ipv-identity-postoffice-start' page response
+        When I submit a 'next' event
+        Then I get a 'claimedIdentity' CRI response
+        When I submit 'kenneth-current' details to the CRI stub
+        Then I get an 'address' CRI response
+        When I submit 'kenneth-current' details to the CRI stub
+        Then I get a 'fraud' CRI response
+        When I submit 'kenneth-score-2' details to the CRI stub
+        Then I get a 'f2f' CRI response
+        When I submit 'kenneth-driving-permit-valid' details with attributes to the async CRI stub
+            | Attribute          | Values                                          |
+            | evidence_requested | {"scoringPolicy":"gpg45","strengthScore":3} |
+        Then I get a 'page-face-to-face-handoff' page response
+
+        # Return journey after popping out to the Post Office
+        When I start a new 'medium-confidence' journey and return to a 'page-ipv-reuse' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P2' identity
+        And an 'IPV_IDENTITY_ISSUED' audit event was recorded [local only]
 
     Scenario: User needs to reprove their identity with F2F pending
         Given I start a new 'medium-confidence' journey
@@ -25,10 +47,12 @@ Feature: Reprove Identity Journey
         Then I get a 'fraud' CRI response
         When I submit 'kenneth-score-2' details to the CRI stub
         Then I get a 'f2f' CRI response
-        When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub
+        When I submit 'kenneth-driving-permit-valid' details with attributes to the async CRI stub
             | Attribute          | Values                                          |
             | evidence_requested | {"scoringPolicy":"gpg45","strengthScore":3} |
         Then I get a 'page-face-to-face-handoff' page response
+
+        # Users been to the Post Office but sadly now has an account intervention
         When I start a new 'medium-confidence' journey with reprove identity
         Then I get a 'reprove-identity-start' page response
         When I submit a 'next' event
--- a/api-tests/features/p2-reprove-identity.feature
+++ b/api-tests/features/p2-reprove-identity.feature
@@ -1,5 +1,32 @@
 @Build
 Feature: Reprove Identity Journey
+    Scenario: User reproves identity
+        Given the subject already has the following credentials
+            | CRI     | scenario                     |
+            | dcmaw   | kenneth-driving-permit-valid |
+            | address | kenneth-current              |
+            | fraud   | kenneth-score-2              |
+        When I start a new 'medium-confidence' journey with reprove identity
+        Then I get a 'reprove-identity-start' page response
+        When I submit a 'next' event
+        Then I get a 'page-ipv-identity-document-start' page response
+        When I submit an 'appTriage' event
+        Then I get a 'dcmaw' CRI response
+        When I submit 'kenneth-passport-valid' details to the CRI stub
+        Then I get a 'page-dcmaw-success' page response
+        When I submit a 'next' event
+        Then I get an 'address' CRI response
+        When I submit 'kenneth-current' details to the CRI stub
+        Then I get a 'fraud' CRI response
+        When I submit 'kenneth-score-2' details to the CRI stub
+        Then I get a 'page-ipv-success' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P2' identity
+        And an 'IPV_IDENTITY_ISSUED' audit event was recorded [local only]
+        And an 'IPV_ACCOUNT_INTERVENTION_START' audit event was recorded [local only]
+        And an 'IPV_ACCOUNT_INTERVENTION_END' audit event was recorded [local only]
 
     Scenario: User reproves with F2F
         Given the subject already has the following credentials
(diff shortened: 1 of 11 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -929,4 +934,20 @@ private void sendProfileMatchedAuditEvent(
                         new AuditRestrictedDeviceInformation(deviceInformation));
         auditService.sendAuditEvent(auditEvent);
     }
+
+    private boolean isReprovingWithF2f(CriResponseItem f2fRequest) {
+        return f2fRequest != null && f2fRequest.isReproveIdentity();
+    }
+
+    private void removeReproveIdentityFlag(boolean isF2FComplete, CriResponseItem f2fResponseItem) {
+        if (isF2FComplete && f2fResponseItem.isReproveIdentity()) {
```
Wynndow: Rather than removing the flag like this, I did consider adding in a new state to the journey map to reset the session, including the CRI response, like we do elsewhere. This proved difficult though due to requiring the VCs to still be in the session store for a reuse journey.
If we wanted to do this we could add a new reset type that purely resets the CRI response. This approach also adds more complexity to the journey maps.
This way seemed easier and faster - but happy to be corrected.

api-tests/features/p2-reprove-identity.feature
```diff
@@ -11,7 +11,30 @@ Feature: Reprove Identity Journey
...
+            | evidence_requested | {"scoringPolicy":"gpg45","strengthScore":3} |
+        Then I get a 'page-face-to-face-handoff' page response
 
+        # Return journey after popping out to the Post Office
+        When I start a new 'medium-confidence' journey and return to a 'page-ipv-reuse' page response
+        When I submit a 'next' event
+        Then I get an OAuth response
+        When I use the OAuth response to get my identity
+        Then I get a 'P2' identity
+        And an 'IPV_IDENTITY_ISSUED' audit event was recorded [local only]
```
Joe-Edwards-GDS, line 62: Perhaps ahead of the game, but the audit events that really matter here are `IPV_ACCOUNT_INTERVENTION_START` and `IPV_ACCOUNT_INTERVENTION_END`

api-tests/features/p2-reprove-identity.feature
```diff
@@ -1,7 +1,7 @@
 @Build
 Feature: Reprove Identity Journey
 
-    Scenario: User needs to reprove their identity
+    Scenario: User reproves with F2F
```
Joe-Edwards-GDS, line 31: Do you think there's value in having a non-F2F example as well? I'm not sure it's that useful, but it feels a bit strange to have F2F only 🤔 

Now we've added COI, I wonder if we should also have some tests to cover a failing COI check too?

api-tests/features/p2-reprove-identity.feature
```diff
@@ -11,7 +11,30 @@ Feature: Reprove Identity Journey
...
+        Then I get a 'fraud' CRI response
+        When I submit 'kenneth-score-2' details to the CRI stub
+        Then I get a 'f2f' CRI response
+        When I submit 'kenneth-driving-permit-valid' details with attributes to the async CRI stub
+            | Attribute          | Values                                          |
+            | evidence_requested | {"scoringPolicy":"gpg45","strengthScore":3} |
+        Then I get a 'page-face-to-face-handoff' page response
 
+        # Return journey after popping out to the Post Office
+        When I start a new 'medium-confidence' journey and return to a 'page-ipv-reuse' page response
```
Joe-Edwards-GDS: Does this need to specify 'with reprove identity'? Otherwise it's as if the intervention was lifted

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: What's the scenario you're trying to avoid by doing this? I guess the inverse risk is that if something blows up afterwards (before the intervention is cleared) then they'll have to start again.

If we add a check for `vcs.isPendingEvcsIdentity()` when skipping, will that help avoid the issue? I.e. so that it only skips if the identity is still pending.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/CriResponseItem.java
```diff
@@ -27,6 +27,7 @@ public class CriResponseItem implements PersistenceItem {
     private String status;
     private long ttl;
     private List<String> featureSet;
+    private boolean reproveIdentity;
```
Joe-Edwards-GDS, line 30: What happens for existing records - will this just default to `false`?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: Do you mean the `isF2FComplete` flag. I feel like I did have a good reason before the weekend, but now I can't for the life of me remember what it was...

As this change checks that `isReproveIdentity` flag on the response item, I'm not sure if also checking `isPendingIdentity` is needed. Although it's made we wonder if we could just check that rather than updating the response item...

api-tests/features/p2-reprove-identity.feature
Wynndow: Do they exist already?

api-tests/features/p2-reprove-identity.feature
Wynndow: There's an API test on the as yet to be raised PR for PYIC-7810 which should address the COI checks.

We could add another test for a non-f2f journey. It won't test loads more, but it is another path (and probably the more commonly used one).

api-tests/features/p2-reprove-identity.feature
Wynndow: Yup

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/CriResponseItem.java
Wynndow: Yeah - if there's no value in DDB, it'll come back as `false`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Wynndow: No - because we want to know if the journey was _started_ in response to an intervention - not if the user happens to come back to core with an pre-existing pending response and a new intervention.

api-tests/features/p2-reprove-identity.feature
Joe-Edwards-GDS: Yep!

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
Joe-Edwards-GDS: Yeah exactly :)

I think maybe the scenario that you're worried about is like:
- User gets account intervention
- User reproves identity using F2F
- User's intervention is cleared
- User gets _another_ intervention
- The pending record from the first reprove is still hanging around (with `isReproveIdentity=true`), and the intervention gets immediately cleared

I don't think it's a massive risk tbh, but nice to solve if we can.

My theory with adding `isPendingEvcsIdentity` is that it'll mean you only get to complete the intervention if you're _actually_ returning from F2F (not that you just happen to have a response record kicking around).

Wynndow: yeah, i think that was what i was worried about. But you're right, adding the pending check should solve it. I'll fix it up.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -599,4 +605,23 @@ private void sendProfileMatchedAuditEvent(
...
     }
+
+    private boolean isReprovingWithF2f(
+            CriResponseItem f2fRequest, VerifiableCredentialBundle vcBundle) {
+        // does the user have a F2F response item that was created in response to an intervention,
+        // and they're returning to core with a pending identity
+        return f2fRequest != null && f2fRequest.isReproveIdentity() && vcBundle.isPendingIdentity();
+    }
+
+    private void removeReproveIdentityFlag(CriResponseItem f2fResponseItem) {
```
Joe-Edwards-GDS: Do we still need to do this if we're enforcing that it's only skipped on the first return after a F2F reprove?

Wynndow: gah. I'm not sure why my brain is struggling with this so much...

I suppose we only care if the user manages to achieve a profile - as we wouln't want any further interventions to be skipped. And if they *do* achieve a profile, then they will no longer have a pending identity. So won't skip...

And if they *don't* achieve a profile, then they'll be on an identity proving journey anyway.

So no?

Joe-Edwards-GDS (approved):
This is super confusing tbh, we'll need to see how fraud use it in practice anyway

PR #2786, opened 2024-12-10
PYIC-7761: Recreate http client with tracing

Wynndow:
### What changed

Recreate http client with tracing

### Why did it change

We introduced logic to recreate the base client in our tracing http client every hour. The new base client wasn't being created with instrumentation though, and so we lost things like spans in Dynatrace. This updates to recreate the client in the same way we create in initially.

### Issue tracking

- [PYIC-7761](https://govukverify.atlassian.net/browse/PYIC-7761)

[PYIC-7761]: https://govukverify.atlassian.net/browse/PYIC-7761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -48,19 +48,47 @@ public class TracingHttpClient extends HttpClient {
     }
 
     public static HttpClient newHttpClient() {
-        return new TracingHttpClient(
-                JavaHttpClientTelemetry.builder(GlobalOpenTelemetry.get())
-                        .build()
-                        .newHttpClient(HttpClient.newHttpClient()));
+        return new TracingHttpClient(getInstrumetedHttpClient());
     }
 
-    // Synchronize to prevent race conditions when multiple threads attempt to recreate client
-    private synchronized void recreateBaseClientIfNecessary() {
-        if (creationTime.plus(MAX_CLIENT_AGE).isBefore(Instant.now())) {
-            LOGGER.info("Recreating base HTTP client due to age");
-            this.baseClient = HttpClient.newHttpClient();
-            this.creationTime = Instant.now(); // Reset creation time
+    @Override
+    public <T> HttpResponse<T> send(
+            HttpRequest request, HttpResponse.BodyHandler<T> responseBodyHandler)
+            throws IOException, InterruptedException {
+        // PYIC-7590: Recreate the HTTP client, to avoid HTTP/2 connection issues.
+        // Check for client is 1hour old, and recreate if older.
+        recreateBaseClientIfNecessary();
+        try {
+            return sendWithTracing(request, responseBodyHandler);
+        } catch (IOException e) {
+            // We see occasional HTTP/2 GOAWAY messages from AWS when connections last
+            // for a long time (>1 hour). Retry them once, to recreate the connection.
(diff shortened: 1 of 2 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -108,6 +136,15 @@ public Optional<Executor> executor() {
         return baseClient.executor();
     }
 
+    // Synchronize to prevent race conditions when multiple threads attempt to recreate client
+    private synchronized void recreateBaseClientIfNecessary() {
+        if (creationTime.plus(MAX_CLIENT_AGE).isBefore(Instant.now())) {
+            LOGGER.info("Recreating base HTTP client due to age");
+            this.baseClient = getInstrumetedHttpClient();
```
Wynndow: This line is the actual functional change here - most of the rest of the diff is just moving public methods above private ones.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -177,4 +174,10 @@ private static void addRequestInformation(Subsegment subsegment, HttpRequest req
 
         subsegment.putHttp("request", requestInformation);
     }
+
+    private static HttpClient getInstrumetedHttpClient() {
```
Joe-Edwards-GDS: Typo `instrumented`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -48,21 +48,49 @@ private TracingHttpClient(HttpClient baseClient) {
...
+        // Check for client is 1hour old, and recreate if older.
+        recreateBaseClientIfNecessary();
+        try {
+            return sendWithTracing(request, responseBodyHandler);
+        } catch (IOException e) {
+            // We see occasional HTTP/2 GOAWAY messages from AWS when connections last
+            // for a long time (>1 hour). Retry them once, to recreate the connection.
+            // In the build environment we see connection resets for idle connections in
+            // the pool. Retrying uses a different connection.
+            if (HTTP_ERROR_MESSAGES_TO_RETRY.contains(e.getMessage())) {
```
Joe-Edwards-GDS: While you're here... one thing I spotted was that your change to include Connection Reset changed this from a substring match to an exact match, which means the GOAWAYs never actually get retried.

I'm not sure retrying helps very much (and recreating the client seems to pretty much eliminate the issue), but ideally we should either make the logic work or remove it.

(Doesn't have to be part of this PR)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Wynndow: Bloomin' words...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Wynndow: Interesting. I'll update it now and pull out the retry stuff.

Joe-Edwards-GDS (approved):

PR #2800, opened 2024-12-11
PYIC-7799: Create traffic when deploying

Wynndow:
**This can't be merged without https://github.com/govuk-one-login/devplatform-deploy/pull/2188**

### What changed

Create traffic when deploying

### Why did it change

This updates the api-tests image to be able to run in a mode that will just generate traffic in the build env. This is to ensure we have traffic to inform the canarys.

It will poll for the status of the deploy action, and stop sending traffic when it's no longer in progress.

### Issue tracking

- [PYIC-7799](https://govukverify.atlassian.net/browse/PYIC-7799)

[PYIC-7799]: https://govukverify.atlassian.net/browse/PYIC-7799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/secure-pipeline/run-tests.sh
+++ b/api-tests/secure-pipeline/run-tests.sh
@@ -2,10 +2,20 @@
 
 set -eo pipefail
 
+get_current_status() {
+  aws codepipeline get-pipeline-state --name "$1" \
+    | jq -r '.stageStates[] | select(.stageName == "Deploy") | .actionStates[] | select(.actionName == "Deploy") | .latestExecution.status'
+}
+
+generate_traffic() {
+  while true; do
+    echo "Running @TrafficGeneration tests"
+    npm run test:build -- --profile trafficGeneration --tags '@TrafficGeneration'
+  done
+}
+
 # Ensure the test report dir exists
 [ -e "$TEST_REPORT_ABSOLUTE_DIR" ] && mkdir -p "$TEST_REPORT_ABSOLUTE_DIR"
-
-echo "Running API tests against the build environment"
 
 CORE_BACK_INTERNAL_API_KEY=$(aws secretsmanager get-secret-value --secret-id CoreBackInternalTestingApiKey | jq -r .SecretString)
 export CORE_BACK_INTERNAL_API_KEY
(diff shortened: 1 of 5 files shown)


api-tests/secure-pipeline/run-tests.sh
```diff
@@ -19,23 +29,48 @@ export CRI_STUB_GEN_CRED_API_KEY
 MANAGEMENT_TICF_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/credentialIssuers/ticf/connections/stub/apiKey | jq -r .SecretString)
 export MANAGEMENT_TICF_API_KEY
 
-MANAGEMENT_CIMIT_STUB_API_KEY=$(aws ssm get-parameter --name /tests/core-back-build/cimit_api_key | jq -r .Parameter.Value)
+MANAGEMENT_CIMIT_STUB_API_KEY=$(aws secretsmanager get-secret-value --secret-id /build/core/cimitManagementApi/apiKey | jq -r .SecretString)
```
Wynndow, line 32: this secret has been created in build. I'm not sure why we're pulling this secret specifically from ssm rather than secrets manager, but better to be consistant. also the traffic codebuild doesn't have perms to pull this ssm param

api-tests/secure-pipeline/run-tests.sh
Joe-Edwards-GDS, line 1: I do love some bash...

If this gets much more complicated we might want to consider a rewrite in JS or Python! Not really worth it now.

Joe-Edwards-GDS (approved):
Looks reasonable to me - I take it you've tested to some extent?

Probably worth keeping an eye on build when we merge :)

Wynndow:
> Looks reasonable to me - I take it you've tested to some extent?

Yep, deployed a test pipeline to the dev account and been running it in there.

api-tests/secure-pipeline/run-tests.sh
Wynndow: yeah we could do - we'd have to invoke it with the bash script though, as the entrypoint invoked by the pipeline is `run-tests.sh`

PR #2803, opened 2024-12-12
Pyic 7650 international address fraud route

DanCorderIPV:
### What changed

Allow international address users to get an M1C profile

### Why did it change

So international address users can get a profile

### Issue tracking

- [PYIC-7650](https://govukverify.atlassian.net/browse/PYIC-7650)

[PYIC-7650]: https://govukverify.atlassian.net/browse/PYIC-7650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -256,6 +266,36 @@ private Optional<Gpg45Profile> findMatchingGpg45Profile(
...
     }
 
+    private boolean allowM1C(List<VerifiableCredential> vcs) {
+        var fraudVcList = vcs.stream().filter(vc -> vc.getCri() == Cri.EXPERIAN_FRAUD).toList();
+
+        if (fraudVcList.isEmpty()) {
+            return false;
+        }
+        if (fraudVcList.size() > 1) {
+            throw new InvalidParameterException("VCs contain more than one fraud VC");
```
DanCorderIPV: Would this be possible after a 6 month fraud check? Would we have both the new and the old fraud VCs here?

DanCorderIPV:
One question is whether we want to update `BuildCriOAuthRequestHandler.getEvidenceRequestForF2F()` to consider M1C. I think leaving it alone for now is safe (but might be "wrong") as we need them to use a passport for M1C anyway and we don't send a minimum validity score to F2F - only the strength score.

api-tests/data/cri-stub-requests/fraud/kenneth-score-0/evidence.json
Joe-Edwards-GDS, line 1: This is not just score 0 - it's a 'no applicable authoritative source' response. If you get a score 0 without that, then you're not eligible for an M1C.

api-tests/package.json
```diff
@@ -14,19 +14,19 @@
...
-    "@cucumber/cucumber": "^11.1.0",
-    "@govuk-one-login/data-vocab": "^1.7.2",
-    "@types/eslint__js": "^8.42.3",
-    "dotenv": "^16.4.5",
-    "eslint": "^8.57.0",
-    "eslint-config-prettier": "^9.1.0",
-    "jose": "^5.9.2",
-    "jsonschema": "^1.4.1",
+    "@cucumber/cucumber": "11.1.0",
+    "@govuk-one-login/data-vocab": "1.8.0",
```
Joe-Edwards-GDS: Note that there is now 1.9.0, which includes `addressRegion` - not required for this PR necessarily!

api-tests/package.json
```diff
@@ -14,19 +14,19 @@
     "lint-fix": "eslint --fix . && prettier . --write"
   },
   "devDependencies": {
-    "@cucumber/cucumber": "^11.1.0",
```
Joe-Edwards-GDS, line 17: Thanks... IMO it doesn't make much difference using absolute version numbers (the lockfile specifies which exact version you're actually using anyway), but we may as well be consistent.

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -224,6 +225,10 @@ private Optional<Gpg45Profile> findMatchingGpg45Profile(
 
             for (Vot requestedVot : gpg45Vots) {
                 var profiles = requestedVot.getSupportedGpg45Profiles();
+                if (requestedVot == Vot.P2 && userIdentityService.allowM1C(vcs)) {
+                    profiles = new ArrayList<>(profiles);
+                    profiles.add(Gpg45Profile.M1C);
+                }
```
Joe-Edwards-GDS: Do you think it might be clearer to invert this? I.e. add to `Vot.P2` and then filter out when matching against a profile.

Something like:
```
var allowZeroFraudScore = userIdentityService.allowZeroFraudScore(vcs);
var profiles = requestedVot.getSupportedGpg45Profiles(allowZeroFraudScore);
```
(You can decide whether it lives here or in `Vot.java`)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -108,6 +111,36 @@ public UserIdentity generateUserIdentity(
         return userIdentityBuilder.build();
     }
 
+    public boolean allowM1C(List<VerifiableCredential> vcs) {
```
Joe-Edwards-GDS: Is this the right place for this? It appears to be effectively static, I wonder if it suits `VcHelper` better - or its own helper?

I'd also suggest that it's phrased more around allowing profiles that don't require a fraud check. Otherwise you're setting yourself up for a rename when we start supporting the equivalent low/high confidence profiles.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/UserIdentityService.java
```diff
@@ -108,6 +111,36 @@ public UserIdentity generateUserIdentity(
...
+        var fraudVcList = vcs.stream().filter(vc -> vc.getCri() == Cri.EXPERIAN_FRAUD).toList();
+
+        if (fraudVcList.isEmpty()) {
+            return false;
+        }
+        if (fraudVcList.size() > 1) {
+            throw new InvalidParameterException("VCs contain more than one fraud VC");
+        }
```
Joe-Edwards-GDS: Could you sidestep these checks by just iterating on the fraud VCs with `anyMatch`?

```
return vcs.stream()
        .filter(vc -> vc.getCri() == EXPERIAN_FRAUD)
        .anyMatch(vc -> {
            ...
        });
```

If you're worried about cases where a user gets multiple fraud VCs, we could still check - perhaps just logging a warning and returning false to avoid blocking the user?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -362,6 +362,9 @@ states:
       enhanced-verification:
         targetJourney: FAILED
         targetState: FAILED
+      fraud-fail-with-no-ci:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS: Can a user still enter an international address if they visit address CRI without the `internationalAddress` context?

We should probably continue to route them to `EVALUATE_SCORES` and let that fail them if they are not eligible.

We'll want to do that when we introduce the fraud resilience work anyway - since it could happen on any journey.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -384,6 +387,9 @@ states:
       enhanced-verification:
         targetJourney: F2F_HAND_OFF
         targetState: START
+      fraud-fail-with-no-ci:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS, line 394: These other states for P1 are more interesting... They won't be valid for M1C, but could be for L1B.

Feels fine to leave this as failed for now, and we can update if/when we support L1B.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -395,6 +395,9 @@ states:
       enhanced-verification:
         targetJourney: FAILED
         targetState: FAILED
+      fraud-fail-with-no-ci:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS: As above, I think this could/should be EVALUATE_SCORES?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/repeat-fraud-check.yaml
```diff
@@ -226,6 +226,9 @@ states:
       enhanced-verification:
         targetJourney: FAILED
         targetState: FAILED_CONFIRM_DETAILS
+      fraud-fail-with-no-ci:
+        targetJourney: FAILED
+        targetState: FAILED
```
Joe-Edwards-GDS: This is a slightly interesting one - my assumption is that we _should_ allow them to get a repeat M1C if Experian isn't usable.

(Not relevant to international address, but will be for fraud resilience)

Joe-Edwards-GDS (commented):
Do we also need to update `CheckExistingIdentity` to correctly handle international address users in repeat fraud checks, or is that covered elsewhere?

api-tests/package.json
DanCorderIPV: I agree, but I believe it's the policy to use fixed numbers

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -371,11 +369,12 @@ private JWEObject signEncryptJar(
     private EvidenceRequest getEvidenceRequestForF2F(
             List<VerifiableCredential> vcs, Vot requestedVot) {
         var gpg45Scores = gpg45ProfileEvaluator.buildScore(vcs);
-        List<Gpg45Scores> requiredEvidences =
+        var isFraudCheckUnavailable = VcHelper.isFraudCheckUnavailable(vcs);
+        var requiredEvidences =
                 gpg45Scores.calculateGpg45ScoresRequiredToMeetAProfile(
-                        requestedVot.getSupportedGpg45Profiles());
+                        requestedVot.getSupportedGpg45Profiles(isFraudCheckUnavailable));
```
Joe-Edwards-GDS: This won't do anything for now (because F2F can never meet M1C), but it might be useful if we introduce L1B

api-tests/features/p2-international-journey.feature
```diff
@@ -43,10 +43,11 @@ Feature: P2 App journey
...
       | context   | "international_user" |
     Then I get a 'fraud' CRI response
-    When I submit 'kenneth-score-2' details to the CRI stub
+    When I submit 'kenneth-no-applicable' details to the CRI stub
     Then I get a 'page-ipv-success' page response
     When I submit a 'next' event
     Then I get an OAuth response
     When I use the OAuth response to get my identity
     Then I get a 'P2' identity
+    And I get 'fraud-check-unavailable' return codes
```
Joe-Edwards-GDS, line 54: (missed this in the other review)

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
```diff
@@ -221,11 +222,12 @@ private Optional<Gpg45Profile> findMatchingGpg45Profile(
 
             var gpg45Vots =
                     requestedVotsByStrength.stream()
-                            .filter(vot -> vot.getSupportedGpg45Profiles() != null)
+                            .filter(vot -> vot.getProfileType() == ProfileType.GPG45)
                             .toList();
 
             for (Vot requestedVot : gpg45Vots) {
-                var profiles = requestedVot.getSupportedGpg45Profiles();
+                var isFraudCheckUnavailable = VcHelper.isFraudCheckUnavailable(vcs);
```
MikeCollingwood: This can be moved outside of this for loop.

The variable could be more easily understood if it were named as (the inverse) `isFraudScoreRequired` like the suggestion for the parameter reference in `getSupportedGpg45Profiles`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
```diff
@@ -33,7 +36,13 @@ public enum Vot {
         this.profileType = profileType;
     }
 
-    public List<Gpg45Profile> getSupportedGpg45Profiles() {
+    public List<Gpg45Profile> getSupportedGpg45Profiles(boolean isFraudCheckUnavailable) {
```
MikeCollingwood: Maybe the parameter could be the inverse `isFraudScoreRequired`? Imo, it would fit the method better, e.g. the use in `fromGpg45Profile`

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
```diff
@@ -38,6 +38,9 @@ nestedJourneyStates:
         exitEventToEmit: next-passport
       enhanced-verification:
         exitEventToEmit: next-passport
+      fraud-fail-with-no-ci:
+        targetJourney: FAILED
+        targetState: FAILED
```
MikeCollingwood, line 43: Are we sending the user to evaluate their scores only if they have completed all of the CRIs up to that point because we can't (yet) tell CRIs about the now-stricter requirements for M1C?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -192,4 +193,26 @@ public static boolean isExpiredFraudVc(VerifiableCredential vc) {
         var now = Instant.now();
         return nbf.plus(expiryPeriod, ChronoUnit.HOURS).isBefore(now);
     }
+
+    public static boolean isFraudCheckUnavailable(List<VerifiableCredential> vcs) {
```
MikeCollingwood, line 197: Might be clearer to name this `isFraudCheckApplicable` and inverse the logic?

MikeCollingwood (commented):
Looks like this covers all the places it needs to. I had a few refactors you can pick and choose from. I also, had a query on when to send to the user to evaluate gpg45 scores in the journey map

lambdas/evaluate-gpg45-scores/src/main/java/uk/gov/di/ipv/core/evaluategpg45scores/EvaluateGpg45ScoresHandler.java
Joe-Edwards-GDS: Ah yes - definitely want to move it out of the loop.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/enums/Vot.java
Joe-Edwards-GDS: I debated the naming of this parameter a bit! I tried to remove the negative and make it `isFraudCheckAvailable`, but that's not _quite_ right in the case where we don't know yet.

`isFraudScoreRequired` perhaps makes the intent clearer (but we'd probably still keep the VC helper method the same)... happy to rename

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
Joe-Edwards-GDS: See comment above, I did try doing this, but then I was worried about the uncertain case, i.e. where you don't yet know.

It feels more natural at that point to say that the fraud check is 'not unavailable', rather than 'is available'? Semantics perhaps...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
Joe-Edwards-GDS: Something like that yes - to attain M1C you need verification score 3, which we can't get from KBV.

There's no point sending the user there if we know they'll fail.

(In future we might be able to set up a more 'self-serve' mechanism where the KBV CRI rejects the request if they know they can't meet it!)

MikeCollingwood: Is it similar for F2F or is that for a separate reason?

MikeCollingwood (dismissed):
Looks good!

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/VotMatcherTest.java
```diff
@@ -137,4 +140,46 @@ void shouldNotMatchGpg45VotIfRequiresAdditionalEvidence() throws Exception {
...
+                        SUPPORTED_VOTS_BY_DESCENDING_STRENGTH, vcs, List.of(), true);
+
+        // Assert
+        verify(mockGpg45ProfileEvaluator).getFirstMatchingProfile(GPG_45_SCORES, expectedProfiles);
+        assertEquals(
+                Optional.of(new VotMatchingResult(P2, Gpg45Profile.M1C, GPG_45_SCORES)), votMatch);
+    }
+
+    @Test
+    void shouldNotMatchM1cIfFraudCheckAvailable() throws Exception {
```
DanCorderIPV, line 167: This test seems wrong now

libs/verifiable-credentials/src/test/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelperTest.java
```diff
@@ -244,6 +246,73 @@ private static Stream<Arguments> UnsuccessfulTestCases() {
         return Stream.of(
                 Arguments.of("VC missing evidence", vcPassportM1aMissingEvidence()),
                 Arguments.of("Failed passport VC", vcPassportM1aFailed()),
-                Arguments.of("Failed fraud check", vcExperianFraudFailed()));
+                Arguments.of("Failed fraud check", vcExperianFraudEvidenceFailed()));
+    }
+
+    @Test
+    void shouldAllowM1cForApplicableAuthoritativeSourceFailedFraudCheck() {
```
DanCorderIPV: These tests could do with renaming

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/service/VotMatcherTest.java
MikeCollingwood: I think it's that it stubs returning M1C from getFirstMatchingProfile when provided a list of achievableProfiles, which exclude M1C. I have updated this. The test still correctly checks for the expected profiles, which are determined by the new logic

DanCorderIPV (commented):
Looks good to me

MikeCollingwood (approved):

Joe-Edwards-GDS (approved):

PR #2804, opened 2024-12-12
PYIC-7746: Embedded metrics for CRI completion

Wynndow:
### What changed

Embedded metrics for CRI completion

### Why did it change
This change starts sending custom metrics to track the rate of users being sent to CRIs and then returning. We can compare these two metrics to help us understand if the completion rates for different CRIs is changing. And alert on it in future.

We're using the logging mechanism, rather than sending custom metrics directly with the cloudwatch SDK to avoid having to have a new AWS client that needs to be initialised. The logs may also be useful for more refined queries.

This uses the embedded metrics mechanism, rather than creating a metric log filter, as it's more explicit and will be harder to break accidentally (by changing a log message for example).

### Screenshots - generated by running the selenium tests

Screenshot 2024-12-12 at 17 42 57
---
Screenshot 2024-12-12 at 17 46 06

### Issue tracking

- [PYIC-7746](https://govukverify.atlassian.net/browse/PYIC-7746)

[PYIC-7746]: https://govukverify.atlassian.net/browse/PYIC-7746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
+++ b/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
@@ -16,7 +16,6 @@ import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport
 import uk.gov.di.ipv.core.library.cimit.exception.CiPostMitigationsException;
 import uk.gov.di.ipv.core.library.cimit.exception.CiPutException;
 import uk.gov.di.ipv.core.library.cimit.exception.CiRetrievalException;
-import uk.gov.di.ipv.core.library.config.EnvironmentVariable;
 import uk.gov.di.ipv.core.library.criapiservice.CriApiService;
 import uk.gov.di.ipv.core.library.criapiservice.exception.CriApiException;
 import uk.gov.di.ipv.core.library.cristoringservice.CriStoringService;
@@ -31,6 +30,7 @@ import uk.gov.di.ipv.core.library.exceptions.IpvSessionNotFoundException;
 import uk.gov.di.ipv.core.library.exceptions.UnrecognisedVotException;
 import uk.gov.di.ipv.core.library.exceptions.VerifiableCredentialException;
 import uk.gov.di.ipv.core.library.helpers.ApiGatewayResponseGenerator;
+import uk.gov.di.ipv.core.library.helpers.EmbeddedMetricHelper;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.helpers.RequestHelper;
 import uk.gov.di.ipv.core.library.helpers.StepFunctionHelpers;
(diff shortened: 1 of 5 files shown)


lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -145,24 +147,6 @@ public ProcessCriCallbackHandler() {
     @Logging(clearState = true)
     public APIGatewayProxyResponseEvent handleRequest(
             APIGatewayProxyRequestEvent input, Context context) {
-        // temporary logging to check which tracing headers are being used by dynatrace
```
Wynndow, line 148: This was leftover from the dynatrace traces ticket. Whoops.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/EmbeddedMetricHelper.java
```diff
@@ -0,0 +1,59 @@
...
+
+        private final String name;
+    }
+
+    public static void createMetrics(
+            Map<Dimension, String> dimensions,
+            Map<Metric, Double> metrics,
+            String govukSigninJourneyId) {
+        METRICS_LOGGER.setNamespace(CORE_BACK_EMBEDDED_METRICS_NAMESPACE);
+        METRICS_LOGGER.putProperty(
```
Wynndow, line 43: properties don't get included in the metric, so it's safe to have values with high cardinality. They are included in the log message that's created.

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
Joe-Edwards-GDS: !

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/EmbeddedMetricHelper.java
```diff
@@ -0,0 +1,61 @@
...
+    @Getter
+    @AllArgsConstructor
+    @ExcludeFromGeneratedCoverageReport
+    public enum Dimension {
+        CRI("cri");
+
+        private final String name;
+    }
+
+    public static void createMetrics(
```
Joe-Edwards-GDS, line 38: A couple of minor points:
- I'm not sure 'create' is the right verb here - the AWS terminology is 'put', but perhaps something like 'record' or 'log' would be clearest?
- Does it need to be plural? At the moment we only record one at a time, so feels a bit YAGNI

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/EmbeddedMetricHelper.java
Joe-Edwards-GDS: I wonder if there's any value in stealing the Log4J context directly - e.g. with something like `ThreadContext.getContext()`, so you automatically get all the other common log properties?

Might risk key clashes, and Journey ID is definitely the most useful, so this is also fine!

Joe-Edwards-GDS (approved):
Happy with this, but a few minor tweaks that we could make - and interested in whether powertools would be of any benefit

Joe-Edwards-GDS:
One other thought I had - does the EMF thing auto-create the metric in CloudWatch? Do we need any CloudFormation to create the metrics before we record them?

Kinda convenient if not, but harder to manage them that way 🤔

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/EmbeddedMetricHelper.java
Wynndow: yeah record works - I left it as plural as we could be recoring multiple, but you're right currently only one. I'll make it singular and future person can update if needed.

Wynndow:
> One other thought I had - does the EMF thing auto-create the metric in CloudWatch? Do we need any CloudFormation to create the metrics before we record them?
> 
> Kinda convenient if not, but harder to manage them that way 🤔

No, it creates them for you. No CloudFormation required.

PR #2821, opened 2024-12-18
PYIC-7532: Use journeyContext to determine internationalAddress user

AmritSidhu:
### What changed

Use journeyContext to determine internationalAddress user
Routing needs to change based on the actual address of the user not just based on the featureFlag
Checked there shouldnt be any overlap with the existing journeyContext for rfc. <- follow up ticket to find a better solution

### Issue tracking

- [PYIC-7532](https://govukverify.atlassian.net/browse/PYIC-7532)

[PYIC-7532]: https://govukverify.atlassian.net/browse/PYIC-7532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MikeCollingwood (approved):
The setting of the journeyContext is in the only place where we know they are an international address user.

It doesn't overlap with `rfc`, since we don't triage for int address users in that journey, therefore don't know when we would want to show them int address-specific content.

PR #2825, opened 2024-12-18
PYIC-7446: Set AWS_EMF_ENVIRONMENT in local

Wynndow:
### What changed

Set AWS_EMF_ENVIRONMENT in local

### Why did it change

When running locally, if you don't override the EMF environment to local, you get a whole bunch of grim looking TCP errors. The metrics are trying to be sent to a sink that doesn't exist.

### Issue tracking

- [PYIC-7446](https://govukverify.atlassian.net/browse/PYIC-7446)

[PYIC-7446]: https://govukverify.atlassian.net/browse/PYIC-7446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/local-running/build.gradle
+++ b/local-running/build.gradle
@@ -44,10 +44,6 @@ dependencies {
 			project(":libs:oauth-key-service")
 }
 
-run.doFirst {
-	environment 'AWS_XRAY_CONTEXT_MISSING', 'IGNORE_ERROR'
-}
-
 java {
 	sourceCompatibility = JavaVersion.VERSION_17
 	targetCompatibility = JavaVersion.VERSION_17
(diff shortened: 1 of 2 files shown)


shivanshuit914 (dismissed):

local-running/build.gradle
```diff
@@ -60,6 +60,10 @@ application {
 	]
 }
 
+tasks.named('run', JavaExec) {
+	environment 'AWS_EMF_ENVIRONMENT', 'Local'
```
Joe-Edwards-GDS, line 60: Should this happen in the same place as `AWS_XRAY_CONTEXT_MISSING` above?

(Not sure which place is better, but it seems weird that they aren't the same?)

Wynndow: Oh, I missed that. No idea which is better. I'll unify them though.

Joe-Edwards-GDS (approved):

PR #2854, opened 2025-01-09
PYIC-7832: Update journey map with new page

AmritSidhu:
### What changed

pyi-cri-escape and pyi-suggest-other-options templates will now be combined into a single dynamic template with URL 

photo-id-security-questions-find-another-way This also remove the no-F2F versions of this page and route to error in the case F2F is disabled

### Issue tracking

- [PYIC-7832](https://govukverify.atlassian.net/browse/PYIC-7832)

[PYIC-7832]: https://govukverify.atlassian.net/browse/PYIC-7832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -336,6 +336,7 @@ states:
         targetState: START
         checkIfDisabled:
           f2f:
+            targetJourney: TECHNICAL_ERROR
             targetState: ERROR
 
   # Common pages
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
@@ -316,10 +316,10 @@ states:
         targetJourney: INELIGIBLE
         targetState: INELIGIBLE_SKIP_MESSAGE
 
-  PYI_CRI_ESCAPE:
+  PYI_KBV_DROPOUT_PHOTO_ID:
     response:
       type: page
-      pageId: pyi-cri-escape
+      pageId: photo-id-security-questions-find-another-way
     events:
       appTriage:
         targetState: APP_DOC_CHECK_PYI_ESCAPE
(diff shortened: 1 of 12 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -328,25 +328,10 @@ states:
       f2f:
         targetJourney: F2F_HAND_OFF
         targetState: START
-
-  PYI_CRI_ESCAPE_NO_F2F:
```
blakeyp: For P2 we should probably keep the old states initially (and remove in a follow-up PR) to stage this out in case there any in flight journeys in these states as the deploy goes through

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -492,26 +464,13 @@ states:
...
         targetState: START
       enhanced-verification:
         targetState: MITIGATION_KBV_FAIL_NO_PHOTO_ID
-        checkIfDisabled:
-          f2f:
-            targetState: MITIGATION_02_OPTIONS
-            auditEvents:
-              - IPV_MITIGATION_START
-            auditContext:
-              mitigationType: enhanced-verification
```
blakeyp: It's probably not explicit in the ticket sorry, but as we're doing this could we also handle `f2f` disabled in `PYI_KBV_DROPOUT_NO_PHOTO_ID` and `MITIGATION_KBV_FAIL_NO_PHOTO_ID` same as photo id?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -358,31 +343,18 @@ states:
...
-        checkIfDisabled:
-          f2f:
-            targetState: PYI_CRI_ESCAPE_NO_F2F
+        targetState: PYI_KBV_DROPOUT_PHOTO_ID
       next:
         targetJourney: EVALUATE_SCORES
         targetState: START
       enhanced-verification:
-        targetState: MITIGATION_02_OPTIONS_WITH_F2F
+        targetState: PYI_KBV_DROPOUT_PHOTO_ID
```
blakeyp: Hmm don't we still need a separate state for this, as the page will be context dependent - like `PYI_KBV_DROPOUT_NO_PHOTO_ID` and `MITIGATION_KBV_FAIL_NO_PHOTO_ID` for no photo id?

blakeyp (approved):

PR #2855, opened 2025-01-09
PYIC-7797 Journey routing for combined lambda

Joe-Edwards-GDS:
### What changed

Added the combined lambda to the step function and routing behind a feature flag.

N.B. does not yet include failed journeys or inherited identities.

### Why did it change

Ability to start using the new combined end-of-journey lambda function.

### Issue tracking

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -78,6 +78,7 @@ import static uk.gov.di.ipv.core.library.enums.CandidateIdentityType.NEW;
 import static uk.gov.di.ipv.core.library.enums.CandidateIdentityType.PENDING;
 import static uk.gov.di.ipv.core.library.enums.CandidateIdentityType.REVERIFICATION;
 import static uk.gov.di.ipv.core.library.enums.CandidateIdentityType.UPDATE;
+import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_CHECK_TYPE;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_COI_CHECK_FAILED_PATH;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_ERROR_PATH;
 import static uk.gov.di.ipv.core.library.journeys.JourneyUris.JOURNEY_NEXT_PATH;
@@ -295,6 +296,9 @@ public class ProcessCandidateIdentityHandler
                     CredentialParseException, ParseException {
         if (COI_CHECK_TYPES.contains(processIdentityType)) {
             var coiCheckType = getCoiCheckType(processIdentityType, clientOAuthSessionItem);
+            LOGGER.info(
+                    LogHelper.buildLogMessage("Performing COI check")
+                            .with(LOG_CHECK_TYPE.getFieldName(), coiCheckType.name()));
             var isCoiCheckSuccessful =
                     checkCoiService.isCoiCheckSuccessful(
                             ipvSessionItem,
@@ -310,6 +314,7 @@ public class ProcessCandidateIdentityHandler
         }
 
         if (PROFILE_MATCHING_TYPES.contains(processIdentityType)) {
+            LOGGER.info(LogHelper.buildLogMessage("Performing profile evaluation"));
             var journey =
                     getJourneyResponseForProfileMatching(
                             ipvSessionItem,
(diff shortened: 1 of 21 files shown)


lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -332,6 +334,16 @@ private Map<String, Object> processCandidateThroughJourney(
                             auditEventUser);
 
             if (journey != null) {
+                // We still store a pending identity - it might be mitigating an existing CI
```
Joe-Edwards-GDS, line 343: This is a slight deviation from existing behaviour - we will store the pending identity regardless of the CI failure, whereas we know that only certain cases (e.g. enhanced verification) are actually able to succeed.

I don't think it should be a significant issue, as a) for now TICF won't issue any meaningful CIs, and b) the pending identity will be ignored anyway

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -178,6 +178,9 @@ states:
             targetState: INELIGIBLE
       returnToRp:
         targetState: RETURN_TO_RP
+        checkFeatureFlag:
+          processCandidateIdentity:
+            targetState: PROCESS_INCOMPLETE_IDENTITY
```
Joe-Edwards-GDS, line 183: This is a slight change - @blakeyp at the moment all these routes go directly to `RETURN_TO_RP`, which means they won't invoke TICF.

I was under the impression that _all_ routes should invoke TICF (even unsuccessful ones) - but if you think that's wrong then we can keep this as-is.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/service/VotMatcher.java
```diff
@@ -30,11 +30,6 @@ public class VotMatcher {
     private final Gpg45ProfileEvaluator gpg45ProfileEvaluator;
     private final CimitUtilityService cimitUtilityService;
 
-    public VotMatcher(
-            UserIdentityService userIdentityService, Gpg45ProfileEvaluator gpg45ProfileEvaluator) {
-        this(userIdentityService, gpg45ProfileEvaluator, null);
```
Joe-Edwards-GDS, line 35: Not sure what this constructor was for, but the `null` broke things!

Joe-Edwards-GDS:
I've checked that this passes all the API tests (with the flag enabled)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -386,15 +406,13 @@ private JourneyResponse getJourneyResponseForProfileMatching(
...
-            LOGGER.info(LogHelper.buildLogMessage("No GPG45 profiles have been met"));
-            return JOURNEY_GPG45_UNMET;
+            return JOURNEY_PROFILE_UNMET;
         }
 
         ipvSessionItem.setVot(votResult.get().vot());
         ipvSessionService.updateIpvSession(ipvSessionItem);
 
         if (votResult.get().vot().getProfileType() == ProfileType.GPG45) {
-            LOGGER.info(LogHelper.buildLogMessage("A GPG45 profile has been met"));
```
Joe-Edwards-GDS, line 397: (Duplicates logging in the VotMatcher)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -332,12 +340,24 @@ private Map<String, Object> processCandidateThroughJourney(
                             auditEventUser);
 
             if (journey != null) {
+                // We still store a pending identity - it might be mitigating an existing CI
+                if (PENDING.equals(processIdentityType)) {
+                    LOGGER.info(LogHelper.buildLogMessage("Storing identity"));
```
thebauSoftwire, line 345: Would we end up storing the identity twice? Since type PENDING is also caught by the if statement on line 359

thebauSoftwire (commented):
Thanks! Looks good, just one question on storing identity for pending identity type.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: actually ignore this I missed the return statement below

thebauSoftwire (approved):

PR #2865, opened 2025-01-13
PYIC-7926: audit events api tests

thebauSoftwire:
### What changed
- adds api tests for audit events
  - this looks like a new features file which contains key journeys to test for all possible audit events
  - we compare these against a set of pre-defined audit events in `api-tests/data/audit-events`
  - new audit events will need to be added to these as well as a new journey if required

### Why did it change
There wasn't a pre-defined way of testing for audit events in the api tests - only testing them in some scenarios and for only a handful of events. This PR tests adds api tests for all possible events.

### Issue tracking

- [PYIC-7926](https://govukverify.atlassian.net/browse/PYIC-7926)

[PYIC-7926]: https://govukverify.atlassian.net/browse/PYIC-7926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/data/audit-events/alternate-doc-mitigation-journey.json
+++ b/api-tests/data/audit-events/alternate-doc-mitigation-journey.json
@@ -67,7 +67,7 @@
       ],
       "successful": false,
       "isUkIssued": true,
-      "age": 59
+      "age": "type[number]"
     },
     "restricted": {
       "device_information": {}
--- a/api-tests/data/audit-events/alternate-doc-mitigation-journey.json
+++ b/api-tests/data/audit-events/alternate-doc-mitigation-journey.json
@@ -1,0 +1,104 @@
+[
+  {
+    "event_name": "IPV_JOURNEY_START",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "reprove_identity": false,
+      "vtr": ["P2"]
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_SUBJOURNEY_START",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "journey_type": "NEW_P2_IDENTITY"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_REDIRECT_TO_CRI",
+    "component_id": "https://identity.local.account.gov.uk",
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_CRI_AUTH_RESPONSE_RECEIVED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "error_code": "access_denied",
+      "error_description": "No error description provided"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_REDIRECT_TO_CRI",
+    "component_id": "https://identity.local.account.gov.uk",
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_VC_RECEIVED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "iss": "https://driving-license-cri.stubs.account.gov.uk",
+      "evidence": [
+        {
+          "activityHistoryScore": 0,
+          "ci": ["NEEDS-ALTERNATE-DOC"],
+          "failedCheckDetails": [
+            {
+              "checkMethod": "data",
+              "identityCheckPolicy": "published"
+            }
+          ],
+          "strengthScore": 3,
+          "type": "IdentityCheck",
+          "validityScore": 0
+        }
+      ],
+      "successful": false,
+      "isUkIssued": true,
+      "age": 59
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_CORE_CRI_RESOURCE_RETRIEVED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "cri": "drivingLicence",
+      "type": "vc"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_MITIGATION_START",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "mitigation_type": "invalid-dl"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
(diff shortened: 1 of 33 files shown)


api-tests/src/utils/object-matchers.ts
```diff
@@ -0,0 +1,42 @@
+export interface ObjectPartialEqualityResult {
+  isPartiallyEqual: boolean;
+  errorMessage?: string;
+}
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export const comparePartialEqualityBetweenObjects = (
+  actualObject: Record<string, any>,
+  expectedObject: Record<string, any>,
```
thebauSoftwire, line 20: this method is a bit weird but I couldn't think of a better way to do it...this method should also work for arrays I think bc they are treated like `{0: something, 1: something}` bc of the "any" in the type expression

lambdas/call-dcmaw-async-cri/src/main/java/uk/gov/di/ipv/core/calldcmawasynccri/service/DcmawAsyncCriService.java
```diff
@@ -40,9 +40,9 @@ public class DcmawAsyncCriService {
     private final CriOAuthSessionService criOAuthSessionService;
 
     @ExcludeFromGeneratedCoverageReport
-    public DcmawAsyncCriService(ConfigService configService) {
+    public DcmawAsyncCriService(ConfigService configService, AuditService auditService) {
         this.configService = configService;
-        this.auditService = AuditService.create(configService);
+        this.auditService = auditService;
```
Joe-Edwards-GDS, line 45: Nice - I think this is really a bug (cc @DanCorderIPV )

We need to use the handler-level audit service or it won't get awaited properly in the lambda call. I'm glad the local runner is exposing this 😄  - I deliberately made it not work if you don't await it.

api-tests/src/utils/object-matchers.ts
Joe-Edwards-GDS: Looks like this is actually a feature in node 23 called `partialDeepStrictEqual`: https://nodejs.org/api/assert.html#assertpartialdeepstrictequalactual-expected-message

Perhaps you could link to that documentation, and suggest switching to it once available? (And maybe use it for inspiration)

api-tests/src/utils/object-matchers.ts
```diff
@@ -0,0 +1,42 @@
...
+): ObjectPartialEqualityResult => {
+  for (const [expectedKey, expectedValue] of Object.entries(expectedObject)) {
+    if (actualObject[expectedKey] === undefined) {
```
Joe-Edwards-GDS: I think you might want `if (expectedKey in actualObject)`?

Depends whether you consider `undefined` to be a valid value of the partial match!

api-tests/src/utils/object-matchers.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (!isEqual.isPartiallyEqual) {
+        return isEqual;
+      }
+      continue;
+    }
+
+    if (expectedValue !== actualObject[expectedKey]) {
+      return {
+        isPartiallyEqual: false,
+        errorMessage: `Expected ${expectedValue} for field ${expectedKey} but got ${actualObject[expectedKey]}`,
```
Joe-Edwards-GDS: This is perfectly sufficient, but if you want extra points, you could pass a 'prefix' argument into the recursion, so the message is something like `Expected "expected" for field foo.bar.baz but got "actual"`. Mainly useful if you end up with lots of nested properties with the same key.

api-tests/features/audit-events.feature
Joe-Edwards-GDS, line 1: Although it might not cover any new events, I wonder if it's worth including a couple more 'happy path' journeys - e.g. 'P2 with app' and 'Reuse'?

If it's lots of extra effort then don't worry too much. This still covers everything and gives us a framework to extend in future, which is the most important thing :)

api-tests/data/audit-events/new-identity-f2f-journey.json
```diff
@@ -0,0 +1,386 @@
...
+          ],
+          "strengthScore": 3,
+          "type": "IdentityCheck",
+          "validityScore": 2,
+          "verificationScore": 3
+        }
+      ],
+      "successful": true,
+      "isUkIssued": true,
+      "age": 59
```
Joe-Edwards-GDS: Is this going to break in a year's time?

api-tests/features/audit-events.feature
Joe-Edwards-GDS: Looks like you have `new-identity-p2-journey.json` but no corresponding test?

Joe-Edwards-GDS (commented):
Thanks, this looks good I think

api-tests/features/audit-events.feature
thebauSoftwire: ahh yeah I had it but thought that a new identity via f2f would cover more events but I'll add both P2 with app and reuse journeys too :)

api-tests/src/utils/object-matchers.ts
thebauSoftwire: great idea! Much easier to debug now, thanks :)

api-tests/data/audit-events/new-identity-f2f-journey.json
thebauSoftwire: good point, I've added a type check here instead so we can still check the property exists and validate the value somehow

api-tests/src/utils/object-matchers.ts
```diff
@@ -3,40 +3,113 @@ export interface ObjectPartialEqualityResult {
...
+    throw new Error(`${expectedType} is unsupported.`);
+  }
+};
+
+const formatKeyHistory = (keyHistory: string[]): string => {
+  const formattedKeyHistory: string[] = [keyHistory[0]];
+  for (let i = 1; i < keyHistory.length; i++) {
+    const key = keyHistory[i];
+    if (isNumeric(key)) {
+      formattedKeyHistory.push(`[${key}]`);
```
Joe-Edwards-GDS, line 107: shiny

api-tests/data/audit-events/alternate-doc-mitigation-journey.json
```diff
@@ -67,7 +67,7 @@
       ],
       "successful": false,
       "isUkIssued": true,
-      "age": 59
+      "age": "type[number]"
```
Joe-Edwards-GDS, line 70: This is neat :)

If we do switch it up in future, it's probably not too bad to omit it - we have unit tests for the detail of this field, but a nice feature to have.

Joe-Edwards-GDS (approved):
Thanks!

PR #2867, opened 2025-01-13
PYIC-7797: route failed + reverification journeys to use process-candidate-identity-lambda

thebauSoftwire:
### What changed
Added routing to failed and reverification journeys to use new process-candidate-identity lambda

### Why did it change
We made a new lambda to consolidate a bunch of other lambdas, this PR uses that lambda on journeys that don't yet use it yet.

### Issue tracking

### DIFF
--- a/api-tests/data/audit-events/delete-pending-f2f-journey.json
+++ b/api-tests/data/audit-events/delete-pending-f2f-journey.json
@@ -204,7 +204,7 @@
     "event_name": "IPV_IDENTITY_STORED",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "identity_type": "PENDING",
+      "identity_type": "pending",
       "vot": null
     },
     "restricted": {
--- a/api-tests/data/audit-events/delete-pending-f2f-journey.json
+++ b/api-tests/data/audit-events/delete-pending-f2f-journey.json
@@ -153,6 +153,27 @@
     }
   },
   {
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_START",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_END",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD",
+      "successful": true
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
     "event_name": "IPV_VC_RECEIVED",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
@@ -180,6 +201,17 @@
     }
   },
   {
+    "event_name": "IPV_IDENTITY_STORED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "identity_type": "PENDING",
+      "vot": null
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
     "event_name": "IPV_REDIRECT_TO_CRI",
     "component_id": "https://identity.local.account.gov.uk",
     "restricted": {
@@ -192,17 +224,6 @@
     "extensions": {
       "cri": "f2f",
       "type": "pending"
-    },
-    "restricted": {
-      "device_information": {}
-    }
-  },
-  {
-    "event_name": "IPV_IDENTITY_STORED",
-    "component_id": "https://identity.local.account.gov.uk",
-    "extensions": {
-      "identity_type": "pending",
-      "vot": null
     },
     "restricted": {
       "device_information": {}
--- a/api-tests/data/audit-events/new-identity-f2f-journey.json
+++ b/api-tests/data/audit-events/new-identity-f2f-journey.json
@@ -153,6 +153,27 @@
     }
   },
   {
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_START",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_END",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD",
+      "successful": true
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
     "event_name": "IPV_VC_RECEIVED",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
@@ -180,6 +201,17 @@
     }
   },
   {
+    "event_name": "IPV_IDENTITY_STORED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "identity_type": "PENDING",
+      "vot": null
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
     "event_name": "IPV_REDIRECT_TO_CRI",
     "component_id": "https://identity.local.account.gov.uk",
     "restricted": {
@@ -192,17 +224,6 @@
     "extensions": {
       "cri": "f2f",
       "type": "pending"
-    },
-    "restricted": {
-      "device_information": {}
-    }
-  },
-  {
-    "event_name": "IPV_IDENTITY_STORED",
-    "component_id": "https://identity.local.account.gov.uk",
-    "extensions": {
-      "identity_type": "pending",
-      "vot": null
     },
     "restricted": {
       "device_information": {}
@@ -329,11 +350,62 @@
     }
   },
   {
-    "event_name": "IPV_IDENTITY_STORED",
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_START",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "identity_type": "new",
-      "vot": "P2"
+      "type": "STANDARD"
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_END",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD",
+      "successful": true
+    },
+    "restricted": {
+      "newName": [
+        {
+          "nameParts": [
+            {
+              "type": "GivenName",
+              "value": "KENNETH"
+            },
+            {
+              "type": "FamilyName",
+              "value": "DECERQUEIRA"
+            }
+          ]
+        }
+      ],
+      "newBirthDate": [
+        {
+          "value": "1965-07-08"
+        }
+      ],
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_GPG45_PROFILE_MATCHED",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "gpg45Profile": "M1B",
+      "gpg45Scores": {
+        "evidences": [
+          {
+            "strength": 3,
+            "validity": 2
+          }
+        ],
+        "activity": 1,
+        "fraud": 2,
+        "verification": 3
+      },
+      "vcTxnIds": []
     },
     "restricted": {
       "device_information": {}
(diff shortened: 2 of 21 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -58,23 +85,64 @@ states:
...
+        targetState: NO_MATCH_PAGE
+      alternate-doc-invalid-dl:
+        targetState: NO_MATCH_PAGE
+      alternate-doc-invalid-passport:
+        targetState: NO_MATCH_PAGE
+      fail-with-ci:
+        targetState: NO_MATCH_PAGE
+      error:
+        targetJourney: TECHNICAL_ERROR
+        targetState: ERROR_NO_TICF
```
thebauSoftwire, line 145: it looks like there's a lot of duplication but it was only with the equivalent states e.g. CRI_TICF_BEFORE_NO_MATCH for this one. I tried to use parent states to reduce duplication for now but didn't think it was worth it if eventually we'll get rid of the equivalent states

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -209,6 +403,27 @@ states:
...
+      alternate-doc-invalid-dl:
+        targetState: NO_MATCH_PAGE_NINO
+      alternate-doc-invalid-passport:
+        targetState: NO_MATCH_PAGE_NINO
+      fail-with-ci:
+        targetState: NO_MATCH_PAGE_NINO
+      error:
+        targetJourney: TECHNICAL_ERROR
+        targetState: ERROR_NO_TICF
+
```
thebauSoftwire, line 464: on line 499, we also call the evaluate-gpg45-scores lambda by itself when we fail an update journey with a valid id e.g. we don't get a ci. Currenlty, we don't have an identityType which supports this so maybe we add another one called `UPDATE_FAILED` which just calls evaluate gpg45?

thebauSoftwire:
We also have operational profile migration/reuse journeys which call only the call-ticf-cri lambda. The only identityType which suits this is the INCOMPLETE one which doesn't it because it's not an incomplete identity. We could instead add a OPERATIONAL identityType?

local-running/core.local.params.yaml
```diff
@@ -389,7 +389,7 @@ core:
     sqsAsync: true
     kidJarHeaderEnabled: true
     drivingLicenceAuthCheck: false
-    processCandidateIdentity: false
+    processCandidateIdentity: true
```
thebauSoftwire: still having difficulty running the api tests consistently locally, fetch is now failing at initialise ipv session :( so testing here and leaving this comment as a note to self to remove this when ready to be merged

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -33,6 +39,20 @@ states:
               - IPV_USER_DETAILS_UPDATE_END
             auditContext:
               successful: false
+            checkFeatureFlag:
+              processCandidateIdentity:
+                targetState: PROCESS_INCOMPLETE_IDENTITY_COULD_NOT_CONFIRM_WITH_DELETION_INVALID_ID
```
Joe-Edwards-GDS: Do we need the 'with deletion' any more? - It's fine, but these super long state names feel a bit awkward!

Sadly, I don't think we can update the existing states without a two-step deploy

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: Yeah agree - it's only worth using parent states if it's between multiple of the new states.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -114,6 +203,27 @@ states:
...
+        identityType: INCOMPLETE
+    events:
+      next:
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
+      enhanced-verification:
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
+      alternate-doc-invalid-dl:
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
+      alternate-doc-invalid-passport:
+        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_WITH_DELETION_VALID_ID
```
Joe-Edwards-GDS: (Happy to leave since it's unchanged but this looks a bit odd - if you've got an outstanding CI here, surely the ID is not valid 🤔)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -152,6 +283,27 @@ states:
...
+        identityType: INCOMPLETE
+    events:
+      next:
+        targetState: COULD_NOT_UPDATE_DETAILS_PAGE
+      enhanced-verification:
+        targetState: COULD_NOT_UPDATE_DETAILS_PAGE
+      alternate-doc-invalid-dl:
+        targetState: COULD_NOT_UPDATE_DETAILS_PAGE
+      alternate-doc-invalid-passport:
+        targetState: COULD_NOT_UPDATE_DETAILS_PAGE
```
Joe-Edwards-GDS, line 338: (ditto)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS: Hmm, good question - I guess we sort of want to use 'EXISTING' identity? Except that won't do GPG45, and will do TICF 😅 

We could probably make it do GPG45 if we want (it shouldn't really do any harm in reuse...), I'm not sure about the duplicate TICF call - it's almost a good thing because the identity is different to the first call. We could try asking the fraud team what they'd prefer to see...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
thebauSoftwire: ah didn't notice this that's a good point 👀

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
thebauSoftwire: I've changed the new ones but haven't updated the others. For my own learning, what would a two step deployment look like for updating the existing states and why do we need it?

Joe-Edwards-GDS: Let's say you had:

```
FIRST:
  events:
    next:
      targetState: SECOND

SECOND:
  events:
    next:
      targetState: THIRD

THIRD:
   ...
```
and you want to rename SECOND...

You'd need to do it in two steps:
```
FIRST:
  events:
    next:
      targetState: SECOND_NEW

SECOND_NEW:
  events:
    next:
      targetState: THIRD

SECOND:
  events:
    next:
      targetState: THIRD

THIRD:
   ...
```

and then
```
FIRST:
  events:
    next:
      targetState: SECOND_NEW

SECOND_NEW:
  events:
    next:
      targetState: THIRD

THIRD:
   ...
```

The reason is that users might be in `SECOND` when you deploy the first change, so the journey map still needs to know about it (otherwise they'll get 'unexpected state' when trying to continue).

If it's a very quick state (e.g. a process state) then the window for this is only a few seconds, and we _might_ be able to get away without the two steps, but if it's a CRI state, then it could cause quite a few errors.

thebauSoftwire: thanks for explaining :) I've made the changes now to do the two-step deployment

thebauSoftwire:
All existing api tests passed 🟢 I've only turned on the `processCandidateIdentity` feature set for the audit events as those changed with the routing changes when introducing the new lambda. Only have one audit event test to fix related to  the process-async-credential lambda

api-tests/data/audit-events/delete-pending-f2f-journey.json
```diff
@@ -152,6 +152,27 @@
...
+    "event_name": "IPV_CONTINUITY_OF_IDENTITY_CHECK_END",
+    "component_id": "https://identity.local.account.gov.uk",
+    "extensions": {
+      "type": "STANDARD",
+      "successful": true
+    },
+    "restricted": {
+      "device_information": {}
+    }
+  },
```
Joe-Edwards-GDS, line 175: Neat how you can see the change very clearly here :)

api-tests/data/audit-events/new-identity-f2f-journey.json
```diff
@@ -329,11 +350,62 @@
...
+      "newBirthDate": [
+        {
+          "value": "1965-07-08"
+        }
+      ],
+      "device_information": {}
+    }
+  },
+  {
+    "event_name": "IPV_GPG45_PROFILE_MATCHED",
```
Joe-Edwards-GDS, line 371: Hmm, this one actually makes me a little nervous - it means we'll record this event twice (you can see it a couple of events up as well)...

I wonder if we should remove the profile matched event from check-existing-identity, and rely on this to do it, when the feature is enabled? - Will that drop it from any other scenarios?

api-tests/data/audit-events/new-identity-p2-app-journey.json
```diff
@@ -222,7 +252,7 @@
     "event_name": "IPV_IDENTITY_STORED",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "identity_type": "new",
+      "identity_type": "NEW",
```
Joe-Edwards-GDS: This is interesting - is this a deliberate difference?

Looks like `IdentityType` has a JsonProperty annotation to make it lowercase, but `CandidateIdentityType` doesn't?

Perhaps we need to add similar annotations, or map one onto the other here...

Joe-Edwards-GDS: Exactly the sort of issue we would have missed without these tests!

api-tests/features/audit-events.feature
```diff
@@ -1,6 +1,7 @@
 Feature: Audit Events
   Scenario: New identity - p2 app journey
-    Given I start a new 'medium-confidence' journey
+    Given I activate the 'processCandidateIdentity' feature set
```
Joe-Edwards-GDS, line 3: (You could consider putting this into `Background`?)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
Joe-Edwards-GDS, line 1: Looking forward to the cleanup here!

Joe-Edwards-GDS (commented):
Looks good - a couple of things flagged up by the audit event tests :)

api-tests/src/steps/ipv-steps.ts
```diff
@@ -111,9 +111,18 @@ const startNewJourney = async (
 };
 
 When(
-  "I activate the {string} feature set(s)",
-  function (this: World, featureSet: string) {
-    this.featureSet = featureSet;
+  /I activate the '([\w,]+)' feature set(?:s)?( along with the existing feature set(?:s)?)?/,
```
Joe-Edwards-GDS: (I sort of wonder if this should be the default, but neat either way!)

api-tests/data/audit-events/new-identity-p2-app-journey.json
```diff
@@ -142,7 +142,7 @@
         }
       ],
       "successful": true,
-      "age": "type[number]"
+      "age": 59
```
Joe-Edwards-GDS: (accidental reversion here)

api-tests/features/audit-events.feature
```diff
@@ -76,7 +79,7 @@ Feature: Audit Events
       | evidence_requested | {"scoringPolicy":"gpg45","strengthScore":3} |
     Then I get a 'page-face-to-face-handoff' page response
 
-    Given I activate the 'pendingF2FResetEnabled' feature set
+    Given I activate the 'pendingF2FResetEnabled' feature set along with the e
```
Joe-Edwards-GDS: eeee

Joe-Edwards-GDS (commented):
I think this is good, just a couple of typos :)

api-tests/src/steps/ipv-steps.ts
thebauSoftwire: good idea 👍

Joe-Edwards-GDS (approved):
Thanks - sorry it ended up being so complicated!

PR #2898, opened 2025-01-21
BAU: investigate flakiness of f2f audit event api test

thebauSoftwire:
Test PR to investigate the flakiness in f2f audit event test

### DIFF
--- a/api-tests/data/audit-events/new-identity-f2f-journey.json
+++ b/api-tests/data/audit-events/new-identity-f2f-journey.json
@@ -170,6 +170,18 @@
       "successful": true
     },
     "restricted": {
+      "newAddress": [
+        {
+          "addressCountry": "GB",
+          "addressLocality": "BATH",
+          "buildingName": "",
+          "buildingNumber": "8",
+          "postalCode": "BA2 5AA",
+          "streetName": "HADLEY ROAD",
+          "subBuildingName": "",
+          "validFrom": "1000-01-01"
+        }
+      ],
       "device_information": {}
     }
   },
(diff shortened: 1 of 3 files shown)


api-tests/data/audit-events/new-identity-f2f-journey.json
```diff
@@ -229,27 +241,6 @@
...
-  {
-    "event_name": "IPV_SUBJOURNEY_START",
-    "component_id": "https://identity.local.account.gov.uk",
-    "extensions": {
-      "journey_type": "F2F_PENDING"
-    },
-    "restricted": {
-      "device_information": {}
-    }
-  },
```
thebauSoftwire, line 252: This was due to the `initialise-ipv-session` lambda being hit before the async VC was fully processed and so was returning an F2F_PENDING journey type. Hopefully, with the delay, we expect this to happen after the async VC is received and processed.

api-tests/src/steps/ipv-steps.ts
```diff
@@ -207,6 +208,13 @@ When(
     reproveIdentity: " with reprove identity" | undefined,
     expectedPage: string,
   ): Promise<void> {
+    // In the case where we were previously on an f2f journey, we want to wait a suitable period of time
+    // to let the request to the process-async-cri lambda to finish before starting a new session. This will
+    // hopefully reduce flakiness with the audit event tests where we expect the events to be in a certain order.
+    if (this.lastCriRequest?.redirectUrl.includes("f2f-cri")) {
+      await delay(NEW_SESSION_DELAY_MILLIS);
```
MikeCollingwood: It's just my preference but I'd rather a separate step for a delay so it can be seen from the top level, as a step. I have this in the strategic app journey to wait for the VC (though it could be replaced by a delay like in this case):
```cucumber
    And I pass on the DCMAW callback in a separate session
    Then I get an error response with message 'Missing ipv session id header' and status code '400'

    # Wait for the VC to be received before continuing. In the usual case the VC will be received well before the user
    # has managed to log back in to the site.
    When I poll for async DCMAW credential receipt # <<<

    And I start a new 'medium-confidence' journey
    Then I get a 'page-dcmaw-success' page response
    When I submit a 'next' event
```

I know the extra step would introduce a fair few lines but I think it makes the delay more easily understood. Happy to be disagreed with

thebauSoftwire: that's a good point, I think it matters less though for this step bc it's used specifically for awaiting async credentials (see comment on line 201) but I agree it could be more clear and it could be good for reducing the time it takes for the api tests to run - we don't need an initial delay for each time this step is ran, it's more so to make sure the audit events are recorded in the correct order. Thanks!

MikeCollingwood (approved):
LGTM

PR #2901, opened 2025-01-22
PYIC-7888 Detailed dynatrace logging in build

Joe-Edwards-GDS:
### What changed

Restore DT for all lambdas in build, and turn on detailed logging.

### Why did it change

DT support have requested detailed DT logging

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -126,6 +126,7 @@ Conditions:
     - !Equals [ !Ref AWS::AccountId, "175872367215" ]
   IsDev01: !Equals [ !Ref AWS::AccountId, "130355686670" ]
   IsNotDevelopment: !Not [ !Condition IsDevelopment ]
+  IsBuild: !Equals [ !Ref Environment, "build" ]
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -752,24 +752,15 @@ Resources:
...
           CLIENT_OAUTH_SESSIONS_TABLE_NAME: !Ref ClientOAuthSessionsTable
-          # PYIC-7888 Temporarily disable DT in build
-          AWS_LAMBDA_EXEC_WRAPPER: !If
-            - UseDynatrace
-            - !If
-              - IsBuild
-              - !Ref AWS::NoValue
-              - /opt/dynatrace
+          # PYIC-7888 Turn on debug logging in build
+          DT_LOGGING_DESTINATION: !If
```
DanCorderIPV, line 756: Is it right that only `BuildClientOauthResponseFunction` has this and `DT_LOGGING_JAVA_FLAGS`

DanCorderIPV: Just seen the slack comment so this makes sense

DanCorderIPV (approved):

PR #2912, opened 2025-01-27
PYIC-7963: Deduplicate shared claims names with case insensitivity

MikeCollingwood:
### What changed

- Deduplicate shared claims names with case insensitivity.

### Why did it change

- To prevent the same name in different formats from triggering TICF to fail.

### Issue tracking

- [PYIC-7963](https://govukverify.atlassian.net/browse/PYIC-7963)

[PYIC-7963]: https://govukverify.atlassian.net/browse/PYIC-7963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/data/cri-stub-requests/dcmaw/kenneth-passport-verification-zero/credentialSubject.json
+++ b/api-tests/data/cri-stub-requests/dcmaw/kenneth-passport-verification-zero/credentialSubject.json
@@ -11,11 +11,11 @@
       "nameParts": [
         {
           "type": "GivenName",
-          "value": "Kenneth"
+          "value": "KENNETH"
         },
         {
           "type": "FamilyName",
-          "value": "Decerqueira"
+          "value": "DECERQUEIRA"
         }
       ]
     }
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
@@ -334,7 +334,6 @@ public class BuildCriOauthRequestHandler
         var sharedClaims =
                 SharedClaimsHelper.generateSharedClaims(
                         ipvSessionItem.getEmailAddress(), vcs, getAllowedSharedClaimAttrs(cri));
-        sharedClaims.setName(userIdentityService.deduplicateNames(sharedClaims.getName()));
 
         if (cri.equals(F2F)) {
             evidenceRequest =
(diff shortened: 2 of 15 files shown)


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
```diff
@@ -326,6 +334,7 @@ private JWEObject signEncryptJar(
         var sharedClaims =
                 SharedClaimsHelper.generateSharedClaims(
                         ipvSessionItem.getEmailAddress(), vcs, getAllowedSharedClaimAttrs(cri));
+        sharedClaims.setName(userIdentityService.deduplicateNames(sharedClaims.getName()));
```
blakeyp: Hmm aren't we stripping disallowed claims in the method above which could in theory include name, so we probably need to do this before that?

api-tests/data/cri-stub-requests/dcmaw/kenneth-passport-verification-zero/credentialSubject.json
```diff
@@ -11,11 +11,11 @@
       "nameParts": [
         {
           "type": "GivenName",
-          "value": "Kenneth"
+          "value": "KENNETH"
         },
         {
           "type": "FamilyName",
-          "value": "Decerqueira"
+          "value": "DECERQUEIRA"
```
blakeyp, line 18: Be good to do the same in the stubs test data - capitalise all the DCMAW names

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/BuildCriOauthRequestHandler.java
blakeyp: Also there's a "Found shared claims" log which we probably want to have after this (and probably after stripping the disallowed claims!)

MikeCollingwood: Ah of course, the points about the stripped claims makes sense! I'll shuffle things around

MikeCollingwood: I don't think the stripping of claims would be affected by the deduplication step but it doesn't look right so will try to move the name logic so it can be accessed by the helper

blakeyp (approved):

PR #2913, opened 2025-01-27
PYIC-7961: pass only dcmaw VC to DL cri

thebauSoftwire:
### What changed
When building the CRI oath request for the DL cri, we don't pass the DL VC to the DL CRI.

### Why did it change
When a user goes through the web journey with a successful driving permit, fails KBV with a CI and tries to mitigate with DCMAW, when passing in a driving permit with different details, we end up sharing both the driving permit from DCMAW and the DL CRI from the beginning of their web journey. We should only be passing in their DCMAW one.

### Issue tracking

- [PYIC-7961](https://govukverify.atlassian.net/browse/PYIC-7961)

[PYIC-7961]: https://govukverify.atlassian.net/browse/PYIC-7961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
@@ -73,7 +73,7 @@ public class SharedClaimsHelper {
     private static void addSharedClaimsFromVc(
             SharedClaims sharedClaims,
             VerifiableCredential vc,
-            EnumMap<Cri, List<DrivingPermitDetails>> drivingPermitsSharedClaims) {
+            EnumMap<Cri, List<DrivingPermitDetails>> existingDrivingPermitSharedClaims) {
         var credentialSubject = vc.getCredential().getCredentialSubject();
         var vcCri = vc.getCri();
 
@@ -97,29 +97,39 @@ public class SharedClaimsHelper {
                     .addAll(personWithDocuments.getSocialSecurityRecord());
         }
         var isExistingDcmawDrivingPermitSharedClaim =
-                drivingPermitsSharedClaims.containsKey(Cri.DCMAW)
-                        || drivingPermitsSharedClaims.containsKey(Cri.DCMAW_ASYNC);
+                existingDrivingPermitSharedClaims.containsKey(Cri.DCMAW)
+                        || existingDrivingPermitSharedClaims.containsKey(Cri.DCMAW_ASYNC);
         if (credentialSubject instanceof PersonWithDocuments personWithDocuments
                 && personWithDocuments.getDrivingPermit() != null
                 // skip adding driving permit from DL VC if there is already a drivingPermit from
                 // DCMAW
                 && !(isExistingDcmawDrivingPermitSharedClaim
                         && Cri.DRIVING_LICENCE.equals(vcCri))) {
-
-            // De-duplicate driving permit shared claims by removing existing DL VC driving permit
-            // shared claim and
-            // replacing with the DCMAW VC driving permit instead.
-            var isDcmawVc = Cri.DCMAW.equals(vcCri) || Cri.DCMAW_ASYNC.equals(vcCri);
-            if (isDcmawVc && drivingPermitsSharedClaims.containsKey(Cri.DRIVING_LICENCE)) {
-                drivingPermitsSharedClaims
-                        .get(Cri.DRIVING_LICENCE)
-                        .forEach(sharedClaims.getDrivingPermit()::remove);
-                drivingPermitsSharedClaims.remove(Cri.DRIVING_LICENCE);
-            }
-
-            drivingPermitsSharedClaims.put(vcCri, personWithDocuments.getDrivingPermit());
-            sharedClaims.getDrivingPermit().addAll(personWithDocuments.getDrivingPermit());
+            deduplicateAndAddDrivingPermitSharedClaim(
+                    sharedClaims,
+                    existingDrivingPermitSharedClaims,
+                    vcCri,
+                    personWithDocuments.getDrivingPermit());
         }
+    }
+
+    private static void deduplicateAndAddDrivingPermitSharedClaim(
+            SharedClaims sharedClaims,
+            EnumMap<Cri, List<DrivingPermitDetails>> drivingPermitsSharedClaims,
+            Cri vcCri,
+            List<DrivingPermitDetails> vcDrivingPermitDetails) {
+        // De-duplicate driving permit shared claims by removing existing DL VC driving permit
+        // shared claim and replacing with the DCMAW VC driving permit instead.
+        var isDcmawVc = Cri.DCMAW.equals(vcCri) || Cri.DCMAW_ASYNC.equals(vcCri);
+        if (isDcmawVc && drivingPermitsSharedClaims.containsKey(Cri.DRIVING_LICENCE)) {
+            drivingPermitsSharedClaims
+                    .get(Cri.DRIVING_LICENCE)
+                    .forEach(sharedClaims.getDrivingPermit()::remove);
+            drivingPermitsSharedClaims.remove(Cri.DRIVING_LICENCE);
+        }
+
+        drivingPermitsSharedClaims.put(vcCri, vcDrivingPermitDetails);
+        sharedClaims.getDrivingPermit().addAll(vcDrivingPermitDetails);
     }
 
     private static void stripDisallowedSharedClaims(
--- a/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
+++ b/lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
@@ -31,7 +31,10 @@ public class SharedClaimsHelper {
     private SharedClaimsHelper() {}
 
     public static SharedClaims generateSharedClaims(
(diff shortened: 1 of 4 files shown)


lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
```diff
@@ -26,14 +27,23 @@ public class SharedClaimsHelper {
...
         sharedClaims.setEmailAddress(emailAddress);
 
         // Other shared claims are added from successful VCs
+        // Filter out the DL VC for DL auth source check as we only want the DCMAW VC being passed
+        // into the DL CRI
         vcs.stream()
+                .filter(
+                        vc ->
+                                !(Cri.DRIVING_LICENCE.equals(targetCri)
+                                        && Cri.DRIVING_LICENCE.equals(vc.getCri())))
```
blakeyp, line 52: Hmm while this will work I think ideally we don't want to filter out the whole VC if we can avoid it, it's just the drivingPermit element that we want to 'de-duplicate' - is there something we can do at a lower level, in `addSharedClaimsFromVc` to filter out the DL drivingPermit if there's also a DCMAW one?

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelperTest.java
```diff
@@ -86,6 +86,75 @@ void generatesSharedClaimsForAllSupportedAttributes() {
...
+                                dlCredentialSubjectBuilder
+                                        .drivingPermit(List.of(updatedTestDrivingPermit))
+                                        .build(),
+                                Cri.DRIVING_LICENCE));
+
+        var sharedClaims =
+                generateSharedClaims(TEST_EMAIL, vcs, List.of(SHARED_CLAIM_ATTR_DRIVING_PERMIT));
+
+        assertEquals(
+                Set.of(TEST_DRIVING_PERMIT, updatedTestDrivingPermit),
```
thebauSoftwire, line 163: I assume we still want to keep the functionality of adding more driving permits if they're from the same CRI (aligns with test on line 247) - though not sure why we'd have multiple VCs from a given CRI?

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
```diff
@@ -30,14 +33,19 @@ private SharedClaimsHelper() {}
     public static SharedClaims generateSharedClaims(
             String emailAddress, List<VerifiableCredential> vcs, List<String> allowedSharedClaims) {
         var sharedClaims = new SharedClaims();
+        var sharedDrivingPermitsMappedToVc =
+                new EnumMap<Cri, List<DrivingPermitDetails>>(Cri.class);
```
MikeCollingwood: I see the advantage of this but I think it makes it harder to understand, as opposed to using sharedClaims throughout

thebauSoftwire: yeah, I agree I'm not entirely a fan of this but I couldn't think of another way to keep track of the CRI that the vc came from as SharedClaims only stores the credential subject portion of the VC. Would appreciate any ideas though!

thebauSoftwire: I could pull out the driving licence and dcmaw vcs and process them separately? So if they both exist, filter out the DL ones and only add dcmaw?

MikeCollingwood: Hmm, good point that makes it quite hard... Since they are the only two ways of getting a driving permit, could it be simpler to do it iteratively like:
- if there is no drivingPermit attribute then add it in
- if there is an existing drivingPermit attribute then replace only if DCMAW

So DCMAW gets priority. Maybe this is too simple?

lambdas/build-cri-oauth-request/src/test/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelperTest.java
blakeyp: Yep we should, it's possible in theory (but I don't think we've ever seen it in practice)

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
thebauSoftwire: Much nicer! But I think the issue with this would be that if we had multiple DCMAW VCs, we wouldn't be appending to the drivingPermit shared claims anymore (see [comment above](https://github.com/govuk-one-login/ipv-core-back/pull/2913#discussion_r1932362535)). So we'd still need a way of knowing that the existing driving permits added were from dcmaw or not :(

MikeCollingwood: I wonder if we need to make the same change to the identityClaim attribute as well?

A scalable way of doing what we want may be to delete the web DL session credential since it's no longer relevant?

This PR looks like it should work but I'm just a bit nervous over the obscurity

MikeCollingwood: I think after a call with Patrick, I understand the difficulties in getting a generic solution for this approach, and the identityClaim shouldn't really matter as well, so am happy to approve this PR

MikeCollingwood (dismissed):
I think after a call with Patrick, I understand the difficulties in getting a generic solution for this approach due to the need for the intact DL VC for the DL auth check, and the identityClaim shouldn't really matter as well, so am happy to approve this PR

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
thebauSoftwire: @blakeyp after speaking to Mike, we realised we could potentially delete the DL VC at the point where we know we require a DL auth source check ie when we map the journeyResponse to `JOURNEY_DL_AUTH_SOURCE_CHECK` (this happens in ProcessCriCallback and CheckMobileAppVcReceipt - as part of `CriCheckingService.checkVcResponse`) after which we wouldn't need the DL VC anymore as we know we're going to be passed into the DL CRI.

That way, we wouldn't need to worry about which shared claim from which VC to use here? What do you think?

blakeyp: I don't think we need to do anything for the identity claim itself, this is really just a workaround for the DL auth check scenario.

Unfortunately we can't just clear the web DL VC because we need it to avoid taking the user on an additional trip to DL CRI for the auth check if they've already used the same DL in web DL CRI originally (in eg. a V03 app mitigation scenario).

I agree it's not ideal (not helped by the fact that us handling the DL auth check is already a 'tactical' solution with the hope that DCMAW can do it themselves in the future). I think it could help to have a really clear comment explaining _why_ we're doing what we're doing here, with reference to DL auth check.

blakeyp: Ah that's an interesting idea, let me double check that we don't need to keep the DL VC for any other reason

blakeyp: Chatted with @thebauSoftwire and @MikeCollingwood and agreed given the complexity of trying to dissect the shared claims, we'll revert to the original simple approach of just excluding the DL VC in shared claims generation for DL CRI - accepting that DL auth check overall is a bit of a bodge

lambdas/build-cri-oauth-request/src/main/java/uk/gov/di/ipv/core/buildcrioauthrequest/helpers/SharedClaimsHelper.java
```diff
@@ -28,14 +29,27 @@ public class SharedClaimsHelper {
...
         sharedClaims.setEmailAddress(emailAddress);
 
         // Other shared claims are added from successful VCs
         vcs.stream()
+                // If the target CRI is the Driving Licence CRI, we filter out
+                // any driving licence VCs so they aren't included in the shared claims.
+                // This is to handle the case where a user has a DL and DCMAW VC with driving permit
+                // details and needs to go through an authoritative source check with the DL CRI. In
+                // this scenario, we only want to share the driving permit details from the DCMAW
+                // VC.
```
blakeyp, line 48: 👍

blakeyp (approved):

PR #2931, opened 2025-02-04
BAU: check for ticf null credentials

thebauSoftwire:
### What changed
Add additional check for null credentials in ticf response


libs/ticf-cri-service/src/main/java/uk/gov/di/ipv/core/library/ticf/TicfCriService.java
```diff
@@ -122,7 +122,7 @@ public List<VerifiableCredential> getTicfVc(
             TicfCriDto ticfCriResponse =
                     OBJECT_MAPPER.readValue(ticfCriHttpResponse.body(), TicfCriDto.class);
 
-            if (ticfCriResponse.credentials().isEmpty()) {
+            if (ticfCriResponse.credentials() == null || ticfCriResponse.credentials().isEmpty()) {
```
blakeyp: Think we use `CollectionUtils.isEmpty` elsewhere

blakeyp (approved):

PR #2933, opened 2025-02-04
PYIC-7872: Add AppConfigService

MikeCollingwood:
### What changed

- Add AppConfigService
  - Extending new, abstract YamlParametersConfigService
  - Using AppConfigProvider powertool

### Why did it change

- Reduced friction config interactions

### Issue tracking

- [PYIC-7872](https://govukverify.atlassian.net/browse/PYIC-7872)

[PYIC-7872]: https://govukverify.atlassian.net/browse/PYIC-7872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deploy/template.yaml
```diff
@@ -1096,6 +1145,15 @@ Resources:
...
         - Statement:
+            - Sid: AppConfigAccess
+              Effect: Allow
+              Action:
+                - appconfig:GetConfiguration
+                - appconfig:StartConfigurationSession
+                - appconfig:GetLatestConfiguration
+              Resource:
+                - !Sub "arn:aws:appconfig:${AWS::Region}:${AWS::AccountId}:application/${AppConfigApplication}"
+                - !Sub "arn:aws:appconfig:${AWS::Region}:${AWS::AccountId}:application/${AppConfigApplication}/*"
```
MikeCollingwood, line 1120: Looked at using IAM Roles instead of repeating this policy statement but Role permissions write over other policies

deploy/template.yaml
```diff
@@ -15,6 +15,31 @@ Globals:
     Environment:
       Variables:
         ENVIRONMENT: !Sub "${Environment}"
+        CONFIGURATION_SOURCE: !If
+          - IsDevelopment
+          - "app-config"
+          - "ssm"
```
MikeCollingwood, line 21: There is a confusing issue where redeploying to dev fails unless this variable is renamed!? Baffling.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    @ExcludeFromGeneratedCoverageReport
+    private String getRawParams(Integer cacheDuration) {
+        var applicationId = getEnvironmentVariable(EnvironmentVariable.APP_CONFIG_ID);
+        var environmentId = getEnvironmentVariable(EnvironmentVariable.APP_CONFIG_ENVIRONMENT_ID);
+        var profileId = getEnvironmentVariable(EnvironmentVariable.APP_CONFIG_PROFILE_ID);
+
+        return ParamManager.getAppConfigProvider(
```
chrisclayson: Not sure if you're aware of [this bug](https://github.com/aws-powertools/powertools-lambda-java/issues/1672) in the default `AppConfigProvider` ... they accepted my PR for the fix, and have merged it, but it has yet to make it into a release 🙈 

Not sure if you doing anything clever elsewhere (maybe in your `AppConfigDataClient` client) to work around the problem?

MikeCollingwood: I wasn't aware but I think it doesn't affect this code since I only fetch the config once per Lambda execution. This does mean I can remove `withMaxAge` though, so thank you

MikeCollingwood:
Tested in dev 🟢 And redeploying has stabilised - maybe there was something more circumstantial causing the earlier issue 🟢

deploy/template.yaml
```diff
@@ -15,6 +15,31 @@ Globals:
     Environment:
       Variables:
         ENVIRONMENT: !Sub "${Environment}"
+        CONFIG_SOURCE: !If
```
blakeyp: we might just want to hardcode this to `ssm` for now?

deploy/template.yaml
```diff
@@ -3942,6 +4165,51 @@ Resources:
...
+      LocationUri: "hosted"
+
+  AppConfigHostedConfigurationVersion:
+    Type: AWS::AppConfig::HostedConfigurationVersion
+    Properties:
+      ApplicationId: !Ref AppConfigApplication
+      ConfigurationProfileId: !Ref AppConfigConfigurationProfile
+      Content: |
+        # Placeholder for configuration values
+      ContentType: "application/x-yaml"
```
blakeyp: Do we need this? we'll keep the source config files presumably in common-infra (private repo) and the config-mgmt lambda will generate the actual AppConfig from the source

MikeCollingwood: Unfortunately it does need a default value to be set, even if it's going to be replaced [link](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appconfig-hostedconfigurationversion.html#:~:text=or%20compressed%20data.-,Required%3A%20Yes,-Type%3A%20String)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
chrisclayson: Once per execution or once per lambda lifetime? If the former, then you may still hit the bug ... I think.

deploy/template.yaml
Joe-Edwards-GDS: What happens if you remove the entire `AppConfigHostedConfigurationVersion` resource? I'd sort of expect that be created by config-mgmt (but it wouldn't be unheard of to need a placeholder to bootstrap the process).

My other question is what happens if you update this (either deliberately or by accident) - is there any risk of overwriting existing config?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Getter @Setter private List<String> featureSet;
+    private final SecretsProvider secretsProvider;
+
+    @ExcludeFromGeneratedCoverageReport
+    public AppConfigService() {
+        var cacheDuration =
+                getEnvironmentVariable(
+                        CONFIG_SERVICE_CACHE_DURATION_MINUTES, DEFAULT_CACHE_DURATION_MINUTES);
+        var paramsRaw = getRawParams();
```
Joe-Edwards-GDS: I think we need to be quite careful about lifecycle here (in relation to @chrisclayson 's comment below)...

This class is currently constructed in the constructor of each lambda handler. This means that it will be called once during the snapstart initialisation step, stored in and never again!

Instead, I think we want to create the `ParamManager` in the constructor (like we do for secrets), and _either_
- Load the config explicitly at the start of each lambda handler
    - This would need some thought about how we enforce this uniformly
    - It would also mean you could end up using config that is <cache duration> + <lambda duration> old
- Load the config every time a parameter is accessed, relying on the AppConfig cache to refetch when appropriate
    - Closer to the original SSM implementation
    - We probably want to do something clever to avoid re-parsing the YAML every time!

(Either way you might need to worry about Chris's bug 😞)

deploy/template.yaml
blakeyp: But could we just omit the entire `AppConfigHostedConfigurationVersion` block?

deploy/template.yaml
Joe-Edwards-GDS: If roles don't work, can you define a customer-managed policy that you reference instead? Not worth spending ages on, but would be nice to avoid quite so much duplication.

(If you do that, you could potentially include some of the other 'common' policies - but still applying the principle of least privilege)

deploy/template.yaml
Joe-Edwards-GDS: Ok Patrick beat me to it 😄

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/EnvironmentVariable.java
```diff
@@ -11,4 +11,8 @@ public enum EnvironmentVariable {
     SESSION_CREDENTIALS_TABLE_NAME,
     SQS_AUDIT_EVENT_QUEUE_URL,
     USER_ISSUED_CREDENTIALS_TABLE_NAME,
+    CONFIG_SOURCE,
+    APP_CONFIG_ID,
+    APP_CONFIG_ENVIRONMENT_ID,
+    APP_CONFIG_PROFILE_ID,
```
Joe-Edwards-GDS: Are we keeping these alphabetical?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -60,6 +66,21 @@ public String getEnvironmentVariable(EnvironmentVariable environmentVariable) {
         return System.getenv(environmentVariable.name());
     }
 
+    public <T> T getEnvironmentVariable(EnvironmentVariable environmentVariable, T defaultValue) {
```
Joe-Edwards-GDS: Would it be easier to just have `getIntegerEnvironmentVariable` to match e.g. `getLongParameter`? Inferring the type from the default is a bit sneaky, and would fail if you used `null` for example.

deploy/template.yaml
```diff
@@ -15,6 +15,28 @@ Globals:
...
       Variables:
         ENVIRONMENT: !Sub "${Environment}"
+        CONFIG_SOURCE: "ssm"
+        APP_CONFIG_ID: !If
+          - IsDevelopment
+          - !If
+            - IsDev01
+            - !Sub "app-config-${Environment}.01.dev.identity.account.gov.uk"
+            - !Sub "app-config-${Environment}.02.dev.identity.account.gov.uk"
+          - !Sub "app-config.identity.${Environment}.account.gov.uk"
```
Joe-Edwards-GDS: Is this a standard convention? (as opposed to something simpler, say `"ipv-core-back-${Environment}"`)

(I genuinely don't know - there are lots of things that use a domain name as an identifier by convention, so it's not unusual)

deploy/template.yaml
```diff
@@ -15,6 +15,28 @@ Globals:
...
+            - !Sub "app-config-environment-${Environment}.01.dev.identity.account.gov.uk"
+            - !Sub "app-config-environment-${Environment}.02.dev.identity.account.gov.uk"
+          - !Sub "app-config-environment.identity.${Environment}.account.gov.uk"
+        APP_CONFIG_PROFILE_ID: !If
+          - IsDevelopment
+          - !If
+            - IsDev01
+            - !Sub "app-config-profile-${Environment}.01.dev.identity.account.gov.uk"
+            - !Sub "app-config-profile-${Environment}.02.dev.identity.account.gov.uk"
+          - !Sub "app-config-profile.identity.${Environment}.account.gov.uk"
```
Joe-Edwards-GDS: Do we need to fully scope all of these - how unique is the profile ID? (If the app config ID is scoped to application and environment, could profile ID just be a simple constant?)

@chrisclayson did you set any convention for these?

deploy/template.yaml
MikeCollingwood: `IPVCoreInternalTestingApiRecord` and `DevDNSRecord` use it for Resource `Name`'s, but I think it makes sense to name it like `"core-back-${Environment}"`. Will update

deploy/template.yaml
MikeCollingwood: I think you're right, it won't need the block which is good news

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
MikeCollingwood: It makes sense with the constructor initialised once!

With Chris' bug, I think it will require the `appConfigProvider` to be remade for each time the config is accessed, but temporarily as it will be fixed in a coming release.

I'll construct the `appConfigProvider` once in the constructor, and to address Chris' bug, will remake it explicitly at the start of each config fetch.

I will look to load the config every time a param is accessed. I'll see if I can - it could be tricky avoiding re-parsing the whole config. I could cache and check hashes to see if I need to parse the YAML?

deploy/template.yaml
MikeCollingwood: I think I cannot attach customer-managed policies to resources directly: https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html#:~:text=attach%20to%20principal%20entities%20(IAM%20users%2C%20IAM%20groups%2C%20and%20IAM%20roles). I will park this for now to focus on the main bug/ testing

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -34,17 +39,35 @@ public class AppConfigService extends YamlParametersConfigService {
...
         var cacheDuration =
-                getEnvironmentVariable(
+                getIntegerEnvironmentVariable(
                         CONFIG_SERVICE_CACHE_DURATION_MINUTES, DEFAULT_CACHE_DURATION_MINUTES);
-        var paramsRaw = getRawParams();
 
-        initializeConfig(paramsRaw);
-        this.secretsProvider =
+        appConfigProvider =
+                (AppConfigProvider)
```
Joe-Edwards-GDS: (Do you need to cast this - could the variable just be a `ParamProvider`?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -34,17 +39,35 @@ public class AppConfigService extends YamlParametersConfigService {
...
+                        ParamManager.getAppConfigProvider(
+                                        AppConfigDataClient.builder()
+                                                .httpClient(UrlConnectionHttpClient.create())
+                                                .build(),
+                                        environmentId,
+                                        applicationId)
+                                .withMaxAge(cacheDuration, MINUTES);
+
+        // Initialise parameters value
+        parseParametersIfNew(appConfigProvider.get(profileId));
```
Joe-Edwards-GDS: N.B. this will happen during the snapstart init phase (so could be long before the config is actually used).

It could be a slight optimisation for cold starts, but I'm a bit nervous that this might cause issues reusing the AppConfig session across multiple lambdas. The AppConfig provider internally stores a session that it uses to know when there have been updates.

I'd probably suggest removing this, and relying on the one in `getParameter`...

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -53,11 +76,59 @@ public AppConfigService() {
...
+        this.appConfigProvider = appConfigProvider;
         this.secretsProvider = secretsProvider;
     }
 
+    @Override
+    public String getParameter(String path) {
+        // Temporary. Delete on next AWS Powertools release:
+        // https://github.com/aws-powertools/powertools-lambda-java/issues/1672
+        appConfigProvider =
+                ParamManager.getAppConfigProvider(
```
Joe-Edwards-GDS: If we construct a new provider every time, will we actually cache values at all? If we end up retrieving config every single call then it's definitely a problem!

I think we probably want to go down the other route, or figure out how to share a cache between providers.

You should be able to see the AppConfig calls in AWS X-Ray, if you want to see if it's working :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlParametersConfigService.java
Joe-Edwards-GDS, line 1: This is called `YamlParametersConfigService`, but doesn't actually do anything with YAML, the methods take JSON!

I wonder if we should either
- Push the YAML parsing (and perhaps also the duplicate detection) down into here, so it just has `updateParameters(String yaml)` or similar
- Pull the YAML object mapper up into the subclasses, and call this `JsonParametersConfigService` or something

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -53,11 +76,59 @@ public AppConfigService() {
...
+                        AppConfigDataClient.builder()
+                                .httpClient(UrlConnectionHttpClient.create())
+                                .build(),
+                        environmentId,
+                        applicationId);
+        parseParametersIfNew(appConfigProvider.get(profileId));
+        return this.getParameterFromStoredValue(path);
```
Joe-Edwards-GDS: You don't necessarily have to rename the superclass method, you could do:

```
@Override
public String getParameter(String path) {
    parseParametersIfNew(appConfigProvider.get(profileId));
    return super.getParameter(path);
}
```

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -53,11 +76,59 @@ public AppConfigService() {
...
+                        AppConfigDataClient.builder()
+                                .httpClient(UrlConnectionHttpClient.create())
+                                .build(),
+                        environmentId,
+                        applicationId);
+        parseParametersIfNew(appConfigProvider.get(profileId));
+        return this.getParametersFromStoredValueByPrefix(path);
+    }
+
+    private void parseParametersIfNew(String paramsRaw) {
```
Joe-Edwards-GDS: I think this is likely the best approach for now, but in future we might want to look at using a 'transformation' (https://docs.powertools.aws.dev/lambda/java/utilities/parameters/#transform-values)

That might mean the transformation is cached automatically by the provider, but probably one to leave for https://govukverify.atlassian.net/browse/PYIC-7878 :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
MikeCollingwood: [This bug](https://github.com/aws-powertools/powertools-lambda-java/issues/1672) means the providers are single-use I believe. So either I remake or I create the custom provider class, or am I missing something?

I was remaking it explicitly here to make it obvious to remove when the fix is released

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CustomAppConfigProvider.java
```diff
@@ -0,0 +1,242 @@
+package uk.gov.di.ipv.core.library.config;
```
MikeCollingwood, line 1: AppConfigProvider is package-private so I couldn't extend as it would need access to the methods of the subclass

deploy/template.yaml
```diff
@@ -657,6 +658,8 @@ Resources:
   IssueClientAccessTokenFunction:
     Type: AWS::Serverless::Function
     DependsOn:
+      - AppConfigEnvironment
+      - AppConfigConfigurationProfile
```
Joe-Edwards-GDS: What happens if this isn't here? It feels a bit suspicious to need this explicitly but willing to accept it :shrug:

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CustomAppConfigProvider.java
Joe-Edwards-GDS: Makes sense - I thought there was a chance that was the case

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CustomAppConfigProvider.java
```diff
@@ -0,0 +1,242 @@
...
+import software.amazon.lambda.powertools.parameters.BaseProvider;
+import software.amazon.lambda.powertools.parameters.cache.CacheManager;
+import software.amazon.lambda.powertools.parameters.transform.TransformationManager;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Modified copy of {@link software.amazon.lambda.powertools.parameters.AppConfigProvider}. getValue
+ * fixed to check for empty configuration string.
```
Joe-Edwards-GDS: Could we create a ticket to replace with the updated official version, and link to it here - `TODO: PYIC-XXXX ...`?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlParametersConfigService.java
```diff
@@ -35,19 +40,34 @@ protected String getParameterFromStoredValue(String path) {
...
                 .collect(
                         Collectors.toMap(
                                 e -> e.getKey().substring(path.length()), Map.Entry::getValue));
     }
 
-    void addJsonConfig(Map<String, String> map, JsonNode tree) {
-        addJsonConfig(map, tree, "");
+    protected void updateParameters(Map<String, String> map, File yamlFile) {
+        try {
+            String yamlContent = Files.readString(yamlFile.toPath());
```
Joe-Edwards-GDS, line 54: Note that this is likely to be slightly less efficient - previously the YAML parser was parsing directly from the file stream, whereas now it is being streamed into memory as a string first and _then_ parsed.

It's not a big deal here - it'll only really make a difference for large files, and the local one is only read once at the start anyway.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/AppConfigService.java
```diff
@@ -83,38 +83,20 @@ public AppConfigService(AppConfigProvider appConfigProvider, SecretsProvider sec
...
-        // Temporary. Delete on next AWS Powertools release:
-        // https://github.com/aws-powertools/powertools-lambda-java/issues/1672
-        appConfigProvider =
-                ParamManager.getAppConfigProvider(
-                        AppConfigDataClient.builder()
-                                .httpClient(UrlConnectionHttpClient.create())
-                                .build(),
-                        environmentId,
-                        applicationId);
         parseParametersIfNew(appConfigProvider.get(profileId));
```
Joe-Edwards-GDS: It's fine, but slightly odd that this parameter is always the same - should the AppConfig call be moved into `parseParametersIfNew`?

(And then that method would be something like `reloadParameters` or similar)

Joe-Edwards-GDS (commented):
Thanks, this is looking good I think - a few small comments but feels nearly there

deploy/template.yaml
MikeCollingwood: Ah it isn't needed. I found "Deployment not found" errors (though these were caused by another issue), but it doesn't try to connect during initilialisation 👍

deploy/template.yaml
```diff
@@ -15,6 +15,10 @@ Globals:
     Environment:
       Variables:
         ENVIRONMENT: !Sub "${Environment}"
+        CONFIG_SOURCE: "ssm"
```
Joe-Edwards-GDS, line 18: To check my understanding - at the moment we won't use AppConfig, but can enable it by flipping this to `app-config` on a per-env basis?

Joe-Edwards-GDS (approved):
Looks good, thanks. Might be worth keeping an eye on deployment into build just to make sure the cloudformation all gets deployed correctly since you had so many issues in dev.

PR #2941, opened 2025-02-07
PYIC-7984: use session security check credential

thebauSoftwire:
### What changed
Use in-session security check credential instead of fetching from CIMIT again in 
- process-candidate-identity lambda
- build-user-identity lambda
Moves `getContraIndicatorsVc` into CimitUtilityService

### Why did it change
Reduce unnecessary calls to CIMIT

### Issue tracking

- [PYIC-7984](https://govukverify.atlassian.net/browse/PYIC-7984)

[PYIC-7984]: https://govukverify.atlassian.net/browse/PYIC-7984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -2,7 +2,6 @@ package uk.gov.di.ipv.core.processcandidateidentity;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
-import com.nimbusds.jwt.SignedJWT;
 import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -394,11 +393,10 @@ public class ProcessCandidateIdentityHandler
             return JOURNEY_VCS_NOT_CORRELATED;
         }
 
-        var jwt = SignedJWT.parse(ipvSessionItem.getSecurityCheckCredential());
-        var cri = configService.getCriByIssuer(jwt.getJWTClaimsSet().getIssuer());
-        var credential =
-                VerifiableCredential.fromValidJwt(clientOAuthSessionItem.getUserId(), cri, jwt);
-        var contraIndicators = cimitService.getContraIndicatorsFromVc(credential);
+        var contraIndicators =
+                cimitService.getContraIndicatorsFromVc(
+                        ipvSessionItem.getSecurityCheckCredential(),
+                        clientOAuthSessionItem.getUserId());
 
         var votResult =
                 votMatcher.matchFirstVot(
@@ -432,7 +430,8 @@ public class ProcessCandidateIdentityHandler
             ClientOAuthSessionItem clientOAuthSessionItem,
             String deviceInformation,
             String ipAddress,
-            AuditEventUser auditEventUser) {
+            AuditEventUser auditEventUser)
+            throws ParseException, NoCriForIssuerException, CredentialParseException {
         try {
             var ticfVcs = ticfCriService.getTicfVc(clientOAuthSessionItem, ipvSessionItem);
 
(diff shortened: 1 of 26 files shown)


lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -375,28 +382,21 @@ private JourneyResponse getJourneyResponseForProfileMatching(
...
-                clientOAuthSessionItem.getVtr().size() == 1
-                        ? List.<ContraIndicator>of()
-                        : cimitService.getContraIndicators(
-                                clientOAuthSessionItem.getUserId(),
-                                clientOAuthSessionItem.getGovukSigninJourneyId(),
-                                ipAddress,
-                                ipvSessionItem);
+                cimitService.getContraIndicatorsFromVc(
+                        ipvSessionItem.getSecurityCheckCredential(),
+                        clientOAuthSessionItem.getUserId());
```
thebauSoftwire: theoretically, we should always have a cimit vc by this point (bc we call it as part of check-existing-identity + after every cri call) but should we still check if there isn't one stored in-session?

In this case, how should it be handled? Should we throw an error?

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CimitService.java
```diff
@@ -132,6 +137,14 @@ public List<ContraIndicator> getContraIndicators(
         return getContraIndicatorsFromVc(vc);
     }
 
+    public List<ContraIndicator> getContraIndicatorsFromVc(String vcString, String userId)
+            throws ParseException, NoCriForIssuerException, CredentialParseException,
+                    CiRetrievalException {
+        var jwt = SignedJWT.parse(vcString);
+        var credential = VerifiableCredential.fromValidJwt(userId, Cri.CIMIT, jwt);
```
thebauSoftwire: hard-coding the cri as `Cri.CIMIT` here as we don't do anything with the CRI type (we're only using it to build a VerifiableCredential obj for the next step) and `getContraIndicatorsFromVc` will check whether this is a valid vc or not. 

I did think about updating `configService.getCriByIssuer` instead but it felt like it didn't match the definition if I updated it to also look for the cimit issuer which isn't a CRI

libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/service/CimitService.java
```diff
@@ -132,6 +136,13 @@ public List<ContraIndicator> getContraIndicators(
         return getContraIndicatorsFromVc(vc);
     }
 
+    public List<ContraIndicator> getContraIndicatorsFromVc(String vcString, String userId)
```
Joe-Edwards-GDS: One thing that occurred to me when writing these tickets - do you think this would be better off in `CimitUtilityService`? (i.e. to separate 'service that invokes CIMIT APIs' from 'service that handles CIMIT business logic')

(not reviewing the rest of this, unless you'd like me to 🙂)

thebauSoftwire: good point thanks :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
blakeyp: It would probably be nice to handle it with a journey error response

thebauSoftwire: thanks :) have added to process-candidate-identity and build-user-identity lambdas

thebauSoftwire:
Manually checked CIMIT VC in following journeys: 🟢  
- New identity via app 
- D02 mitigation via passport 
- D02 mitigation via passport then failed with V03
- V03 mitigation (via dcmaw + f2f) same-session
- V03 mitigation (via dcmaw) separate session
- Failed V03 mitigation (via dcmaw) same-session
- Successful reverification
- Update name + address

E2E tests passing in dev env

blakeyp (approved):

PR #2961, opened 2025-02-14
PYIC-8069: use app doc check nested journey for reverification

thebauSoftwire:
### What changed
- use nested journey for the reverification journey map
- should now allow users to retry scanning their DL details if they're incorrect
- add more api tests

### Issue tracking

- [PYIC-8069](https://govukverify.atlassian.net/browse/PYIC-8069)

[PYIC-8069]: https://govukverify.atlassian.net/browse/PYIC-8069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/data/audit-events/reverification-failed-journey.json
+++ b/api-tests/data/audit-events/reverification-failed-journey.json
@@ -70,7 +70,7 @@
     "event_name": "IPV_SUBJOURNEY_START",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "journey_type": "FAILED"
+      "journey_type": "INELIGIBLE"
     },
     "restricted": {
       "device_information": {}
--- a/api-tests/features/mfa-reset-journey.feature
+++ b/api-tests/features/mfa-reset-journey.feature
@@ -116,7 +116,7 @@ Feature: MFA reset journey
         | context   | "check_details" |
       Then I get a 'uk-driving-licence-details-not-correct' page response
       When I submit a 'end' event
-      Then I get a 'prove-identity-another-way' page response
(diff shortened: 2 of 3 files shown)


api-tests/data/audit-events/reverification-failed-journey.json
```diff
@@ -70,7 +70,7 @@
     "event_name": "IPV_SUBJOURNEY_START",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "journey_type": "FAILED"
+      "journey_type": "INELIGIBLE"
```
thebauSoftwire, line 73: When a user gets a CI e.g. through DCMAW during a reverification journey, [`process-cri-callback` will actually return a `fail_with_no_ci` event](https://github.com/govuk-one-login/ipv-core-back/blob/59b48891b5c562eff918621c85d6706b1e984c37/lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java#L271). This is because `fail_with_ci` is only returned from the cimit utility service which we don't call as part of reverification because we don't care about whether the user has a ci - just whether the vc is successful or not.
This doesn't seem right, but I guess technically they aren't failing bc of a CI was checked...

blakeyp: Hmm yeah, maybe we should just rename the `fail_with_no_ci` event to something more appropriate (but not for this ticket)

blakeyp (approved):

PR #2970, opened 2025-02-19
BAU fix kbv entry event

Joe-Edwards-GDS:
### What changed

- Specify the correct entry event when transitioning from an alternate doc mitigation on passport. 
- Update journey map unit test to catch this issue
- Update journey rendering to display correctly

### Why did it change

It was broken

### DIFF
--- a/journey-map/public/helpers.mjs
+++ b/journey-map/public/helpers.mjs
@@ -63,25 +63,20 @@ export const resolveEventTargets = (
     (k) => disabledCris.includes(k),
   );
   if (disabledResolution) {
-    return resolveEventTargets(
-      {
-        ...definition.checkIfDisabled[disabledResolution],
-        journeyContext: definition.journeyContext,
-      },
-      formData,
-      resolvedTargets,
-    );
+    // Resolve target and propagate journeyContext property
+    const resolved = definition.checkIfDisabled[disabledResolution];
+    resolved.journeyContext = definition.journeyContext;
+
+    return resolveEventTargets(resolved, formData, resolvedTargets);
   }
 
   Object.keys(definition.checkJourneyContext || {}).forEach(
     (journeyContext) => {
-      const targets = resolveEventTargets(
-        {
-          ...definition.checkJourneyContext[journeyContext],
-          journeyContext: journeyContext,
-        },
-        formData,
-      );
+      // Resolve target and set journeyContext property
+      const resolved = definition.checkJourneyContext[journeyContext];
+      resolved.journeyContext = journeyContext;
+
+      const targets = resolveEventTargets(resolved, formData);
       resolvedTargets.push(...targets);
     },
   );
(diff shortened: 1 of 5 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -263,6 +263,7 @@ states:
         targetState: MITIGATION_CHECK_FRAUD_AFTER_DL
       alternate-doc-next-passport:
         targetState: MITIGATION_KBVS
+        targetEntryEvent: next
```
Joe-Edwards-GDS, line 266: This wasn't actually causing a problem in production because the underlying event inside the nested state is actually still `next`:

- `CRI_FRAUD` emits `next` within `ADDRESS_AND_FRAUD`
- This is mapped to `next` as an exit event within `ADDRESS_AND_FRAUD`
- This is mapped to `alternate-doc-next-passport` as an exit event within `WEB_DL_OR_PASSPORT`
- This maps to the actual target in `NEW_P2_IDENTITY`

The journey engine uses the original event (i.e. `next`), even though it's actually two layers removed from the actual target 😞 

It's quite hard to unpick this in the Java, so I might leave it - but there's a small risk here: if the underlying event is not a valid entry event but the exit event is, the unit test and the journey visualisation wouldn't spot the issue.

The best way to fix this is to _always_ specify the entry event if there's any ambiguity!

lambdas/process-journey-event/src/test/java/uk/gov/di/ipv/core/processjourneyevent/JourneyMapTest.java
```diff
@@ -106,34 +106,40 @@ void shouldMatchNestedJourneyEntryEvents(IpvJourneyTypes journeyType) throws IOE
                 for (var sourceKey : stateMachineKeys) {
                     var sourceState = stateMachine.get(sourceKey);
 
-                    if (sourceState instanceof BasicState sourceBasicState) {
-                        var sourceEvents = sourceBasicState.getEvents();
```
Joe-Edwards-GDS, line 110: This means we check the events on basic states, but not the exit events of nested states.

I'm a little nervous that the tests below could have similar gaps, but it's pretty hard to digest..!

journey-map/public/helpers.mjs
```diff
@@ -63,25 +63,20 @@ export const resolveEventTargets = (
...
-      {
-        ...definition.checkIfDisabled[disabledResolution],
-        journeyContext: definition.journeyContext,
-      },
-      formData,
-      resolvedTargets,
-    );
+    // Resolve target and propagate journeyContext property
+    const resolved = definition.checkIfDisabled[disabledResolution];
+    resolved.journeyContext = definition.journeyContext;
```
Joe-Edwards-GDS, line 68: I'm not totally happy that this method has an unexpected side-effect (setting this property), but I'm not that keen to refactor everything..!

journey-map/public/render.mjs
```diff
@@ -27,19 +27,13 @@ const expandParents = (journeyStates, otherStates) => {
...
-  }
-
-  if (eventDef.checkJourneyContext) {
-    const journeyCtx = Object.keys(eventDef.checkJourneyContext)[0];
-    return mapTargetStateToExpandedState(
-      eventDef.checkJourneyContext[journeyCtx],
-      subJourneyState,
-    );
-  }
+  resolveEventTargets(eventDef, formData).forEach((targetDef) => {
```
Joe-Edwards-GDS, line 32: We should map target states on the actual target(s) of the event, not just the top-level one

DanCorderIPV (approved):

PR #2976, opened 2025-02-20
PYIC-8032: Update core-back-internal to allow error codes return

MikeCollingwood:
### What changed

The important change is adding the option to return an object with `required: [ "errorCode" ]` attribute, instead of a Client/ Cri/ etc response, allowing for error objects.

I changed the reference from "code" to "errorCode" to make it obvious this object being referenced in the openApi spec is an error reponse.

Dev-testing this, I could see the error response message come through in the core-front logs with code & message.

The error referenced in the ticket was due to sending `profile-unmet` (from `ProcessCandidateIdentity`, which wasn't allowed in the journey map) after `CheckExistingIdentity` had said the user had met a profile.

(This was likely due to a race condition on saving credentials to session credentials before `ProcessCandidateIdentity` picks them up. After 30 successful repeats of the reuse journey, the period between the log for the submission of the credential and the log for the incorrect event was `0.64s`)

### Issue tracking

I wrote more details on the investigation into the error on the ticket:
- [PYIC-8032](https://govukverify.atlassian.net/browse/PYIC-8032)

[PYIC-8032]: https://govukverify.atlassian.net/browse/PYIC-8032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/src/types/internal-api.ts
+++ b/api-tests/src/types/internal-api.ts
@@ -64,15 +64,14 @@ export const isClientResponse = (
 };
 
 export interface ErrorResponse {
-  code: number;
-  message: string;
-  statusCode: number;
+  errorCode: number;
+  errorMessage: string;
 }
 
 export const isErrorResponse = (
   journeyEngineResponse: JourneyEngineResponse,
 ): journeyEngineResponse is ErrorResponse => {
-  return !!(journeyEngineResponse as ErrorResponse).code;
+  return !!(journeyEngineResponse as ErrorResponse).errorCode;
 };
 
 export interface ProcessCriCallbackRequest {
--- a/api-tests/src/types/internal-api.ts
+++ b/api-tests/src/types/internal-api.ts
@@ -64,6 +64,7 @@ export const isClientResponse = (
 };
 
 export interface ErrorResponse {
+  statusCode: number;
   errorCode: number;
   errorMessage: string;
 }
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
@@ -6,8 +6,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class StepFunctionHelpers {
-    private static final String CODE = "code";
-    private static final String MESSAGE = "message";
+    private static final String ERROR_CODE = "errorCode";
+    private static final String ERROR_MESSAGE = "errorMessage";
     private static final String STATUS_CODE = "statusCode";
     private static final String TYPE = "type";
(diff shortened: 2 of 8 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/StepFunctionHelpers.java
```diff
@@ -20,8 +20,8 @@ public static Map<String, Object> generateErrorOutputMap(
             int statusCode, ErrorResponse errorResponse) {
         Map<String, Object> output = new HashMap<>();
         output.put(STATUS_CODE, statusCode);
-        output.put(MESSAGE, errorResponse.getMessage());
-        output.put(CODE, errorResponse.getCode());
+        output.put(ERROR_CODE, errorResponse.getCode());
+        output.put(ERROR_MESSAGE, errorResponse.getMessage());
```
Joe-Edwards-GDS, line 24: Could we be clearer in the PR description what this change is and what it's achieving? We just seem to be changing the shape of this error response, but not including any additional information?

MikeCollingwood:
Ah, the OpenAPI spec may not be enforced. I still think it's a good change (clearer documentation) and in dev-testing I did see the response object

api-tests/src/types/internal-api.ts
```diff
@@ -64,15 +64,14 @@ export const isClientResponse = (
 };
 
 export interface ErrorResponse {
-  code: number;
-  message: string;
-  statusCode: number;
```
blakeyp, line 69: Why are we dropping statusCode here?

MikeCollingwood:
Adding changes to the velocity script allows us to use the status code in the response to core-front. I will add the statusCode references back in.

Screenshot 2025-02-25 at 10 48 35
Screenshot 2025-02-25 at 10 48 28

api-tests/src/types/internal-api.ts
MikeCollingwood: I was considering dropping it altogether as it didn't get picked up, until more recently, I've updated the Velocity script to pick it up and override the response status code. I will add the references back in

openAPI/core-back-internal.yaml
```diff
@@ -126,6 +126,8 @@ paths:
                 #set ($bodyObj = $util.parseJson($input.body))
 
                 #if ($bodyObj.status == "SUCCEEDED")
+                  #set($output = $util.parseJson($bodyObj.output))
+                  #set($context.responseOverride.status = $output.statusCode)
```
MikeCollingwood, line 130: This overrides the response status code

blakeyp (approved):

PR #2979, opened 2025-02-21
PYIC 7808 dl auth check strategic app

DanCorderIPV:
### What changed

Routing for strategic app to include driving licence authoritative source check steps

### Why did it change

We're adding authoritative source checks

### Issue tracking

- [PYIC-7808](https://govukverify.atlassian.net/browse/PYIC-7808)

[PYIC-7808]: https://govukverify.atlassian.net/browse/PYIC-7808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV:
I think this could really do with more API tests to cover:
- The P1 new identity journey
- P1 and P2 mitigation journeys
- P1 and P2 international address journeys
However, it's taken a long time so far to get the current tests working so I think these other tests are probably better put into a separate ticket - there are currently no strategic app versions of these tests at all so it's not just a case of adding DL auth check paths.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-handle-result.yaml
```diff
@@ -0,0 +1,77 @@
...
+      criId: drivingLicence
+      context: check_details
+    parent: CRI_STATE
+    events:
+      next:
+        exitEventToEmit: next
+      access-denied:
+        targetState: REMOVE_DCMAW_FROM_SESSION
+      alternate-doc-invalid-dl:
+        targetState: RESET_SESSION_INVALID_DL
```
thebauSoftwire, line 46: do we need to emit a `IPV_MITIGATION_START` audit event here?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -917,6 +983,37 @@ states:
...
+    nestedJourney: STRATEGIC_APP_TRIAGE
+    exitEvents:
+      next:
+        targetState: POST_APP_DOC_CHECK_INTERNATIONAL_SUCCESS_PAGE
+      anotherWay:
+        targetState: STRATEGIC_APP_NON_UK_NO_APP_PAGE
+      incompleteDlAuthCheckInvalidDl:
+        targetState: STRATEGIC_APP_NON_UK_NO_APP_PAGE
+      failedDlAuthCheckInvalidDl:
+        targetState: STRATEGIC_APP_NON_UK_NO_APP_PAGE
```
thebauSoftwire, line 996: why do we not go through the web route for intl address users?

thebauSoftwire (commented):
Nice! 🙌  just two questions

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
DanCorderIPV: International address users are only allowed to prove their identity with a chipped passport through the app, so we can't offer them any other routes.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-handle-result.yaml
DanCorderIPV: I'm pretty sure we shouldn't emit that event exactly here - this exit event is interpreted in lots of different ways depending on where this nested journey is used.
This is working exactly the same as the `alternate-doc-invalid-dl` event in the non-strategic app routing.
It is a good question as to whether we should emit that event when this subjourney (and the non-strategic version) is being used at the start of the new P1 and P2 identity journeys when we go to the `WEB_DL_OR_PASSPORT` subjourney. I guess we aren't getting a CI here so it's not really a mitigation journey.
@blakeyp could you confirm whether the above is correct?

thebauSoftwire: that's a good point, in that case we might need to update the `app-doc-check` nested journey as part of another ticket. I can see it's emitting the audit event there and we should potentially move it to the parent journey map? 
I think the event should be emitted though as it was a bug raised previously [here](https://govukverify.atlassian.net/browse/PYIC-8023)?

DanCorderIPV: Yes, that bug fix looks wrong to me - I think we'll be emitting the event in places [like here](https://github.com/govuk-one-login/ipv-core-back/blob/b5a30dc1ff96a96f2e2ffd85ee63e161af43f252/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml#L582) which seems wrong

thebauSoftwire (approved):

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/strategic-app-handle-result.yaml
blakeyp: Failing the DL auth check with a CI should start a mitigation journey where `IPV_MITIGATION_START` is fired (and the same for 'separate session' mitigations at the start of new P1/P2 journeys).

Yeah I think we're not being consistent with what we define as the 'start' of a mitigation journey and where we emit this event, I think we need a separate ticket to review this as I'm getting confused myself.

PR #2985, opened 2025-02-24
PYIC-8097 Add CloudFormation and Dockerfile for journey map

Joe-Edwards-GDS:
### What changed

- Add Dockerfile for running the journey map
- Add CloudFormation for deploying the journey map

### Why did it change

GitHub pages is good for hosting, but has a few limitations
- We are restricted to a purely static site
- We cannot add any authentication/ip whitelisting
- We have limited control over the number of deployments and URLs

This will be slightly more overhead/cost to maintain, but allow us to develop more flexibly, and bring it in line with the rest of our infrastructure.

### Issue tracking

- [PYIC-8097](https://govukverify.atlassian.net/browse/PYIC-8097)

[PYIC-8097]: https://govukverify.atlassian.net/browse/PYIC-8097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


journey-map.Dockerfile
```diff
@@ -0,0 +1,43 @@
+# Dockerfile for journey map deployment
+# This must be in the root directory so it can access the journey map files in ./lambdas/process-journey-event
```
Joe-Edwards-GDS, line 2: This is a little bit annoying...

A couple of alternatives
- We could move the journey map yaml into here, and update the process-journey-event lambda build process to pull them in from here (faff)
- We could symlink from here to the lambdas (symlinks are poorly supported on non-unix systems, and potentially surprising)

journey-map/deploy/template.yaml
Joe-Edwards-GDS, line 1: This is basically just an API GW -> ALB -> ECS Fargate service (with a single container).

The domains look like
- `dev-joee-journey-map.02.core.dev.stubs.account.gov.uk` in dev
- `journey-map.build.stubs.account.gov.uk` in build stubs
- `journey-map.stubs.account.gov.uk` in prod stubs

I suggest putting it in the stubs account so that it's insulated from all our production services. (The flipside is that if we want to access any production config, metrics, etc. it will be much harder!)

Joe-Edwards-GDS: (The first of these has been deployed but has a password attached, ask me if you want it)

blakeyp (approved):

PR #3001, opened 2025-02-28
PYIC-7888 disable xray

Joe-Edwards-GDS:
### What changed

- Removed direct consumption of transitive dependency on apache-http-components (transitive dependency of powertools-tracing)
- Removed powertools-tracing

**Recommend reviewing these commits separately**

### Why did it change

Dynatrace support have requested we try and remove AWS X-Ray as a possible contributing factor towards segmentation faults.

In any case, AWS X-Ray gets limited use, and should be largely superseded by dynatrace distributed traces.


gradle/libs.versions.toml
```diff
@@ -7,6 +7,7 @@ pact = "4.6.5"
 powertools = "1.18.0"
 
 [libraries]
+apacheHttpCore = "org.apache.httpcomponents.core5:httpcore5:5.3.3"
```
Joe-Edwards-GDS, line 10: IMO importing transitive dependencies is a bit of a code smell!

We had a load of places that used Apache HTTP components, but it was only available as a transitive dependency of powertools tracing. This meant that removing powertools broke a bunch of unrelated code!

Instead, we should declare a direct dependency on things we need :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/StateMachineInitializer.java
```diff
@@ -190,8 +190,18 @@ private String createNestedJourneyStateName(
         return String.format("%s/%s", state.getName(), nestedJourneyStateName);
     }
 
-    private String getJourneyConfig(IpvJourneyTypes journeyType) throws IOException {
-        return readFileToString(journeyType.getPath());
+    private InputStream getJourneyConfig(String path) throws IOException {
```
Joe-Edwards-GDS: This is the only vaguely meaningful change - we pass the input stream directly into the YAML parser, rather than streaming to a string first.

This is likely to be more efficient anyway :)

blakeyp (dismissed):

Joe-Edwards-GDS:
Checked in dev and it all seems to work. You still get x-ray traces but they include no detail at all:
image

blakeyp (approved):

PR #3010, opened 2025-03-03
PYIC-8127 Rethrow fatal IOExceptions on all HTTP requests

blakeyp:
### What changed

Throw unchecked exception for 'fatal' IOExceptions on all HTTP requests to force lambda instance to crash. If timeout exception, just rethrow. If "Connection reset" exception or HttpRetryException, retry once and then throw unchecked exception.

### Why did it change

We've seen a further instance in prod of a lambda spawned with apparently no network connectivity, raising IOExceptions on HTTP requests, this time in process-cri-callback, and again eventually crashing with a seg fault. As a further mitigation, we should rethrow fatal IOExceptions on all HTTP requests, not just EVCS requests as previously.

### Issue tracking
- [PYIC-8127](https://govukverify.atlassian.net/browse/PYIC-8127)

[PYIC-8127]: https://govukverify.atlassian.net/browse/PYIC-8127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
@@ -11,12 +11,15 @@ import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLParameters;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.Authenticator;
 import java.net.CookieHandler;
+import java.net.HttpRetryException;
 import java.net.ProxySelector;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.net.http.HttpTimeoutException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Optional;
@@ -51,15 +54,25 @@ public class TracingHttpClient extends HttpClient {
         try {
             return baseClient.send(request, responseBodyHandler);
         } catch (IOException e) {
-            // In the build environment we see connection resets for idle connections in
-            // the pool. Retrying uses a different connection.
-            if (e.getMessage().contains("Connection reset")) {
-                LOGGER.warn(
-                        LogHelper.buildErrorMessage("Retrying after HTTP IOException", e)
-                                .with("host", request.uri().getHost()));
-                return baseClient.send(request, responseBodyHandler);
+            LOGGER.error("HTTP request failed with IOException", e);
+            if (e instanceof HttpTimeoutException) {
+                throw e;
             }
-            throw e;
+            // In the build environment we see connection resets for idle connections in the
+            // pool. Retrying uses a different connection.
+            if (e.getMessage().contains("Connection reset") || e instanceof HttpRetryException) {
+                LOGGER.warn(
+                        LogHelper.buildErrorMessage("Retrying after non-fatal HTTP IOException", e)
+                                .with("host", request.uri().getHost()));
+                try {
+                    return baseClient.send(request, responseBodyHandler);
+                } catch (IOException ex) {
+                    throw new UncheckedIOException(ex);
+                }
+            }
+            // Rethrow any other IOException as unchecked exception to force a crash (see PYIC-8058
+            // and linked incident INC0014124)
+            throw new UncheckedIOException(e);
         }
     }
 
--- a/libs/common-services/src/test/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClientTest.java
+++ b/libs/common-services/src/test/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClientTest.java
@@ -7,11 +7,14 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.net.http.HttpTimeoutException;
 
+import static org.junit.jupiter.api.Assertions.assertThrowsExactly;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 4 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -51,15 +54,16 @@ public <T> HttpResponse<T> send(
...
-            // the pool. Retrying uses a different connection.
-            if (e.getMessage().contains("Connection reset")) {
+            LOGGER.error("HTTP request failed with IOException", e);
+            if (isNonFatalIoException(e)) {
                 LOGGER.warn(
-                        LogHelper.buildErrorMessage("Retrying after HTTP IOException", e)
+                        LogHelper.buildErrorMessage("Retrying after non-fatal HTTP IOException", e)
                                 .with("host", request.uri().getHost()));
                 return baseClient.send(request, responseBodyHandler);
             }
```
Joe-Edwards-GDS, line 72: This logic doesn't seem quite right - I think this means timeouts will be retried, which probably isn't what you want?

I suspect there are 3(?) paths:
- Timeouts -> throw IOException as-is
- Connection reset -> retry once, and throw unchecked IO if it fails?
- Any other IO Exception -> throw unchecked IO immediately?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
```diff
@@ -51,15 +54,16 @@ public <T> HttpResponse<T> send(
         try {
             return baseClient.send(request, responseBodyHandler);
         } catch (IOException e) {
-            // In the build environment we see connection resets for idle connections in
-            // the pool. Retrying uses a different connection.
-            if (e.getMessage().contains("Connection reset")) {
+            LOGGER.error("HTTP request failed with IOException", e);
```
Joe-Edwards-GDS, line 57: Do we need this? If it's rethrown then we should log it where its caught, and if it's retried then we get the warning.

(Doesn't matter too much in the fatal case, but e.g. if we recover gracefully then maybe it's only a warning?)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Joe-Edwards-GDS: If you can solve the connection reset some other way then that would be ideal, because we can reduce down to two cases... - I know Chris looked into it a bit and didn't really get anywhere :(

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
blakeyp: I wanted to put this in for now so we can easily see exactly what IOExceptions we're getting and AWS Support have asked for stack traces (not that they'll be much use..)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
blakeyp: I had thought be no harm in retrying on timeouts, now I've remembered TICF 🤦‍♂️

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
Joe-Edwards-GDS: Ok fair enough - I don't think our error logging includes stack traces at the moment though.

I can't remember all the history tbh. I came very close to adding them in a while back, but it all got a bit muddied and I ended up not including it. Something about not including (possibly sensitive) stack traces in logs, but IMO the value they add is a lot greater than the risk.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
blakeyp: I don't have much context on the "Connection reset" errors other than the comment 😕

Joe-Edwards-GDS: Some context on https://github.com/govuk-one-login/ipv-core-back/pull/2704 (including a stack overflow link blaming the NAT gateway...)

blakeyp: I've updated, it's not particularly pretty but I think does what we want for now

Joe-Edwards-GDS (approved):
This looks ok to me.. Might want to monitor the logs a little bit once it hits to make sure we're not crashing on anything unexpected.

As mentioned, I don't think our error logs include stack traces at all, but that can be done as a separate change.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java
blakeyp: `LOGGER.error(message, ex)` should print the stack trace, at least it does in unit tests but since it's not a structured message it might come out a bit funky in cloudwatch
> Logs a message at the ERROR level including the stack trace of the Throwable throwable passed as parameter.

Joe-Edwards-GDS: Ah sorry, I'm not looking closely enough - this is distinct from `buildErrorMessage`!

Yes, the underlying logger should be able to log stack traces (though as you say, I don't know exactly what it'll look like in CW/splunk).

I do still wonder if we should make our default error logging include the stack trace too, so we can get it in our standard log format.

PR #3017, opened 2025-03-04
BAU: Avoid possible race conditions in local audit service

Joe-Edwards-GDS:
### What changed

Use concurrent methods directly to avoid mutating the class variable.

### Why did it change

AWS identified a possible race condition here - even though it's unused in production, it's still worth fixing.

### DIFF
--- a/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/LocalAuditService.java
+++ b/libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/LocalAuditService.java
@@ -17,7 +17,7 @@ public class LocalAuditService implements AuditService {
 
     private static final Deque<AuditEvent> AUDIT_EVENTS = new ConcurrentLinkedDeque<>();
 
-    private Deque<AuditEvent> pendingAuditEvents = new ConcurrentLinkedDeque<>();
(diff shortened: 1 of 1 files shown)


libs/audit-service/src/main/java/uk/gov/di/ipv/core/library/service/LocalAuditService.java
```diff
@@ -35,15 +35,18 @@ public static List<AuditEvent> getAuditEvents(String userId) {
...
-    public synchronized void awaitAuditEvents() {
-        AUDIT_EVENTS.addAll(pendingAuditEvents);
-        pendingAuditEvents = new ConcurrentLinkedDeque<>();
+    public void awaitAuditEvents() {
+        var next = pendingAuditEvents.pollFirst();
+        while (next != null) {
+            AUDIT_EVENTS.addLast(next);
+            next = pendingAuditEvents.pollFirst();
+        }
     }
```
Joe-Edwards-GDS, line 51: The risk was that (previously) a thread could call `sendAuditEvent` between the `addAll` and replacement. This could cause the event to get lost, although in practice it shouldn't be possible.

In the new version, the new event will either get caught during the poll loop, or not (in which case it'll stay pending).

blakeyp (approved):

PR #3029, opened 2025-03-10
BAU: Fix compiler warnings

Joe-Edwards-GDS:
### What changed

Fix various compiler warnings
- Deprecation warning from a Nimbusds update
- Some code using unchecked casts
- Class sharing warning with the way we run tests
- Unused aspect warning in various libs


build.gradle
```diff
@@ -40,6 +40,9 @@ subprojects {
 		environment "LAMBDA_TASK_ROOT", "handler"
 		environment 'AWS_EMF_ENVIRONMENT', 'Local'
 		environment 'AWS_XRAY_CONTEXT_MISSING', 'IGNORE_ERROR'
+
+		// Class data sharing doesn't work with the tests anyway, and this suppresses the warning
+		jvmArgs('-Xshare:off')
```
Joe-Edwards-GDS, line 45: This is a feature for class sharing between multiple JVM instances - it improves performance by sharing some class data between processes.

I'd prefer to fix the underlying issue here (i.e. what's causing the warning), but I don't fully understand it tbh - I assume it's something about the way gradle/junit starts the subprocess to run the tests.

Turning the feature off entirely fixes the warning and doesn't seem to hurt test performance as far as I can tell... But if we're worried we could leave it as-is and deal with the noisy output 🤷

lambdas/issue-client-access-token/src/main/java/uk/gov/di/ipv/core/issueclientaccesstoken/validation/TokenRequestValidator.java
```diff
@@ -95,6 +96,7 @@ private ClientAuthenticationVerifier<Object> getClientAuthVerifier(
             ConfigService configService) {
         return new ClientAuthenticationVerifier<>(
                 new OAuthKeyServiceClientCredentialsSelector(oAuthKeyService),
-                Set.of(new Audience(configService.getParameter(COMPONENT_ID))));
+                Set.of(new Audience(configService.getParameter(COMPONENT_ID))),
+                JWTAudienceCheck.STRICT);
```
Joe-Edwards-GDS, line 100: This is actually a change - the other possible value is `LEGACY` which allows clients to supply a multivalued `aud` value, which I don't believe we need.

libs/common-services/build.gradle
Joe-Edwards-GDS, line 1: Here (and elsewhere) - we'd get a warning for including aspects that didn't apply any changes (in this case, we included powertools metrics, but don't use any of the annotations in this lib, so there's nothing for them to do)

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/persistence/DynamoDataStoreTest.java
```diff
@@ -84,13 +84,14 @@ void shouldPutItemIntoDynamoDbTable() {
         assertEquals(
                 authorizationCodeItem.getIpvSessionId(),
                 authorizationCodeItemArgumentCaptor.getValue().getIpvSessionId());
-        assertNotNull(Long.valueOf(authorizationCodeItemArgumentCaptor.getValue().getTtl()));
+        assertNotEquals(0L, authorizationCodeItemArgumentCaptor.getValue().getTtl());
```
Joe-Edwards-GDS, line 87: (This test was actually broken - we were boxing a value, so it could never possibly be null!)

lambdas/issue-client-access-token/src/test/java/uk/gov/di/ipv/core/issueclientaccesstoken/service/ClientAuthJwtIdServiceTest.java
```diff
@@ -21,14 +20,14 @@
 @ExtendWith(MockitoExtension.class)
 class ClientAuthJwtIdServiceTest {
     @Mock private DataStore<ClientAuthJwtIdItem> mockDataStore;
-    @InjectMocks private ClientAuthJwtIdService clientAuthJwtIdService;
```
Joe-Edwards-GDS, line 24: This is a subtle problem with `@InjectMocks` - it takes the 'largest' constructor (by arguments), but `ClientAuthJwtIdService` has two one-arg constructors. It was basically chance that previously it used the correct one, and now it doesn't.

We can fix by constructing explicitly instead (which is perhaps clearer anyway!)

blakeyp (approved):

PR #3030, opened 2025-03-10
BAU: Add permissions for _staging queue

MikeCollingwood:
### What changed

-  Enable ProcessAsyncCriCredentialFunction to receive from stubQueue_criResponseQueue_staging queue

### Why did it change

- To allow us to stub VC creation in Staging


deploy/template.yaml
```diff
@@ -2124,6 +2143,17 @@ Resources:
               - ReportBatchItemFailures
       AutoPublishAlias: live
 
+  StagingAsyncCriResponseMapping:
```
blakeyp: Could this be configured in`Events` above like the original mapping, probably as a new event?

PR #3031, opened 2025-03-10
PYIC-7982: add support for checkMitigation in journey map

thebauSoftwire:
### What changed
- Updates process-journey-event to handle new `checkMitigation` on events

| mitigation on the journey map |
|-|
| Screenshot 2025-03-10 at 21 14 12 |

### Why did it change
This will help de-duplicate states in our journey map which specifically handle mitigations as well as let us handle mitigations from any state type

### Issue tracking

- [PYIC-7982](https://govukverify.atlassian.net/browse/PYIC-7982)

[PYIC-7982]: https://govukverify.atlassian.net/browse/PYIC-7982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
```diff
@@ -73,12 +84,53 @@ public TransitionResult resolve(JourneyContext journeyContext) throws UnknownEve
...
+                return checkFeatureFlag.get(featureFlagValue).resolve(resolveParameters);
             }
         }
+        if (checkMitigation != null) {
+            var matchedMitigationEvent = getMitigationEvent(resolveParameters);
+
+            if (matchedMitigationEvent.isPresent()) {
+                var mitigationEvent = matchedMitigationEvent.get();
+                LOGGER.info("Mitigation found. Using alternative event.");
+                return mitigationEvent.resolve(resolveParameters);
```
thebauSoftwire: I tried a couple of ways to include the mitigation name in this log, the best I could come up with without increasing cognitive complexity was updating `getMitigationEvent` below to return an `Optional<String>` and moving `validMitigation.map(s -> checkMitigation.get(s));` in this method so that:
```
var validMitigation = getMitigationEvent(resolveParameters);
var matchedMitigationEvent = validMitigation.map(s -> checkMitigation.get(s));

if (matchedMitigationEvent.isPresent()) {
    var mitigationEvent = matchedMitigationEvent.get();
    LOGGER.info("Mitigation '{}' found. Using alternative event.", validMitigation.get());
    return mitigationEvent.resolve(resolveParameters);
end
```
Where we can safely call `validMitigation.get()` without an `isPresent()` as `matchedMitigationEvent.isPresent()` can only be true if there is a value for `validMitigation`.  We could also replace it with `validMitigation.orElseThrow()` but sonar wasn't happy with either.

So settled with removing it from this log instead of suppressing a sonar error

journey-map/public/render.mjs
```diff
@@ -225,6 +231,35 @@ const renderTransitions = (journeyStates, formData) => {
...
+
+  if (journeyContext) {
+    return labelWithClass(
+      JOURNEY_CONTEXT_TRANSITION_CLASSNAME,
+      eventLabel + journeyContextLabel,
+    );
+  }
+
+  if (mitigationLabel) {
+    return labelWithClass(MITIGATIONS_TRANSITION_CLASSNAME, mitigationLabel);
```
Joe-Edwards-GDS: Do you think there's any value in having the original event as well? I'm not sure - I like keeping it as concise as possible, so long as there's no ambiguity.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -106,6 +106,16 @@ private boolean isScoreBreachingCiThreshold(int score, Vot vot) {
                 > Integer.parseInt(configService.getParameter(CI_SCORING_THRESHOLD, vot.name()));
     }
 
+    public Optional<String> getMitigationJourneyEvent(
+            List<ContraIndicator> cis, Vot confidenceRequested) throws ConfigException {
+        var journeyResponse = getMitigationJourneyIfBreaching(cis, confidenceRequested);
+        if (journeyResponse.isPresent()) {
+            var journey = journeyResponse.get().getJourney();
+            return Optional.of(journey.substring(journey.lastIndexOf("/") + 1));
```
Joe-Edwards-GDS, line 114: (A bit annoying that we have to do this string manipulation - this is kinda related to https://govukverify.atlassian.net/browse/PYIC-3927 but we're unlikely to do anything about it any time soon..)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -545,4 +593,12 @@ private boolean isBackEventDefinedOnState(JourneyState journeyState) {
                         instanceof BasicState basicState
                 && basicState.getEvents().containsKey(BACK_EVENT);
     }
+
+    private void logErrorWithJourneyEventAndJourneyType(
```
Joe-Edwards-GDS: Should this include the state as well as the event and journey?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -277,21 +287,53 @@ private StepResponse executeJourneyEvent(
...
+                    journeyEvent,
+                    ipvSessionItem.getState().subJourney().name());
+            throw new JourneyEngineException();
+        } catch (MissingSecurityCheckCredential e) {
+            logErrorWithJourneyEventAndJourneyType(
+                    "Missing security check credential from session.",
+                    e,
+                    journeyEvent,
+                    ipvSessionItem.getState().subJourney().name());
+            throw new JourneyEngineException();
```
Joe-Edwards-GDS: Our exception handling is really a bit suspect 🤔

Feels like we need a shared exception base class that contains all the properties we need - i.e. a 'log message' as well as the detail message. Perhaps linked to `ErrorResponse` as well. That way all these catch blocks could be combined into a single one that just logs the exception with some extra detail.

That's way out of scope though, so no need to worry for this PR.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/EventResolveParameters.java
Joe-Edwards-GDS, line 1: The `EventResolveParameters` abstraction here feels a little bit weird, because it includes a service (which is not really a 'parameter'). That was kinda already true with the `configService` inside the journey context, so not really a change.

It feels ok to leave, and I can't think an obviously better name (without overloading 'context' even further...)

If you want to explore a slightly more involved alternative, you could try inverting the relationship, and make this an `EventResolver` class, something like:

```
public class EventResolver {
  private CimitUtilityService cimitUtilityService;
  // etc.

  public TransitionResult resolve(Event event) {
    if (<is disabled>) {
      return resolve(<get disabled event>);
    }
    // etc.
    return new TransitionResult(...);
  }
}
```
I haven't thought through the full implications, but it could be a more natural way of doing it?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/EventResolveParameters.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.di.ipv.core.processjourneyevent.statemachine.events;
+
+import uk.gov.di.ipv.core.library.persistence.item.ClientOAuthSessionItem;
+import uk.gov.di.ipv.core.library.persistence.item.IpvSessionItem;
+import uk.gov.di.ipv.core.library.service.CimitUtilityService;
+import uk.gov.di.ipv.core.processjourneyevent.statemachine.stepresponses.JourneyContext;
+
+public record EventResolveParameters(
+        JourneyContext journeyContext,
```
Joe-Edwards-GDS: Should we flatten this out, so there is a `String journeyContext` and a `ConfigService configService`?

(This 'journey context' actually predates the current 'journeyContext', and was meant to be a general container for stuff the journey engine needed to resolve events - as you've created!)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
Joe-Edwards-GDS: If these are all thrown from the same place (the `resolve` method), then one option would be to catch there and rethrow a `JourneyEngineException` with the original exception as a cause, to avoid propagating all the checked exceptions. Fine to leave as-is.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
Joe-Edwards-GDS: Hmm, I see the problem - Sonar doesn't understand the relationship between the two optionals.

I'd be inclined to do almost what you describe, but with the map access explicit:
```
var mitigation = getMitigation(resolveParameters);

if (mitigation.isPresent() && checkMitigation.containsKey(mitigation.get())) {
    var mitigationEvent = checkMitigation.get(mitigation.get());
    LOGGER.info("Mitigation '{}' found. Using alternative event.", mitigation.get());
    return mitigationEvent.resolve(resolveParameters);
}
```

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
```diff
@@ -73,12 +84,53 @@ public TransitionResult resolve(JourneyContext journeyContext) throws UnknownEve
...
+
+        var contraIndicators =
+                cimitUtilityService.getContraIndicatorsFromVc(
+                        securityCheckCredential, clientOAuthSessionItem.getUserId());
+
+        var lowestGpg45ConfidenceRequested =
+                clientOAuthSessionItem
+                        .getParsedVtr()
+                        .getLowestStrengthRequestedGpg45Vot(
+                                resolveParameters.journeyContext().configService());
```
Joe-Edwards-GDS: Elsewhere (e.g. `CriCheckingService`) we're using `ipvSession.getTargetVot()` for this. They should generally be equivalent, but I think there's an edge case somewhere (@DanCorderIPV do you recall?)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
```diff
@@ -32,8 +38,13 @@ public class BasicEvent implements Event {
     private LinkedHashMap<String, Event> checkJourneyContext;
     private List<AuditEventTypes> auditEvents;
     private LinkedHashMap<String, String> auditContext;
+    private LinkedHashMap<String, Event> checkMitigation;
```
Joe-Edwards-GDS, line 25: Do we need to update `JourneyMapTest` to cover these events (e.g. in `checkTargetStatesExist` and other places that read into the conditional events)

Joe-Edwards-GDS (commented):
Thanks, this looks good!

I've left a few suggestions, but they're mostly optional tidying up :)

journey-map/public/render.mjs
thebauSoftwire: I assumed that the original event would always be `fail-with-ci` (can't think of a case where it wouldn't be) so decided to omit it from the label

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
thebauSoftwire: Thanks for this!! 🙏 Moved `checkMitigation.containsKey(mitigation.get())` into `getMitigationEvent` below to keep cognitive complexity below the threshold

journey-map/public/render.mjs
Joe-Edwards-GDS: Yeah, I think it probably will be from a CRI state. It might not be from non-CRI states (e.g. if we want to divert the user further down the line) - but it might still be obvious.

Happy to leave it and we can always change it later as we understand the impact more, up to you :)

thebauSoftwire: that's a fair point, will add back in :)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
thebauSoftwire: ahh missed this, thanks!

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/BasicEvent.java
Joe-Edwards-GDS: Feels a bit weird to move the `containsKey` away from where the check is needed, but happy to work around.

(Also - if you ever think Sonar is directing you to make worse code, we can mark as a false positive/wontfix, it's only there to help us :) )

thebauSoftwire: yeah I agree, it was more so to stop sonar from complaining :') but I don't think marking this as a false positive/won't fix would be good as it might be worth refactoring this method at some point when we start adding other checks which push it past the threshold

Joe-Edwards-GDS: Sure - I'd love to do a complete rewrite of this stuff anyway (e.g. see my other comment), perhaps even into Typescript!

Not a priority yet, but could be later in the year...

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/events/EventResolveParameters.java
thebauSoftwire: I'm having a hard time picturing what this would look like, would this EventResolver class not act like a service essentially? So we'd pass in an event to it's `resolve` method but in order to initialise the service we'd either have to:
1. initialise EventResolver at the State level which means we'd need to either pass in a config service from higher up into the State classes or initialise a config service for every instance of the State class
2. initialise EventResolver at the ProcessJourneyEventHandler level which would mean passing it down to where the events are similarly to how we're passing down CimitUtilityService

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -106,6 +106,16 @@ private boolean isScoreBreachingCiThreshold(int score, Vot vot) {
                 > Integer.parseInt(configService.getParameter(CI_SCORING_THRESHOLD, vot.name()));
     }
 
+    public Optional<String> getMitigationJourneyEvent(
+            List<ContraIndicator> cis, Vot confidenceRequested) throws ConfigException {
+        var journeyResponse = getMitigationJourneyIfBreaching(cis, confidenceRequested);
```
Joe-Edwards-GDS: Sorry I missed this the first time - does this cover the second condition on the ticket:
> The user is not breaching the CI threshold and has a mitigated CI

It's a weird one, but I think we need it to (e.g.) ensure we still route the user on a mitigation route even if they've done a partial journey that mitigated the CI.

Joe-Edwards-GDS (commented):
One comment, but otherwise I think this looks good - assuming that we're leaving a refactor of the journey engine for later.

Joe-Edwards-GDS (dismissed):
Looks good, but happy to review again after the refactor, if you think it's an improvement :)

Joe-Edwards-GDS (approved):
Thanks, this is neat 😄 

As discussed, we could probably go a lot further (e.g. to do the same thing at the state level), but I think we should save that for another day!

PR #3039, opened 2025-03-13
BAU: Remove target VOT in favour of a computed value

Joe-Edwards-GDS:
### What changed

- Remove `targetVot` from IPV session
- Introduce `VotHelper.getThresholdVot` to compute the VOT to use when determining thresholds for (e.g.) CI scoring and evidence requests to CRIs

### Why did it change

`targetVot` is almost exactly synonymous with 'lowest VOT in the VTR' and is often muddled with the actual achieved VOT on the session.

Moreover, eliminating state mutation on the session helps make our service easier to understand and less prone to bugs. E.g. we have a bug where errors from EVCS prevented redirect back to the RP because `targetVot` hadn't been set properly (https://govukverify.atlassian.net/browse/PYIC-8062).


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -259,18 +260,7 @@ private JourneyResponse getJourneyResponse(
...
-            var lowestGpg45ConfidenceRequested =
-                    clientOAuthSessionItem
-                            .getParsedVtr()
-                            .getLowestStrengthRequestedGpg45Vot(configService);
-
-            // As almost all of our journeys are proving or mitigating a GPG45 vot we set the
-            // target vot here as a default value. It will be overridden for identity reuse.
-            ipvSessionItem.setTargetVot(lowestGpg45ConfidenceRequested);
-            ipvSessionService.updateIpvSession(ipvSessionItem);
+            var targetVot = VotHelper.getThresholdVot(ipvSessionItem, clientOAuthSessionItem);
```
Joe-Edwards-GDS, line 263: At this stage the VOT on the session is P0, so this is equivalent to 'lowest strength requested GPG45 vot'

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -325,12 +315,6 @@ private JourneyResponse getJourneyResponse(
                             areGpg45VcsCorrelated,
                             contraIndicators);
             if (profileMatchResponse.isPresent()) {
-                // We are re-using an existing Vot, so it might not be a GPG45 vot
-                ipvSessionItem.setTargetVot(
-                        clientOAuthSessionItem
-                                .getParsedVtr()
-                                .getLowestStrengthRequestedVot(configService));
-                ipvSessionService.updateIpvSession(ipvSessionItem);
```
Joe-Edwards-GDS, line 333: Conversely, at this point the VOT on the session is the matched VOT, which will be used instead.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/Vtr.java
Joe-Edwards-GDS, line 1: This class was really just a helper class masquerading as a data class - its only public members were helpers.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VotHelper.java
```diff
@@ -0,0 +1,45 @@
...
+                .filter(vot -> clientOAuthSessionItem.getVtr().contains(vot.name()))
+                .toList();
+    }
+
+    /**
+     * Gets the VOT to be used as the minimum threshold for passing checks. This will either be the
+     * VOT of the identity that has been achieved, or the minimum requested by the client.
+     */
+    public static Vot getThresholdVot(
+            IpvSessionItem ipvSessionItem, ClientOAuthSessionItem clientOAuthSessionItem) {
```
Joe-Edwards-GDS, line 26: It's a little awkward that we've got `vot` on the session and `vtr` on the client session - otherwise we could have considered putting this on one of the session items instead.

I don't feel too bad about it - means we take a more functional approach instead.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -446,31 +446,34 @@ public JourneyResponse getJourneyResponseFromTicfCall(
...
-
-            var journeyResponse =
-                    cimitUtilityService.getMitigationJourneyIfBreaching(cis, thresholdVot);
-            if (journeyResponse.isPresent()) {
-                LOGGER.info(
-                        LogHelper.buildLogMessage(
-                                "CI score is breaching threshold - setting VOT to P0"));
-                ipvSessionItem.setVot(Vot.P0);
-                ipvSessionService.updateIpvSession(ipvSessionItem);
+            if (!clientOAuthSessionItem.isReverification()) {
```
Joe-Edwards-GDS, line 449: Small, but possibly significant change here - we ignore VCs from TICF CRI because we can't score them without a target VOT.

We still submit them to CIMIT, so they will still block the user from accessing their identity on any subsequent journeys after the MFA reset.

This is an existing issue, but isn't manifesting because of https://govukverify.atlassian.net/browse/PYIC-8010

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VotHelper.java
```diff
@@ -0,0 +1,45 @@
...
+        if (ipvSessionItem.getVot() != Vot.P0) {
+            // If we've met a VOT already, that is the threshold we should use
+            return ipvSessionItem.getVot();
+        }
+
+        if (clientOAuthSessionItem.getVtr() == null) {
+            // This may happen on (e.g.) a reverification request
+            throw new IllegalStateException("Attempted to calculate threshold VOT with no VTR");
```
Joe-Edwards-GDS, line 34: At the moment this is _just_ for reverification. We could provide a 'sensible' default here instead (e.g. P2) but IMO it makes more sense to force you to handle the issue directly.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VotHelper.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        var gpg45Vots =
+                Vot.SUPPORTED_VOTS_BY_DESCENDING_STRENGTH.stream()
+                        .filter(vot -> clientOAuthSessionItem.getVtr().contains(vot.name()))
+                        .filter(vot -> vot.getProfileType() == ProfileType.GPG45)
+                        .toList();
+
+        return gpg45Vots.get(gpg45Vots.size() - 1);
```
Joe-Edwards-GDS, line 43: Java 21 has `List.getLast()`... soon! 😄

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: ohh good spot!!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VotHelper.java
thebauSoftwire: Should this also check for an empty vtr for the case you've added in [ClientOauthSessionDetailsService.getEnabledVtr](https://github.com/govuk-one-login/ipv-core-back/pull/3039/files#diff-0384087424c1800866db67985f333eed3e5ab7c662b4f0af4da055d24cea30f7) when p1 journeys are disabled (though it's live now, but for future features)?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/helpers/VotHelperTest.java
```diff
@@ -0,0 +1,65 @@
...
+        var clientOAuthSessionItem =
+                ClientOAuthSessionItem.builder().vtr(List.of("P1", "P2", "PCL250")).build();
+
+        var thresholdVot = VotHelper.getThresholdVot(ipvSessionItem, clientOAuthSessionItem);
+
+        assertEquals(Vot.P1, thresholdVot);
+    }
+
+    @Test
+    void getThresholdVotThrowsIfNoRequestedVots() {
```
thebauSoftwire, line 57: similar test if vtr was an empty list? (see comment above)

thebauSoftwire (commented):
nice stuff! Just a small comment on handling an empty vtr

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/helpers/VotHelper.java
Joe-Edwards-GDS: Ah good point - so what happens if we get sent a VTR of `["P1"]` but low confidence is disabled?

Awkward, we probably need another check in `ClientOauthSessionDetailsService` as well!

Joe-Edwards-GDS: Actually we're probably getting rid of that feature flag soon, so I'll just add the check here as a backstop to avoid  overcomplicating for something that'll never get used.

thebauSoftwire (approved):

PR #3056, opened 2025-03-18
PYIC-7983: update lambdas + journey maps to stop emitting old cimit events

thebauSoftwire:
### What changed
ProcessCriCallback: updated routing to no longer emit mitigation events
CheckExistingIdentity: updated to no longer emit mitigation events
Journey maps updated with `checkMitigation`

### Why did it change
We can simplify the journey map by using the new `checkMitigation` feature so we can change the routing based on mitigation from anywhere on the journey map instead of just CRI states.

### Issue tracking


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -14,7 +14,7 @@ states:
   ENHANCED_VERIFICATION:
```
thebauSoftwire, line 27: Is there a reason we have these entry events for P1? We don't currently use them. Is it in place for future mitigation journeys?
- ENHANCED_VERIFICATION
- ALTERNATE_DOC_PASSPORT
-  ALTERNATE_DOC_DL
- ENHANCED_VERIFICATION_F2F_FAIL

api-tests/features/dl-auth-source-checks/dl-auth-source-checks.feature
```diff
@@ -116,16 +116,16 @@ Feature: Authoritative source checks with driving licence CRI
     Then I get a 'photo-id-security-questions-find-another-way' page response
     When I submit an 'appTriage' event
     Then I get a 'dcmaw' CRI response
-    When I submit 'kenneth-driving-permit-valid' details to the CRI stub that mitigate the 'NEEDS-ENHANCED-VERIFICATION' CI
+    When I submit 'kenneth-driving-permit-valid' details to the CRI stub
     Then I get a 'drivingLicence' CRI response
-    When I submit '<dl_details>' details with attributes to the CRI stub
+    When I submit '<dl_details>' details with attributes to the CRI stub that mitigate the 'NEEDS-ENHANCED-VERIFICATION' CI
```
thebauSoftwire, line 121: This wasn't testing mitigation via the APP + DL properly

api-tests/features/dl-auth-source-checks/dl-auth-source-checks.feature
```diff
@@ -116,16 +116,16 @@ Feature: Authoritative source checks with driving licence CRI
...
       | Attribute | Values          |
       | context   | "check_details" |
     Then I get a '<page_response>' page response
     Examples:
       | dl_details                                 | page_response    |
       | kenneth-driving-permit-valid               | page-ipv-success |
-      | kenneth-driving-permit-needs-alternate-doc | pyi-no-match     |
+#      | kenneth-driving-permit-needs-alternate-doc | pyi-no-match     |
```
MikeCollingwood: How come this needs commenting?

deploy/template.yaml
```diff
@@ -1727,6 +1727,8 @@ Resources:
             ParameterName: !Sub ${Environment}/core/*
         - AWSSecretsManagerGetSecretValuePolicy:
             SecretArn: !Sub arn:aws:secretsmanager:eu-west-2:*:secret:/${Environment}/core/evcs/apiKey-*
+        - AWSSecretsManagerGetSecretValuePolicy:
+            SecretArn: !Sub arn:aws:secretsmanager:eu-west-2:*:secret:/${Environment}/core/self/ciConfig-*
```
MikeCollingwood, line 1729: Does this also mean we can remove these permissions from other Lambdas?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
```diff
@@ -30,7 +30,7 @@ states:
       invalid-request:
         targetJourney: FAILED
         targetState: FAILED
-      enhanced-verification:
+      enhanced-verification: # To delete - processCriCallback no longer emits this
```
MikeCollingwood, line 33: I think you can delete as part of this PR and it'd be useful to confirm the Lambdas don't use them

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
MikeCollingwood: I probably put them there, and would guess in an effort to reflect P2, during the early stages of setting it up. Not worth keeping if YAGNI

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -291,14 +289,15 @@ private JourneyResponse getJourneyResponse(
             // That might cause an issue if a user needs to mitigate a P2 journey but comes back to
             // us with a P1 request that doesn't need mitigation. This is out of scope for the MVP
             // though.
-            var ciScoringCheckResponse =
-                    cimitUtilityService.getMitigationJourneyIfBreaching(
```
MikeCollingwood, line 295: Are we able to delete this method, and similar like `getMitigationJourneyEvent`? Would help confirm that it only relies on your new logic

api-tests/features/dl-auth-source-checks/dl-auth-source-checks.feature
Joe-Edwards-GDS: 😱 I thought we fixed this, but must have slipped through the net

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
blakeyp: I think this is a known gap - see Joe's comment on https://govukverify.atlassian.net/browse/PYIC-7983?focusedCommentId=206127

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -206,6 +198,23 @@ public List<ContraIndicator> getContraIndicatorsFromVc(String vcString, String u
...
+        var cimitConfig = configService.getCimitConfig();
+        return !cis.isEmpty()
+                && cis.stream()
+                        .anyMatch(
+                                ci ->
+                                        cimitConfig.containsKey(ci.getCode())
+                                                && !isBreachingCiThresholdIfMitigated(
+                                                        ci, cis, confidenceRequested));
+    }
+
```
MikeCollingwood, line 212: Could these methods move up or down one method to keet the `getContraIndicatorsFromVc` methods together

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -241,40 +245,34 @@ public JourneyResponse checkVcResponse(
...
         var isReverification = scopeClaims.contains(ScopeConstants.REVERIFICATION);
         if (!isReverification) {
+            // Get CIs from old CIMIT VC to check if we have new CIs
+            var oldCis =
+                    cimitUtilityService.getContraIndicatorsFromVc(
+                            ipvSessionItem.getSecurityCheckCredential(),
+                            clientOAuthSessionItem.getUserId());
+
             var contraIndicatorsVc =
                     cimitService.getContraIndicatorsVc(
```
MikeCollingwood: Minor, but I think this could be more easily understood renamed to `fetchContraIndicatorsVc`. I initially assumed it was getting from the session we pass into it

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -291,14 +289,15 @@ private JourneyResponse getJourneyResponse(
...
+                if (cimitUtilityService.areMitigationsAvailableForBreachingCi(
+                        contraIndicators, targetVot)) {
+                    if (asyncCriStatus.isAwaitingVc()) {
+                        return asyncCriStatus.getJourneyForAwaitingVc(false);
+                    }
+                    return getNewIdentityJourney(targetVot);
                 }
-                return ciScoringCheckResponse.get();
+                return JOURNEY_FAIL_WITH_CI;
             }
```
Joe-Edwards-GDS, line 297: This feels a bit radical, but do you think we can simplify this so the 'non fail' cases just fall through below? i.e.:
```
if (cimitUtilityService.isBreachingCiThreshold(contraIndicators, targetVot) &&
        !cimitUtilityService.areMitigationsAvailableForBreachingCi(contraIndicators, targetVot)) {
    return JOURNEY_FAIL_WITH_CI;
}

// if there are available mitigations, just continue!
```

If the user is breaching then profile matching will return empty, and the remaining conditions cover async and new journey responses.

The benefit of what you've got is that we'll try matching profiles even if we know it won't succeed. I don't think that's a major issue (it should be fast!), but if you want to keep it and make it explicit, perhaps wrap the profile matching in an 'if not breaching' check?

api-tests/features/dl-auth-source-checks/dl-auth-source-checks.feature
blakeyp: should this be removed?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
```diff
@@ -211,7 +213,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                     HttpStatusCode.INTERNAL_SERVER_ERROR,
                     ErrorResponse.FAILED_TO_SEND_AUDIT_EVENT,
                     Level.ERROR);
-        } catch (UnrecognisedVotException e) {
+        } catch (UnrecognisedVotException | CredentialParseException | ParseException e) {
```
MikeCollingwood: These changes (well, only for ParseException as CredentialParseException is handled) may also be needed in`CheckMobileAppVcReceiptHandler` since it also uses the `CriCheckingService`

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -241,40 +245,34 @@ public JourneyResponse checkVcResponse(
...
-            if (isReverification) {
-                setFailedIdentityCheckOnIpvSessionItem(ipvSessionItem);
+            var newCis = cimitUtilityService.getContraIndicatorsFromVc(contraIndicatorsVc);
+
+            // If there are new CIs or the CIs don't have any available mitigations, we return
+            // fail-with-ci
+            var targetVot = VotHelper.getThresholdVot(ipvSessionItem, clientOAuthSessionItem);
+            if (cimitUtilityService.isBreachingCiThreshold(newCis, targetVot)
+                    && (!cimitUtilityService.areContraIndicatorsTheSame(oldCis, newCis)
+                            || !cimitUtilityService.areMitigationsAvailableForBreachingCi(
```
MikeCollingwood: Could we invert as it's easier to understand that we send JOURNEY_FAIL_WITH_CI 
 when breaching CI except when it's the already mitigated. A comment, or separated variable, could help clarify that to the reader
 
EDIT:
Ah sorry, I missed the comment you left above, please ignore my mention of adding a comment

MikeCollingwood (commented):
Looks great! Would it make sense to delete the old mitigation logic except for the states, to let us test when there is only your logic

api-tests/features/dl-auth-source-checks/dl-auth-source-checks.feature
thebauSoftwire: oops yep! 🤦‍♀️ Missed from testing

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -206,6 +198,23 @@ public List<ContraIndicator> getContraIndicatorsFromVc(String vcString, String u
         return getContraIndicatorsFromVc(credential);
```
Joe-Edwards-GDS: (You could consider catching the `ParseException` and rethrowing as a `CiExtractionException` if you wanted fewer things to handle)

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/ProcessCriCallbackHandler.java
thebauSoftwire: Has already been added [here](lambdas/check-mobile-app-vc-receipt/src/main/java/uk/gov/di/ipv/core/checkmobileappvcreceipt/CheckMobileAppVcReceiptHandler.java), unless you meant something else?

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
Joe-Edwards-GDS: I think there's a subtle edge case here - if the user picks up a _non-breaching_ CI while on a mitigation route. E.g. if the threshold is 10 and we have:

oldCis - X (score 15, with mitigation of -10)
newCis - X (score 15, with mitigation), Y (score 2)

The CIs are not equal, but the same mitigation route is still 'valid'.

I wonder if it's something like:
```
var oldMitigation = getMitigationEvent(oldCis);
var newMitigation = getMitigationEvent(newCis);
// Fail if breaching and there is no mitigation available, or a new mitigation is required
if (isBreaching(newCis) && (newMitigation == null || newMitigation != oldMitigation)) {
  return JOURNEY_FAIL_WITH_CI;
}
```
(you might be able to refactor the logic a bit, to make the methods in `CimitUtilityService` less repetitive - perhaps even commonising some of the logic in `EventResolver#getMitigationEvent`)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -23,6 +23,16 @@ states:
       ipv-gpg45-medium:
         targetJourney: NEW_P2_IDENTITY
         targetState: START
+        checkMitigation:
```
Joe-Edwards-GDS: There is an interesting question of responsibility here... Should this check happen here, or in the `START` state of `new-p2-identity`?

I don't think it makes any difference to the behaviour, so it's a question of what makes most sense in the journey map - would you rather see all the possible combinations on the initial page, with different entry points in P2, or are they all just 'P2' at that point, and the start state diverges into mitigations where necessary.

What do you think? I think this is a question for you, since it's what you'll be using :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -53,19 +63,23 @@ states:
       f2f-fail:
         targetJourney: F2F_FAILED
         targetState: FAILED
+        checkMitigation:
+          enhanced-verification:
+            targetJourney: NEW_P2_IDENTITY
+            targetState: ENHANCED_VERIFICATION_F2F_FAIL
```
Joe-Edwards-GDS, line 59: Do we actually need this any more? If they go into the standard F2F failed journey then the mitigation route could be applied there instead - and if the check mitigation is in the START state of new p2 identity then it might 'just work'!

Happy to leave that as a follow-up (so you don't mess with any states in this PR), but might be another simplification that this enables 😄

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -23,6 +23,16 @@ states:
...
+            targetState: ALTERNATE_DOC_PASSPORT
+          alternate-doc-invalid-passport:
+            targetJourney: NEW_P2_IDENTITY
+            targetState: ALTERNATE_DOC_DL
+          enhanced-verification:
+            targetJourney: NEW_P2_IDENTITY
+            targetState: ENHANCED_VERIFICATION
       ipv-gpg45-low:
         targetJourney: NEW_P1_IDENTITY
         targetState: START
```
Joe-Edwards-GDS: We will need mitigation routing for P1 as well - either here or in the START event of new-p1-identity.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/address-and-fraud.yaml
```diff
@@ -17,8 +17,14 @@ nestedJourneyStates:
...
         targetState: CRI_FRAUD
-      enhanced-verification:
+      enhanced-verification: # To delete - processCriCallback no longer emits this
         targetState: CRI_FRAUD
+      fail-with-ci:
+        targetJourney: FAILED
+        targetState: FAILED
+        checkMitigation:
+          enhanced-verification:
+            targetState: CRI_FRAUD
```
Joe-Edwards-GDS: These checks shouldn't be needed any more 😄 (exactly what my fiddly logic in the process CRI callback is meant to achieve)

The reason we had this `enhanced-verification` path before, was in case you were visiting address CRI with an existing enhanced verification mitigation, we want the user to be able to continue. The logic you've added means we should now emit a 'normal' `next` event in that scenario! (And if we get a `fail-with-ci`, that means this CRI has failed with a CI we did not expect!)

Let me know via slack if you want me to explain with a more concrete example...

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/address-and-fraud.yaml
```diff
@@ -28,8 +34,14 @@ nestedJourneyStates:
...
```
Joe-Edwards-GDS: (Same with all of these - in fact the only cases we should need to keep are the CRIs that we expect to start a mitigation action from)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -37,12 +37,22 @@ nestedJourneyStates:
...
+      fail-with-ci:
+        targetJourney: FAILED
+        targetState: FAILED
+        checkMitigation:
+          alternate-doc-invalid-dl:
+            targetState: RESET_SESSION_INVALID_DL
+            auditEvents:
+              - IPV_MITIGATION_START
+            auditContext:
+              mitigationType: invalid-dl
```
Joe-Edwards-GDS, line 55: For example, this one is required, because this is where that mitigation journey begins 🙂

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
thebauSoftwire: _Very_ unlikely and I might also be wrong, but I kept these because if someone was still on an old instance of the processCriCallback/checkExistingIdentity lambdas they'd still emit these old events then the process-journey-event lambda wouldn't know what to do with them. But I can delete if you think I'm being overly cautious/don't think this would actually happen

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
```diff
@@ -19,12 +19,18 @@ nestedJourneyStates:
...
       invalid-request:
         targetState: DWP_THIN_FILE_PAGE
       access-denied:
         targetState: PRE_EXPERIAN_PAGE
+      fail-with-ci:
+        targetJourney: FAILED
+        targetState: FAILED
+        checkMitigation:
+          enhanced-verification:
+            exitEventToEmit: enhanced-verification
```
Joe-Edwards-GDS: There is another option here too, now we've decoupled the mitigation check from the event. No particular need to do it, but you _could_ shift the mitigation check onto the parent, if it's cleaner to have there.

e.g. this could just be:
```
fail-with-ci:
  exitEventToEmit: fail-with-ci
```
and the parent (say, `KBV_PHOTO_ID`) could have:
```
exitEvents:
  fail-with-ci:
    targetJourney: FAILED
    targetState: FAILED
    checkMitigation:
      enhanced-verification:
        targetState: MITIGATION_KBV_FAIL_PHOTO_ID
        # and all the audit stuff
```

As with the start of journey stuff, the functional behaviour should be the same, so it's a question of semantics and responsibilities. Which sounds more correct:
- The KBV nested journey checks whether you can start a mitigation journey, and emits `enhanced-verification`
- The New P2 identity sees the event and handles the mitigation

or

- The KBV nested journey fails and emits `fail-with-ci`
- The New P2 Identity journey checks whether you can start a mitigation journey and handles the mitigation

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
Joe-Edwards-GDS, line 1: Something a bit weird is going on with nested journeys and the mitigations in the journey visualisation - `KBV_PHOTO_ID` is only showing one `next` event, which seems wrong? 🤔

Happy to leave that as a separate bug if the routing itself is working correctly, but IMO it's quite high priority - if the visualisation doesn't reflect reality it will confuse a lot of people!

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
thebauSoftwire: The `getMitigationJourneyIfBreaching` has _slightly_ different logic (it returns the mitigation journey only for breaching CIs) and is still used in `process-candidate-identity` but I get it's not quite clear so have renamed the methods and did a bit of refactoring to make clearer, thanks :)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
Joe-Edwards-GDS, line 1: There's a lot going on in here!

Is the aim to update in the next revision so that we only have _one_ copy of each of DL and Passport CRIs, and use the mitigation checks to route appropriately?

That's neat 😄 - though I slightly wonder if it'd be better to stick with the current layout in this PR (i.e. continue to route via the 'mitigation specific' variants) and then make that change in a follow-up?

If you're confident you've got it mapped out in your head I'm happy to try and combine them together.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -139,8 +139,19 @@ states:
         checkIfDisabled:
           dcmaw:
             targetState: NINO_START_PAGE
+            checkMitigation:
+              enhanced-verification:
+                targetJourney: TECHNICAL_ERROR
+                targetState: ERROR
```
Joe-Edwards-GDS, line 158: nice!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
Joe-Edwards-GDS: In fact it looks like you've partially done this already, e.g. you're handling mitigation events on the identity start page.

Joe-Edwards-GDS (commented):
This is awesome work - it's not easy to get your head around all these journeys!

I somewhat lost steam once I started looking at `new-p2-identity.yaml`, but if all the API tests continue to pass then it ought to be working as expected.

I do wonder in future if it would be worth trying to do this in two (or more) PRs:
- First just replacing like-for-like the mitigation events with mitigation checks, but no changes in the states
- Then apply the simplifications - e.g. cutting out duplicate segments of mitigation journeys etc.

However, I'm happy for you to combine some of the latter if you've already worked it all out.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
Joe-Edwards-GDS: I think our canary deployment means that:
- Each individual API call into core-back uses a consistent set of lambdas (via the step function)
- Different API calls into core-back may use a mixture

Consequences of this are roughly:
- step-function transitions, e.g. `check-existing-identity` -> `process-journey-event` will definitely use a consistent set
- other transitions, including `process-cri-callback` -> `process-journey-event` (because core-front makes these as separate API calls) could use different versions

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
thebauSoftwire: that's a good point, I guess to simplify the initial journey map, it would be nice that these lived in the p2 map instead since it's a mitigation under this identity. The only thing would then be that we have other p2 entry events on that initial map e.g. reprove-identity (could raise the question - why is this not handled within the p2 map instead too? Other than the fact that we don't have a mechanism to handle it similarly to mitigations...) but I am in favour of tidying up that initial journey map

Joe-Edwards-GDS: Sounds good - and a very good point about reprove-identity!

It's a bit of a weak justification, but if you want to feel better about it you could say that 'reprove identity' is slightly more distinct - whereas the mitigation routes really are 'new p2 identity' but with some conditions attached.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
thebauSoftwire: ohh nice spot! I didn't realise they could be routed back to new-p2 from f2f-failed, I've taken note and will do this as part of a separate PR to tidy up the journey map

deploy/template.yaml
thebauSoftwire: good shout! Have removed where necessary (except for from lambdas that are unused e.g. call-ticf-cri lambda that are getting removed soon)

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
thebauSoftwire: ahhh it looks like it's hiding it because the checkMitigation is pointing to failed journeys and only appears when you toggle showing the failed states in the visualiser

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
thebauSoftwire: yep! I've commented them out for now and will put back in a future PR

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
Joe-Edwards-GDS: Hmm - but shouldn't we see the dropout and mitigation routes.. (which aren't failures)?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
thebauSoftwire: good idea, I'll move it out into the parent state bc I think then we won't have to check for an enhanced-verification exitEvent in certain mitigation cases, just the fail-with-ci

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -107,32 +106,41 @@ private boolean isScoreBreachingCiThreshold(int score, Vot vot) {
...
+        var mitigatedCi = hasMitigatedContraIndicator(cis);
+        if (mitigatedCi.isPresent()) {
+            var cimitConfig = configService.getCimitConfig();
+
+            return getMitigationJourneyResponse(
+                    cimitConfig.get(mitigatedCi.get().getCode()), mitigatedCi.get().getDocument());
         }
+
         return Optional.empty();
     }
```
Joe-Edwards-GDS, line 146: The method below here (getMitigationJourneyIfBreaching) is only really used in process candidate identity after TICF CRI - which is probably a bit of an oversight by me!

Do you think that's still the right thing to use there - I wonder if it should follow the same logic as in `CriCheckingService` instead... That might mean we can clean up these methods even more, because we'll never actually use the `JourneyResponse` (which makes sense - they are no longer journey responses, but mitigations to check).

Happy for you to do that in a separate PR if you want to focus on the main CRI callback for now!

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/kbvs.yaml
Joe-Edwards-GDS: Seems to be fixed now :)

Joe-Edwards-GDS (approved):
It's quite hard to follow all the branches but this looks good as far as I can tell, thanks for working through it!

If we can get the selenium tests working then that might be nice to check, but the API tests should be much more comprehensive (and the selenium tests will run in build). In any case, I don't think it should be possible accidentally let people through with CIs, so the risk of anything major should be low. 🤞

I've suggested one other place we probably want to change - handling (potential) CIs from TICF, but that can be done in a follow up if you want to keep it in manageable chunks.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
thebauSoftwire: yeah that makes sense, thanks! I've raised a separate PR [here](https://github.com/govuk-one-login/ipv-core-back/pull/3068) for this. Just tidying this up now, will ping when ready

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
thebauSoftwire: sorry Mike! Turns out you were right and they should be using the same logic (see comment [here](https://github.com/govuk-one-login/ipv-core-back/pull/3056#discussion_r2006080715))

PR #3065, opened 2025-03-20
PYIC-8131: Allow any CRI VC to source passport/ dl claims

MikeCollingwood:
### What changed

- Instead of selecting allowed VCs by CRI, allow any VC to produce a passport/ dl/ nino claim and use that in the user identity claims

### Why did it change

- It produces those claims

### Issue tracking

- [PYIC-8131](https://govukverify.atlassian.net/browse/PYIC-8131)

[PYIC-8131]: https://govukverify.atlassian.net/browse/PYIC-8131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -66,8 +67,8 @@
 import static uk.gov.di.ipv.core.library.helpers.LogHelper.LogField.LOG_MESSAGE_DESCRIPTION;
 
 public class UserIdentityService {
-    private static final List<Cri> PASSPORT_CRI_TYPES = List.of(PASSPORT, DCMAW);
-    private static final List<Cri> DRIVING_PERMIT_CRI_TYPES = List.of(DCMAW, DRIVING_LICENCE);
+    private static final List<Cri> PASSPORT_CRI_TYPES = List.of(PASSPORT, DCMAW, F2F);
+    private static final List<Cri> DRIVING_PERMIT_CRI_TYPES = List.of(DRIVING_LICENCE, DCMAW, F2F);
```
MikeCollingwood, line 71: This is the important bit

MikeCollingwood:
Actually, like suggested, I'll just refactor to allow any Vc with those claims

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -496,114 +487,25 @@ public Optional<List<PostalAddress>> generateAddressClaim(List<VerifiableCredent
...
+        for (var vc : vcs) {
+            if (vc.getCredential() instanceof IdentityCheckCredential identityCheckCredential) {
+                var credentialSubject = identityCheckCredential.getCredentialSubject();
 
-            if (isNullOrEmpty(drivingPermit)) {
-                return Optional.empty();
+                var claim = getClaim.apply(credentialSubject);
+                if (!isNullOrEmpty(claim)) {
+                    return Optional.of(claim);
+                }
```
thebauSoftwire: nice!

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -496,114 +487,25 @@ public Optional<List<PostalAddress>> generateAddressClaim(List<VerifiableCredent
...
             }
-
-            drivingPermit.forEach(
-                    permit -> {
-                        permit.setFullAddress(null);
-                        permit.setIssueDate(null);
-                    });
```
thebauSoftwire, line 586: I guess we don't want to remove the fullAddress and issueDate from the DL permit details in the final identity bundle anymore? It does make this refactor a lot neater

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -496,114 +487,25 @@ public Optional<List<PostalAddress>> generateAddressClaim(List<VerifiableCredent
...
-            var drivingPermit = credentialSubject.getDrivingPermit();
+    public <T> Optional<List<T>> getFirstClaim(
+            List<VerifiableCredential> vcs, Function<IdentityCheckSubject, List<T>> getClaim) {
+        for (var vc : vcs) {
+            if (vc.getCredential() instanceof IdentityCheckCredential identityCheckCredential) {
+                var credentialSubject = identityCheckCredential.getCredentialSubject();
 
-            if (isNullOrEmpty(drivingPermit)) {
-                return Optional.empty();
+                var claim = getClaim.apply(credentialSubject);
```
thebauSoftwire, line 508: Based on the [schema](https://vocab.account.gov.uk/v1/json-schemas/IdentityCheckCredential.json), it looks like the credential subject can be null. Does this mean that we can have no passport or DL claims? If so, should we log something? And would `getClaim.apply` still work on a null argument?

thebauSoftwire (commented):
Nice! Just a couple of questions

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
thebauSoftwire: We were previously handling this by throwing an error since we needed the credential subject to get the required details but not sure if that's needed since we don't do the same for the `identityClaim` cc @blakeyp @Joe-Edwards-GDS

MikeCollingwood: It does need to be able to work with `null` `credentialSubject`'s, good spot thank you! We might want logging, though I thought throwing an error wasn't consistent with the earlier checks which instead return `Optional.empty()`

blakeyp: We wouldn't expect a DL/passport VC without credentialSubject, we should probably still check and throw if it's missing

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
MikeCollingwood: Ah I'm not sure. I think I will add this logic back in, but outside of the method as I don't want to obscure it. Thanks for the spot

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
MikeCollingwood: Should they also produce claims? If so, that would be a more obvious check to use

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -271,18 +263,28 @@ private void addUserIdentityClaims(
...
 
-            Optional<List<PassportDetails>> passportClaim = generatePassportClaim(vcs);
+            Optional<List<PassportDetails>> passportClaim =
+                    getFirstClaim(vcs, IdentityCheckSubject::getPassport, PASSPORT);
             passportClaim.ifPresent(userIdentityBuilder::passportClaim);
 
             Optional<List<DrivingPermitDetails>> drivingPermitClaim =
-                    generateDrivingPermitClaim(vcs);
-            drivingPermitClaim.ifPresent(userIdentityBuilder::drivingPermitClaim);
+                    getFirstClaim(vcs, IdentityCheckSubject::getDrivingPermit, DRIVING_LICENCE);
```
thebauSoftwire: Does `DCMAW` and `DMAW_ASYNC` also need to be on here as valid CRIs for passport and DL permits?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -271,18 +263,28 @@ private void addUserIdentityClaims(
...
+                                    permit.setFullAddress(null);
+                                    permit.setIssueDate(null);
+                                });
+                        userIdentityBuilder.drivingPermitClaim(drivingPermit);
+                    });
         }
 
-        Optional<List<SocialSecurityRecordDetails>> ninoClaim = generateNinoClaim(vcs, profileType);
+        Optional<List<SocialSecurityRecordDetails>> ninoClaim =
+                getFirstClaim(vcs, IdentityCheckSubject::getSocialSecurityRecord, NINO);
```
thebauSoftwire: It looks like we can also get a nino claim from `Cri.HMRC_MIGRATION`

thebauSoftwire (commented):
Looking good, just one comment on other valid CRIs

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
MikeCollingwood: This parameter is just to address the comment thread above, where if they receive a Driving Licence, but do not have a credentialSubject, then it throws. Maybe it isn't a useful check though, as it feels a bit "tacked on"

thebauSoftwire: Wouldn't we get an IdentityCheckCredential from DCMAW too containing drivingPermit details e.g. from an app only journey? In that case we'd also need to check that that contains a non-null credential subject.

If you were to keep that parameter, maybe make it a set of Cri enums instead and re-name it to be something like `vcCrisRequiringCredentialSubject` for self-documentation then similar to what you're already doing:
`vcCrisRequiringCredentialSubject.contains(vc.getCri())`?

Joe-Edwards-GDS: What is the purpose of checking for a credential subject here?

If we are worried about receiving malformed credentials, we should check them when validating credentials as they enter the system (e.g. in the CriCheckingService or VcHelper.isSuccessfulVc), not when performing other unrelated tasks.

It doesn't hurt exactly, but I wouldn't spend too long worrying about it!

MikeCollingwood: I just chatted with Patrick about it and he agrees as well. We were doing it before but I think it doesn't add much value, so have just removed the check

blakeyp (approved):

PR #3068, opened 2025-03-21
PYIC-7983: match cri checking service logic

thebauSoftwire:
### What changed
Update process-candidate-identity handling of ticf VCs to match criCheckingService.checkVcResponse

### Why did it change
In response to these comments about making the logic more consistent:
https://github.com/govuk-one-login/ipv-core-back/pull/3056#discussion_r2006080715
https://github.com/govuk-one-login/ipv-core-back/pull/3056#discussion_r2002960219

### Issue tracking

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
@@ -119,14 +119,9 @@ states:
       fail-with-ci:
         targetJourney: FAILED
         targetState: FAILED_NO_TICF
-      enhanced-verification:
-        targetState: CRI_F2F
-      alternate-doc-invalid-dl:
-        targetJourney: FAILED
-        targetState: FAILED_NO_TICF
-      alternate-doc-invalid-passport:
-        targetJourney: FAILED
-        targetState: FAILED_NO_TICF
+        checkMitigation:
+          enhanced-verification:
+            targetState: CRI_F2F
       error:
         targetJourney: TECHNICAL_ERROR
         targetState: ERROR
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
@@ -119,9 +119,6 @@ states:
       fail-with-ci:
         targetJourney: FAILED
         targetState: FAILED_NO_TICF
-        checkMitigation:
-          enhanced-verification:
-            targetState: CRI_F2F
       error:
         targetJourney: TECHNICAL_ERROR
         targetState: ERROR
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
@@ -132,12 +132,6 @@ states:
     events:
       next:
         targetState: NO_MATCH_PAGE
-      enhanced-verification:
-        targetState: NO_MATCH_PAGE
-      alternate-doc-invalid-dl:
-        targetState: NO_MATCH_PAGE
-      alternate-doc-invalid-passport:
-        targetState: NO_MATCH_PAGE
       fail-with-ci:
         targetState: NO_MATCH_PAGE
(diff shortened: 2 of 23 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/f2f-hand-off.yaml
```diff
@@ -119,14 +119,9 @@ states:
...
         targetJourney: FAILED
         targetState: FAILED_NO_TICF
-      enhanced-verification:
-        targetState: CRI_F2F
-      alternate-doc-invalid-dl:
-        targetJourney: FAILED
-        targetState: FAILED_NO_TICF
-      alternate-doc-invalid-passport:
-        targetJourney: FAILED
-        targetState: FAILED_NO_TICF
```
thebauSoftwire, line 129: Assuming these are fine to delete as part of this PR as `process-candidate-identity` is called as part of a step function transition

Joe-Edwards-GDS: Yep I think that should be fine

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -212,14 +200,15 @@ states:
...
-        targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_VALID_ID
       fail-with-ci:
         targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_INVALID_ID
+        checkMitigation:
+          enhanced-verification:
+            targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_VALID_ID
+          alternate-doc-invalid-dl:
+            targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_VALID_ID
+          alternate-doc-invalid-passport:
+            targetState: COULD_NOT_CONFIRM_DETAILS_PAGE_VALID_ID
```
Joe-Edwards-GDS: Do we need these? I think this probably applies in a lot of these cases - the fancy logic means that these should only occur if TICF itself causes the failure, which probably should be considered the same as any other CI failure. E.g. in this case, you actually _don't_ have a valid ID if you've just triggered an `enhanced-verification`.

...not that it can actually happen, because TICF will never trigger those interventions!

thebauSoftwire:
Just tidying this up now! Will ping when ready!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -115,46 +112,26 @@ public Optional<String> getMitigationEventIfBreachingOrActive(
 
     public Optional<String> getMitigationEventIfBreachingOrActive(
             List<ContraIndicator> cis, Vot confidenceRequested) throws ConfigException {
-        var journeyResponse = getMitigationJourneyIfBreachingOrActive(cis, confidenceRequested);
```
Joe-Edwards-GDS, line 118: I love a review with a lot of red 😍

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/CimitUtilityService.java
```diff
@@ -115,46 +112,26 @@ public Optional<String> getMitigationEventIfBreachingOrActive(
...
-        return Optional.empty();
-    }
-
-    public Optional<JourneyResponse> getMitigationJourneyIfBreaching(
-            List<ContraIndicator> cis, Vot confidenceRequested) throws ConfigException {
         if (isBreachingCiThreshold(cis, confidenceRequested)) {
             return Optional.of(
-                    getCiMitigationJourneyResponse(cis, confidenceRequested)
-                            .orElse(JOURNEY_FAIL_WITH_CI));
+                    getCiMitigationEvent(cis, confidenceRequested).orElse(FAIL_WITH_CI_EVENT));
```
Joe-Edwards-GDS: It doesn't really do any harm, but is this default ever used?

Joe-Edwards-GDS (commented):
Nice - we could merge this as-is, but I think we might be able to drop a few more of the `checkMitigation` blocks...

Joe-Edwards-GDS (approved):

PR #3069, opened 2025-03-21
BAU: manually update di-vocab

thebauSoftwire:
### What changed
Manually updating di-vocab because dependabot doesn't check for its udpates

### Why did it change
There's a new di-vocab version

### DIFF
--- a/api-tests/package-lock.json
+++ b/api-tests/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "devDependencies": {
         "@cucumber/cucumber": "11.2.0",
-        "@govuk-one-login/data-vocab": "1.9.0",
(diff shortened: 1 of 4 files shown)


Joe-Edwards-GDS:
Damn, I thought we'd made this work, but I guess it's broken 😞  (e.g. https://github.com/govuk-one-login/ipv-core-back/pull/2813)

Joe-Edwards-GDS:
Ah - no it's because we've got patch versions disabled in our dependabot config, so we'll only pick it up if we get a minor or major version update.

thebauSoftwire:
ahh, that's why it didn't get updated, I thought we still had to manually check for di-vocab updates

DanCorderIPV (approved):

PR #3074, opened 2025-03-21
PYIC-7983: tidying journey map

thebauSoftwire:
### What changed
Simplified WEB_DL_OR_PASSPORT:
- reduced exitEvents emitted
- reduced duplicated states
Simplified handling of f2f_failed enhanced-verification by using F2F_FAILED journey map

### Why did it change
We now have a way of tidying up the journey map with checkMitigations. This is a follow-up PR to tidy up the journey map.

### Issue tracking

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
@@ -53,10 +53,6 @@ states:
       f2f-fail:
         targetJourney: F2F_FAILED
         targetState: FAILED
-        checkMitigation: # Will remove in future PR to handle this mitigation in F2F_FAILED
-          enhanced-verification:
-            targetJourney: NEW_P2_IDENTITY
-            targetState: ENHANCED_VERIFICATION_F2F_FAIL
       error:
         targetJourney: TECHNICAL_ERROR
         targetState: ERROR
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
@@ -23,14 +23,11 @@ nestedJourneyStates:
     events:
       next:
         targetState: ADDRESS_AND_FRAUD_AFTER_PASSPORT
-#        checkMitigation:
-#          alternate-doc-invalid-dl:
-#            targetState: MITIGATION_PP_ADDRESS_AND_FRAUD
       access-denied:
         targetState: PROVE_ANOTHER_WAY_AFTER_PASSPORT
-#        checkMitigation:
-#          alternate-doc-invalid-dl:
-#            targetState: PP_NO_OTHER_PHOTO_ID
+        checkMitigation:
+          alternate-doc-invalid-dl:
+            targetState: PP_NO_OTHER_PHOTO_ID
       alternate-doc-invalid-passport: # To delete - processCriCallback no longer emits this
         targetState: MITIGATION_PP_NO_MATCH_ANOTHER_WAY
         auditEvents:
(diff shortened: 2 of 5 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
```diff
@@ -53,10 +53,6 @@ states:
       f2f-fail:
         targetJourney: F2F_FAILED
         targetState: FAILED
-        checkMitigation: # Will remove in future PR to handle this mitigation in F2F_FAILED
-          enhanced-verification:
-            targetJourney: NEW_P2_IDENTITY
-            targetState: ENHANCED_VERIFICATION_F2F_FAIL
```
thebauSoftwire, line 59: why do we not handle these with p1 journeys?

Joe-Edwards-GDS: That's a great question - I expect this is another oversight...

Does this mean there's a slightly weird edge case where
- You go on a low confidence journey, end up on `enhanced-verification` and mitigate w/ F2F
- F2F fails
- You start another low confidence journey, and get routed to F2F_FAILED
- F2F failed routes you to new P2 identity (not P1!)

Is the fix to route to `INITIAL_JOURNEY_SELECTION` instead, like we plan to with the pending F2F reset? We'd want to double-check audit events don't get duplicated.

cc @blakeyp this might be better as a bug ticket, since it's not strictly related to this change.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
```diff
@@ -167,17 +158,17 @@ nestedJourneyStates:
...
-#      type: page
-#      pageId: prove-identity-no-other-photo-id
-#      context: drivingLicence
-#    events:
-#      back:
-#        targetState: CRI_DRIVING_LICENCE
-#      returnToRp:
-#        exitEventToEmit: return-to-rp
-  MITIGATION_DL_PROVE_ANOTHER_WAY:
+  DL_NO_OTHER_PHOTO_ID:
```
Joe-Edwards-GDS, line 161: I'm a tiny bit nervous about these new states...

It's a real pain, but I think our canary deployments mean introducing new states is also potentially problematic - a user could be routed to this state by the new lambda, and then hit an old lambda when they proceed from it. However, if we can check how many users _actually_ visit this page within our canary window, we might be able to sneak it inside our error budget (e.g. if its <1) 🤞 

If not, it makes moving states a 3-step process (🤮)
- Add new state
- Update routing to new state
- Delete old state

I'd love to have a better solution for this - we can't solve it perfectly (because we have to tolerate a canary rollback), but e.g. if it _only_ broke in a rollback scenario that would be a lot better.

cc @blakeyp what we discussed on wed

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
thebauSoftwire: yep, at the moment, they'd be routed to P2 and get a P2 identity at the end if they successfully mitigate

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/web-dl-or-passport.yaml
thebauSoftwire: Note from slack convo: We only get a handful of visits to these pages a day (~4 visits) so probably safe to make this change as one PR and delete the unused states in another

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml
thebauSoftwire: should we have some api tests for p1 mitigation journeys too?

blakeyp: Yep, just recreated it in build and raised [PYIC-8190](https://govukverify.atlassian.net/browse/PYIC-8190). Good spot!

Joe-Edwards-GDS (approved):

PR #3077, opened 2025-03-24
PYIC-7983: remove unused states from journey map

thebauSoftwire:
## Proposed changes
Remove unused states from journey map where we've simplified it as part of [this PR](https://github.com/govuk-one-login/ipv-core-back/pull/3074)

### What changed
Removing unused states and tidying up the journey map more

### Issue tracking

### DIFF
--- a/api-tests/features/strategic-app/p1-strategic-app-dl-auth-source-check-enhanced-verification-mitigation.feature
+++ b/api-tests/features/strategic-app/p1-strategic-app-dl-auth-source-check-enhanced-verification-mitigation.feature
@@ -42,7 +42,7 @@ Feature:  Mitigating CIs with enhanced verification using the async DCMAW CRI an
       Then I get a 'pyi-triage-select-smartphone' page response with context 'mam'
       When I submit an 'iphone' event
       Then I get a 'pyi-triage-mobile-download-app' page response with context 'iphone'
-      When the async DCMAW CRI produces a 'kenneth-driving-permit-valid' VC that mitigates the 'NEEDS-ENHANCED-VERIFICATION' CI
+      When the async DCMAW CRI produces a 'kenneth-driving-permit-valid' VC
     # And the user returns from the app to core-front
       And I pass on the DCMAW callback
       Then I get a 'check-mobile-app-result' page response
@@ -50,7 +50,7 @@ Feature:  Mitigating CIs with enhanced verification using the async DCMAW CRI an
       Then the poll returns a '201'
       When I submit the returned journey event
       Then I get a 'drivingLicence' CRI response
-      When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub
+      When I submit 'kenneth-driving-permit-valid' details with attributes to the CRI stub that mitigate the 'NEEDS-ENHANCED-VERIFICATION' CI
         | Attribute | Values          |
         | context   | "check_details" |
       Then I get a 'page-ipv-success' page response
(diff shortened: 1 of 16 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -176,16 +162,12 @@ states:
...
-            auditEvents:
-              - IPV_MITIGATION_START
-            auditContext:
-              mitigationType: invalid-dl
+        targetState: WEB_DL_OR_PASSPORT
+        targetEntryEvent: alternate-doc-invalid-dl-another-way
+        auditEvents:
+          - IPV_MITIGATION_START
+        auditContext:
+          mitigationType: invalid-dl
```
thebauSoftwire, line 170: if we reach this exitEvent, it means that we have a `alternate-doc-invalid-dl` mitigation and so I didn't need to use checkMitigation here

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
```diff
@@ -23,8 +23,6 @@ nestedJourneyStates:
         exitEventToEmit: incomplete
       fail-with-no-ci:
         exitEventToEmit: incomplete
-      enhanced-verification: # To delete - processCriCallback no longer emits this
-        exitEventToEmit: incomplete
```
thebauSoftwire, line 27: Is it possible that a real CRI wouldn't mark a CI as mitigated with CIMIT even if the details passed in are valid? In our stub, we have to manually indicate that the details are mitigating a specific CI.

Removing this event and with the other changes, means that processCriCallback will  return a `next` event (as we still breach the threshold with the existing CI and we have no new CIs) which will route them to the dcmaw success page - previously this would route them to the post-office page. 

They would still not get an identity with a breaching CI as they wouldn't meet a profile (as part of processCandidateIdentity) and so would get a pyi-no-match page _but_ they wouldn't get routed to try the post office.

thebauSoftwire: Similar issue with the strategic app

api-tests/features/strategic-app/p1-strategic-app-dl-auth-source-check-enhanced-verification-mitigation.feature
```diff
@@ -42,15 +42,15 @@ Feature:  Mitigating CIs with enhanced verification using the async DCMAW CRI an
       Then I get a 'pyi-triage-select-smartphone' page response with context 'mam'
       When I submit an 'iphone' event
       Then I get a 'pyi-triage-mobile-download-app' page response with context 'iphone'
-      When the async DCMAW CRI produces a 'kenneth-driving-permit-valid' VC that mitigates the 'NEEDS-ENHANCED-VERIFICATION' CI
+      When the async DCMAW CRI produces a 'kenneth-driving-permit-valid' VC
```
thebauSoftwire, line 45: the CI would get mitigated at the DL CRI if dl-auth-source-check is required

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/app-doc-check.yaml
blakeyp: Chatted with Thea on this, it _should_ never happen so we don't really need to worry about it and as you say they'd still ultimately fail anyway

blakeyp (approved):

PR #3084, opened 2025-03-27
PYIC-8194: Add workflow linter

MikeCollingwood:
### What changed

- Add workflow linter

### Why did it change

- Workflows fail silently on syntax errors

### Issue tracking

- [PYIC-8194](https://govukverify.atlassian.net/browse/PYIC-8194)

[PYIC-8194]: https://govukverify.atlassian.net/browse/PYIC-8194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/link-workflows.yaml
+++ b/.github/workflows/link-workflows.yaml
@@ -1,5 +1,7 @@
 # DO NOT CHANGE THIS FILE
-# Breaking the  syntax of this file would silently allow other workflows to break.
+# This action should be kept as small as possible and only do linting of workflow actions as syntax errors in
+# actions cause them to silently fail. If you do need to change this file, make sure that you watch it execute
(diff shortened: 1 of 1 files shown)


.github/workflows/link-workflows.yaml
```diff
@@ -0,0 +1,19 @@
+# DO NOT CHANGE THIS FILE
```
DanCorderIPV, line 1: Maybe add a bit more context here :) e.g.
This action should kept as small as possible and only do linting of other workflow actions as syntax errors in actions cause them to silently fail.
If you do need to change this file make sure that you watch it execute successfully in github actions as you will not be warned if it fails.

blakeyp (approved):

PR #3089, opened 2025-03-28
PYIC-8135: remove getting ipAddress + update AuditEventUser in user-reverification lambda

thebauSoftwire:
### What changed
Updates user-reverification lambda to not get the ipAddress as it's not available to it

### Why did it change
We were getting warning logs from `user-reverification` saying it had an empty ipAddress. This is expected as it's not available to the lambda so we remove the call to get it from the request headers.

### Issue tracking

- [PYIC-8135](https://govukverify.atlassian.net/browse/PYIC-8135)

[PYIC-8135]: https://govukverify.atlassian.net/browse/PYIC-8135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/user-reverification/src/main/java/uk/gov/di/ipv/core/userreverification/UserReverificationHandler.java
+++ b/lambdas/user-reverification/src/main/java/uk/gov/di/ipv/core/userreverification/UserReverificationHandler.java
@@ -124,7 +124,7 @@ public class UserReverificationHandler extends UserIdentityRequestHandler
                                     userId,
                                     ipvSessionItem.getIpvSessionId(),
                                     clientOAuthSessionItem.getGovukSigninJourneyId(),
-                                    RequestHelper.getIpAddress(input)),
+                                    null),
                             new AuditExtensionReverification(
                                     response.success(), response.failureCode()),
                             new AuditRestrictedDeviceInformation(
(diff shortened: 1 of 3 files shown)


thebauSoftwire:
Tested in dev and we're no longer getting warning logs from the `user-reverification` lambda and the `IPV_REVERIFY_END` audit event no longer contains the ipAddress

Screenshot 2025-03-28 at 12 04 10

lambdas/user-reverification/src/main/java/uk/gov/di/ipv/core/userreverification/UserReverificationHandler.java
```diff
@@ -124,7 +124,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                                     userId,
```
blakeyp, line 122: I think we should also just be using `createWithoutDeviceInformation` here? (as we don't get the cloudfront header from the FE which provides this)

thebauSoftwire: good point, thanks! Have also updated the event-catalogue to reflect this

blakeyp (approved):

PR #3107, opened 2025-04-07
PYIC-8223 Temporary mitigation to force lambda restart on fatal IOException

blakeyp:
### What changed

Temporary mitigation to force lambda restart on fatal IOException

### Why did it change

The previous mitigation [PYIC-8127](https://govukverify.atlassian.net/browse/PYIC-8127) doesn't work in the way we expected - it turns out throwing a runtime exception from the root lambda handler _doesn't_ cause a lambda instance to crash and restart (and we have confirmed this with AWS Support).

This takes things to the next level, calling System.exit() to force the program to crash and restart the lambda instance (reproducible in dev). This is really not nice and we'll want to remove it as soon as we can, but is a temporary mitigation while we get to the bottom of the root cause of the issue, so that if this happens again in prod in the meantime only a single invocation of an unhealthy lambda instance will fail before respawning.

### Issue tracking
- [PYIC-8223](https://govukverify.atlassian.net/browse/PYIC-8223)

[PYIC-8127]: https://govukverify.atlassian.net/browse/PYIC-8127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PYIC-8223]: https://govukverify.atlassian.net/browse/PYIC-8223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
+++ b/lambdas/build-client-oauth-response/src/main/java/uk/gov/di/ipv/core/buildclientoauthresponse/BuildClientOauthResponseHandler.java
@@ -38,6 +38,7 @@ import uk.gov.di.ipv.core.library.service.ClientOAuthSessionDetailsService;
 import uk.gov.di.ipv.core.library.service.ConfigService;
 import uk.gov.di.ipv.core.library.service.IpvSessionService;
 
+import java.io.UncheckedIOException;
 import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


blakeyp:
I've tested this in dev and it produces the desired behaviour, causing a crash and cycling the lambda instance
image

blakeyp:
Sonar complaining about code duplication - I'm not intending to address this given this is temporary code that we want to remove as soon as we've resolved the issue

DanCorderIPV (approved):

MikeCollingwood (approved):

PR #3127, opened 2025-04-14
PYIC-8227: handle null security check credential prior to ticf call

thebauSoftwire:
### What changed
Add call to fetch security check credential if null if calling ticf

### Why did it change
A user had failed their journey because they got a NullPointerException from the `process-candidate-identity` lambda. This is because the lambda was attempting to parse a null security check credential on the session. Typically, by the time the user gets to this lambda, we expect there to already be a security check credential attached to the session (as it would have been done as part of `check-existing-identity`. However, in the case for the user, the lambda threw an error prior to making this call so it was null when the call to `process-candidate-identity` was made.

To handle this, prior to making a call to ticf, we must check if the security check credential is null. If so, make a call to get it.

### Issue tracking

- [PYIC-8227](https://govukverify.atlassian.net/browse/PYIC-8227)

[PYIC-8227]: https://govukverify.atlassian.net/browse/PYIC-8227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thebauSoftwire:
Tested in dev (throwing an error from `check-existing-identity` prior to storing the security check credential) and no longer getting NullPointerException from `process-candidate-identity`. I can also see in the logs that we're making a call to fetch the CIMIT VC twice 🍏

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -445,6 +445,23 @@ public JourneyResponse getJourneyResponseFromTicfCall(
...
+            String previousSecurityCheckCredential = ipvSessionItem.getSecurityCheckCredential();
+            if (StringUtils.isBlank(previousSecurityCheckCredential)) {
+                previousSecurityCheckCredential =
+                        cimitService
+                                .fetchContraIndicatorsVc(
+                                        clientOAuthSessionItem.getUserId(),
+                                        clientOAuthSessionItem.getGovukSigninJourneyId(),
+                                        ipAddress,
+                                        ipvSessionItem)
+                                .getVcString();
```
DanCorderIPV, line 501: As we expect the security check credential to be saved in the session here, should we also save it to the session if we have to retrieve it?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -445,6 +445,23 @@ public JourneyResponse getJourneyResponseFromTicfCall(
             String ipAddress,
             AuditEventUser auditEventUser) {
         try {
+            // We must check if the security check credential on the session is empty.
+            // This can happen when an error occurs prior to the first call to get the
+            // security check credential e.g. in the check-existing-identity lambda.
```
DanCorderIPV, line 488: Are there any other effects we should consider/fix in other lambdas from having a missing security check credential?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: It's saved on the session as part of [fetchContraIndicatorsVc](https://github.com/govuk-one-login/ipv-core-back/blob/36536d041f54d6cfdb761ea8172b5290a848ac65/libs/cimit-service/src/main/java/uk/gov/di/ipv/core/library/cimit/service/CimitService.java#L113). The reason I reassign the output instead of just using `ipvSessionItem.getSecurityCheckCredential()` directly further down is because I wanted to make it clearer we're using it just for the old mitigations. It's not immediately obvious that `fetchContraIndicatorsVc` will save it onto the session and we also use `fetchContraIndicatorsVc` again below so wanted to avoid the confusion of using `ipvSessionItem.getSecurityCheckCredential()`

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: I don't think so, except from here where we might expect a `null` security check credential because a user goes down the technical_error journey prior to fetching their CIMIT VC, everywhere else (buildUserIdentity, criCheckingService, eventResolver) we expect there to be a non-null credential and so handle this by throwing an error if we do

thebauSoftwire: actually I just realised that `criCheckingService` isn't handling it properly either 🤦‍♀️ Thanks!

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -473,6 +473,25 @@ public JourneyResponse getJourneyResponseFromTicfCall(
...
+            // If it is, we need to make a call to CIMIT prior to getting the TICF VC
+            // in order to get the mitigation information unaffected by this new VC.
+            String previousSecurityCheckCredential = ipvSessionItem.getSecurityCheckCredential();
+            if (!StringUtils.isBlank(clientOAuthSessionItem.getScope())
```
thebauSoftwire: This is to handle the case where claims set validation fails as part of `initialise-ipv-session` where we will have a null scope, vtr etc. Have added here in case we still want to call ticf in this scenario, which is how it's currently being handled (cc @blakeyp, want to double check we should still be calling ticf in the case of a failed claims set validation? Or should we skip calling ticf altogether?)

blakeyp: @Joe-Edwards-GDS with the original TICF work was there a line between errors that we don't need to call TICF for vs ones that we do, or is it just all errors that we handle should go through TICF?

Joe-Edwards-GDS: That's a great question... The original guidance was to call TICF in every scenario (success or failure), but it's likely to be less important for technical errors like JAR validation issues.

If it'll make our lives easier, we could skip those cases (maybe double-check with someone like @johnstark-gds that it's not a problem)

thebauSoftwire: note from Slack convo with John: since we won't have some of the required fields to make a TICF call when we fail to parse the JAR request (e.g. sub will be null) and the failure happens before we emit any audit events (e.g. IPV_JOURNEY_START), we don't need to make a call to TICF

Joe-Edwards-GDS: OK cool, that's an option we can take then (but emphasis on the _if it'll make our lives easier_!)

thebauSoftwire: hmm not sure if it's technically _easier_ but I think it would save us a call to ticf, clean up our logs as the call to ticf would result in a 400 anyway and I think makes the code a bit more readable than leaving the above check in 🤔

lambdas/process-cri-callback/src/main/java/uk/gov/di/ipv/core/processcricallback/service/CriCheckingService.java
```diff
@@ -259,10 +259,15 @@ public JourneyResponse checkVcResponse(
             IpvSessionItem ipvSessionItem,
             List<VerifiableCredential> sessionVcs)
             throws CiRetrievalException, HttpResponseExceptionWithErrorBody, CiExtractionException,
-                    CredentialParseException, ConfigException {
-        var scopeClaims = clientOAuthSessionItem.getScopeClaims();
-        var isReverification = scopeClaims.contains(ScopeConstants.REVERIFICATION);
+                    CredentialParseException, ConfigException, MissingSecurityCheckCredential {
+        var isReverification = clientOAuthSessionItem.isReverification();
         if (!isReverification) {
+
```
blakeyp: stray line break?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -473,6 +473,34 @@ public JourneyResponse getJourneyResponseFromTicfCall(
             String ipAddress,
             AuditEventUser auditEventUser) {
         try {
+            // If we have an invalid ClientOauthSessionItem (e.g. as a result of failed JAR request
+            // validation), we cannot make a request to TICF as we will have missing required
+            // properties.
+            if (TRUE.equals(clientOAuthSessionItem.isErrorClientSession())) {
```
blakeyp: can this be just `if (clientOAuthSessionItem.isErrorClientSession())` ?

thebauSoftwire: ah good point 🤦‍♀️

blakeyp (approved):

PR #3130, opened 2025-04-14
Pyic 8180 nino mitigation

DanCorderIPV:
### What changed

Mitigation for NINO

### Why did it change

So users can carry on after a failed NINO check

### Issue tracking

- [PYIC-8180](https://govukverify.atlassian.net/browse/PYIC-8180)

[PYIC-8180]: https://govukverify.atlassian.net/browse/PYIC-8180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -420,6 +420,41 @@ states:
       fail-with-ci:
         targetJourney: FAILED
         targetState: FAILED_NINO
+        checkMitigation:
+          alternate-doc-invalid-nino:
+            targetState: MITIGATION_NINO_NO_MATCH
+            auditEvents:
+              - IPV_MITIGATION_START
+            auditContext:
+              mitigationType: invalid-nino
```
DanCorderIPV, line 429: In the event catalogue `mitigationType` is just a string so this shouldn't break anything, but do we need to tell anyone about the new possible value?

local-running/package.json
DanCorderIPV, line 1: This should never have been committed in the first place

api-tests/data/cri-stub-requests/nino/kenneth-with-breaching-ci/evidence.json
DanCorderIPV, line 1: Just making this a bit more realistic

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
thebauSoftwire: There's a [list of consumers](https://event-catalogue.internal.account.gov.uk/events/IPV_MITIGATION_START/#consumers) on the IPV_MITIGATION_START audit event - might be good to let them know?

I wonder if we should also update the event-catalogue to use an enum of expected values instead of just a string?

thebauSoftwire (approved):
Nice! One question around maybe updating event-catalogue

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
blakeyp: Yes it would be worth raising a catalogue PR with an enum and descriptions of each value and flag that we'll be sending this new value for NINO mitigation

DanCorderIPV:
Ticket has been abandoned

PR #3157, opened 2025-05-06
PYIC-8252: update core-back to use new PUT endpoint on EVCS

thebauSoftwire:
### What changed
Updates core-back to use new PUT endpoint for EVCS when the `storedIdentityServiceEnabled` feature flag is enabled.

### Why did it change
There will be a new PUT endpoint for EVCS which allows storing a stored identity object (containing information about the user's identity and their highest achieved VOT currently supported by ipv core). Eventually, we'll use this endpoint instead of the POST + UPDATE ones.

### Issue tracking

- [PYIC-8252](https://govukverify.atlassian.net/browse/PYIC-8252)

## Checklists

[PYIC-8252]: https://govukverify.atlassian.net/browse/PYIC-8252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -327,9 +327,7 @@ public class ProcessCandidateIdentityHandler
 
         // These identity types require the VCs from EVCS. To save multiple calls,
         // we call for them once here.
-        if (COI_CHECK_TYPES.contains(processIdentityType)
-                || STORE_IDENTITY_TYPES.contains(processIdentityType)
-                || PENDING.equals(processIdentityType)) {
+        if (requiresExistingVcsFromEvcs(processIdentityType)) {
             evcsUserVcs =
                     evcsService.getUserVCs(
                             userId,
@@ -359,15 +357,33 @@ public class ProcessCandidateIdentityHandler
             }
         }
 
+        boolean areVcsCorrelated = false;
+        VotMatchingResult votMatchingResult = null;
+        if (requiresVotMatchingResult(processIdentityType)) {
+            if (StringUtils.isBlank(ipvSessionItem.getSecurityCheckCredential())) {
+                return new JourneyErrorResponse(
+                                JOURNEY_ERROR_PATH,
+                                HttpStatusCode.INTERNAL_SERVER_ERROR,
+                                MISSING_SECURITY_CHECK_CREDENTIAL)
+                        .toObjectMap();
+            }
+
+            areVcsCorrelated = userIdentityService.areVcsCorrelated(sessionVcs);
+            votMatchingResult =
+                    getVotMatchingResult(
+                            ipvSessionItem, clientOAuthSessionItem, sessionVcs, areVcsCorrelated);
+        }
+
         if (PROFILE_MATCHING_TYPES.contains(processIdentityType)) {
             LOGGER.info(LogHelper.buildLogMessage("Performing profile evaluation"));
             var journey =
                     getJourneyResponseForProfileMatching(
                             ipvSessionItem,
-                            clientOAuthSessionItem,
                             deviceInformation,
                             sessionVcs,
-                            auditEventUser);
+                            auditEventUser,
+                            areVcsCorrelated,
+                            votMatchingResult);
 
             if (journey != null) {
                 return journey.toObjectMap();
@@ -388,14 +404,15 @@ public class ProcessCandidateIdentityHandler
                 // We still store a pending identity - it might be mitigating an existing CI
                 if (PENDING.equals(processIdentityType)) {
                     LOGGER.info(LogHelper.buildLogMessage("Storing identity"));
-                    storeIdentity(
+                    storeIdentityService.storeIdentity(
                             ipvSessionItem,
                             clientOAuthSessionItem,
                             processIdentityType,
                             deviceInformation,
                             sessionVcs,
                             auditEventUser,
-                            evcsUserVcs);
+                            evcsUserVcs,
+                            null);
                 }
                 return journey.toObjectMap();
             }
@@ -404,45 +421,46 @@ public class ProcessCandidateIdentityHandler
 
         if (STORE_IDENTITY_TYPES.contains(processIdentityType)) {
             LOGGER.info(LogHelper.buildLogMessage("Storing identity"));
-            storeIdentity(
+            storeIdentityService.storeIdentity(
                     ipvSessionItem,
                     clientOAuthSessionItem,
                     processIdentityType,
                     deviceInformation,
                     sessionVcs,
                     auditEventUser,
-                    evcsUserVcs);
+                    evcsUserVcs,
+                    votMatchingResult);
         }
 
         return JOURNEY_NEXT.toObjectMap();
     }
 
+    private boolean requiresVotMatchingResult(CandidateIdentityType processIdentityType) {
+        return (STORE_IDENTITY_TYPES.contains(processIdentityType)
+                        && !PENDING.equals(processIdentityType))
+                || PROFILE_MATCHING_TYPES.contains(processIdentityType);
+    }
+
+    private boolean requiresExistingVcsFromEvcs(CandidateIdentityType processIdentityType) {
+        return COI_CHECK_TYPES.contains(processIdentityType)
+                || STORE_IDENTITY_TYPES.contains(processIdentityType)
+                || PENDING.equals(processIdentityType);
+    }
+
     private JourneyResponse getJourneyResponseForProfileMatching(
             IpvSessionItem ipvSessionItem,
-            ClientOAuthSessionItem clientOAuthSessionItem,
             String deviceInformation,
             List<VerifiableCredential> sessionVcs,
-            AuditEventUser auditEventUser)
-            throws HttpResponseExceptionWithErrorBody, ParseException, CredentialParseException,
-                    CiExtractionException {
-        if (StringUtils.isBlank(ipvSessionItem.getSecurityCheckCredential())) {
-            return new JourneyErrorResponse(
-                    JOURNEY_ERROR_PATH,
-                    HttpStatusCode.INTERNAL_SERVER_ERROR,
-                    MISSING_SECURITY_CHECK_CREDENTIAL);
-        }
-
-        var areVcsCorrelated = userIdentityService.areVcsCorrelated(sessionVcs);
+            AuditEventUser auditEventUser,
+            boolean areVcsCorrelated,
+            VotMatchingResult votMatchingResult) {
 
         if (!areVcsCorrelated) {
             return JOURNEY_VCS_NOT_CORRELATED;
         }
 
-        var votMatches =
-                getVotMatchingResult(
-                        ipvSessionItem, clientOAuthSessionItem, sessionVcs, areVcsCorrelated);
+        var strongestRequestedMatch = votMatchingResult.strongestRequestedMatch();
 
-        var strongestRequestedMatch = votMatches.strongestRequestedMatch();
         if (strongestRequestedMatch.isEmpty()) {
             return JOURNEY_PROFILE_UNMET;
         }
@@ -459,7 +477,7 @@ public class ProcessCandidateIdentityHandler
             }
             sendProfileMatchedAuditEvent(
                     profile.get(),
-                    votMatches.gpg45Scores(),
+                    votMatchingResult.gpg45Scores(),
                     VcHelper.filterVCBasedOnProfileType(sessionVcs, ProfileType.GPG45),
                     auditEventUser,
                     deviceInformation);
@@ -572,36 +590,6 @@ public class ProcessCandidateIdentityHandler
                     HttpStatusCode.INTERNAL_SERVER_ERROR,
                     ERROR_PROCESSING_TICF_CRI_RESPONSE);
         }
-    }
-
-    private void storeIdentity(
-            IpvSessionItem ipvSessionItem,
-            ClientOAuthSessionItem clientOAuthSessionItem,
-            CandidateIdentityType processIdentityType,
-            String deviceInformation,
-            List<VerifiableCredential> sessionVcs,
-            AuditEventUser auditEventUser,
-            List<EvcsGetUserVCDto> evcsUserVcs)
-            throws EvcsServiceException, HttpResponseExceptionWithErrorBody, CiExtractionException,
-                    CredentialParseException, ParseException {
-        var areVcsCorrelated = userIdentityService.areVcsCorrelated(sessionVcs);
-
-        VotMatchingResult matchingVot = null;
-        if (!PENDING.equals(processIdentityType)) {
-            matchingVot =
-                    getVotMatchingResult(
-                            ipvSessionItem, clientOAuthSessionItem, sessionVcs, areVcsCorrelated);
-        }
-
-        storeIdentityService.storeIdentity(
-                ipvSessionItem,
-                clientOAuthSessionItem,
-                processIdentityType,
-                deviceInformation,
-                sessionVcs,
-                auditEventUser,
-                evcsUserVcs,
-                matchingVot);
     }
 
     private void sendProfileMatchedAuditEvent(
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -59,6 +59,7 @@ import uk.gov.di.ipv.core.library.ticf.TicfCriService;
 import uk.gov.di.ipv.core.library.ticf.exception.TicfCriServiceException;
 import uk.gov.di.ipv.core.library.useridentity.service.UserIdentityService;
 import uk.gov.di.ipv.core.library.useridentity.service.VotMatcher;
+import uk.gov.di.ipv.core.library.useridentity.service.VotMatchingResult;
 import uk.gov.di.ipv.core.library.verifiablecredential.helpers.VcHelper;
 import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredentialsService;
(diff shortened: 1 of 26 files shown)


thebauSoftwire:
**Some notes:**
`storePendingVc` (which is used by processAsyncCriCredential) doesn't use the PUT endpoint and still uses the old one as we want to just add the single pending VC to EVCS without updating any existing ones which is what the PUT endpoint now does.

`storeInheritedIdentity` also doesn't use the new endpoint. When storing an inherited identity to EVCS we also update existing inherited identity VCs to HISTORIC. To minimise changes around inherited identities across the stub and core-back, `storeInheritedIdentity` still uses the old endpoints.

local-running/core.local.params.yaml
```diff
@@ -78,6 +78,8 @@ core:
     apiBaseUrl: "https://cimit-api.stubs.account.gov.uk"
   evcs:
     applicationUrl: "https://evcs.stubs.account.gov.uk"
+  storedIdentityService:
+    componentId: "https://reuse-identity.build.account.gov.uk"
```
thebauSoftwire, line 82: the SIS identifier hasn't been confirmed yet but they expect it'll look something like this

thebauSoftwire:
API tests all passing with feature flag turned on 🍏

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -311,6 +312,7 @@ private CoiCheckType getCoiCheckType(
         return CoiCheckType.STANDARD;
     }
 
+    @SuppressWarnings("java:S3776") // Cognitive Complexity of methods should not be too high
```
DanCorderIPV: I'm not keen on suppressing these errors. If we think it's too strict we should consider upping the limit, otherwise we should look at making this method easier to follow.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -326,9 +328,7 @@ private Map<String, Object> processCandidateThroughJourney(
 
         // These identity types require the VCs from EVCS. To save multiple calls,
         // we call for them once here.
-        if (COI_CHECK_TYPES.contains(processIdentityType)
-                || STORE_IDENTITY_TYPES.contains(processIdentityType)
-                || PENDING.equals(processIdentityType)) {
+        if (requiresExistingVcsFromEvcs(processIdentityType)) {
```
DanCorderIPV, line 334: Nice

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -358,15 +358,33 @@ private Map<String, Object> processCandidateThroughJourney(
             }
         }
 
+        boolean areVcsCorrelated = false;
+        VotMatchingResult votMatchingResult = null;
+        if (requiresVotMatchingResult(processIdentityType)) {
+            if (StringUtils.isBlank(ipvSessionItem.getSecurityCheckCredential())) {
```
DanCorderIPV: Would this be neater as `throw new HttpResponseExceptionWithErrorBody()` inside `getVotMatchingResult()`?

Here it's not clear why we care about the `SecurityCheckCredential` being present

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/StoreIdentityService.java
```diff
@@ -45,19 +47,22 @@ public StoreIdentityService(
         this.evcsService = evcsService;
     }
 
+    @SuppressWarnings("java:S107") // Methods should not have too many parameters
```
DanCorderIPV: Do we really need to suppress this warning? Can we do something else?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
DanCorderIPV: This would also bring the method complexity down below the threshold :)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -389,12 +407,13 @@ private Map<String, Object> processCandidateThroughJourney(
                     LOGGER.info(LogHelper.buildLogMessage("Storing identity"));
                     storeIdentityService.storeIdentity(
                             ipvSessionItem,
-                            userId,
+                            clientOAuthSessionItem,
```
DanCorderIPV: Why change this? AFAICS the only thing the clientOAuthSessionItem is ever used for is to get the userId

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/StoreIdentityService.java
```diff
@@ -45,19 +47,22 @@ public StoreIdentityService(
...
             IpvSessionItem ipvSessionItem,
-            String userId,
+            ClientOAuthSessionItem clientOAuthSessionItem,
             CandidateIdentityType identityType,
             String deviceInformation,
             List<VerifiableCredential> credentials,
             AuditEventUser auditEventUser,
-            List<EvcsGetUserVCDto> evcsVcs)
+            List<EvcsGetUserVCDto> evcsVcs,
+            VotMatchingResult matchingVot)
```
DanCorderIPV: It looks like the VotMatchingResult here is only ever used to get the strongest match. Can we just pass that in instead to minimise the reach of VotMatchingResult?

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
```diff
@@ -1700,6 +1700,54 @@ void findIdentityReturnsIdentityClaimForOperationalVcWithNoEvidence() throws Exc
...
+        // Assert
+        assertEquals(3, result.size());
+        // First element in the array is the identity claim
+        assertEquals("KENNETH DECERQUEIRA", ((IdentityClaim) result.get(0)).getFullName());
+        // Second element in the array is the address claim if it exists
+        assertEquals("IDSWORTH ROAD", ((PostalAddress) result.get(1)).getStreetName());
+        // Third element is the passport claim if it exists
+        assertEquals("321654987", ((PassportDetails) result.get(2)).getDocumentNumber());
+    }
+
```
DanCorderIPV, line 1753: Assuming that the claims stay in this form we should have a test that checks what happens with VCs with multiple claims.

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -285,6 +285,19 @@ private void addUserIdentityClaims(
         ninoClaim.ifPresent(userIdentityBuilder::ninoClaim);
     }
 
+    public List<Object> getUserClaimsForStoredIdentity(
+            Vot achievedVot, List<VerifiableCredential> vcs)
+            throws HttpResponseExceptionWithErrorBody, CredentialParseException {
+        // Substituting arbitrary values into vcs, sub and vtm as they are required
+        // properties for the builder. We just need the builder to add claims to.
+        var userIdentityBuilder = UserIdentity.builder().vcs(List.of()).sub("").vtm("");
```
DanCorderIPV: As discussed, we're not very happy with this, but we'll wait until we have an example SI object before working out exactly what to do

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/dto/EvcsStoredIdentityDto.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.di.ipv.core.library.evcs.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import uk.gov.di.ipv.core.library.enums.Vot;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record EvcsStoredIdentityDto(String jwt, Vot vot, Object metadata) {}
```
DanCorderIPV: Presumably the metadata is going to be defined in another ticket?

If so I'd like to either not include it here at all, or to just create an empty metadata class so that we can us it instead of `Object`

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
```diff
@@ -83,4 +87,15 @@ public UserIdentity(
         this.vtm = vtm;
         this.returnCode = returnCode;
     }
+
+    @JsonIgnore
+    public List<Object> getAllClaims() {
```
DanCorderIPV: Using `Object` in Java is a bit of a smell. The claims objects currently don't have a common base class, but I'd be inclined to add an abstract base class that they could all share and you could use here.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java
```diff
@@ -0,0 +1,111 @@
...
+        this.configService = configService;
+    }
+
+    private JWTClaimsSet createStoredIdentityJwt(
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            List<VerifiableCredential> vcs,
+            VotMatchingResult votMatchingResult)
+            throws HttpResponseExceptionWithErrorBody, CredentialParseException,
+                    FailedToCreateStoredIdentityForEvcsException {
+        Instant now = Instant.now();
```
DanCorderIPV: As a general thing, any time that you have a class using `now()` or similar it's worth considering passing in a `Clock` object in the constructor so that in testing we can pass in a fixed clock and get a predictable output

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    public EvcsStoredIdentityDto getStoredIdentityForEvcs(
+            ClientOAuthSessionItem clientOAuthSessionItem,
+            List<VerifiableCredential> vcs,
+            VotMatchingResult votMatchingResult,
+            Object metadata)
+            throws FailedToCreateStoredIdentityForEvcsException {
+        var strongestMatchedVot = votMatchingResult.strongestMatch();
```
DanCorderIPV: Not sure it's worth checking this here, you also test it in `createStoredIdentityJwt()` where it's actually used

DanCorderIPV (changes requested):

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/dto/EvcsStoredIdentityDto.java
blakeyp: I don't think we're going to be using `metadata` at all, in which case let's leave it off for now

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: ahh thanks :) not sure why I put that outside in the function in the first place 🫠

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: ahh this was a remnant of an initial change I had made and forgot to undo 🤦‍♀️ thanks!

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/service/StoreIdentityService.java
thebauSoftwire: The strongest requested vot is used to build the stored identity object in `StoredIdentityService.createStoredIdentityJwt`

thebauSoftwire: oh actually BuildUserIdentity uses the vot on the session, so to keep it consistent I'll change it to that in case the way it's set changes, thanks

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/UserIdentity.java
thebauSoftwire: don't need this anymore as claims should be structured similar to how it is for UserIdentity response so marking this as resolved :)

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
thebauSoftwire: having spoken to Chris and David, we've clarified that the claims should look like this:
```
{
  ...
  "claims": {
    "https://vocab.account.gov.uk/v1/passport" : [ {
      "expiryDate" : "2030-01-01",
      "icaoIssuerCode" : "GBR",
      "documentNumber" : "321654987"
    } ],
    "https://vocab.account.gov.uk/v1/coreIdentity" : {
      "name" : [ {
        "nameParts" : [ {
          "type" : "GivenName",
          "value" : "KENNETH"
        }, {
          "type" : "FamilyName",
          "value" : "DECERQUEIRA"
        } ]
      } ],
      "birthDate" : [ {
        "value" : "1965-07-08"
      } ]
    },
    "https://vocab.account.gov.uk/v1/address" : [ {
      "addressCountry" : "GB",
      "buildingName" : "",
      "streetName" : "HADLEY ROAD",
      "postalCode" : "BA2 5AA",
      "buildingNumber" : "8",
      "addressLocality" : "BATH",
      "validFrom" : "2000-01-01"
    } ]
  }
}
```

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java
thebauSoftwire: the vot in the stored identity jwt is the achieved _requested_ vot by the user (the same as what we have in the UserIdentity response) - have updated this though to take in this vot separately so it is more consistent with how BuildUserIdentity assigns the vot in its response

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
Joe-Edwards-GDS: FWIW the principle behind the SIS record design was to make it as similar as possible to the `UserIdentity` object returned to orchestration.

I don't think it's too late to tweak that to make things easier (e.g. adding a `vtm`) - any differences were probably unintentional!

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
thebauSoftwire: claims are no longer an array and should match what the claims look like in UserIdentity response (see [here](https://github.com/govuk-one-login/ipv-core-back/pull/3157#discussion_r2079778704))

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
thebauSoftwire: have also confirmed that `vtm` and `returnCode` are not needed in the stored identity object

DanCorderIPV (approved):

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
```diff
@@ -188,31 +211,66 @@ private void persistUserVCs(
...
+                                                                : CURRENT,
+                                                        null,
+                                                        ONLINE))
+                                .toList();
+                if (!CollectionUtils.isEmpty(existingEvcsUserVCs))
+                    updateExistingUserVCs(
+                            userId, credentials, existingEvcsUserVCs, isPendingIdentity);
+                if (!userVCsToStore.isEmpty()) evcsClient.storeUserVCs(userId, userVCsToStore);
+            }
+        } catch (FailedToCreateStoredIdentityForEvcsException e) {
```
blakeyp, line 270: Are we failing open here as per AC scenario 3? i.e. if anything goes wrong in the construction of the SIS record request or we get an error response, we should fall back to the POST request with no impact to the user journey.

I think this will also impact how we set the `sis_record_created` extension in #3162 (which should only be set to true if the PUT request succeeds) cc @undermad

thebauSoftwire: ah I forgot about this AC 🤦‍♀️ thanks for the spot!

PR #3161, opened 2025-05-09
PYIC-8284 Break out mermaid rendering steps

Joe-Edwards-GDS:
### What changed

Pull out the code that actually renders the mermaid graph

### Why did it change

So it can be tested separately to the full rendering engine.

### Issue tracking

- [PYIC-8284](https://govukverify.atlassian.net/browse/PYIC-8284)

## Checklists

[PYIC-8284]: https://govukverify.atlassian.net/browse/PYIC-8284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/src/helpers/contract-nested.ts
+++ b/journey-map/src/helpers/contract-nested.ts
@@ -1,0 +1,18 @@
+import { JourneyState } from "../types.js";
+
+// Turn a nested journey state into a standard response state
+export const contractNestedJourneys = (
+  journeyMap: Record<string, JourneyState>,
+): void => {
+  Object.values(journeyMap).forEach((state) => {
+    if (state.nestedJourney) {
+      state.response = {
+        type: "nestedJourney",
+        nestedJourney: state.nestedJourney,
+      };
+      delete state.nestedJourney;
+      state.events = state.exitEvents;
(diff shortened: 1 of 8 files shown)


journey-map/src/helpers/contract-nested.ts
```diff
@@ -0,0 +1,18 @@
...
+): void => {
+  Object.values(journeyMap).forEach((state) => {
+    if (state.nestedJourney) {
+      state.response = {
+        type: "nestedJourney",
+        nestedJourney: state.nestedJourney,
+      };
+      delete state.nestedJourney;
+      state.events = state.exitEvents;
+      delete state.exitEvents;
```
Joe-Edwards-GDS, line 15: I'm somewhat of the opinion that `exitEvents` should just be `events` on nested journeys anyway! Unfortunately it's not totally trivial to do because Jackson gets upset about deducing the difference between nested/basic states.

It would simplify a lot of logic along the lines of 'use events, unless it's a nested journey and then use exit events'

blakeyp (approved):

PR #3165, opened 2025-05-15
BAU: Add dynatrace trace id to logs

Joe-Edwards-GDS:
### What changed

Add trace id (and span id, for completeness) to the logs

### Why did it change

So we can link traces in dynatrace with specific logs

We should also update https://team-manual.account.gov.uk/teams/IPV-Core-team/How-IPV-Core-Works/distributed-tracing/


lambdas/build-user-identity/src/main/java/uk/gov/di/ipv/core/builduseridentity/BuildUserIdentityHandler.java
```diff
@@ -97,6 +97,8 @@ public BuildUserIdentityHandler() {
     @Metrics(captureColdStart = true)
     public APIGatewayProxyResponseEvent handleRequest(
             APIGatewayProxyRequestEvent input, Context context) {
+        LogHelper.attachTraceId();
+        LogHelper.attachComponentId(configService);
```
Joe-Edwards-GDS, line 101: This didn't include the component id previously - assuming that wasn't intentional, but happy to remove if it is.

thebauSoftwire (approved):

PR #3168, opened 2025-05-16
Pyic 8246 initial ais call

DanCorderIPV:
### What changed

Add code to check whether the user needs to reprove their identity by calling AIS instead of using the value from orchestration

### Why did it change

At some point orchestration will stop sending their value.

### Issue tracking

- [PYIC-8246](https://govukverify.atlassian.net/browse/PYIC-8246)

## Checklists

- [ x ] API/unit/contract tests have been written/updated

[PYIC-8246]: https://govukverify.atlassian.net/browse/PYIC-8246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -202,6 +208,19 @@ public APIGatewayProxyResponseEvent handleRequest(
                                     getJarUserInfo(claimsSet).map(JarUserInfo::evcsAccessToken),
                                     claimsSet));
 
+            if (configService.enabled(AIS_ENABLED)) {
```
DanCorderIPV, line 212: I've kept this in `InitialiseIpvSession` as the flag is used in audit events raised here.

DanCorderIPV:
It might be nice to move the reproveIdentity flag out of the ClientOAuthSessionItem but if we do that it probably deserves its own ticket as it will need to be multi step process to not break inflight journeys.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/client/AisClient.java
```diff
@@ -0,0 +1,162 @@
...
+                    RETRY_DELAY_MILLIS,
+                    () -> {
+                        LOGGER.info(LogHelper.buildLogMessage("Sending HTTP request to AIS"));
+                        try {
+                            var response =
+                                    httpClient.send(
+                                            aisHttpRequest, HttpResponse.BodyHandlers.ofString());
+                            checkResponseStatusCode(response);
+                            return response;
+                        } catch (IOException e) {
```
DanCorderIPV, line 109: The EvcsClient code that is similar to this treats IOExceptions as non-retriable, and only retries on 401 (unauthorised) and 429 (too many requests).
I would have though IOExceptions are likely to be transitory things and worth retrying whereas 401 seems unlikely to be fixed by retrying.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/dto/AccountInterventionStatusDto.java
```diff
@@ -0,0 +1,62 @@
...
+import uk.gov.di.ipv.core.library.ais.enums.AisInterventionType;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class AccountInterventionStatusDto {
+    @JsonProperty private Intervention intervention;
```
MikeCollingwood: It is ok to omit the `@JsonProperty` tag when the attribute name matches the json attribute.

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
```diff
@@ -341,6 +346,57 @@ void shouldReturnIpvSessionIdAndSendAuditEventWhenProvidedValidReproveRequest()
...
+        Map<String, Object> responseBody =
+                OBJECT_MAPPER.readValue(response.getBody(), new TypeReference<>() {});
+
+        assertEquals(HttpStatusCode.OK, response.getStatusCode());
+        assertEquals(ipvSessionItem.getIpvSessionId(), responseBody.get("ipvSessionId"));
+
+        ArgumentCaptor<AuditEvent> auditEventCaptor = ArgumentCaptor.forClass(AuditEvent.class);
+        verify(mockAuditService, times(2)).sendAuditEvent(auditEventCaptor.capture());
+        var capturedValues = auditEventCaptor.getAllValues();
+        assertEquals(AuditEventTypes.IPV_JOURNEY_START, capturedValues.get(0).getEventName());
```
thebauSoftwire, line 380: there's a `reprove_identity` on the extensions of `IPV_JOURNEY_START` audit event. Should we check that too?

api-tests/src/steps/ipv-steps.ts
```diff
@@ -206,7 +208,7 @@ When(
 
 // Variant of the journey start that retries, e.g. to wait for an async F2F request
 When(
-  /I start a new '([\w-]+)' journey( with reprove identity)? and return to a '([\w-]+)' page response$/,
+  /I start new '([\w-]+)' journeys( with reprove identity)? until I get a '([\w-]+)' page response$/,
```
MikeCollingwood, line 211: I agree the phrase needed an update though I think it would make more sense implying you're attempting to start the journey.
```suggestion
  /I attempt to start a new '([\w-]+)' journey( with reprove identity)? until I get a '([\w-]+)' page response$/,
```

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood:
```suggestion
            if (!isReverification && configService.enabled(AIS_ENABLED)) {
```
If reverification, can we skip over all of this logic?

api-tests/src/steps/ais-steps.ts
```diff
@@ -0,0 +1,28 @@
...
+  "AIS_FORCED_USER_IDENTITY_VERIFY",
+  "AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY",
+];
+
+When(
+  "The AIS stub will return an {string} result",
+  async function (this: World, desiredApiResult: string): Promise<void> {
+    this.userId = this.userId ?? getRandomString(16);
+
+    if (!VALID_RESPONSE_TYPES.includes(desiredApiResult)) {
```
MikeCollingwood, line 22: Minor but could this instead use the enum `AisInterventionType`?

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/enums/AisAuditLevel.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.di.ipv.core.library.ais.enums;
+
+// Suppress enums should be upper case rule. These have to match the values used on the AIS API
+@SuppressWarnings("java:S115")
+public enum AisAuditLevel {
+    standard,
+    enahnced
```
MikeCollingwood:
```suggestion
    enhanced
```

Could this enum not be put into the Dto class?

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/enums/AisInterventionType.java
```diff
@@ -0,0 +1,12 @@
...
+public enum AisInterventionType {
+    AIS_NO_INTERVENTION,
+    AIS_ACCOUNT_SUSPENDED,
+    AIS_ACCOUNT_UNSUSPENDED,
+    AIS_ACCOUNT_BLOCKED,
+    AIS_ACCOUNT_UNBLOCKED,
+    AIS_FORCED_USER_PASSWORD_RESET,
+    AIS_FORCED_USER_IDENTITY_VERIFY,
+    AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY
+}
```
thebauSoftwire, line 12: do we use this anywhere else? Can we just put this in the dto?

api-tests/src/steps/ais-steps.ts
```diff
@@ -0,0 +1,28 @@
...
+const VALID_RESPONSE_TYPES = [
+  "AIS_NO_INTERVENTION",
+  "AIS_ACCOUNT_SUSPENDED",
+  "AIS_ACCOUNT_UNSUSPENDED",
+  "AIS_ACCOUNT_BLOCKED",
+  "AIS_ACCOUNT_UNBLOCKED",
+  "AIS_FORCED_USER_PASSWORD_RESET",
+  "AIS_FORCED_USER_IDENTITY_VERIFY",
+  "AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY",
+];
```
thebauSoftwire: Minor thing but could we make this an enum and use that enum to type the argument for `primeResponseForUser`?
```
export enum ValidResponseTypes {
"AIS_NO_INTERVENTION",
  "AIS_ACCOUNT_SUSPENDED",
  "AIS_ACCOUNT_UNSUSPENDED",
  "AIS_ACCOUNT_BLOCKED",
  "AIS_ACCOUNT_UNBLOCKED",
  "AIS_FORCED_USER_PASSWORD_RESET",
  "AIS_FORCED_USER_IDENTITY_VERIFY",
  "AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY",
}

When(
  "The AIS stub will return an {string} result",
  async function (this: World, desiredApiResult: string): Promise<void> {
  ...
  if (!(desiredApiResult.toUpperCase() in ValidResponseTypes)) {...}
  ...  
}

# ais-management-api.ts
export const primeResponseForUser = async (
  userId: string,
  desiredResponse: ValidResponseType,
)
```

thebauSoftwire:
sorry for the double review! Didn't realise Mike had already started taking a look at it

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
DanCorderIPV: I considered that, but when we turn the flag on we don't want the behaviour to be influenced by the orch request, otherwise when that gets turned off our behaviour may change.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/client/AisClient.java
DanCorderIPV: @blakeyp If you agree with this, is it worth raising a tech debt ticket for the EvcsClient code?

lambdas/initialise-ipv-session/src/test/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandlerTest.java
DanCorderIPV: We could, but the full details of the audit events are checked in the audit event API tests so I don't think it's vital we do it here.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/enums/AisAuditLevel.java
DanCorderIPV: It could, but it is used in the TestData class and this is more in line with the other services we have
It's marginal, but I think the DTO class is already a bit big with all the subclasses

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/enums/AisInterventionType.java
DanCorderIPV: We could, but it is used in the TestData class and this is more in line with the other services we have.
It's marginal, but I think the DTO class is already a bit big with all the subclasses

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/dto/AccountInterventionStatusDto.java
MikeCollingwood: I think this is still the case

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
MikeCollingwood: I guess, though with isReverification the clientOAuthSession updates would be redundant & a change from what we had originally

MikeCollingwood (approved):
Looks good to me, thank you for the clarifications

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/client/AisClient.java
blakeyp: @DanCorderIPV we have some special handling for IOExceptions in the base http client where we retry on certain types of IOException and fail hard on anything else (due to the recent incidents where lambda instances appear to lose network connectivity and all requests fail with IOException)
https://github.com/govuk-one-login/ipv-core-back/blob/c8c8a91da1fb2b6a64f2f0b176503846729b38da/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/tracing/TracingHttpClient.java#L56

PR #3171, opened 2025-05-19
PYIC-6241 Configure lambda triggers for dcmaw async queues

blakeyp:
### What changed

In dev and build we use a single shared stub queue hosted in our stubs account for testing both F2F and DCMAW async journeys. In staging/int/prod we need to support separate queues owned by F2F (existing) and DCMAW (new) in their own accounts. Set up new conditional EventSourceMappings to support this and set required KMS permissions on lambda function.

### Why did it change

Ahead of staging testing and go-live of the new v2 app journey.

### Issue tracking
- [PYIC-6241](https://govukverify.atlassian.net/browse/PYIC-6241)

[PYIC-6241]: https://govukverify.atlassian.net/browse/PYIC-6241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -2291,10 +2291,9 @@ Resources:
                 - IsDevOrBuild
                 - !If
                   - IsDevelopment
-                  - !Join
-                    - ''
-                    - - !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
-                      - !Sub "_${Environment}"
+                  - !Sub
+                    - "${BaseArn}_${Environment}"
+                    - BaseArn: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
                   - !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
                 - "arn:aws:sqs:region:000000000000:not-used" # to satisfy linter - never actually used
             - Sid: f2fAsyncCriResponseQueuePermission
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -2254,39 +2281,66 @@ Resources:
...
+                - sqs:GetQueueAttributes
+                - sqs:ReceiveMessage
+              Resource: !If
+                - IsDevOrBuild
+                - !If
+                  - IsDevelopment
+                  - !Join
+                    - ''
+                    - - !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
+                      - !Sub "_${Environment}"
```
MikeCollingwood:
```suggestion
                  - !Sub
                    - "${BaseArn}_${Environment}"
                    - BaseArn: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
```

Would this be better than a join operator?

deploy/template.yaml
```diff
@@ -2254,39 +2281,66 @@ Resources:
...
+                  - !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, sharedStubAsyncCriResponseQueueArn ]
+                - ""
```
MikeCollingwood: I think `!Ref AWS::NoValue` can be used for this case

blakeyp: The linter still complains not a valid ARN, so I've had to set dummy ARN values 🤦‍♂️

MikeCollingwood (approved):

PR #3179, opened 2025-05-22
PYIC-8327: Remove kidJarHeaderEnabled feature flag.

undermad:
### What changed

Features Set - KID_JAR_HEADER removed and cleaned places that depends on it.

### Why did it change

As kidJarHeaderEnabled feature flag is enabled across all environments, it can be removed from the project.

### Issue tracking

- [PYIC-8327](https://govukverify.atlassian.net/browse/PYIC-8327)
- [ipv-core-common-infra](https://github.com/govuk-one-login/ipv-core-common-infra/pull/1228)

## Checklists

[PYIC-8327]: https://govukverify.atlassian.net/browse/PYIC-8327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
+++ b/libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
@@ -52,7 +52,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Objects;
 
-import static uk.gov.di.ipv.core.library.config.CoreFeatureFlag.KID_JAR_HEADER;
 import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.ENVIRONMENT;
 import static uk.gov.di.ipv.core.library.domain.Cri.DCMAW;
 import static uk.gov.di.ipv.core.library.domain.Cri.DWP_KBV;
@@ -196,10 +195,7 @@ public class CriApiService {
                             .build();
 
             var signedClientJwt =
-                    JwtHelper.createSignedJwt(
-                            clientAuthClaimsSet,
-                            signerFactory.getSigner(),
-                            configService.enabled(KID_JAR_HEADER));
+                    JwtHelper.createSignedJwt(clientAuthClaimsSet, signerFactory.getSigner(), true);
             var clientAuthentication = new PrivateKeyJWT(signedClientJwt);
             var redirectionUri = criConfig.getClientCallbackUrl();
             var authorizationGrant = new AuthorizationCodeGrant(authorizationCode, redirectionUri);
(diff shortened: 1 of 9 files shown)


libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
```diff
@@ -196,10 +195,7 @@ private HTTPRequest buildAccessTokenRequestWithJwtAuthenticationAndAuthorization
                             .build();
 
             var signedClientJwt =
-                    JwtHelper.createSignedJwt(
-                            clientAuthClaimsSet,
-                            signerFactory.getSigner(),
-                            configService.enabled(KID_JAR_HEADER));
+                    JwtHelper.createSignedJwt(clientAuthClaimsSet, signerFactory.getSigner(), true);
```
AmritSidhu: I think the `true` can to be removed here. If its always `true` we can remove the `boolean includeKid` from the method?

AmritSidhu (commented):
`kidJarHeaderEnabled` can be removed from the `/local-running/core.local.params.yaml` file too

libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
undermad: `StoredIdentityService.Java` actually use this method - set to `false` - and doesn't depend on the feature flag.
https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java#L97

undermad: StoredIdentityService.java actually use this method - set to false - and doesn't depend on the feature flag.
https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java#L97

Joe-Edwards-GDS: That sounds like a mistake to me - IMO we should always be including a `kid` (otherwise it's impossible to keep track of the key used)

undermad: I removed this boolean value then and change StroredIdentityService to include kid in jwt header.

blakeyp (approved):

PR #3202, opened 2025-06-04
PYIC-8361: update api tests for when strategic app is turned on

thebauSoftwire:
### What changed and why
Updating tests that don't use the strategic app to explicitly use the `disableStrategicApp` feature set so all existing tests can co-exist when the V2 app goes live.

### Issue tracking

- [PYIC-8361](https://govukverify.atlassian.net/browse/PYIC-8361)

## Checklists

[PYIC-8361]: https://govukverify.atlassian.net/browse/PYIC-8361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/account-intervention/journey-ending-interventions.feature
+++ b/api-tests/features/account-intervention/journey-ending-interventions.feature
@@ -2,7 +2,7 @@
 Feature: Journey ending interventions
 
   Scenario Outline: <intervention> intervention at <when> of identity proving journey
-    Given I activate the 'accountInterventions' feature set
+    Given I activate the 'accountInterventions,disableStrategicApp' feature set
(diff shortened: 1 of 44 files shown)


DanCorderIPV (approved):
This makes sense and means that we keep testing what's in prod.
Do you know what the plan is for these tests once the v2 app goes live in prod? Will they all be updated with the new routing?

blakeyp:
@DanCorderIPV once v2 app is permanently switched on and we've exited the ELS (Early Live Support) period we'll want to decommission the old v1 app routes, which should include updating/removing old tests

PR #3206, opened 2025-06-06
PYIC-8258: store reuse identities with PUT evcs endpoint

thebauSoftwire:
### What changed
Store reuse identities (inc inherited identities)

### Why did it change
Returning users will have valid credentials that will need to have a corresponding Stored Identity Record. During phase 1, we aren't able to query SIS so IPV Core must assume that a returning user does not have an existing record and write one proactively so that when SIS is rolled out, returning users do not have to go through an identity proofing journey again.

### Issue tracking

- [PYIC-8258](https://govukverify.atlassian.net/browse/PYIC-8258)

## Checklists

[PYIC-8258]: https://govukverify.atlassian.net/browse/PYIC-8258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/stored-identity/p2-journeys.feature
+++ b/api-tests/features/stored-identity/p2-journeys.feature
@@ -120,6 +120,7 @@ Feature: Stored Identity - P2 journeys
       | dcmaw   | kenneth-passport-valid |
       | address | kenneth-current        |
       | fraud   | kenneth-score-2        |
+    And I don't have a stored identity in EVCS
 
     When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-reuse' page response
--- a/api-tests/features/stored-identity/p2-journeys.feature
+++ b/api-tests/features/stored-identity/p2-journeys.feature
@@ -113,3 +113,18 @@ Feature: Stored Identity - P2 journeys
       When I use the OAuth response to get my identity
       Then I get a 'P2' identity
       And I have a 'GPG45' stored identity record type with a 'P2' vot
+
+  Scenario: Reuse journey - identity is stored to EVCS
+    Given the subject already has the following credentials
+      | CRI     | scenario               |
+      | dcmaw   | kenneth-passport-valid |
+      | address | kenneth-current        |
+      | fraud   | kenneth-score-2        |
+
+    When I start a new 'medium-confidence' journey
+    Then I get a 'page-ipv-reuse' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+    And I have a 'GPG45' stored identity record type with a 'P2' vot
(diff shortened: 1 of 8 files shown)


thebauSoftwire:
Manually tested the stored identities for reuse journeys with and w/out inherited identities 🍏

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -570,17 +563,25 @@ private void storeCandidateIdentity(
...
+        return (shouldStoreIdentity(processIdentityType) && !PENDING.equals(processIdentityType))
                 || PROFILE_MATCHING_TYPES.contains(processIdentityType);
     }
 
     private boolean requiresExistingVcsFromEvcs(CandidateIdentityType processIdentityType) {
         return COI_CHECK_TYPES.contains(processIdentityType)
-                || STORE_IDENTITY_TYPES.contains(processIdentityType)
+                || shouldStoreIdentity(processIdentityType)
                 || PENDING.equals(processIdentityType);
     }
```
undermad: This become hard to read boolean logic to me. I would suggest something like this.
   
```
private boolean requiresVotMatchingResult(CandidateIdentityType processIdentityType) {
        if (configService.enabled(STORED_IDENTITY_SERVICE) && EXISTING.equals(processIdentityType)) {
            return true;
        }

        var combinedList = Stream.concat(PROFILE_MATCHING_TYPES.stream(), STORE_IDENTITY_TYPES.stream())
                .filter(identityType -> !identityType.equals(PENDING))
                .distinct()
                .toList();

        return combinedList.contains(processIdentityType);
    }

    private boolean requiresExistingVcsFromEvcs(CandidateIdentityType processIdentityType) {
        if (configService.enabled(STORED_IDENTITY_SERVICE) && EXISTING.equals(processIdentityType)) {
            return true;
        }

        return COI_CHECK_TYPES.contains(processIdentityType) || STORE_IDENTITY_TYPES.contains(processIdentityType);
    }
```

api-tests/features/stored-identity/p2-journeys.feature
```diff
@@ -113,3 +113,18 @@ Feature: Stored Identity - P2 journeys
...
       Then I get a 'P2' identity
       And I have a 'GPG45' stored identity record type with a 'P2' vot
+
+  Scenario: Reuse journey - identity is stored to EVCS
+    Given the subject already has the following credentials
+      | CRI     | scenario               |
+      | dcmaw   | kenneth-passport-valid |
+      | address | kenneth-current        |
+      | fraud   | kenneth-score-2        |
+
```
undermad, line 124: Can we add additional line that check the SIS record doesn't exist. This may be more verbose and indicate intention of that test.

undermad (changes requested):
Looking great, just few comments.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: thanks for this :) have made similar changes but changed `configService.enabled(STORED_IDENTITY_SERVICE) && EXISTING.equals(processIdentityType)` to use the function below (`shouldStoreExistingIdentity`) as I thought it was more readable. With this, `requiresExistingVcsFromEvcs` could also be simplified, thanks!

undermad (approved):

PR #3212, opened 2025-06-10
PYIC-8305: Build new lambda for manual pending F2F resets

AmritSidhu:
### What changed

Build new lambda for manual pending F2F resets
- input event should be a string containing a single user id 
- it should not be possible to process more than one user id at a time 
- should error if user id empty/not provided
- should delete only the F2F pending record(s) for the given user id (from the cri-response table) 
- should error if no F2F pending records exist for given user id response 
- should be json with ‘success’/'error' result and friendly error message if error 
- should log on success and error (do not log user id as PII)

### Why did it change

Currently pending F2F resets are carried out locally using a basic bash script. Until we have self-service pending F2F reset in place we want to migrate the manual resets process to Technical Service Desk (TSD). To do this we should build a lambda for TSD to use via the AWS Console, making the process more robust, accessible and auditable.

### Issue tracking

- [PYIC-8305](https://govukverify.atlassian.net/browse/PYIC-8305)

## Checklists

- [ ] All READMEs and documentation updated where necessary

- [ ] No risk of PII, credentials or anything else sensitive being exposed through logs

- [ ] API/unit/contract tests have been written/updated

[PYIC-8305]: https://govukverify.atlassian.net/browse/PYIC-8305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DanCorderIPV (dismissed):
Would be nice to have "Arrange", "Act", "Assert" comments in the unit tests.

They are simple at the moment, but if we don't add the comments now they probably won't be added even if things get more complex.

DanCorderIPV (approved):

PR #3223, opened 2025-06-12
Revert "Revert "PYIC-8254: Add EVCS PUT contract tests""

MikeCollingwood:
Reverts govuk-one-login/ipv-core-back#3220


MikeCollingwood:
SI not validated by EVCS so have removed the invalid SI case for now

chrisclayson:
I wouldn't put this back in, we're going to change the shape of proposed slightly, so you might want to hold off.

PR #3227, opened 2025-06-16
PYIC-8363: Check TICF VC for relevant interventions

MikeCollingwood:
### What changed

- Check TICF VC for relevant interventions

### Why did it change

- TICF can return account interventions we must deal with

### Issue tracking

- [PYIC-8363](https://govukverify.atlassian.net/browse/PYIC-8363)

## Checklists

[PYIC-8363]: https://govukverify.atlassian.net/browse/PYIC-8363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/src/types/ticf-management-api.ts
+++ b/api-tests/src/types/ticf-management-api.ts
@@ -2,6 +2,7 @@ export interface TicfManagementParameters {
   evidence: {
     type: string;
     ci: string[] | undefined;
+    intervention: { interventionCode: string } | undefined;
     txn: string | undefined;
   };
   responseDelay: number;
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -133,6 +133,16 @@ public class ProcessCandidateIdentityHandler
             new JourneyResponse(JOURNEY_FAIL_WITH_CI_PATH);
     private static final JourneyResponse JOURNEY_ACCOUNT_INTERVENTION =
             new JourneyResponse(JOURNEY_ACCOUNT_INTERVENTION_PATH);
+    private static final Map<String, AisInterventionType> interventionCodeTypes =
+            Map.of(
+                    "00", AIS_NO_INTERVENTION,
+                    "01", AIS_ACCOUNT_SUSPENDED,
+                    "02", AIS_ACCOUNT_UNSUSPENDED,
+                    "03", AIS_ACCOUNT_BLOCKED,
+                    "04", AIS_FORCED_USER_PASSWORD_RESET,
+                    "05", AIS_FORCED_USER_IDENTITY_VERIFY,
+                    "06", AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY,
+                    "07", AIS_ACCOUNT_UNBLOCKED);
 
     private final ConfigService configService;
     private final ClientOAuthSessionDetailsService clientOAuthSessionDetailsService;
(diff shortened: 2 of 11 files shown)


libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
```diff
@@ -19,9 +20,34 @@ public AisService(AisClient aisClient) {
...
+                return new AccountInterventionState(false, false, false, false);
+            }
+            case AIS_ACCOUNT_SUSPENDED -> {
+                return new AccountInterventionState(false, true, false, false);
+            }
+            case AIS_ACCOUNT_BLOCKED -> {
+                return new AccountInterventionState(true, false, false, false);
+            }
+            case AIS_FORCED_USER_PASSWORD_RESET -> {
+                return new AccountInterventionState(false, true, false, true);
```
MikeCollingwood, line 41: TICF doesn't know the difference between the following due to only getting an intervention code:
AIS_FORCED_USER_PASSWORD_RESET (suspended ✅ & password reset ✅)
AIS_FORCED_USER_PASSWORD_RESET (suspended ❌ & password reset ✅)

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -723,6 +747,41 @@ public JourneyResponse getJourneyResponseFromTicfCall(
...
+        Map<String, AisInterventionType> interventionCodeTypes =
+                Map.of(
+                        "00", AIS_NO_INTERVENTION,
+                        "01", AIS_ACCOUNT_SUSPENDED,
+                        "02", AIS_ACCOUNT_UNSUSPENDED,
+                        "03", AIS_ACCOUNT_BLOCKED,
+                        "04", AIS_FORCED_USER_PASSWORD_RESET,
+                        "05", AIS_FORCED_USER_IDENTITY_VERIFY,
+                        "06", AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY,
+                        "07", AIS_ACCOUNT_UNBLOCKED);
```
thebauSoftwire: could this just be a static property of the handler?

api-tests/src/types/ticf-management-api.ts
```diff
@@ -2,6 +2,7 @@ export interface TicfManagementParameters {
   evidence: {
     type: string;
     ci: string[] | undefined;
+    intervention: { interventionCode: string } | undefined;
```
thebauSoftwire, line 5: should we update the docs on the ticf stub to include this? We might also need to update some of the types e.g. `TicfEvidenceItem`

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
thebauSoftwire: is the bottom one possible?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -723,6 +747,41 @@ public JourneyResponse getJourneyResponseFromTicfCall(
...
+                                        .getEvidence().stream())
+                .map(RiskAssessment::getIntervention)
+                .filter(Objects::nonNull)
+                .map(Intervention::getInterventionCode)
+                .filter(Objects::nonNull)
+                .map(
+                        interventionCode ->
+                                aisService.getStateByIntervention(
+                                        interventionCodeTypes.getOrDefault(
+                                                interventionCode, AIS_NO_INTERVENTION)))
```
thebauSoftwire: do we need to handle the case where ticf were to add a new intervention code that we still haven't added to the `interventionCodeTypes`? If they return a valid intervention that should block the user's journey but we default to `AIS_NO_INTERVENTION` here

thebauSoftwire (commented):
Nice set of tests 🙌  Just some questions

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
MikeCollingwood: I think throwing a runtime error would be the alternative? Blocking in a different sense

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
MikeCollingwood: I think the difference is when it raised during a journey? This is based on Dan's test cases

MikeCollingwood: Ah, on talking about it with Dan, it sounds like it might not be possible. So it shouldn't matter for TICF, we could remove the AIS cases but I don't think we need to as it's treated as unexpected

api-tests/src/types/ticf-management-api.ts
MikeCollingwood: I have updated https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3763505535/IPV+TICF+CRI+Interaction. Oh yeah looks like the stub needs an update

thebauSoftwire (approved):

PR #3229, opened 2025-06-16
PYIC-8335: upgrade dompurify and mermaid and fix journey map styling

thebauSoftwire:
### What changed
- upgraded dompurify and mermaid.js 
- updated styling to be compatible with new mermaid.js version so that:
  - state and event names aren't cut off
  - can differentiate between different events
  - arrow highlighting works again
 
Screenshot 2025-06-16 at 16 48 04

### Why did it change
We have a security alert in core-back related to an old dompurify. This PR resolves that alert by upgrading mermaid.js. However, this new major version has some breaking changes such as:
- event + state names being cut off
- state and arrow highlights not working
- edgeLabel classes not being applied

### Issue tracking

- [PYIC-8335](https://govukverify.atlassian.net/browse/PYIC-8335)

## Checklists

[PYIC-8335]: https://govukverify.atlassian.net/browse/PYIC-8335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/public/style.css
+++ b/journey-map/public/style.css
@@ -77,7 +77,6 @@ body {
 #diagram > * {
   height: 100%;
   width: 100%;
-  max-width: unset !important;
 }
 form {
   flex: 0 1 525px;
@@ -227,4 +226,7 @@ form fieldset p {
 }
 .mitigationTransition {
   background-color: #e4c4ff;
+}
+span.edgeLabel > p {
+  text-wrap: wrap;
 }
--- a/journey-map/public/style.css
+++ b/journey-map/public/style.css
@@ -218,15 +218,22 @@ form fieldset p {
 #state-search {
   font-size: 1rem;
 }
-.journeyCtxTransition {
+#diagramSvg .edgeLabel p.journeyCtxTransition {
   background-color: #f5daa4;
 }
 .journeyCtxList {
   padding: 4px 24px 0;
 }
-.mitigationTransition {
-  background-color: #e4c4ff;
+#diagramSvg .edgeLabel p.mitigationTransition {
+  background-color: #ffbda8;
 }
-span.edgeLabel > p {
+.defaultEdgeLabel {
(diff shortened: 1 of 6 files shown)


journey-map/public/style.css
```diff
@@ -77,7 +77,6 @@ body {
 #diagram > * {
   height: 100%;
   width: 100%;
-  max-width: unset !important;
```
thebauSoftwire, line 80: I don't think this is doing anything in this new version...

journey-map/src/index.ts
```diff
@@ -36,6 +36,10 @@ mermaid.initialize({
   // Required to enable links and callbacks
   // This is (relatively) safe, as we only run on our own generated mermaid charts
   securityLevel: "loose",
+  flowchart: {
+    useMaxWidth: false,
+    wrappingWidth: 800,
```
thebauSoftwire, line 41: new major version enforces a max node width of 200px :( instead of wrapping the text, I've set it to a large enough number so that the state names in snake case aren't wrapped which makes it easier to read

journey-map/public/style.css
```diff
@@ -219,12 +218,22 @@ form fieldset p {
...
 .journeyCtxList {
   padding: 4px 24px 0;
 }
-.mitigationTransition {
-  background-color: #e4c4ff;
+#diagramSvg .edgeLabel p.mitigationTransition {
+  background-color: #ffbda8;
+}
+.defaultEdgeLabel {
+  text-wrap: wrap;
```
thebauSoftwire, line 231: unfortunately there isn't a way to configure the max-width of an edge label so I force it to wrap and added some styling (e.g. borders between different events) so it's easier to read

blakeyp (approved):

PR #3238, opened 2025-06-18
BAU Flatten asyncCriErrorResponse metric dimensions for DT filtering

blakeyp:
### What changed

Flatten asyncCriErrorResponse metric dimensions into a single composite dimension

### Why did it change

Dynatrace dashboards don't seem to support multi-dimensional filtering in CloudWatch metrics, so need to flatten into a single composite dimension we can filter on


MikeCollingwood (approved):
Makes sense!

PR #3239, opened 2025-06-18
BAU: Switch to apache client

hugh-emerson:
### What changed

Switch from `UrlConnectionHttpClient` (software.amazon.awssdk:url-connection-client) to `ApacheHttpClient` (software.amazon.awssdk:apache-client) as a speculative fix for the segfault issue.

I didn't reproduce a segfault in the dev environment. I believe previously, attempts at fixing the issue were made on the build environment. I don't think it's worth trying to feature flag this or anything though there is a slight risk of performance degradation. 

### Why did it change

Because Apache is the only http client that they claim to support. See known limitations at the bottom of this page: https://docs.dynatrace.com/docs/ingest-from/amazon-web-services/integrate-into-aws/aws-lambda-integration/aws-lambda-extension

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,8 @@ uuid_results
 
 libs/test-helpers/bin
 local-running/bin/
+libs/common-services/bin
+libs/ticf-cri-service/bin
 
 # User-specific secrets
 secrets.gradle
--- a/build.gradle
+++ b/build.gradle
@@ -31,9 +31,6 @@ spotless {
 
 subprojects {
 	task allDeps(type: DependencyReportTask) {}
-	configurations.all {
-		exclude group: 'software.amazon.awssdk', module: 'apache-client'
-	}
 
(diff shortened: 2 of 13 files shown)


blakeyp (approved):

hugh-emerson:
Looking at it again, we use two other HTTP clients. Netty for async AWS SDK requests, and the native Java HttpClient for non-sdk requests (wrapped in TracingHttpClient defined by us). The IOExceptions you found before are mostly from the TracingHttpClient but I hadn't noticed. So it's unlikely that this PR would fix it if using the wrong HttpClient really is the cause. Assuming those IOExceptions are linked to the segfaults. We could potentially migrate those requests to Apache client (and then we also wouldn't need to wrap it in a class to enable tracing), but it's a bigger change than this PR, which currently only updates AWS SDK http clients.

PR #3247, opened 2025-06-23
PYIC-8427: use evcs /identity endpoint

thebauSoftwire:
### What changed

- use EVCS /identity POST endpoint to store stored identity record
  - fails open when failing to create SI object, EVCS returns non-202 response or fails to send/process http request
- removes use of PUT `/vcs` request to store VCs + SI
  - removes references to this endpoint but keeps some of the logic so we don't have to put the same logic back in in phase 2

### Why did it change
Trust & Reuse team have re-estimated the size of their work to instead now implement a POST /identity endpoint. As part of phase 1, we only need to use this endpoint to store the SI record. In phase 2, the endpoint will be updated to store the VCs as well.

### Issue tracking

- [PYIC-8427](https://govukverify.atlassian.net/browse/PYIC-8427)

## Checklists

[PYIC-8427]: https://govukverify.atlassian.net/browse/PYIC-8427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
+++ b/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
@@ -10,7 +10,6 @@ import uk.gov.di.ipv.core.library.evcs.client.EvcsClient;
 import uk.gov.di.ipv.core.library.evcs.dto.EvcsCreateUserVCsDto;
 import uk.gov.di.ipv.core.library.evcs.dto.EvcsGetUserVCDto;
 import uk.gov.di.ipv.core.library.evcs.dto.EvcsPostIdentityDto;
-import uk.gov.di.ipv.core.library.evcs.dto.EvcsPutUserVCsDto;
 import uk.gov.di.ipv.core.library.evcs.dto.EvcsUpdateUserVCsDto;
 import uk.gov.di.ipv.core.library.evcs.enums.EvcsVCState;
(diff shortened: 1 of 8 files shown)


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
```diff
@@ -189,25 +189,25 @@ public List<EvcsGetUserVCDto> getUserVCs(
...
             Vot achievedVot)
             throws FailedToCreateStoredIdentityForEvcsException, EvcsServiceException {
-        var putUserVcsDto =
-                createCompletedPutUserVcsDto(
+        var postIdentityDto =
+                createCompletedPostIdentityDto(
                         userId, credentials, strongestAchievedVot, achievedVot);
-        return evcsClient.storeUserVCs(putUserVcsDto);
+        return evcsClient.storeUserIdentity(postIdentityDto);
     }
```
thebauSoftwire, line 208: I've updated these methods to use the new `/identity` POST endpoint. Though these methods write a `vcs` object (and an undefined `si` object - which is required in the evcs stub) in the request body to the new endpoint, they aren't used anywhere in the code (except for some tests). I've just kept these here as they will eventually be needed in phase 2

DanCorderIPV (approved):

PR #3268, opened 2025-07-02
PYIC-8457: move process-candidate-identity-build lambda to use opente…

hugh-emerson:
…lemetry layer instead of dynatrace

### What changed

Just updating a single lambda in build environment for now. I picked process-candidate-identity because it seemed to have a lot of hits in cloudwatch for "SIGTERM"

Note some minor issues currently:
* It seems to not get put in the correct Management Zone in Dynatrace
* Not all metrics from before are present
* I'll manually put the "dynatrace_api_token_otlp" secret into Secrets Manager, but eventually we'd want the observability team to put it in the same place as the normal api token. This is an api token that they created for me with necessary permissions for the /otlp endpoints.

I think we should test if it even works and solves the issue before worrying about tidying up issues like those

### Issue tracking

- [PYIC-8457](https://govukverify.atlassian.net/browse/PYIC-8457)

[PYIC-8457]: https://govukverify.atlassian.net/browse/PYIC-8457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -55,7 +55,7 @@ pre-commit autoupdate && pre-commit install
 ```
 
 ### Secrets
-To set up secrets, copy the `secrets.gradle.template` to `secrets.gradle` and enter values for each variable.
+To set up secrets, copy the `secrets.template.gradle` to `secrets.gradle` and enter values for each variable.
```
DanCorderIPV: I think it would be better to rename the template file to match this - then there's no danger of it being used by mistake (and it matches our template pattern for other files)

deploy/template.yaml
```diff
@@ -2896,18 +2908,36 @@ Resources:
...
+          DT_CONNECTION_AUTH_TOKEN_OTLP: !If
+            - IsBuild
+            - !Sub
+              - '{{resolve:secretsmanager:${SecretArn}:SecretString}}' # pragma: allowlist secret
+              - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref AWS::AccountId, dynatraceOtlpSecretArn ]
+            - !Ref AWS::NoValue
           AWS_LAMBDA_EXEC_WRAPPER: !If
             - UseDynatrace
-            - /opt/dynatrace
+            - /opt/otel-handler
```
DanCorderIPV: I think this needs an `!If IsBuild` ?

DanCorderIPV (changes requested):

DanCorderIPV (dismissed):

DanCorderIPV (approved):

PR #3276, opened 2025-07-07
PYIC-8470: invalidate si for all journeys except reuse and reverification

thebauSoftwire:
### What changed

- invalidate SI in check-existing-identity lambda
- invalidate SI in reset-session-identity lambda to cover update journeys
- added new api tests 
- added new api test step to create an SI without having to repeat steps for a ID prov. journey

### Why did it change

- so that we can invalidate a user's identity when they are redirected to core-back (on a identity proofing journey or update journey) and it stays invalidated if they abandon or fail

### Issue tracking

- [PYIC-8470](https://govukverify.atlassian.net/browse/PYIC-8470)

## Checklists

[PYIC-8470]: https://govukverify.atlassian.net/browse/PYIC-8470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
thebauSoftwire, line 1: Looks big but the majority of the changes here are to move tests into nested blocks. See [c29993b](https://github.com/govuk-one-login/ipv-core-back/pull/3276/commits/c29993b8d882ae5bf6e9de6e310f52c5872c2eb3) commit

api-tests/features/stored-identity/p1-reuse-existing-identity.feature
```diff
@@ -67,3 +68,33 @@ Feature: Stored Identity - Update Existing Identity
...
+      When I start a new 'low-confidence' journey
+      Then I get a 'page-ipv-reuse' page response
+      When I submit an 'update-details' event
+      Then I get a 'update-details' page response
+      When I submit a 'given-names-only' event
+      Then I get a 'page-update-name' page response
+      When I submit a 'update-name' event
+      Then I get a 'dcmaw' CRI response
+      # SI record invalidated as part of reset-session-identity lambda
+      And I have a 'GPG45' stored identity record type with a 'P1' vot that is 'invalid'
```
aleemriaz: The ticket mentions invalidating "at the very start," and correct me if I'm wrong, but for this reuse journey, the "start" seems to be when the user decides to update their details. 🤔 

To match that, maybe we could move the check for the invalid state to happen right after the 'Then I get a 'update-details' page response' step?

thebauSoftwire: yeah, we still invalidate at the start of a journey but I realised that with a reuse journey, we actually call `process-candidate-identity` just before we reach the [reuse screen](https://identity.build.account.gov.uk/dev/template/page-ipv-reuse/en) which will create the same SI record for the user again (see [8258](https://govukverify.atlassian.net/jira/software/c/projects/PYIC/boards/382/backlog?selectedIssue=PYIC-8258)). For reuse journeys, we should be invalidating the SI record after this time but before the user makes any changes to their VCs bc of an update. The most sensible place we could see to put this was in `reset-session-identity` lambda which is called once a user starts an update journey, for update name this is after the [update-name](https://identity.build.account.gov.uk/dev/template/page-update-name/en) page which is why I call it at a weird place :/

thebauSoftwire: I think for update address it happens right after the update-details screen though, I'll check if there are places I need to add that

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClient.java
```diff
@@ -190,8 +192,38 @@ public void updateUserVCs(String userId, List<EvcsUpdateUserVCsDto> evcsUserVCsT
...
+        } catch (URISyntaxException e) {
+            throw new EvcsServiceException(
+                    HTTPResponse.SC_SERVER_ERROR, ErrorResponse.FAILED_TO_CONSTRUCT_EVCS_URI);
+        } catch (JsonProcessingException e) {
+            throw new EvcsServiceException(
+                    HTTPResponse.SC_SERVER_ERROR, ErrorResponse.FAILED_TO_PARSE_EVCS_REQUEST_BODY);
+        }
+    }
+
+    private URI getIdentityUri(String identitySubPaths) throws URISyntaxException {
```
MikeCollingwood:
```suggestion
    private URI getIdentityUri(String path) throws URISyntaxException {
```
(V minor) It's sub-IDENTITY_SUB_PATH and the plural is slightly misleading, so might just be better to call it path as it's implied it's getting /identity at least from the method name

lambdas/reset-session-identity/src/main/java/uk/gov/di/ipv/core/resetsessionidentity/ResetSessionIdentityHandler.java
```diff
@@ -108,6 +109,10 @@ public Map<String, Object> handleRequest(ProcessRequest input, Context context)
             String govukSigninJourneyId = clientOAuthSessionItem.getGovukSigninJourneyId();
             LogHelper.attachGovukSigninJourneyIdToLogs(govukSigninJourneyId);
 
+            if (configService.enabled(STORED_IDENTITY_SERVICE)) {
+                evcsService.invalidateStoredIdentityRecord(clientOAuthSessionItem.getUserId());
+            }
```
MikeCollingwood, line 115: Do we definitely want to invalidate here every time? In the case they have failed/ abandoned V2 App for MFA reset, I think we might not want to force them to reprove? If they change details, then it makes sense - but I think that's ProcessCandidateIdentity?

api-tests/features/stored-identity/p2-journeys.feature
```diff
@@ -122,18 +122,50 @@ Feature: Stored Identity - P2 journeys
...
+      Then I get a 'fraud' CRI response
+      When I submit 'kenneth-changed-given-name-score-2' details with attributes to the CRI stub
+        | Attribute          | Values                   |
+        | evidence_requested | {"identityFraudScore":1} |
+      Then I get a 'page-ipv-success' page response with context 'updateIdentity'
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I get a 'P2' identity
+      And I have a 'GPG45' stored identity record type with a 'P2' vot that is 'valid'
```
MikeCollingwood: Might be worth defaulting asserting it is valid for other uses of the same step?

api-tests/features/stored-identity/p2-journeys.feature
```diff
@@ -122,18 +122,50 @@ Feature: Stored Identity - P2 journeys
...
+      # New reuse p2 journey with update name
+      When I start a new 'medium-confidence' journey
+      Then I get a 'page-ipv-reuse' page response
+      When I submit an 'update-details' event
+      Then I get a 'update-details' page response
+      When I submit a 'given-names-only' event
+      Then I get a 'page-update-name' page response
+      When I submit a 'update-name' event
+      Then I get a 'dcmaw' CRI response
+      # SI record invalidated as part of reset-session-identity lambda
```
MikeCollingwood, line 186: Is it not initially invalidated as part of check-existing-identity?

I think ProcessCandidateIdentity wouldn't recreate the stored identity because it's EXISTING in all cases except when there's a floating async VC.

lambdas/reset-session-identity/src/main/java/uk/gov/di/ipv/core/resetsessionidentity/ResetSessionIdentityHandler.java
thebauSoftwire: ah thanks! I missed adding the reverification check

api-tests/features/stored-identity/p2-journeys.feature
thebauSoftwire: Yeah we invalidate it there but we also need to invalidate it as part of reset-session-identity to cover reuse update journeys as we also create SI records for every reuse journey (see [8258](https://govukverify.atlassian.net/jira/software/c/projects/PYIC/boards/382/backlog?selectedIssue=PYIC-8258)) so that for users who already have an identity prior to SIS going live, we can create an SI record for them. More details in [this comment](https://github.com/govuk-one-login/ipv-core-back/pull/3276#discussion_r2191811035)

api-tests/features/stored-identity/p1-reuse-existing-identity.feature
aleemriaz: Ah that makes sense, thanks for explaining.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClient.java
```diff
@@ -190,8 +192,38 @@ public void updateUserVCs(String userId, List<EvcsUpdateUserVCsDto> evcsUserVCsT
...
+                            .POST(
+                                    HttpRequest.BodyPublishers.ofString(
+                                            OBJECT_MAPPER.writeValueAsString(
+                                                    new EvcsInvalidateStoredIdentityDto(userId))))
+                            .header(
+                                    X_API_KEY_HEADER,
+                                    configService.getSecret(ConfigurationVariable.EVCS_API_KEY))
+                            .header(CONTENT_TYPE, ContentType.APPLICATION_JSON.toString());
+
+            sendHttpRequest(httpRequestBuilder.build());
```
MikeCollingwood: Could it return the result of this, so it can be tested more easily? I think updateUserVCs could do with it as well, but not necc in scope.

api-tests/features/stored-identity/p2-journeys.feature
MikeCollingwood: Ah, I didn't see there was a check for `EXISTING` as well:
```
private boolean shouldStoreIdentity(CandidateIdentityType identityType) {
    return STORE_IDENTITY_TYPES.contains(identityType)
            || shouldStoreExistingIdentity(identityType);
}
```

MikeCollingwood (dismissed):
LGTM!

MikeCollingwood (approved):

PR #3277, opened 2025-07-07
PYIC-8464: add support for H1A profiles

thebauSoftwire:
### What changed

- added H1A as supported profile (P3)
- added api tests to cover support of new profile

### Why did it change

- Supporting the H1A profile enables Relying Parties like IBCA to receive a P3 identity where users have sufficient evidence. It allows for tactical adoption of high confidence identity journeys without requiring major changes to the current IPV implementation, and unlocks the ability to prove identity at a higher assurance level for those users who are able to meet the required criteria.

### Issue tracking

- [PYIC-8464](https://govukverify.atlassian.net/browse/PYIC-8464)

## Checklists

[PYIC-8464]: https://govukverify.atlassian.net/browse/PYIC-8464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/cimit/p3-enhanced-verification-mitigation.feature
hugh-emerson, line 1: Are these the same as p2 features just with 'high-medium-confidence' at the start and 'P3' at the end? In which case can they reuse the same code instead, via "scenario outline" and "examples"?

api-tests/features/p3-journeys.feature
hugh-emerson, line 1: Does it make sense to split this up into multiple files the same way the p2 journeys are? Just for consistency / easier comparison. 

Or, similar to my other comment, can these be combined with p2 scenarios? I'd also prefer that (if possible) because this is fundamentally still the P2 journey. We might at a future date have an actual P3-specific journey. The difference for now is just be the requested and returned VOT following the same P2 journey? I guess you wouldn't be able to keep some of the commonised "background" steps because `I start a new 'medium-confidence' journey` would need to instead be parameterised with the example data... but feels like maybe still less repetitive than what you have now?

api-tests/features/p3-journeys.feature
```diff
@@ -0,0 +1,204 @@
...
+      | Attribute          | Values                                          |
+      | evidence_requested | {"scoringPolicy":"gpg45","verificationScore":2} |
+    Then I get a 'page-ipv-success' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
+
+  Scenario: P2 identity met for high-medium confidence journey via f2f journey
+    Given I start a new 'high-medium-confidence' journey
```
hugh-emerson: Why is this one "given" rather than "when"?

thebauSoftwire: I don't think there was a reason! I've moved this test elsewhere now

hugh-emerson (approved):
👍

PR #3292, opened 2025-07-14
BAU: Create a bakeless AppConfig deployment strategy

MikeCollingwood:
Reverts govuk-one-login/ipv-core-back#3236


undermad:
You can hide it behind the condition to create resource only in dev environment

undermad (approved):

PR #3294, opened 2025-07-14
PYIC-8506: Modify intervention-handling to use exceptions

MikeCollingwood:
### What changed

- Break from ProcessCandidateIdentityHandler when account intervention detected.

### Why did it change

- So TICF intervention skips following code, like for AIS in ProcessCandidateIdentityHandler

### Issue tracking

- [PYIC-8506](https://govukverify.atlassian.net/browse/PYIC-8506)

## Checklists

[PYIC-8506]: https://govukverify.atlassian.net/browse/PYIC-8506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -11,6 +11,7 @@ import software.amazon.awssdk.http.HttpStatusCode;
 import software.amazon.lambda.powertools.logging.Logging;
 import software.amazon.lambda.powertools.metrics.Metrics;
 import uk.gov.di.ipv.core.library.ais.enums.AisInterventionType;
+import uk.gov.di.ipv.core.library.ais.exception.AccountInterventionException;
 import uk.gov.di.ipv.core.library.ais.exception.AisClientException;
 import uk.gov.di.ipv.core.library.ais.service.AisService;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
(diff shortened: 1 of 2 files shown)


MikeCollingwood:
I have not written an API test since it would be for F2F only to check the precursor CRI VCs are not saved, which would involve a new management endpoint to fetch EVCS VCs for a user. We need a bearer token to get EVCS VCs

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -292,7 +294,9 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
                     ipAddress,
                     sessionVcs,
                     auditEventUser);
-
+        } catch (AccountInterventionException e) {
+            updateIpvSessionWithIntervention(ipvSessionItem);
+            return JOURNEY_ACCOUNT_INTERVENTION.toObjectMap();
```
MikeCollingwood, line 299: Using an exception allows us to be consistent with how we deal with interventions

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/exception/AccountInterventionException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.di.ipv.core.library.ais.exception;
+
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+public class AccountInterventionException extends Exception {
+    public AccountInterventionException() {
+        super("Account intervention.");
```
thebauSoftwire: It looks like this exception is only thrown when we detect an intervention, could we make this message more descriptive to reflect this? It could help with debugging

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
thebauSoftwire: hmm, would detecting an intervention count as an exception? I feel like an exception indicates something in the programme going wrong e.g. failed api call, failed parsing or invalid/unexpected params etc. This is just an opinion though, so happy to also go with this since it consolidates intervention handling

MikeCollingwood: I agree with you that it doesn't feel like the right use of an exception. That being said, it is pretty close to a failure, since it causes a user's journey to break, and it substantially helps with breaking the code to only do these two lines. I think the consolidation is the main reason I chose to do this

thebauSoftwire (approved):

PR #3296, opened 2025-07-15
PYIC-8457: move all lambdas to open telemetry in build environment

hugh-emerson:
### What changed

Extend earlier change that was made process-candidate-identity to all other lambdas, so in the build environment they all use the aws otel lambda layer.

There is the option of using a common collector.yaml in S3, but I think it's simpler to just have a copy in each lambda. They're relatively simple files.


MikeCollingwood (approved):
I count all the lambdas and see there is nothing new, just extended

PR #3301, opened 2025-07-15
BAU: Fix null reference exception by not calling AIS when we don't have a user ID

DanCorderIPV:
### What changed

Don't call AIS when we don't have a user ID

### Why did it change

It doesn't make sense and was causing a null reference exception

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -267,10 +267,11 @@ public class ProcessCandidateIdentityHandler
 
             String userId = clientOAuthSessionItem.getUserId();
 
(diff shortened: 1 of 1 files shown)


lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -269,7 +269,8 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
 
             // We skip AIS checks for reverification journeys
             if (configService.enabled(AIS_ENABLED)
-                    && !SKIP_AIS_TYPES.contains(processIdentityType)) {
+                    && !SKIP_AIS_TYPES.contains(processIdentityType)
+                    && userId != null) {
```
thebauSoftwire: It's not really possible but should we use StringUtils.isEmpty instead to also cover the empty string case?

thebauSoftwire (dismissed):
LGTM! Just one suggestion

thebauSoftwire (approved):

PR #3347, opened 2025-07-30
PYIC-7877: Update ConfigService to correctly refresh parameters

MikeCollingwood:
### What changed

- Update config service to correctly refresh parameters

### Why did it change

- Parameters were previously being added to a map, instead of replacing it. This meant a Lambda, while it was alive, would just gain attributes. In this case, it failed as we tried to handle an old parameter in a new way:
https://gds.slack.com/archives/C05TJ52FZ8B/p1753883130036029

### Issue tracking

- [PYIC-7877](https://govukverify.atlassian.net/browse/PYIC-7877)
- https://gds.slack.com/archives/C05TJ52FZ8B/p1753883130036029

## Checklists

[PYIC-7877]: https://govukverify.atlassian.net/browse/PYIC-7877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
@@ -31,14 +31,14 @@ public class YamlConfigService extends YamlParametersConfigService {
     }
 
     public YamlConfigService(File parametersFile, File secretsFile) {
-        setParameters(updateParameters(parametersFile));
-        secrets = updateParameters(secretsFile);
+        setParameters(parseParameters(parametersFile));
+        secrets = parseParameters(secretsFile);
     }
 
-    private Map<String, String> updateParameters(File yamlFile) {
+    private Map<String, String> parseParameters(File yamlFile) {
         try {
             String yamlContent = Files.readString(yamlFile.toPath());
-            return updateParameters(yamlContent);
+            return parseParameters(yamlContent);
         } catch (IOException e) {
             throw new IllegalArgumentException("Could not read parameters yaml file", e);
         }
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
@@ -24,21 +24,21 @@ public class YamlConfigService extends YamlParametersConfigService {
         this.featureSet.remove();
     }
 
-    private final Map<String, String> secrets = new HashMap<>();
+    private Map<String, String> secrets = new HashMap<>();
 
     public YamlConfigService() {
         this(PARAMETERS_FILE, SECRETS_FILE);
     }
 
     public YamlConfigService(File parametersFile, File secretsFile) {
-        updateParameters(parameters, parametersFile);
-        updateParameters(secrets, secretsFile);
+        setParameters(updateParameters(parametersFile));
+        secrets = updateParameters(secretsFile);
     }
 
-    private void updateParameters(Map<String, String> map, File yamlFile) {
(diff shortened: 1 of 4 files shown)


undermad (dismissed):

undermad (dismissed):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlParametersConfigService.java
```diff
@@ -64,6 +64,7 @@ public Map<String, String> getParametersByPrefixYaml(String path) {
     }
 
     protected void updateParameters(Map<String, String> map, String yaml) {
+        map.clear();
```
DanCorderIPV: I think this is safe in this case, but as a general rule this makes me uncomfortable - there will be a period of time where the config is being rebuilt and is incomplete. It's generally safer to do things like updating config atomically, so there's just a single assignment to new object that was built earlier.

undermad: Yes, actually I though about it, Map could be set as protected or if needed outside the class we can provide getter with unmodifiable map return.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlConfigService.java
```diff
@@ -24,21 +24,21 @@ public void removeFeatureSet() {
...
 
     public YamlConfigService() {
         this(PARAMETERS_FILE, SECRETS_FILE);
     }
 
     public YamlConfigService(File parametersFile, File secretsFile) {
-        updateParameters(parameters, parametersFile);
-        updateParameters(secrets, secretsFile);
+        setParameters(updateParameters(parametersFile));
+        secrets = updateParameters(secretsFile);
```
MikeCollingwood: secrets are an attribute of this class, no the class it inherits from, hence why we can do this

undermad (dismissed):
LGTM!

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/YamlParametersConfigService.java
```diff
@@ -63,10 +67,12 @@ public Map<String, String> getParametersByPrefixYaml(String path) {
         return lookupParams;
     }
 
-    protected void updateParameters(Map<String, String> map, String yaml) {
+    protected Map<String, String> updateParameters(String yaml) {
```
DanCorderIPV: Can these methods to be renamed to `parseParamaters` or similar?

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/AppConfigServiceTest.java
```diff
@@ -221,6 +221,9 @@ void getParameterReturnsUpdatedParameters() {
 
         // Assert
         assertEquals("different-component-id", param);
+        assertThrows(
+                ConfigParameterNotFoundException.class,
+                () -> configService.getParameter(ConfigurationVariable.BEARER_TOKEN_TTL));
```
DanCorderIPV, line 228: I think it would be good to also show that you can retrieve this parameter before the update

DanCorderIPV (changes requested):

libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/AppConfigServiceTest.java
MikeCollingwood: Ah I did that test myself but I can show you

DanCorderIPV (approved):

PR #3348, opened 2025-07-31
PYIC-7877 Allow getCriIssuers to receive it for multiple connections

undermad:
### What changed

- Allow getIssuersCris in yaml format to return multiple issuers for each connection.

### Why did it change

- Staging environment uses multiple connections for cris

### Issue tracking

- [PYIC-7877](https://govukverify.atlassian.net/browse/PYIC-7877)

## Checklists

[PYIC-7877]: https://govukverify.atlassian.net/browse/PYIC-7877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
@@ -285,15 +285,19 @@ public abstract class ConfigService {
             if (cri.getId().equals(Cri.CIMIT.getId())) {
                 continue;
             }
+
+            var connectionsPath =
+                    String.format(
+                            ConfigurationVariable.CREDENTIAL_ISSUER_CONNECTION_PREFIX.getPath(),
+                            cri.getId());
+
+            var pattern = Pattern.compile("[^/]+/componentId");
             try {
-                var connection = getActiveConnection(cri);
-                var path =
-                        String.format(
-                                ConfigurationVariable.CREDENTIAL_ISSUER_COMPONENT_ID.getPath(),
-                                cri.getId(),
-                                connection);
-                var componentId = getParameter(path);
-                issuerToCri.put(componentId, cri);
+                var connections =
+                        getParametersByPrefixYaml(connectionsPath).entrySet().stream()
+                                .filter(entry -> pattern.matcher(entry.getKey()).find())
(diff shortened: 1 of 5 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/ConfigService.java
```diff
@@ -285,15 +285,19 @@ private Map<String, Cri> getIssuerCrisYaml() {
...
-                                ConfigurationVariable.CREDENTIAL_ISSUER_COMPONENT_ID.getPath(),
-                                cri.getId(),
-                                connection);
-                var componentId = getParameter(path);
-                issuerToCri.put(componentId, cri);
+                var connections =
+                        getParametersByPrefixYaml(connectionsPath).entrySet().stream()
+                                .filter(entry -> pattern.matcher(entry.getKey()).find())
+                                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+                connections.values().forEach(value -> issuerToCri.put(value, cri));
```
blakeyp, line 299: Rather than matching with regex could we do something more like the existing method ie. deserialise the yaml into `CriConfig`? (appreciate this may get superseded by the POJO ticket)

undermad:
```
                var connections =
                        getParametersByPrefixYaml(connectionsPath).entrySet().stream()
                                .filter(entry -> entry.getKey().endsWith("/componentId"))
                                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```

I came up with this solution instead of regex. 

Creation of CriConfig doesn't make any sense as only one value is required (componentId).
Previously it was logical because it was easy to use OBJECT MAPPER to create instance from json.

undermad: `getParametersByPrefixYaml(connectionsPath)`
this call will return key value map that will looks like this 

```
stub/componentId -> issuer url
stub/somethingElse -> something else
main/componentId -> issuer url
main/somethingElse -> something else
```

then its filtered to contain only keys that ents with `/componentId` of specified cri that is our lookup
This will add to the map issuer for all connections for given cri

blakeyp (approved):

PR #3350, opened 2025-08-01
PYIC-7878: Create Config class

MikeCollingwood:
### What changed

1. Update existing tests to use valid Cri configs
2. Update Cri config classes to be strict
3. Add Config class

### Why did it change

- Config is consumed using first-class accessors, rather than parameterised paths
- Keeps the config as a single object which is easier to pass around, cache etc. in code
- Makes the config a single, validatable object.

### Issue tracking

- [PYIC-7878](https://govukverify.atlassian.net/browse/PYIC-7878)

## Checklists

- [ ] Check Cri config `NON_NULL` assertions. It has only been based on what we've seen.
- [ ] IF it is not true in production, Production will break if it tries to use this Config

[PYIC-7878]: https://govukverify.atlassian.net/browse/PYIC-7878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/CriConnectionWrapper.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/CriConnectionWrapper.java
@@ -1,0 +1,26 @@
+package uk.gov.di.ipv.core.library.config.domain;
+
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.extern.jackson.Jacksonized;
+import uk.gov.di.ipv.core.library.dto.CriConfig;
+
+import java.util.Map;
+
+@Data
+@Builder
+@Jacksonized
+public class CriConnectionWrapper<T extends CriConfig> {
+    @NonNull final String id;
+    @NonNull final String name;
+    @NonNull final String enabled;
+    @NonNull final String unavailable;
+    final String allowedSharedAttributes;
+    @NonNull final String activeConnection;
+    @NonNull final Map<String, @NonNull T> connections;
+
+    public T getActiveConfig() {
+        return connections.get(activeConnection);
+    }
+}
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
@@ -1,0 +1,39 @@
+package uk.gov.di.ipv.core.library.config.domain;
+
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.extern.jackson.Jacksonized;
+import uk.gov.di.ipv.core.library.domain.ContraIndicatorConfig;
+
+import java.util.List;
+import java.util.Map;
+
+@Data
+@Builder
+@Jacksonized
+public class InternalOperationsConfig {
+    final String configFormat;
+    @NonNull final String componentId;
+    @NonNull final String signingKeyId;
+    @NonNull final String sisSigningKeyId;
+    @NonNull final String audienceForClients;
+    @NonNull final Integer jwtTtlSeconds;
+    @NonNull final Integer maxAllowedAuthClientTtl;
+    @NonNull final Integer fraudCheckExpiryPeriodHours;
+    @NonNull final Integer dcmawAsyncVcPendingReturnTtl;
+    @NonNull final String clientJarKmsEncryptionKeyAliasPrimary;
+    @NonNull final String clientJarKmsEncryptionKeyAliasSecondary;
+    @NonNull final String coreVtmClaim;
+    @NonNull final Integer backendSessionTimeout;
+    @NonNull final Integer backendSessionTtl;
+    @NonNull final Integer bearerTokenTtl;
+    @NonNull final Integer criResponseTtl;
+    @NonNull final Integer sessionCredentialTtl;
+    @NonNull final Integer authCodeExpirySeconds;
+    @NonNull final Integer oauthKeyCacheDurationMins;
+    @NonNull final List<ContraIndicatorConfig> ciScoringConfig;
+    @NonNull final VotCiThresholdsConfig ciScoringThresholdByVot;
+    @NonNull final Map<String, @NonNull String> returnCodes;
+    @NonNull final CoiConfig coi;
+}
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/VotCiThresholdsConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/VotCiThresholdsConfig.java
@@ -1,0 +1,24 @@
+package uk.gov.di.ipv.core.library.config.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.extern.jackson.Jacksonized;
+
+@Data
+@Builder
+@Jacksonized
+public class VotCiThresholdsConfig {
+    @NonNull
+    @JsonProperty("P1")
+    final Integer p1;
+
+    @NonNull
+    @JsonProperty("P2")
+    final Integer p2;
+
+    @NonNull
+    @JsonProperty("P3")
+    final Integer p3;
+}
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
@@ -4,6 +4,7 @@ import com.nimbusds.jose.jwk.RSAKey;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.NonNull;
 import lombok.Setter;
 import lombok.experimental.SuperBuilder;
 import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
@@ -18,9 +19,9 @@ import java.text.ParseException;
 @EqualsAndHashCode(callSuper = true)
 @ExcludeFromGeneratedCoverageReport
 public class OauthCriConfig extends RestCriConfig {
-    private URI tokenUrl;
+    @NonNull private URI tokenUrl;
+    @NonNull private String clientId;
     private URI authorizeUrl;
-    private String clientId;
     private String encryptionKey;
     private URI clientCallbackUrl;
     private boolean requiresAdditionalEvidence;
--- a/libs/common-services/src/test/resources/test-parameters.yaml
+++ b/libs/common-services/src/test/resources/test-parameters.yaml
@@ -1,19 +1,416 @@
 ---
 core:
   self:
-    componentId: "test-component-id"
+    configFormat: "yaml"
+    componentId: "https://identity.local.account.gov.uk"
+    signingKeyId: "some-signin-id"
+    audienceForClients: "https://identity.local.account.gov.uk"
+    jwtTtlSeconds: 3600
+    maxAllowedAuthClientTtl: 3600
+    fraudCheckExpiryPeriodHours: 720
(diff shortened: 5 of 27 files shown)


MikeCollingwood:
For reviewers: Please give feedback on Cri config NON_NULL assertions. It has only been based on what we've seen.

undermad:
I know it may be demanding but I would suggest using records instead of classes as it's immutable in default. That would remove lots of redundant `final` keywords. Records are perfect usage for this kind of data.

undermad (changes requested):
Could you make sure test-parameters doesn't have any sensitive keys included in the configuration? It seems to be copied from one of the config files from core-common-infra which is originally private repository, and this core-back is publicly available.

libs/common-services/src/test/resources/test-parameters.yaml
MikeCollingwood, line 1: These are taken from the `core.local.params.yaml`

undermad: Ah, okay!

undermad (dismissed):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
```diff
@@ -18,10 +19,10 @@
...
 @ExcludeFromGeneratedCoverageReport
 public class OauthCriConfig extends RestCriConfig {
-    private URI tokenUrl;
-    private URI authorizeUrl;
-    private String clientId;
-    private String encryptionKey;
+    @NonNull private URI tokenUrl;
+    @NonNull private URI authorizeUrl;
+    @NonNull private String clientId;
+    @NonNull private String encryptionKey;
```
undermad, line 25: Can we remove those annotation if we are not sure which are mandatory? If there is a risk to fail the production that may be not detected through the pipelines we shouldn't include them. Can we find another solution for it?

undermad (dismissed):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
```diff
@@ -18,10 +19,10 @@
 @EqualsAndHashCode(callSuper = true)
 @ExcludeFromGeneratedCoverageReport
 public class OauthCriConfig extends RestCriConfig {
-    private URI tokenUrl;
-    private URI authorizeUrl;
-    private String clientId;
-    private String encryptionKey;
+    @NonNull private URI tokenUrl;
```
blakeyp, line 22: `dcmawAsync` doesn't have an `authorizeUrl` or `encryptionKey` (since there's no synchronous JAR request like the other CRIs)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/CriConnectionWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Map;
+
+@Data
+@Builder
+@Jacksonized
+public class CriConnectionWrapper<T extends CriConfig> {
+    @NonNull final String id;
+    @NonNull final String name;
+    @NonNull final String enabled;
+    @NonNull final String unavailable;
```
blakeyp, line 18: We don't actually use `unavailable` at all so this be a good opportunity to clean it up entirely

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    @NonNull final Integer backendSessionTtl;
+    @NonNull final Integer bearerTokenTtl;
+    @NonNull final Integer criResponseTtl;
+    @NonNull final Integer sessionCredentialTtl;
+    @NonNull final Integer authCodeExpirySeconds;
+    @NonNull final Integer oauthKeyCacheDurationMins;
+    @NonNull final VotCiThresholdsConfig ciScoringThresholdByVot;
+    @NonNull final Map<String, @NonNull String> returnCodes;
+    @NonNull final CoiConfig coi;
+}
```
blakeyp, line 39: We'll need to add `sisSigningKeyId` (string) - new key has just been added

blakeyp, line 39: I think we're missing `journey` in here? see https://github.com/govuk-one-login/ipv-core-common-infra/blob/1976553a5233f336011f37e3599879c41dbba526/utils/config-mgmt/app/configs/core.build.params.yaml#L175 but I don't think `journey/type` is actually used anywhere 🤔  if that's the case we should just clean it up

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/VotCiThresholdsConfig.java
```diff
@@ -0,0 +1,24 @@
...
+    final Integer p1;
+
+    @NonNull
+    @JsonProperty("P2")
+    final Integer p2;
+
+    @NonNull
+    @JsonProperty("P3")
+    final Integer p3;
+}
```
blakeyp, line 24: in config we still have `PCL200` and `PCL250` 
https://github.com/govuk-one-login/ipv-core-common-infra/blob/1976553a5233f336011f37e3599879c41dbba526/utils/config-mgmt/app/configs/core.production.params.yaml#L121 which are not used so we should remove
(we probably just missed these in [PYIC-8411](https://govukverify.atlassian.net/browse/PYIC-8411))

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
```diff
@@ -18,10 +19,10 @@
...
+    @NonNull private URI authorizeUrl;
+    @NonNull private String clientId;
+    @NonNull private String encryptionKey;
     private URI clientCallbackUrl;
```
blakeyp, line 26: I'd expect all to have a callback URL

blakeyp (commented):
This is really nice 🎉  just a few comments and note this is likely to clash a little with [PYIC-7900](https://govukverify.atlassian.net/browse/PYIC-7900) as CI scoring config will be moving into AppConfig cc @thebauSoftwire

blakeyp:
> I think I am to picky with requested changes, my approach about writing clean code my differ from others. I don't want to stop progress of the ticket.

@undermad suggestions are always welcome, we can always raise follow-up tickets to address things if we think we can't do them in scope of the current ticket :)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    @NonNull final String clientJarKmsEncryptionKeyAliasSecondary;
+    @NonNull final String coreVtmClaim;
+    @NonNull final Integer backendSessionTimeout;
+    @NonNull final Integer backendSessionTtl;
+    @NonNull final Integer bearerTokenTtl;
+    @NonNull final Integer criResponseTtl;
+    @NonNull final Integer sessionCredentialTtl;
+    @NonNull final Integer authCodeExpirySeconds;
+    @NonNull final Integer oauthKeyCacheDurationMins;
+    @NonNull final VotCiThresholdsConfig ciScoringThresholdByVot;
```
thebauSoftwire, line 36: have recently added a `ciScoringConfig` here (sorry for the conflict!) too. There's already an existing POJO for it [here](https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/domain/ContraIndicatorConfig.java)

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
arthurs-gds: Agreed it's not in use.  I'm raising a separate BAU PR with all these cleanups.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
arthurs-gds: Do you mean that `ciScoringConfig` should replace `VotCiThresholdsConfig` (with some tweaking of the POJOs) or that they should exists alongside each other?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Map;
+
+@Data
+@Builder
+@Jacksonized
+public class InternalOperationsConfig {
+    final String configFormat;
+    @NonNull final String componentId;
+    @NonNull final String signingKeyId;
```
undermad, line 18: This should be added to local params as "dummy-value". I think is not there, otherwise it will throw an error.

undermad: Not sure if this is will be relevant now as this is not used anywhere apart from test, but there are little differences in configs between local and other environments. It's worth to be aware of it that this may cause some conflicts in the code on next PRs.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
arthurs-gds: ticf is the only exception in the various `*.params.yaml`s we looked at, in not having one.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/InternalOperationsConfig.java
arthurs-gds: That's an excellent shout, thanks.  I fixed this one but need to be on the lookout for more.

local-running/core.local.params.yaml
```diff
@@ -136,7 +137,6 @@ core:
           credentialUrl: https://dcmaw-cri.stubs.account.gov.uk/credentials/issue
           clientId: ipv-core-local
           signingKey: '{\"kty\":\"EC\",\"crv\":\"P-256\",\"x\":\"RBXnILIdExUEWUJMlYeD6agE8u9gGgA3InKrd5TKhhY\",\"y\":\"kKtt9v_xq9oqvv5_E8AHcV77IYQfyNwaTQyTYxdO_UM\"}'
-          encryptionKey: '{\"kty\":\"RSA\",\"e\":\"AQAB\",\"n\":\"vyapkvJXLwpYRJjbkQD99V2gcPEUKrO3dwjcAA9TPkLucQEZvYZvb7-wfSHxlvJlJcdS20r5PKKmqdPeW3Y4ir3WsVVeiht2iOZUreUO5O3V3o7ImvEjPS_2_ZKMHCwUf51a6WGOaDjO87OX_bluV2dp01n-E3kiIl6RmWCVywjn13fX3jsX0LMCM_bt3HofJqiYhhNymEwh39oR_D7EE5sLUii2XvpTYPa6L_uPwdKa4vRl4h4owrWEJaJifMorGcvqhCK1JOHqgknN_3cb_ns9Px6ynQCeFXvBDJy4q71clkBq_EZs5227Y1S222wXIwUYN8w5YORQe3M-pCIh1Q\"}'
```
undermad: encryptionKey does exist in dcmaw CRI, but doesn't exist in dcmawAsync. Could you bring this one back and remove it from dcmawAsync below?

undermad (commented):
LGTM! Just this last little fix, and should be ok!

local-running/core.local.params.yaml
arthurs-gds: Good catch.  Done.

undermad (approved):

PR #3356, opened 2025-08-05
BAU: update local running to pass query params to processJourneyEvent…

thebauSoftwire:
… handler

### What changed

- allow query params to be passed to the process-journey-event lambda when the `/journey/{event}` endpoint is hit running locally

### Why did it change

- there's some extra validation to check we aren't dealing with mismatching page states by passing back a `currentPage` query param to core-back from the FE. This wasn't working when running locally so I've added in mapping similar to that in the core-back-internal.yaml openAPI spec

### DIFF
--- a/local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
+++ b/local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
@@ -148,10 +148,7 @@ public class JourneyEngineHandler {
                 .deviceInformation(ctx.header(ENCODED_DEVICE_INFORMATION))
                 .clientOAuthSessionId(ctx.header(CLIENT_SESSION_ID))
                 .featureSet(ctx.header(FEATURE_SET))
-                .journey(
-                        !journeyEvent.startsWith("/journey/")
-                                ? "/journey/" + journeyWithQuery
-                                : journeyWithQuery)
+                .journey(journeyWithQuery)
                 .build();
     }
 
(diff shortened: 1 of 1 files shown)


local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
```diff
@@ -133,14 +134,24 @@ private Map<String, Object> processJourneyStep(
...
                 .ipvSessionId(ctx.header(IPV_SESSION_ID))
                 .ipAddress(ctx.header(IP_ADDRESS))
                 .deviceInformation(ctx.header(ENCODED_DEVICE_INFORMATION))
                 .clientOAuthSessionId(ctx.header(CLIENT_SESSION_ID))
                 .featureSet(ctx.header(FEATURE_SET))
-                .journey(journey)
+                .journey(
+                        !journeyEvent.startsWith("/journey/")
+                                ? "/journey/" + journeyWithQuery
+                                : journeyWithQuery)
```
DanCorderIPV: This doesn't seem related to adding `currentPage` is it necessary?

DanCorderIPV (changes requested):

local-running/src/main/java/uk/gov/di/ipv/coreback/handlers/JourneyEngineHandler.java
thebauSoftwire: ah you're right not necessary, thanks! I was just matching the openAPI spec mapping

DanCorderIPV (approved):

PR #3371, opened 2025-08-14
PYIC-8260: Support after key from EVCS

thebauSoftwire:
### What changed

- adding support for `afterKey` attribute from EVCS when getting user VCs

### Why did it change

- EVCS returns an `afterKey` attribute which breaks a user's journey because it's not an attribute we expect. There was an instance of this happening when a user was coming back with >25 VCs. See [this thread](https://gds.slack.com/archives/C02SZLNC3A5/p1744619651262269).

### Issue tracking

- [PYIC-8260](https://govukverify.atlassian.net/browse/PYIC-8260)

## Checklists

[PYIC-8260]: https://govukverify.atlassian.net/browse/PYIC-8260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/features/p2-reuse-journey.feature
```diff
@@ -75,3 +75,17 @@ Feature: P2 Reuse journey
...
+      | dcmaw      | kenneth-passport-valid | 25             |
+      | address    | kenneth-current        | 1              |
+      | fraud      | kenneth-score-2        | 1              |
+
+    When I start a new 'high-medium-confidence' journey
+    Then I get a 'page-ipv-reuse' page response
+    When I submit a 'next' event
+    Then I get an OAuth response
+    When I use the OAuth response to get my identity
+    Then I get a 'P2' identity
```
MikeCollingwood: Probably not enough as VCs sorted by signature in EVCS so a strong chance the address & fraud VCs will be in the first batch

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClient.java
```diff
@@ -80,10 +81,44 @@ public EvcsGetUserVCsDto getUserVcs(
...
+
+        try {
+            var afterKey = evcsGetUserVcsDto.afterKey();
+            while (!StringUtils.isBlank(afterKey)) {
+                LOGGER.info("Attempting additional request to get the rest of the user's VCs");
+                var additionalUserVcs =
+                        attemptGetUserVcsRequest(
+                                userId, evcsAccessToken, vcStatesToQueryFor, afterKey);
+                evcsUserVcs.addAll(additionalUserVcs.vcs());
+                afterKey = additionalUserVcs.afterKey();
```
MikeCollingwood, line 107: Would be nice to have a way of protecting from infinite loops. Maybe check afterKey is increasing & limit loops to 100 as an extreme upper limit

MikeCollingwood:
Tested against a real instance of EVCS, with lots of VCs, these logs occur when testing against these changes 🍏 
Screenshot 2025-08-18 at 16 26 34

MikeCollingwood:
When testing real EVCS with the same test, but without afterKey handling:
Screenshot 2025-08-18 at 17 31 21

It fails because on reuse journeys it's not managing to fetch the right VCs to create an identity to reuse

MikeCollingwood:
In summary, the API test (will remove now testing complete) only passes with the afterKey logic. Have tested 3 times with & without.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClient.java
blakeyp: yes let's add this in, add a counter and cap at 100 to be safe

blakeyp (approved):

PR #3375, opened 2025-08-19
PYIC-8573: Invalidate session on first AIS call. 

undermad:
### What changed

- Moved first AIS call to CheckExistingIdentity lambda
- First AIS call invalidate session if account intervention is detected
- It allow reprove journeys to go through the process

- Refactored logic in mid journey ais call to use enum instead of states
 
### Why did it change

Currently, IPV allows users to continue through their journey after the first AIS call, even if that call returns a fraud intervention (e.g., permanent suspension, temporary suspension, or password reset). The session is only terminated after the second AIS call. This results in:

- Unnecessary CRI usage, incurring costs for journeys that cannot succeed.
- Poor user experience due to late termination.
- Delay in fraud risk enforcement.

### Issue tracking

- [Clean-up PR](https://github.com/govuk-one-login/ipv-core-back/pull/3386/files)

- [PYIC-8573](https://govukverify.atlassian.net/browse/PYIC-8573)

## Checklists

[PYIC-8573]: https://govukverify.atlassian.net/browse/PYIC-8573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/data/audit-events/reprove-identity-journey.json
```diff
@@ -3,7 +3,7 @@
     "event_name": "IPV_JOURNEY_START",
     "component_id": "https://identity.local.account.gov.uk",
     "extensions": {
-      "reprove_identity": true,
+      "reprove_identity": false,
       "vtr": ["P2"]
```
undermad: As I moved the first AIS call to the CheckExistingIdentity lambda, there may be cases where the JAR from the auth team comes without the reprove_identity flag. In that case, it will trigger the IPV_JOURNEY_START event with this property set to false. Later, it may be detected in the next step, which will then fire the IPV_ACCOUNT_INTERVENTION_START event containing this information.

Also, since this property is marked as not required in the extension in the event-catalogue, we could remove it from the launched event if it’s set to false in the auth JAR. I’m going to confirm this with the data team, but I’d appreciate any advice on it in the meantime.

thebauSoftwire: I agree, we can probably remove this extension from the `IPV_JOURNEY_START` audit event. It looks like it was [added a while ago](https://github.com/govuk-one-login/ipv-core-back/pull/1451) to keep track of identity proofing journeys started due to account interventions but since then, we have added separate [account_intervention audit events](https://github.com/govuk-one-login/ipv-core-back/pull/2076) which includes the type of intervention

thebauSoftwire: It would also mean we can clean up any remaining `initialAccountInterventionState` logic from `inititialise-ipv-session`

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/exception/AccountInterventionException.java
```diff
@@ -4,6 +4,7 @@
 
 @ExcludeFromGeneratedCoverageReport
 public class AccountInterventionException extends Exception {
+
     public AccountInterventionException() {
         super(
                 "Account intervention. This is thrown when an intervention has been discovered in ProcessCandidateIdentity.");
```
thebauSoftwire: Should we update this description since this exception is now being used in `check-existing-identity`?

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -0,0 +1,122 @@
...
+public final class AccountInterventionEvaluator {
+
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private AccountInterventionEvaluator() {
+        // prevent initialisation
+    }
+
+    public static boolean isInitialAccountInterventionDetected(
```
thebauSoftwire: I think this function name is a little misleading as `reprove_identity` is still an intervention, maybe we could re-name to something like `hasUnsupportedAccountIntervention` (or similar, I'm not good at coming up with names 😅 )?

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -0,0 +1,122 @@
...
+                && initialAccountInterventionState.isSuspended()
+                && initialAccountInterventionState.isReproveIdentity()) {
+            return false;
+        }
+
+        if (!initialAccountInterventionState.isBlocked()
+                && !initialAccountInterventionState.isResetPassword()
+                && !initialAccountInterventionState.isSuspended()) {
+            return false;
+        }
```
thebauSoftwire: it looks like the `AccountInterventionStatusDto` has an `Intervention.description` property which contains the AIS intervention type. Since we only support the `AIS_FORCED_USER_IDENTITY_VERIFY`, perhaps we could simplify by checking that the returned intervention from this initial call is this type and throw if not? 

Since `AIS_NO_INTERVENTION, AIS_ACCOUNT_UNSUSPENDED, AIS_ACCOUNT_UNBLOCKED` is already covered by `hasNoInterventionFlag(...)`. This could mean in future, we could extend this by checking against an EnumSet of interventions we support, but let me know what you think!

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
thebauSoftwire, line 1: nice!

api-tests/features/account-intervention/journey-ending-interventions.feature
```diff
@@ -94,29 +91,23 @@ Feature: Journey ending interventions
...
-      | Password reset and reprove identity | end   | AIS_FORCED_USER_IDENTITY_VERIFY                    | AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY |
+      | intervention                        | first_ais_response                                 | second_ais_response                                |
+      | Blocked                             | AIS_FORCED_USER_IDENTITY_VERIFY                    | AIS_ACCOUNT_BLOCKED                                |
+      | Suspended                           | AIS_FORCED_USER_IDENTITY_VERIFY                    | AIS_ACCOUNT_SUSPENDED                              |
+      | Password reset                      | AIS_FORCED_USER_IDENTITY_VERIFY                    | AIS_FORCED_USER_PASSWORD_RESET                     |
+      | Password reset and reprove identity | AIS_FORCED_USER_IDENTITY_VERIFY                    | AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY |
 
-  Scenario Outline: <intervention> intervention <when> of update identity journey
+  Scenario: <intervention> intervention of update identity journey
```
thebauSoftwire: I think the `<intervention>` can be removed from here now since you've simplified it :)

thebauSoftwire (commented):
Nice! Just a few small things and some suggestions

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -353,76 +350,6 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
         }
     }
 
-    private void updateIpvSessionWithIntervention(IpvSessionItem ipvSessionItem) {
```
thebauSoftwire, line 356: nice refactor!

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
undermad: I agree, we could avoid inconvenient boolean logic checks and simplify our code. As `AisInterventionType` determine what action we should take it may be unnecessary to check it against boolean values in this case.

undermad: I crafted this method that is much cleaner, but it return value based on intervention type not a state - where in the code we also create a state based on the intervention type.

```
    public static boolean hasInvalidAccountIntervention(AisInterventionType interventionType) {
        return switch (interventionType) {
            case AIS_NO_INTERVENTION,
                    AIS_FORCED_USER_IDENTITY_VERIFY,
                    AIS_ACCOUNT_UNSUSPENDED,
                    AIS_ACCOUNT_UNBLOCKED ->
                    false;
            default -> {
                LOGGER.info(
                        LogHelper.buildLogMessage(
                                "Intervention detected at the start of the journey. Intervention type: %s"
                                        .formatted(interventionType)));
                yield true;
            }
        };
    }
```

It seems to be very reasonable to go with this approach but Jira ticked also specify that if flag isReproveIdentity is set to true and all other flags are false we should also allow user to proceed. Currently there is no code implemented that would create this state based on the AccountInterventionType (see [getStateByIntervention](https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java#L38) in `AisService`).

I assume this is simply not possible as very likely isSuspended will also be set to true when isReproveIdentity is true.
Please advice
cc: @thebauSoftwire , @blakeyp

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
```diff
@@ -1399,8 +1428,123 @@ void shouldNotReturnJourneyRepeatFraudCheckResponseIfNotExpiredFraudAndFlagIsTru
...
+                                    AisInterventionType.AIS_FORCED_USER_IDENTITY_VERIFY)),
+                    Arguments.of(
+                            new AccountInterventionStateWithType(
+                                    AccountInterventionState.builder()
+                                            .isBlocked(false)
+                                            .isSuspended(false)
+                                            .isReproveIdentity(true)
+                                            .isResetPassword(false)
+                                            .build(),
+                                    AisInterventionType.AIS_FORCED_USER_IDENTITY_VERIFY)));
```
undermad: If we use check first ais call based on enum, is this outcome possible? I assume it may come also with isSuspended flag set to true in this case.

api-tests/data/audit-events/reprove-identity-journey.json
undermad: I think we should keep default `initialAccountInterentionState` initialisation in `intialise-ipv-session` lambda as if we switch off `AIS_ENABLED` and then switch it on it may throw null pointer exception in mid journey AIS call. 
It could be worth to remove this once we decide to remove `AIS_ENABLED` flag.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
thebauSoftwire: ah I was under the impression that we could only get one of the interventions specified in [this table](https://govukverify.atlassian.net/wiki/spaces/AIS/pages/4579591069/Account+Status+Table) and if more combinations of states were added, an associated type would also be added for it. But if this is the case:
```
Jira ticked also specify that if flag isReproveIdentity is set to true and all other flags are false we should also allow user to proceed
```
then I guess this won't work unless there's an extra check to cover it. Probably one for @blakeyp

api-tests/data/audit-events/reprove-identity-journey.json
thebauSoftwire: That's a fair point! I think it would be fine to eventually move it within `check-existing-identity` so all the initial AIS logic is in one place (if/when we remove `extensions.reprove_identity` from the `IPV_JOURNEY_START` event). I don't think it would cause a null pointer exception if `initialAccountInterventionState` got moved to `check-existing-identity` as we would still save the reprove result from the orch JAR request onto the session if the AIS flag is disabled

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
thebauSoftwire: I just realised that there is one place where the state is like this and that's when we first initialise the interventionState based on just the reprove flag on the orch JAR request [here](https://github.com/govuk-one-login/ipv-core-back/blob/806586ef6e0ca119f5f52409e3ae43c39332eb73/lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java#L178-L184). (Though I wonder if we should be marking it as suspended too?) cc @blakeyp

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
```diff
@@ -218,7 +194,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                             ipAddress);
 
             AuditExtensionsIpvJourneyStart extensionsIpvJourneyStart =
-                    new AuditExtensionsIpvJourneyStart(isReproveIdentity, vtr);
+                    new AuditExtensionsIpvJourneyStart(isReproveIdentity ? true : null, vtr);
```
blakeyp, line 205: I think we should remove the `reprove_identity` flag from this event entirely, but we'll need to raise a catalogue PR and check Data are happy before we merge this ⚠️ 
- This flag is really just duplicating the `type` field on the `IPV_ACCOUNT_INTERVENTION_START` event
- Reprove intervention should now always be handled by the AIS API call rather than the old claim received in the Orch JAR (we probably want a new ticket to clean up this and the feature flag entirely)

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
blakeyp: This approach looks good 👍 that's correct a reprove intervention will have both reprove and suspended flags set to true

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
blakeyp: Yeah if we're accurately mirroring a reprove intervention here then suspended should be set to true too (though this will get factored out soon as we'll stop checking the claim in the Orch JAR)

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -0,0 +1,102 @@
...
+
+        // If the user is currently reproving their identity and it has been detected
+        if (notBlockedAndNotPasswordReset(
+                        initialAccountInterventionState, currentAccountInterventionState)
+                && initialAccountInterventionState.isSuspended()
+                && !currentAccountInterventionState.isSuspended()
+                && initialAccountInterventionState.isReproveIdentity()
+                && !currentAccountInterventionState.isReproveIdentity()) {
+            return false;
+        }
```
blakeyp, line 100: Do we really need to check states here anymore? I think this now boils down to:
- if AIS_FORCED_USER_IDENTITY_VERIFY at start *and* end of journey, allow user through (intervention not lifted yet)
- if no intervention at start but AIS_FORCED_USER_IDENTITY_VERIFY at end, stop user
- if any other not allowed intervention at end, stop user

?

We're not going to be switching off the AIS feature flag so I don't think we need to worry about that (i.e. the AIS call for reprove at the start will always override the Orch JAR claim now) - but cleaning this up isn't in scope of this ticket so we can defer this to a follow-up ticket

api-tests/data/audit-events/reprove-identity-journey.json
undermad: Hi @thebauSoftwire, attempted to move `initialAccountInterventionState/initialAisInterventionType` initialisation to `check-existing-identity` but currently this is not possible as MFA journeys doesn't go through this lambda and later on in `process-candidate-identity` it uses it to compare against TICF response.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
undermad: Refactored mid journey ais call to use enum instead of states, as staging is required to avoid interrupting in flight journeys I going to raise another clean up PR once this one will be approved. Already started working on it.

lambdas/initialise-ipv-session/src/main/java/uk/gov/di/ipv/core/initialiseipvsession/InitialiseIpvSessionHandler.java
blakeyp: We're deferring this to a new ticket [PYIC-8616](https://govukverify.atlassian.net/browse/PYIC-8616) and I'm checking with Orch about if/when they actually plan to stop sending us the reprove_identity JAR claim

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -268,12 +265,28 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
...
                     && !SKIP_AIS_TYPES.contains(processIdentityType)
                     && !StringUtils.isBlank(userId)) {
-                var interventionState = aisService.fetchAccountState(userId);
-                if (midJourneyInterventionDetected(
-                        ipvSessionItem.getInitialAccountInterventionState(), interventionState)) {
-                    throw new AccountInterventionException();
+                var interventionStateWithType = aisService.fetchAccountStateWithType(userId);
+
+                var initialInterventionState = ipvSessionItem.getInitialAccountInterventionState();
+                var currentInterventionState = interventionStateWithType.accountInterventionState();
```
blakeyp, line 273: A nice addition would be to add a comment on the bits to be removed in the clean up PR following staging out

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -0,0 +1,141 @@
...
+        // Otherwise interventions are invalid
+        LOGGER.info(
+                LogHelper.buildLogMessage(
+                        "Mid journey intervention detected. Initial intervention: %s Final intervention: %s"
+                                .formatted(
+                                        initialAisInterventionType, currentAisInterventionType)));
+        return true;
+    }
+
+    private static boolean isValidIntervention(AisInterventionType aisInterventionType) {
```
blakeyp, line 66: could this be reused in `hasInvalidAccountIntervention` ? and then maybe `isMidJourneyInterventionDetected` could be simplified?

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -0,0 +1,141 @@
...
+public final class AccountInterventionEvaluator {
+
+    private static final Logger LOGGER = LogManager.getLogger();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private AccountInterventionEvaluator() {
+        // prevent initialisation
+    }
+
+    public static boolean hasInvalidAccountIntervention(AisInterventionType interventionType) {
```
blakeyp: To align with the naming of the other public method and given the log refers to 'start of journey', could rename this `isStartOfJourneyInterventionDetected` ?

undermad: Fixed!

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
undermad: Great point, reused that in `isStartOfJourneyInterventionDetected` method also refactored `isMidOfJourneyInterventionDetected` to improve readability.

blakeyp (dismissed):

blakeyp (approved):

PR #3379, opened 2025-08-20
PYIC-8595: Add VOT info to IPV_IDENTITY_REUSE_COMPLETE event

AmritSidhu:
### What changed

- Add VOT info to IPV_IDENTITY_REUSE_COMPLETE event

### Why did it change

- Existing event data contains the VTR, the returned VOT and the technically achieved VOT in the stored identity for the current journey. But the data team would like to track whether stored identities change between P2 and P3 during reuse, so we need to emit extra data

### Issue tracking

- [PYIC-8595](https://govukverify.atlassian.net/browse/PYIC-8595)

## Checklists

[PYIC-8595]: https://govukverify.atlassian.net/browse/PYIC-8595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -18,6 +18,7 @@ import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionAccountIntervention;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionPreviousAchievedVot;
 import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensionPreviousIpvSessionId;
+import uk.gov.di.ipv.core.library.auditing.extension.AuditExtensions;
 import uk.gov.di.ipv.core.library.auditing.restricted.AuditRestrictedDeviceInformation;
 import uk.gov.di.ipv.core.library.cimit.exception.CiRetrievalException;
 import uk.gov.di.ipv.core.library.cimit.service.CimitService;
@@ -574,11 +575,11 @@ public class CheckExistingIdentityHandler
         var previousIpvSessionItem =
                 ipvSessionService.getIpvSessionByClientOAuthSessionId(
                         criOAuthSessionItem.getClientOAuthSessionId());
-        sendAuditEventWithPreviousIpvSessionId(
+        sendAuditEventWithExtension(
                 AuditEventTypes.IPV_APP_SESSION_RECOVERED,
                 auditEventUser,
                 deviceInformation,
-                previousIpvSessionItem.getIpvSessionId());
+                new AuditExtensionPreviousIpvSessionId(previousIpvSessionItem.getIpvSessionId()));
 
         sessionCredentialsService.persistCredentials(
                 credentialBundle.credentials, auditEventUser.getSessionId(), true);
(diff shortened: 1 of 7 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -638,6 +648,17 @@ private void sendAuditEventWithPreviousIpvSessionId(
                         new AuditRestrictedDeviceInformation(deviceInformation)));
     }
 
+    private void sendAuditEventWithPreviousAchievedVot(
```
thebauSoftwire: There's a similar method above this `sendAuditEventWithPreviousIpvSessionId` which does something pretty similar. I wonder if we could commonise these and just pass in an `AuditExtensions`? That way, if someone wanted to create an event with another type of extension, it doesn't create a precedent of creating another function. This does mean we do lose some readability from the function name though, so just a suggestion!

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -663,4 +684,17 @@ private boolean isReprovingWithF2f(
                 && f2fStatus.isReproveIdentity()
                 && vcBundle.isPendingReturn();
     }
+
+    private Vot getStrongestAchievableVotFromBundle(List<VerifiableCredential> credentials) {
+        try {
+            boolean correlated = userIdentityService.areVcsCorrelated(credentials);
+            var result =
+                    votMatcher.findStrongestMatches(
+                            List.of(Vot.P2, Vot.P1), credentials, List.of(), correlated);
```
thebauSoftwire: Might have missed something from the ticket, is there a reason we aren't including P3 here? Should this be `SUPPORTED_VOTS_BY_DESCENDING_STRENGTH` instead?

thebauSoftwire (commented):
Just one question around the vots we check against

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
hugh-emerson: The first parameter passed to votMatcher.findStrongestMatches doesn't really matter here because we aren't looking at result.strongestRequestedMatch. You could even pass in an empty list, but technically SUPPORTED_VOTS_BY_DESCENDING_STRENGTH does let the for loop complete slightly sooner... so I we may as well pass that in.

Either way, List.of(Vot.P2, Vot.P1) is a bit confusing.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
hugh-emerson: Yeah I agree a common sendAuditEventWithExtensions function is probably slightly tidier.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
thebauSoftwire: ahh good point

hugh-emerson (dismissed):
Looks good to me 👍

thebauSoftwire (dismissed):

thebauSoftwire (approved):

PR #3386, opened 2025-08-27
PYIC-8573: Clean up

undermad:
### What changed

- Removed initialAccountInterventionState from IpvSessionItem as we now using enum instead of states.
 
### Why did it change

As we decided to change mid journey ais call to check it against enum instead of states, we had to stage out those changes to not interrupt in-flight journeys. Once previous PR will be merged, this will clean up remaining code. 

https://github.com/govuk-one-login/ipv-core-back/pull/3375

### Issue tracking

- [PYIC-8573](https://govukverify.atlassian.net/browse/PYIC-8573)

## Checklists

[PYIC-8573]: https://govukverify.atlassian.net/browse/PYIC-8573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
```diff
@@ -25,16 +27,6 @@ public AisService(AisClient aisClient) {
...
-        try {
-            var interventionDetails = aisClient.getAccountInterventionStatus(userId);
-            return interventionDetails.getState();
-        } catch (Exception e) {
-            LOGGER.error(AIS_FAIL_OPEN_ERROR_DESCRIPTION, e);
-            return getStateByIntervention(AisInterventionType.AIS_NO_INTERVENTION);
-        }
-    }
-
     public AccountInterventionStateWithType fetchAccountStateWithType(String userId) {
```
DanCorderIPV: Is this still needed? It seems to only be used in test files

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
```diff
@@ -50,55 +42,17 @@ public AccountInterventionStateWithType fetchAccountStateWithType(String userId)
...
-                            .build();
-            default ->
-                    AccountInterventionState.builder()
-                            .isBlocked(false)
-                            .isSuspended(false)
-                            .isReproveIdentity(false)
-                            .isResetPassword(false)
-                            .build();
-        };
+    public AisInterventionType fetchAisInterventionType(String userId) {
```
DanCorderIPV, line 28: It's probably a bit late, and hopefully this was all discussed earlier on, but I'm a little concerned about moving to the enum value. At the moment it's not a problem, but the logic there could potentially get messy in the future - e.g. I don't think there's any actual difference between `AIS_NO_INTERVENTION`, `AIS_ACCOUNT_UNSUSPENDED`, and `AIS_ACCOUNT_UNBLOCKED`

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
DanCorderIPV: If not then I think we can also delete `AccountInterventionStateWithType`

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -266,30 +266,14 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
...
-                } else {
-                    if (AccountInterventionEvaluator.isMidOfJourneyInterventionDetected(
-                            initialInterventionType, currentInterventionType)) {
-                        throw new AccountInterventionException();
-                    }
-                    ipvSessionItem.setAisInterventionType(currentInterventionType);
-                }
+                var currentInterventionType = aisService.fetchAisInterventionType(userId);
 
+                if (AccountInterventionEvaluator.isMidOfJourneyInterventionDetected(
```
DanCorderIPV: I know that it's often recommended that methods returning boolean value start with `is` or `has` but here it reads really strangely to me. Also MidOfJourney should really be MidJourney or MiddleOfJourney to sound natural.
I think `midJourneyInterventionDetected()` reads nicely, but `hasMidJourneyIntervention()` would also work

DanCorderIPV (changes requested):
Have we considered whether we actually need to store anything in the session now that we are immediately acting on initial interventions?
The only case I can see where we care about the initial intervention value is if we have gone down a reprove identity journey and the final AIS check also returns REPROVE_IDENTITY.
Is there another way we could tell that we were on the reprove identity journey and just remove all AIS information from the session?

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
undermad: I'm not best in naming variables or methods but this is important to keep them accurate and descriptive as possible. Amended this method name and also that used at beginning of the journey to keep them similar.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
undermad: After I removed `aisInterventionType` from `IpvSessionItem` I also removed this record and relevant method in `AisService`

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
undermad: Yes, there is no flags differences between those interventions. In overall I think enum approach is much cleaner. Previously we mapped enum values to state objects which has 4 flags and compared them using boolean logic which quickly grow to inconvenient to read piece of code. I believe its easier to act based on intervention type and if in future we will need state for specific reason we could always implement this along the enum values.

undermad:
@DanCorderIPV 
Indeed you are right, we don't need to store anything now in `IpvSessionItem`, some additional refactoring was required but looks very promising. I could reuse `isReproveIdentity` from `ClientOAuthSessionItem`.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/service/AisService.java
DanCorderIPV: My concern is mostly theoretical, and at the moment this code is much easier to understand, but the flags are the real source of truth as far as I can tell.
Anyway, fine to leave this for now.

libs/ais-service/src/test/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluatorTest.java
```diff
@@ -50,66 +47,58 @@ private static Stream<Arguments> getValidAccountInterventionState() {
...
-                Arguments.of(
-                        AIS_ACCOUNT_UNSUSPENDED,
-                        AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_ACCOUNT_BLOCKED),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_ACCOUNT_SUSPENDED),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_FORCED_USER_PASSWORD_RESET),
-                Arguments.of(
-                        AIS_FORCED_USER_IDENTITY_VERIFY,
-                        AIS_FORCED_USER_PASSWORD_RESET_AND_IDENTITY_VERIFY));
+                Arguments.of(false, AIS_ACCOUNT_BLOCKED),
```
DanCorderIPV, line 58: There are now some duplicates cases here that can be deleted

libs/ais-service/src/test/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluatorTest.java
```diff
@@ -50,66 +47,58 @@ private static Stream<Arguments> getValidAccountInterventionState() {
...
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_FORCED_USER_IDENTITY_VERIFY),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_ACCOUNT_UNBLOCKED),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_ACCOUNT_UNSUSPENDED),
-                Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_NO_INTERVENTION));
+                Arguments.of(false, AIS_NO_INTERVENTION),
+                Arguments.of(false, AIS_ACCOUNT_UNSUSPENDED),
+                Arguments.of(false, AIS_ACCOUNT_UNBLOCKED),
+                Arguments.of(false, AIS_ACCOUNT_UNBLOCKED),
+                Arguments.of(false, AIS_NO_INTERVENTION),
+                Arguments.of(false, AIS_ACCOUNT_UNSUSPENDED),
```
DanCorderIPV: More duplicates here

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
```diff
@@ -42,7 +41,6 @@ public class IpvSessionItem implements PersistenceItem {
     private long ttl;
     private String emailAddress;
     private ReverificationStatus reverificationStatus;
-    private AisInterventionType aisInterventionType;
```
DanCorderIPV, line 47: Is it safe to just delete this? Will there be any issues with journeys that started before this code goes live but finish after it's live?

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -787,30 +783,138 @@ void shouldNotInterruptProcessingIfNoRelevantMidJourneyAccountInterventionIsRece
...
-                            "00",
-                            AIS_NO_INTERVENTION,
-                            AIS_NO_INTERVENTION),
-                    // Reprove identity not cleared after reproved
-                    Arguments.of(
-                            AIS_FORCED_USER_IDENTITY_VERIFY,
-                            "05",
-                            AIS_FORCED_USER_IDENTITY_VERIFY,
-                            AIS_FORCED_USER_IDENTITY_VERIFY));
+                    Arguments.of("00", AIS_NO_INTERVENTION, AIS_NO_INTERVENTION));
```
DanCorderIPV: I think this should be `Arguments.of("05", AIS_FORCED_USER_IDENTITY_VERIFY, AIS_FORCED_USER_IDENTITY_VERIFY));`

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -787,30 +783,138 @@ void shouldNotInterruptProcessingIfNoRelevantMidJourneyAccountInterventionIsRece
...
+        }
+
+        @Test
+        void shouldNotInterruptProcessingIfReproveJourneyHasAccountInterventionReproveIdentity()
```
DanCorderIPV: Isn't this covered in the tests above?
Maybe add a case `Arguments.of("00", AIS_FORCED_USER_IDENTITY_VERIFY, AIS_FORCED_USER_IDENTITY_VERIFY));` to rule out TICF

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -787,30 +783,138 @@ void shouldNotInterruptProcessingIfNoRelevantMidJourneyAccountInterventionIsRece
...
+                            eq(List.of()),
+                            eq(P2),
+                            eq(STRONGEST_MATCHED_VOT),
+                            eq(CandidateIdentityType.UPDATE),
+                            any());
+        }
+
+        @Test
+        void
+                shouldNotInterruptProcessingIfReproveJourneyHasMidJourneyAccountInterventionReproveIdentityReceivedFromTicf()
```
DanCorderIPV: I think this is also covered above

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -835,31 +939,26 @@ void shouldInterruptProcessingIfRelevantMidJourneyAccountInterventionIsReceivedF
...
+                    Arguments.of(AIS_ACCOUNT_SUSPENDED),
                     // Finally reset password
-                    Arguments.of(AIS_NO_INTERVENTION, AIS_FORCED_USER_PASSWORD_RESET),
+                    Arguments.of(AIS_FORCED_USER_PASSWORD_RESET),
                     // Reprove identity that has been triggered during the journey
-                    Arguments.of(AIS_NO_INTERVENTION, AIS_FORCED_USER_IDENTITY_VERIFY),
+                    Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY),
                     // Reprove identity that cleared during the journey but got re-suspended
-                    Arguments.of(AIS_FORCED_USER_IDENTITY_VERIFY, AIS_ACCOUNT_SUSPENDED),
+                    Arguments.of(AIS_ACCOUNT_SUSPENDED),
```
DanCorderIPV: Couple of duplicates here

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
DanCorderIPV: Oh, actually, I think I see what's happened.
Instead of deleting the first test argument here maybe just change it to `isReproveIdentity` then you can cover the test cases below

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -908,63 +1007,85 @@ private static Stream<Arguments> createRelevantTicfInterventionStates() {
...
-                            AIS_FORCED_USER_IDENTITY_VERIFY,
-                            AIS_NO_INTERVENTION,
-                            "04",
-                            AIS_FORCED_USER_PASSWORD_RESET),
+                    Arguments.of(AIS_NO_INTERVENTION, "04"));
+        }
+
+        @ParameterizedTest
+        @MethodSource("createRelevantTicfInterventionStatesForReproveJourneys")
+        void
```
DanCorderIPV: Similar to above, can you add `isReproveIdentity` to the test parameters instead of adding this test?

DanCorderIPV (changes requested):

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
undermad: Yes, shouldn't interrupt any in-flight journeys as it will be checked against `reproveIdentity` flag and new fetched account intervention type where the flag is always presented in `ClientOAuthSessionItem`.

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
undermad: This is great idea, included that in test cases.

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -974,111 +894,31 @@ void shouldInterruptProcessingIfRelevantMidJourneyAccountInterventionIsReceivedF
...
-                    Arguments.of(
-                            new AccountInterventionStateWithType(
-                                    new AccountInterventionState(false, true, true, false),
-                                    AIS_FORCED_USER_IDENTITY_VERIFY),
-                            new AccountInterventionStateWithType(
-                                    new AccountInterventionState(false, false, false, false),
-                                    AIS_NO_INTERVENTION),
-                            "05",
-                            AIS_FORCED_USER_IDENTITY_VERIFY),
+                    Arguments.of(true, AIS_NO_INTERVENTION, "05"),
```
undermad, line 916: Current implementation looking great but this is a case that I didn't think before. There may be a scenario where reprove identity journey clean the intervention on second ais call but we receive reprove identity intervention from TICF. 

Not sure yet how to resolve that, maybe we could actually keep `aisInterventionType` in `ipvSessionItem` or apply different solution. I could provide additional method specifically for TICF check that would take intervention type as parameter but that will require drill it through methods chain in process-candidate-identity lambda where there is already many parameters in each of them.

DanCorderIPV: I believe the reason that we check the TICF return value is that is quicker to update than AIS so should be more up to date.
So if we are on a reprove identity journey, the second AIS call says no intervention, but TICF says reprove identity then I guess we're in a very edge case of we successfully reproved, but there was a new reprove identity set on the user immediately afterwards.
I'd suggest that we block the user in this case and they will have to reprove their identity again. But in practice I don't think this should happen.
@blakeyp do you have any further insight on this?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
DanCorderIPV: I was more worried about desrialising the object from the data base and it having extra data that we no longer have a field for.
Googling hasn't found a definitive bit of documentation, but it looks like it should be safe and the extra data will be silently ignored.

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
blakeyp: Agreed and happy to keep this as is for now but I've asked Akhil to confirm the exact behaviour of AIS/TICF in this case

undermad: Just simply included previously implemented method for TICF intervention check. It require current intervention from AIS but as it is called in same lambda I could pass it down through methods. This solve the issue and I could completely remove any data related to account intervention from `IpvSessionItem`.

libs/ais-service/src/main/java/uk/gov/di/ipv/core/library/ais/helper/AccountInterventionEvaluator.java
```diff
@@ -48,6 +48,35 @@ public static boolean hasMidJourneyIntervention(
...
+        var reproveToValid = isIdentityVerifyToValid(current, ticfIntervention);
+
+        if (bothValid || bothReprove || reproveToValid) {
+            return false;
+        }
+
+        LOGGER.info(
+                LogHelper.buildLogMessage(
+                        "TICF intervention detected. Current intervention: %s TICF intervention: %s"
+                                .formatted(current, current)));
```
DanCorderIPV: One of these shouldn't be `current`

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
```diff
@@ -796,50 +791,23 @@ void shouldNotInterruptProcessingIfNoRelevantMidJourneyAccountInterventionIsRece
...
-                    Arguments.of(
-                            new AccountInterventionStateWithType(
-                                    new AccountInterventionState(false, false, false, false),
-                                    AIS_NO_INTERVENTION),
-                            "00",
-                            AIS_NO_INTERVENTION,
-                            new AccountInterventionStateWithType(
-                                    new AccountInterventionState(false, false, false, false),
-                                    AIS_NO_INTERVENTION)),
+                    Arguments.of(false, "00", AIS_NO_INTERVENTION, AIS_NO_INTERVENTION),
```
DanCorderIPV: I don't think the last parameters here are used? The test method only expects 3 parameters

DanCorderIPV (changes requested):

lambdas/process-candidate-identity/src/test/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandlerTest.java
undermad: This method source was originally written this way and used in two different unit tests where in each different parameters was used. I split that arguments provider method to avoid any unused variables and confusion around it.

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/persistence/item/IpvSessionItem.java
undermad: I tried to look for an answer but couldn't find either. Previously I deleted `AccountInterventionState` and nothing breaks so it should behave in this case exactly same.

DanCorderIPV (approved):

PR #3414, opened 2025-09-05
Pyic 8393 compare sis identity

DanCorderIPV:
### What changed

Add comparison of SIS identity and EVCS stored VCs

### Why did it change

So we can validate that SIS is working as expected

### Issue tracking

- [PYIC-8393](https://govukverify.atlassian.net/browse/PYIC-8393)

[PYIC-8393]: https://govukverify.atlassian.net/browse/PYIC-8393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -296,8 +296,9 @@ public Map<String, Object> handleRequest(JourneyRequest event, Context context)
             }
 
             if (configService.enabled(SIS_VERIFICATION)) {
-                // PYIC-8393 Make use of the results of this call
-                sisService.getStoredIdentity(clientOAuthSessionItem);
+                // Call this before we invalidate the stored identity!
+                sisService.compareStoredIdentityWithStoredVcs(
+                        clientOAuthSessionItem, auditEventUser);
```
blakeyp, line 301: Can we save a call to EVCS by fetching the VCs first and passing the relevant ones (only `CURRENT`) here?

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,341 @@
...
+            if (hasValidPendingReturnVcs) {
+                // + pending return VCs
+                evcsIdentityVcs.addAll(pendingReturnVcs);
+            } else {
+                // + all vcs
+                evcsIdentityVcs.addAll(vcs.getOrDefault(CURRENT, List.of()));
+            }
+
+            return evcsIdentityVcs;
+        } catch (Exception e) {
```
blakeyp, line 365: Worth logging the exceptions (here and above) ?

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,341 @@
...
+                    "Exception caught calculating VOT from EVCS VCs");
+        }
+    }
+
+    private List<VerifiableCredential> getEvcsVerifiableCredentials(
+            String userId, String evcsAccessToken) throws SisMatchException {
+        try {
+            var vcs =
+                    evcsService.fetchEvcsVerifiableCredentialsByState(
+                            userId, evcsAccessToken, true, CURRENT, PENDING_RETURN);
```
blakeyp, line 347: We should only need `CURRENT` VCs (that make up a complete identity) ?

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,341 @@
...
+            evcsVcSignatures =
+                    evcsCredentials.stream()
+                            .map(credential -> getVcSignature(credential.getVcString()))
+                            .toList();
+
+            maximumMatchedVot = calculateMaximumMatchedVot(evcsCredentials, userId);
+
+            // Compare VOTs
+            var evcsVot = maximumMatchedVot.isPresent() ? maximumMatchedVot.get().vot() : null;
+            if (storedIdentityResult.identityDetails().vot() != evcsVot) {
```
blakeyp: Need to check we're comparing the right `vot` here - believe we should be comparing with the `vot` calculated by SIS which is a claim of the JWT itself (not the top-level `vot` returned in the response)

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,341 @@
...
+                        userIdentityService, new Gpg45ProfileEvaluator(), cimitUtilityService);
     }
 
     public SisGetStoredIdentityResult getStoredIdentity(
             ClientOAuthSessionItem clientOAuthSessionItem) {
+        // We want to get any identity in SIS so we ask for all vots.
         return sisClient.getStoredIdentity(
                 clientOAuthSessionItem.getEvcsAccessToken(),
-                clientOAuthSessionItem.getVtrAsVots(),
+                List.of(Vot.P0, Vot.P1, Vot.P2, Vot.P3),
```
blakeyp: I might be missing something but I thought we would pass the VTR as is? since we want to verify SIS's calculated VOT is the same as our calculated VOT for this reuse of identity

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,341 @@
...
+                    storedIdentityResult,
+                    VerificationOutcome.FAILURE,
+                    FailureCode.UNEXPECTED_ERROR,
+                    e.getMessage(),
+                    evcsVcSignatures,
+                    sisVcSignatures);
+        }
+    }
+
+    private void sendComparisonAuditEvent(
```
blakeyp, line 215: Does this fully cover the audit event ticket? [PYIC-8394](https://govukverify.atlassian.net/browse/PYIC-8394) if so we should unblock and bring in to sprint too

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: I did originally try that but it got a bit fiddly.
Things have changed a bit since then so I can take another look

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
DanCorderIPV: I think that ticket is to create the event in the event catalogue and Thea is waiting for the various reviews that it needs.

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
DanCorderIPV: Good point, will do

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
DanCorderIPV: Leaving this as discussed - pending shouldn't matter, but if they do it's an indication that something's not working as expected

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
DanCorderIPV: Leaving this as discussed (unless the EVCS team think the extra requests will cause problems)

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
DanCorderIPV: As discussed the ticket says compare max vot, but calculated vot is more interesting. Will compare both.

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/dto/SisStoredIdentityCheckDto.java
```diff
@@ -8,6 +8,8 @@ public record SisStoredIdentityCheckDto(
         String content,
         boolean isValid,
         boolean expired,
+        // This is the maximum VoT for the user, the calculated VoT for this request is stored in
+        // the content.
```
DanCorderIPV, line 13: @thebauSoftwire is this also what you understand the SIS behaviour to be?

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,366 @@
...
+                throw new SisMatchException(
+                        FailureCode.SIS_ERROR,
+                        "Call to SIS service failed, no stored identity comparison can be made");
+            }
+
+            if (!storedIdentityResult.identityWasFound()) {
+                LOGGER.info(
+                        LogHelper.buildLogMessage(
+                                "No credential found in SIS so no comparison to do"));
+                return;
```
DanCorderIPV, line 123: @blakeyp just to highlight this - we won't log anything even if we think that there should be an existing identity. This is because people that created their account before SIS was switched on won't have IDs in SIS so we'd get a lot of false positives.

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,366 @@
...
+
+            String sisJwt = sisIdFound ? storedIdentityResult.identityDetails().content() : "";
+
+            var auditEvent =
+                    AuditEvent.createWithoutDeviceInformation(
+                            AuditEventTypes.IPV_STORED_IDENTITY_CHECKED,
+                            configService.getParameter(ConfigurationVariable.COMPONENT_ID),
+                            auditEventUser,
+                            new AuditExtensionsSisComparison(
+                                    sisRequestedVot,
```
DanCorderIPV, line 237: Re-reading the ticket, I think this is meant to be the SIS calculated VoT rather than the max VoT

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
```diff
@@ -1,29 +1,366 @@
...
+                new VotMatcher(
+                        userIdentityService, new Gpg45ProfileEvaluator(), cimitUtilityService);
     }
 
     public SisGetStoredIdentityResult getStoredIdentity(
             ClientOAuthSessionItem clientOAuthSessionItem) {
+        // We want to get any identity in SIS so we ask for all vots.
         return sisClient.getStoredIdentity(
                 clientOAuthSessionItem.getEvcsAccessToken(),
                 clientOAuthSessionItem.getVtrAsVots(),
```
DanCorderIPV, line 95: @thebauSoftwire @blakeyp do we know what SIS's behaviour is if it has a record for a user but we ask it for a VoT that it doesn't think it can provide?
I would guess that it won't return an identity and we'll behave as if there is no record in SIS

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
blakeyp: Trust & Reuse are happy with it and they'll organise another perf test to be sure

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
blakeyp: Just checking on this

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
blakeyp: I think it will also cover QA of the audit event (?) QA might need to be across all 3 tickets at this point

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/dto/SisStoredIdentityCheckDto.java
thebauSoftwire: from what I understand [based off this SPT ticket](https://govukverify.atlassian.net/browse/SPT-1620) is that it's the maximum VOT for the user based off the VTRs we send in the request. It's also a P0 if the user has an invalid SI record(ie missing vot or storedIdentityJwt)

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/service/SisService.java
blakeyp: So as it stands they'll still return the record but with content.vot = P0 and isValid set to false

libs/sis-service/src/main/java/uk/gov/di/ipv/core/library/sis/dto/SisStoredIdentityCheckDto.java
blakeyp: In that ticket:
> N.B. Confusingly, for phase 2 ONLY, there are two vot fields on the returned payload. The vot at the root level should ALWAYS be the vot stored on the identity record. The content.vot field is the one we are calculating in this story.

thebauSoftwire: have spoken in Slack and have clarified that the comment you've added is correct :) (ignore the comment I've made above!)

blakeyp (approved):

PR #3420, opened 2025-09-08
PYIC-8391: Add consumer tests for sis /user-identity endpoint.

undermad:
### What changed

- Added consumer tests for /user-identity endpoint that include 200/401/403/404/500 responses

### Why did it change

- To establish integration between IpvCore and StoredIdentityService api

### Issue tracking

- [PYIC-8391](https://govukverify.atlassian.net/browse/PYIC-8391)

## Checklists

[PYIC-8391]: https://govukverify.atlassian.net/browse/PYIC-8391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/sis-service/src/test/java/uk/gov/di/ipv/core/library/sis/pact/ContractTest.java
```diff
@@ -0,0 +1,233 @@
...
+        when(mockConfigService.getParameter(ConfigurationVariable.SIS_APPLICATION_URL))
+                .thenReturn("http://localhost:" + mockServer.getPort());
+        sisClient = new SisClient(mockConfigService);
+    }
+
+    @Pact(provider = "StoredIdentityServiceProvider", consumer = "IpvCoreBack")
+    public RequestResponsePact validGetStoredIdentityRequestReturns200(
+            PactDslWithProvider builder) {
+        return builder.given(String.format("%s is a valid vtr list", TEST_VOTS))
+                .given(String.format("%s is a valid journey id", TEST_JOURNEY_ID))
```
thebauSoftwire: it might also be useful to include what we expect the vot the SI record should have here so SIS know that for this test, when we request vtr=["P1", "P2"], that the SI record should have a vot that is either P2 or P3 in order for the endpoint to return `response.vot=P2`

thebauSoftwire (commented):
Nice! We could also add a couple more test cases based on previous ACs:
- 400: `vtr` and `govukSigninJourneyId` are required properties in the [request body](https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/12bac6383cfe76b74f738cd4e6998cc86a9439f8/src/specs/api.yaml#L59-L78), we could add a test for when either one of these isn't present
- 200 Malformed JSON: returned when SIS get an SI record that doesn't have a required property such as the `vot` so this wouldn't be returned to us (see last AC on [this ticket](https://govukverify.atlassian.net/browse/PYIC-8381) for more info) but we'd still get a 200

undermad:
Changed also `.given` descriptions to be more accurate

thebauSoftwire (approved):
LGTM!

PR #3440, opened 2025-09-16
PYIC-8624: set FMS tags on core-back resources

thebauSoftwire:
### What changed

- set FMS tags on core-back resources

### Why did it change

- As part of transitioning FMS WAF rule enforcement from COUNT to BLOCK, Security have set up custom policies for us to apply to our resources in scope. In COUNT mode FMS merely logs the rule violations, in BLOCK mode violating requests are actively blocked.
- Setting the tag values to neither true/false means they won't get picked up at all

### Issue tracking

- [PYIC-8624](https://govukverify.atlassian.net/browse/PYIC-8624)

## Checklists

[PYIC-8624]: https://govukverify.atlassian.net/browse/PYIC-8624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -395,6 +395,9 @@ Resources:
               Resource:
                 - 'execute-api:/*'
       StageName: !Sub ${Environment}
+      Tags:
+        FMSRegionalPolicy: !If [ IsDevOrBuildOrStaging, "false", "none" ]
+        CustomPolicy: !If [ IsDevOrBuildOrStaging, "true", "none" ]
       TracingEnabled: true
       AccessLogSetting:
         DestinationArn: !GetAtt IPVCorePrivateAPILogGroup.Arn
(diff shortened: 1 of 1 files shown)


deploy/template.yaml
```diff
@@ -870,6 +879,9 @@ Resources:
           Name: "AWS::Include"
           Parameters:
             Location: "../openAPI/core-back-analytics.yaml"
+      Tags:
+        FMSRegionalPolicy: false
+        CustomPolicy: true
```
blakeyp: While this APIGW is only used in the lower envs by the journey map, it is deployed up to the int/prod accounts so we should probably still conditionalise it?

blakeyp (approved):

PR #3445, opened 2025-09-18
PYIC-7655 Tidy up CRI API request headers

blakeyp:
### What changed

Remove workarounds for Content-Type header and set correct Accept header for CRI API requests

### Why did it change

While testing the integration with DWP KBV CRI last year it was discovered that for the synchronous CRI credential request we were explicitly setting the `Content-Type` header with value empty string. This was seemingly introduced because NINO CRI rejected our requests with an "Unsupported media type" error because the nimbus lib set a Content-Type header `application/x-www-form-urlencoded` by default in its underlying HTTP request even with no request body. However, DWP KBV CRI was found to be rejecting the `Content-Type` header empty string, understandably because empty string is not a valid content type.

Ideally we wouldn't be sending a `Content-Type` at all in these requests (since we aren't providing a body) but there didn't seem to be a way to omit this header through nimbus. However, earlier in the year we switched the underlying HTTP client for these requests to our own tracing client, which means we can now do away with these header workarounds (Content-Type header won't be set at all now).

Also we should make sure we're setting the correct `Accept` header: `application/jwt, application/json` because we support both responses - JWT for a synchronous CRI response and JSON for asynchronous.

### Issue tracking

- [PYIC-7655](https://govukverify.atlassian.net/browse/PYIC-7655)

## Checklists

[PYIC-7655]: https://govukverify.atlassian.net/browse/PYIC-7655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
+++ b/libs/cri-api-service/src/main/java/uk/gov/di/ipv/core/library/criapiservice/CriApiService.java
@@ -52,9 +52,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Objects;
 
-import static uk.gov.di.ipv.core.library.config.EnvironmentVariable.ENVIRONMENT;
 import static uk.gov.di.ipv.core.library.domain.Cri.DCMAW;
-import static uk.gov.di.ipv.core.library.domain.Cri.DWP_KBV;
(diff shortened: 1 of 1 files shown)


hugh-emerson (approved):
LGTM.

Maybe unit tests should cover the headers, but not a big deal

PR #3449, opened 2025-09-22
PYIC-8298: Allow InitialiseIpvSession to decrypt with any KMS key

MikeCollingwood:
### What changed

- Allow InitialiseIpvSession to decrypt with any KMS key

### Why did it change

- So we can rotate keys in future. At the moment we don't allow decryption with different keys, which means this would have to be updated exactly when we update the key, which we cannot do.

### Issue tracking

- [PYIC-8298](https://govukverify.atlassian.net/browse/PYIC-8298)

## Checklists

[PYIC-8298]: https://govukverify.atlassian.net/browse/PYIC-8298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1345,7 +1345,7 @@ Resources:
                 - 'kms:Decrypt'
                 - 'kms:GenerateDataKey'
               Resource:
-                - !ImportValue CoreEncryptionKeyArn
+                - !Sub "arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/*"
             - Sid: kmsAuditEventQueueEncryptionKeyPermission
               Effect: Allow
               Action:


deploy/template.yaml
```diff
@@ -1345,7 +1345,7 @@ Resources:
                 - 'kms:Decrypt'
                 - 'kms:GenerateDataKey'
               Resource:
-                - !ImportValue CoreEncryptionKeyArn
+                - !Sub "arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/*"
```
blakeyp, line 1348: Is there any prefix we can add to restrict it further rather than allowing any key in our account?

MikeCollingwood: Unfortunately not, the arns look like:
```
arn:aws:kms:eu-west-2:335257547869:key/897xyz34b
```

blakeyp (approved):

PR #3468, opened 2025-09-30
PYIC-7697: Clean up fallback key configuration

MikeCollingwood:
### What changed

- Clean up fallback key configuration

### Why did it change

- Redundant logic now

### Issue tracking

- [PYIC-7697](https://govukverify.atlassian.net/browse/PYIC-7697)

## Checklists

[PYIC-7697]: https://govukverify.atlassian.net/browse/PYIC-7697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/dto/OauthCriConfig.java
```diff
@@ -24,13 +20,7 @@ public class OauthCriConfig extends RestCriConfig {
     @NonNull private URI tokenUrl;
     @NonNull private String clientId;
     private URI authorizeUrl;
-    @RemoveEscapedQuotationMarks private String encryptionKey;
```
blakeyp, line 27: Do we need to keep this as an optional initially and then remove it later once we've merged this PR _and_ removed all the keys from config?

otherwise
- if we merge this PR first, config validation will fail because encryption key is removed from Config class (?)
- if we merge config PRs first, core-back will fail trying to parse an empty key (?)

🤔

thebauSoftwire: good point! Have put it back

blakeyp (approved):

PR #3485, opened 2025-10-09
BAU: refactor f2f queue contract tests to use data vocab classes

thebauSoftwire:
### What changed

- refactor f2f queue contract tests to use data vocab classes

### Why did it change

- neaten up the contract tests so we're not traversing VC with JSON nodes


DanCorderIPV (approved):
Nice :D

PR #3498, opened 2025-10-15
PYIC-8724: Update Manual F2F Pending Reset Lambda to throw errors on failure

AmritSidhu:
### What changed

- updates the manual-f2f-pending-reset Lambda so that execution failures are surfaced as red (failed) in the AWS Lambda console. 
- Now, when a failure occurs (e.g., missing user ID, no F2F record found, or an internal exception), the handler throws a RuntimeException, ensuring the Lambda execution result correctly reflects a failure.

Screenshot 2025-10-15 at 13 22 46
Screenshot 2025-10-15 at 13 23 00

### Why did it change

- Previously, even when an operation failed, the Lambda reported a “Succeeded” status because errors were returned in the response map instead of being thrown.

### Issue tracking

- [PYIC-8724](https://govukverify.atlassian.net/browse/PYIC-8724)

## Checklists

[PYIC-8724]: https://govukverify.atlassian.net/browse/PYIC-8724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/manual-f2f-pending-reset/src/main/java/uk/gov/di/ipv/core/manualf2fpendingreset/ManualF2fPendingResetHandler.java
+++ b/lambdas/manual-f2f-pending-reset/src/main/java/uk/gov/di/ipv/core/manualf2fpendingreset/ManualF2fPendingResetHandler.java
@@ -12,6 +12,7 @@ import uk.gov.di.ipv.core.library.auditing.AuditEventTypes;
 import uk.gov.di.ipv.core.library.auditing.AuditEventUser;
 import uk.gov.di.ipv.core.library.criresponse.service.CriResponseService;
 import uk.gov.di.ipv.core.library.domain.Cri;
+import uk.gov.di.ipv.core.library.exceptions.ManualF2fPendingResetException;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.persistence.item.CriResponseItem;
 import uk.gov.di.ipv.core.library.service.AuditService;
@@ -26,7 +27,6 @@ public class ManualF2fPendingResetHandler implements RequestHandler<String, Map<
 
     private static final String RESULT_KEY = "result";
     private static final String MESSAGE_KEY = "message";
-    private static final String RESULT_ERROR = "error";
     private static final String RESULT_SUCCESS = "success";
 
     private final CriResponseService criResponseService;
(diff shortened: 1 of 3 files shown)


lambdas/manual-f2f-pending-reset/src/main/java/uk/gov/di/ipv/core/manualf2fpendingreset/ManualF2fPendingResetHandler.java
```diff
@@ -85,14 +72,46 @@ public Map<String, Object> handleRequest(String input, Context context) {
...
         return response;
     }
+
+    private void validateInput(String input) {
+        if (input == null || input.isBlank()) {
+            throw new ManualF2fPendingResetException("Missing or empty userId in input");
+        }
+    }
+
+    private CriResponseItem findPendingRecord(String userId) {
```
blakeyp: Should this be void? we're returning the item but not actually using it in the calling code

lambdas/manual-f2f-pending-reset/src/main/java/uk/gov/di/ipv/core/manualf2fpendingreset/ManualF2fPendingResetHandler.java
```diff
@@ -85,14 +71,44 @@ public Map<String, Object> handleRequest(String input, Context context) {
...
+        try {
+            if (criResponseService.getCriResponseItem(userId, Cri.F2F) == null) {
+                throw new ManualF2fPendingResetException("No F2F pending record found.");
+            }
+        } catch (ManualF2fPendingResetException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new ManualF2fPendingResetException("Failed to look up F2F record", e);
+        }
+    }
```
undermad: Hi Amrit, since this method doesn't return any value, maybe we should rename it to "checkIfPendingRecordExist" or something similar. Word "find" indicate it may return a value.

Also, this Throw e is somehow hard to understood and `getCriResponseItem` doesn't throw any exception

```
        var response = criResponseService.getCriResponseItem(userId, Cri.F2F);
        if(Objects.isNull(response)) {
            throw new ManualF2fPendingResetException("Failed to look up F2F record");
        }
```

undermad (commented):
Can we also bring back Arrange/Act/Assert to follow the pattern that we use in other places?

lambdas/manual-f2f-pending-reset/src/main/java/uk/gov/di/ipv/core/manualf2fpendingreset/ManualF2fPendingResetHandler.java
```diff
@@ -85,14 +71,44 @@ public Map<String, Object> handleRequest(String input, Context context) {
...
+
+    private void deletePendingRecord(String userId) {
+        try {
+            criResponseService.deleteCriResponseItem(userId, Cri.F2F);
```
undermad: Same here, `deleteCriResponseItem` doesn't throw any exception so try catch is not required.
If for some reason we face any Exception it should be handled in the main handler, where we catch Exception and re-throw.

undermad (approved):

PR #3530, opened 2025-10-23
PYIC-8698: highlight edges and their labels on the journey map

thebauSoftwire:
### What changed

- adds click handlers to edges and edge labels so that they turn black/bold to highlight associated edges + their label
- makes the default colour for the edges darker so it's easier to see them
- sets the min edge width to 2 instead of 1, again for visibility and to make it easier to select them

| Example | Screenshot |
|-|-|
| Selected edge is also a highlighted edge | Screenshot 2025-10-23 at 13 50 10 |
| Selected edge is non-highlighted edge | Screenshot 2025-10-23 at 13 50 39 |

### Why did it change

- Atm, it's quite difficult which edge each edge label (event) belongs to. This updates the journey map so that clicking on either the edge/edge label highlights the associated edge label/edge

### Issue tracking

- [PYIC-8698](https://govukverify.atlassian.net/browse/PYIC-8698)

## Checklists

[PYIC-8698]: https://govukverify.atlassian.net/browse/PYIC-8698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/public/style.css
+++ b/journey-map/public/style.css
@@ -236,3 +236,6 @@ form fieldset p {
 span.edgeLabel .defaultEdgeLabel:last-child {
   border-bottom: none;
 }
+span.edgeLabelIdentifier {
+  text-shadow: 0.5px 0px 0px #555, -0.5px 0px 0px #555;
+}
--- a/journey-map/public/style.css
+++ b/journey-map/public/style.css
@@ -237,5 +237,5 @@ span.edgeLabel .defaultEdgeLabel:last-child {
   border-bottom: none;
 }
 span.edgeLabelIdentifier {
-  text-shadow: 0.5px 0px 0px #555, -0.5px 0px 0px #555;
+  text-shadow: 0.3px 0px 0px #555, -0.3px 0px 0px #555;
 }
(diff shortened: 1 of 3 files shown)


journey-map/public/style.css
```diff
@@ -236,3 +236,6 @@ form fieldset p {
 span.edgeLabel .defaultEdgeLabel:last-child {
   border-bottom: none;
 }
+span.edgeLabelIdentifier {
+  text-shadow: 0.45px 0px 0px #333, -0.45px 0px 0px #333;
+}
```
thebauSoftwire, line 241: When making the text bold, it would get cropped as the labels are a fixed size after being rendered. This creates a "bold" look to prevent the text from being cropped as it's not actually making them bigger. I did also try and use pseudo-elements to set the size of the label to as if the text was originally bold but that also didn't work and formatted it weirdly :(

undermad: In general attaching shadows may create performance issues but it should be fine as only one label will be emphasised.

undermad (approved):
It looking awesome!

PR #3542, opened 2025-10-31
PYIC-8734: Update documentation on the journey engine + map

thebauSoftwire:
### What changed

- Moved old journey engine docs to `process-journey-event`
- Added docs for journey map syntax

There is a team manual PR [here](https://github.com/govuk-one-login/team-manual/pull/1130) to act as the reminder to update these READMEs.

### Why did it change

- The journey map has evolved and new features and have been added  and we had no docs for them

### Issue tracking

- [PYIC-8734](https://govukverify.atlassian.net/browse/PYIC-8734)

## Checklists

[PYIC-8734]: https://govukverify.atlassian.net/browse/PYIC-8734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+## Sub-Journey States
+Journey States represent a point in the user's journey.
+A basic state is defined by a name, a list of `events` that state accepts and optionally, a `response` type.
+```yaml
+BASIC_STATE_NAME:
+  response: # This defines the Step Response from the State Machine when transitioning to this state
+    type: process
+    lambda: reset-session-identity
+  events:
+    event_name_1: # These are the events the state supports, which define what will happen if the event is received from this state
```
arthurs-gds: "... if the event is received from this state"
It's not clear to me from this wording where the event comes from (how it's generated) or where it goes.
Is it this:
"...what will happen if the State Machine receives event_name_1 while in this state"

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+      ...
+    event_name_2:
+      ...
+```
+
+### Sub-Journey Entry Point State
+A state can be used as an entry point to another sub-journey if the `response` is omitted. To enter a new sub-journey, we must specify the `targetJourney` and `targetState` on the event:
```
arthurs-gds: The wording here wasn't clear to me at first.  I had to think pretty hard to understand where the `response` had to be omitted.

Is this accurate?
"A state can be used as an entry point to a sub-journey if that state doesn't have a `response`."

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+    ENTRY_POINT_STATE_NAME:
+      events:
+        event_name_1:
+          ...
+        event_name_2:
+          ...
+```
+
+### State Inheritance
+It is also possible to define a common state for other states to inherit from. These also omit the `response` type and a `parent` must be defined within the state.
```
arthurs-gds: "These also omit..."  I'm not totally sure what "these" refers to here.  If it's child-states, could we say so explicitly?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+
+When a child state inherits from a parent state, it inherits all the events from the parent state without having to explicitly define the same events.
```
arthurs-gds: If the defined parent state doesn't exist in the sub-journey file, where does it come from?
For example, `CRI_STATE` in `kbvs.yaml`.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+      ...
+    exitEvent2:
+      ...
+```
+
+All exit events specified on a nested journey entry state must cover all exit events emitted from the [Nested Journey](#nested-journeys).
+
+## Response Types
+The Step Response returned by the State Machine when transitioning to this state is reflected by the `response` block. A `response` can be three types:
+* `process`: the state represents a lambda
```
arthurs-gds: Isn't it a call to a lambda?

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
```diff
@@ -0,0 +1,313 @@
...
+    type: cri
+    criId: dcmaw # this is the CRI identifier
+  events:
+    ...
+```
+
+## Events
+The `events` define what the state accepts and how the State Machine should handle when it receives a particular event for the current state.
+
+> The listed `events` must cover all of the events that could possibly be emitted by the state.
```
arthurs-gds: It feels like this line and the one before it contradict each other.  One seems to say that the events are incoming, the other that the events are outgoing.
I thought that the events defined in a state are events it can accept (eg. while in state `SORRY_CRI_TECHNICAL_PROBLEM_CRI_UK_PASSPORT` the engine can accept event `try-again` triggered by the user clicking that button)

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
thebauSoftwire: yes exactly that 😅 much better worded, thanks!

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
thebauSoftwire: ah yes, actioning a state can result in different journey events being sent to the State Machine. These events might not necessarily be handled by the State that was actioned which would cause an error which is what I was trying to warn for, in a not so well-worded way 😅 Hopefully I've worded it better now but please let me know otherwise!

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
arthurs-gds: My confusion on this one was less on where the `targetJourney` and `targetState` had to be, and more on where to omit the `response`.  I've figured out that's the entry-point-state, but it's ambiguous in line 40.

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
arthurs-gds: I'm not totally sure what actioning a state is, or having executed a state.
Let me see whether I can paraphrase what you're saying:
When the State Machine is in a given state, receiving an event not defined for that state will trigger an error.
Is that it?

thebauSoftwire: yeah, so if we created a page e.g. `live-in-uk` to send a `/journey/notSupportedEvent` when a user clicks a button but we defined a state for this in core-back like:
```
LIVE_IN_UK:
  response:
      type: page
      pageId: live-in-uk
   events:
       next:
           targetState: STATE_2
```
The user would get an error if they clicked that button because that event isn't supported

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
thebauSoftwire: ahh gotcha thanks

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
thebauSoftwire: Is this better?
> The events listed must uniquely correspond to all possible journey events that the State Machine can receive for a given state. If the State Machine receives an event that isn’t defined for its current state, it will throw an error.

arthurs-gds: Yep, excellent, thanks!

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/journey-map-syntax.md
arthurs-gds: A suggestion for line 40:
"A state that does not have a `response` defined can be used as an entry point to a sub-journey."
Is that accurate?

arthurs-gds (approved):

PR #3551, opened 2025-11-04
PYIC: 8694 Improve analytics error handling

DanCorderIPV:
### What changed

Better error messages in the journey map

### Why did it change

Confusing experience when CloudWatch is slow to respond.

### Issue tracking

- [PYIC-8694](https://govukverify.atlassian.net/browse/PYIC-8694)

[PYIC-8694]: https://govukverify.atlassian.net/browse/PYIC-8694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/fetch-journey-transitions/src/main/java/uk/gov/di/ipv/core/fetchjourneytransitions/domain/ErrorResponseBody.java
+++ b/lambdas/fetch-journey-transitions/src/main/java/uk/gov/di/ipv/core/fetchjourneytransitions/domain/ErrorResponseBody.java
@@ -1,17 +1,22 @@
 package uk.gov.di.ipv.core.fetchjourneytransitions.domain;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.Getter;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+@Getter
+@JsonInclude(value = JsonInclude.Include.NON_NULL)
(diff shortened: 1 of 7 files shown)


lambdas/fetch-journey-transitions/src/main/java/uk/gov/di/ipv/core/fetchjourneytransitions/domain/ErrorResponseBody.java
```diff
@@ -0,0 +1,34 @@
...
+        this.message = message;
+        this.code = code.getCode();
+    }
+
+    public String toString() {
+        try {
+            return OBJECT_MAPPER.writeValueAsString(this);
+        } catch (JsonProcessingException e) {
+            LOGGER.error("Error serialising error response", e);
+            return String.format("{\"message\": \"Unexpected error\", \"code\": \"%s\"}", ErrorCode.UNEXPECTED_ERROR.getCode());
```
thebauSoftwire: should the message be more specific here since we know the likely cause? `UNEXPECTED_ERROR` sounds like it has the potential to be used anywhere so it might be helpful in differentiating between different errors

thebauSoftwire (dismissed):
Nice 🎉

lambdas/fetch-journey-transitions/src/main/java/uk/gov/di/ipv/core/fetchjourneytransitions/domain/ErrorResponseBody.java
DanCorderIPV: I don't think this is a useful one to expose to consumers/end users as it most likely means that we have a bug that they can't do anything about.

thebauSoftwire (approved):

PR #3552, opened 2025-11-04
PYIC-8696: fix transition count not working for sub-journey entry states

thebauSoftwire:
### What changed

- fix transition count not working for sub-journey entry states
- tidy up code for calculating transition numbers for edges

### Why did it change

- the transition count was incorrect from some entry events because we weren't taking into account all relevant transition logs.

### Issue tracking

- [PYIC-8696](https://govukverify.atlassian.net/browse/PYIC-8696)

[PYIC-8696]: https://govukverify.atlassian.net/browse/PYIC-8696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


undermad:
Hi Thea, 
It looking much better than before but there are few issues I found. Those changes breaking transition from nested journeys. See the attached screenshots.

Also, I would recommend to extract this logic to separate file same as I did for nested journeys traffic, and also we should write tests to make sure it working as intended. Something similar for main journeys will guarantee expected behaviour.
https://github.com/govuk-one-login/ipv-core-back/blob/main/journey-map/src/helpers/nested-journey-traffic.test.ts


Screenshot 2025-11-04 at 16 58 24
Screenshot 2025-11-04 at 16 58 58

undermad (commented):
I think unit tests are mandatory to achieve required behaviour, it is quite complex to write it so including them could be great idea.

journey-map/src/helpers/sub-journey-traffic.test.ts
```diff
@@ -0,0 +1,673 @@
...
+          from: BASIC_PAGE_STATE,
+          toJourney: FIRST_SUB_JOURNEY,
+          to: BASIC_CRI_STATE,
+        }),
+      ],
+      expectedCount: 2,
+    },
+    {
+      testCase:
+        "the sourceState is an entry state for a first sub-journey and the targetState is a basic state within the same sub-journey",
```
thebauSoftwire, line 57: This test covers the bug identified in the ticket

undermad:
Looking much better, but I found another edge case we should also cover as part of this ticket. There is some transitions from ENTRY node to the nested journey. I created screenshot indicating this scenario.
Screenshot 2025-11-05 at 14 53 39

undermad (approved):
Looking great!

PR #3557, opened 2025-11-06
Pyic 8697: Add system settings dropdown selection and emphasise p3 journey in nav.

undermad:
### What changed

- Added dropdown selection for target environment to fetch feature flags and enabled/disabled CRIs.
- Added information about P3 identity to navigation bar (New P2/P3 Identity).
- Changed html to be more semantic for better readability. 

### Why did it change

- To allow fetch system settings from specific environments
- To emphasise that P3 journeys going through same P2 route.

### Issue tracking

- [PYIC-8697](https://govukverify.atlassian.net/browse/PYIC-8697)

## Checklists

[PYIC-8697]: https://govukverify.atlassian.net/browse/PYIC-8697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/journey-map/src/constants.ts
+++ b/journey-map/src/constants.ts
@@ -1,7 +1,7 @@
 export const JOURNEY_TYPES: Record<string, string> = {
   INITIAL_JOURNEY_SELECTION: "Initial journey selection",
-  NEW_P1_IDENTITY: "New P1 identity",
-  NEW_P2_IDENTITY: "New P2/P3 identity",
+  NEW_P1_IDENTITY: "New at least P1 identity",
+  NEW_P2_IDENTITY: "New at least P2 identity",
   REUSE_EXISTING_IDENTITY: "Reuse existing identity",
   UPDATE_NAME: "Update name",
   UPDATE_ADDRESS: "Update address",
--- a/journey-map/src/constants.ts
+++ b/journey-map/src/constants.ts
@@ -1,7 +1,7 @@
 export const JOURNEY_TYPES: Record<string, string> = {
   INITIAL_JOURNEY_SELECTION: "Initial journey selection",
   NEW_P1_IDENTITY: "New P1 identity",
-  NEW_P2_IDENTITY: "New P2 identity",
+  NEW_P2_IDENTITY: "New P2/P3 identity",
   REUSE_EXISTING_IDENTITY: "Reuse existing identity",
   UPDATE_NAME: "Update name",
   UPDATE_ADDRESS: "Update address",
--- a/journey-map/src/constants.ts
+++ b/journey-map/src/constants.ts
@@ -54,3 +54,11 @@ export const CRI_NAMES: Record<string, string> = {
 export const TOP_DOWN_JOURNEYS = ["INITIAL_JOURNEY_SELECTION"];
 export const ERROR_JOURNEYS = ["TECHNICAL_ERROR"];
 export const FAILURE_JOURNEYS = ["INELIGIBLE", "FAILED"];
+
+export enum Environment {
+  PRODUCTION = "production",
+  INTEGRATION = "integration",
+  STAGING = "staging",
+  BUILD = "build",
+  SHARED_DEV = "shared",
+}
--- a/journey-map/src/constants.ts
+++ b/journey-map/src/constants.ts
@@ -60,5 +60,5 @@ export enum Environment {
   INTEGRATION = "integration",
   STAGING = "staging",
   BUILD = "build",
-  SHARED_DEV = "shared",
+  SHARED = "shared",
 }
--- a/journey-map/src/index.ts
+++ b/journey-map/src/index.ts
@@ -16,7 +16,8 @@ import { JourneyTransition, setJourneyTransitionsData } from "./data/data.js";
 import {
   getJourneyTransitions,
   getSystemSettings,
-  TargetEnvironment,
+  Environment,
+  mapStringToEnvironment,
 } from "./service/analyticsService.js";
 import { parseTransitionsApiForm } from "./helpers/analytics.js";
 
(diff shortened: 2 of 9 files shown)


undermad:
Screenshot 2025-11-06 at 14 28 34

journey-map/src/constants.ts
```diff
@@ -54,3 +54,11 @@ export const CRI_NAMES: Record<string, string> = {
...
 export const ERROR_JOURNEYS = ["TECHNICAL_ERROR"];
 export const FAILURE_JOURNEYS = ["INELIGIBLE", "FAILED"];
+
+export enum Environment {
+  PRODUCTION = "production",
+  INTEGRATION = "integration",
+  STAGING = "staging",
+  BUILD = "build",
+  SHARED_DEV = "shared",
+}
```
thebauSoftwire, line 64: nice!

journey-map/src/service/analyticsService.ts
```diff
@@ -1,22 +1,43 @@
...
+      return Environment.STAGING;
+    case "build":
+      return Environment.BUILD;
+    case "shared":
+      return Environment.SHARED_DEV;
+    default:
+      alert(`Unknown environment: ${environment}. Fetching from production`);
+      return Environment.PRODUCTION;
+  }
+};
```
thebauSoftwire, line 21: Nice and readable :) another way of doing this is using something like: 
```
const env  = Environment[environmentKey as keyof typeof Environment];
```
and checking if `env` is undefined or not. This better aligns with SOLID principles (the open/closed principle) since if we were to ever add a new environment (unlikely tbf), we would only need to update the enum and not have to touch this function. Let me know what you think though!

journey-map/src/index.ts
```diff
@@ -288,7 +332,10 @@ const displayJourneyContextInfo = (ctxOptions: string[]): void => {
 };
 
 const updateView = async (): Promise<void> => {
-  const options = parseOptions(new FormData(form));
+  const options = parseOptions(
+    new FormData(form),
+    new FormData(journeyMapForm),
```
thebauSoftwire: what's the difference between `form` and `journeyMapForm` here?

journey-map/src/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const JOURNEY_TYPES: Record<string, string> = {
   INITIAL_JOURNEY_SELECTION: "Initial journey selection",
   NEW_P1_IDENTITY: "New P1 identity",
-  NEW_P2_IDENTITY: "New P2 identity",
+  NEW_P2_IDENTITY: "New P2/P3 identity",
```
thebauSoftwire: I'm not sure we should change the label for the P2 journey map to `P2/P3`. Whilst it's correct that a user can go through the P2 journey and get a P3 at the end, it's not possible for all the routes to return a P3. I think these journeys represent the journey required by a user to meet the minimum identity within their VTR.

For example, if a user has VTR=["P1", "P2"], they'll still go down the P1 journey but can still get a P2 at the end, provided they have strong enough documents. The only scenario atm where a user can actually get a P3 is if their VTR=["P2","P3"]. We don't currently support VTR=["P3"] since this is likely to need its own P3 journey map. 

Not sure where the best place is to put this but maybe a section in the description section of the dropdown when a journey is selected? Let me know if this doesn't make sense

thebauSoftwire (commented):
UI looks great ✨ 🎉  One suggestion and a comment on the P2/P3 label

journey-map/src/index.ts
undermad: Oh, you right. The "form" is core-back configuration (feature flags and disabled CRIs) and "journeyMapForm" is other options (show failures, expand nested journeys).
I renamed them to be more accurate.

journey-map/src/service/analyticsService.ts
undermad: I tried to do it as you suggested but this code returning undefined as Environment is an Enum and it's doesn't map directly to the string. I really like idea of open close principle therefore I modified this code to respect that rule. Hopefully it will be okay!

journey-map/src/constants.ts
undermad: As far as I know, we only support IBCA for P3, and the journey goes through the "P2" routes. That’s why I believe it’s correct to indicate in the navigation bar that this level of confidence can be achieved on this map, even if not all routes provide that possibility.

Additionally, in the NEW_P2_IDENTITY description, it’s mentioned that it allows reaching at least P2, but that information isn’t very visible. The goal is to make it a bit clearer and provide more hints about it.

journey-map/src/service/analyticsService.ts
thebauSoftwire: Nice :) I think the issue with it always returning `undefined` was because the string was being converted to lower-case first. If we used `const normalized = environment.toUpperCase().trim();` instead it should return the enum since the "key" of the enum is upper-case not lower-case

journey-map/src/constants.ts
thebauSoftwire: Then by that logic, we should also update the P1 journey label to `P1/P2` since a user can still go down the P1 journey and receive a P2 (I think they should also be able to receive a P3, if they use a passport in the doc app and get fraud=2) 🤔 Maybe we could update the labels to `Minimum P1 identity`, `Minimum P2 identity`? And clarify in the description that the P3 identity is supported within the P2 journey (we can make this bold to stand out more). It might also help to move the journey description section to the far left of the screen so it's the first thing a user reads opening the dropdown.

Another option, though more complicated and might not be worth the effort: we could have some sort of dropdown or section, which can filter the P1/P2 subjourneys by level of confidence? So if the LoC was set to P3, the user should be displayed the options to view the P1 or P2 journeys, if it was set to P1, the user should only be displayed the P1 subjourney.

journey-map/src/service/analyticsService.ts
undermad: Oh, my bad. Indeed you are right!

journey-map/src/constants.ts
undermad: You may be right, description even indicate that. I updated the label for P1 journeys as well and moved node description to the left. Updated screenshot below:
Screenshot 2025-11-07 at 14 48 08

thebauSoftwire (approved):

PR #3579, opened 2025-11-19
BAU: ignore test only changes when running deploy job

thebauSoftwire:
### What changed

- prevent deploying to aws if the PR only has a test change

### Why did it change

- Putting up a PR which changed just the contract tests triggered deployment to AWS even though it wasn't really changing any application code. Ignoring test only changes means we don't have to run the deployment

## Checklists

### DIFF
--- a/.github/workflows/secure-post-merge.yml
+++ b/.github/workflows/secure-post-merge.yml
@@ -6,7 +6,7 @@ on:
       - main
     paths-ignore:
       - '**/README.md'
-      - '**/test/**'
+      - '/lambdas/*/src/test/**'
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


hugh-emerson (dismissed):
Seems sensible. Maybe you could be more precise than `**/test/**`? Say `/lambdas/*/src/test/**`? But practically speaking, your version is fine.

While you're at it, how about adding `journey-map/**` to this ignore list too? It has its own workflow, right?

thebauSoftwire:
good point, thanks @hugh-emerson!

hugh-emerson (approved):
LGTM 👍

PR #3584, opened 2025-11-21
PYIC-8798: Return 400 from check-mobile-app-vc-receipt if IPV Session is expired.

undermad:
### What changed

- Added method to evaluate if IPV Session is expired 
- Created exception and error response message 
- Added check for expired IPV Session in check-mobile-app-vc-receipt and return 400 if it is.

### Why did it change

- Trust & Reuse have reported seeing a daily number of EVCS ‘get VCs’ requests using an expired access token which they reject with a 401 response (apparently it also triggers some alarms on their side). We believe the vast majority if not all of these requests are coming from the check-mobile-app-vc-receipt lambda which the frontend v2 app polling calls to check if an app VC has been received yet or not. It seems, perhaps from users abandoning the journey at the app polling pages, that calls are being made beyond the 1 hour expiry of the EVCS access token (which Orch sends us at the start of a journey). 

Currently the only place we explicitly check our (IPV) session expiry (also 1 hour) is in the journey engine when transitioning between states. The actual session item in DynamoDB has a longer TTL of 24 hours.

We should add a check in the check-mobile-app-vc-receipt lambda to reject requests (with a 400 error) where the user's IPV session has expired. In turn this should drastically reduce, hopefully diminish calls to EVCS being made with an expired access token.

### Issue tracking

- [PYIC-8798](https://govukverify.atlassian.net/browse/PYIC-8798)

## Checklists

[PYIC-8798]: https://govukverify.atlassian.net/browse/PYIC-8798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
@@ -199,6 +199,13 @@ public class IpvSessionService {
         dataStore.update(ipvSessionItem);
     }
 
+    public boolean checkIfSessionExpired(IpvSessionItem ipvSessionItem) {
+        var creationDateTime = Instant.parse(ipvSessionItem.getCreationDateTime());
+        var expirationDateTime =
+                creationDateTime.plusSeconds(configService.getBackendSessionTimeout());
+        return Instant.now().isAfter(expirationDateTime);
+    }
+
     private String toExpiryDateTime(long expirySeconds) {
         return Instant.now().plusSeconds(expirySeconds).toString();
     }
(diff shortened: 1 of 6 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
```diff
@@ -199,6 +199,13 @@ public void invalidateSession(IpvSessionItem ipvSessionItem, String errorDescrip
         dataStore.update(ipvSessionItem);
     }
 
+    public boolean checkIfSessionExpired(IpvSessionItem ipvSessionItem) {
+        var creationDateTime = Instant.parse(ipvSessionItem.getCreationDateTime());
+        var expirationDateTime =
+                creationDateTime.plusSeconds(configService.getBackendSessionTimeout());
+        return Instant.now().isAfter(expirationDateTime);
+    }
+
```
thebauSoftwire, line 208: nice! Should we also reuse this in `process-journey-event`? [here](https://github.com/govuk-one-login/ipv-core-back/blob/67eada2780aee73a72131be7c85d650c6ffe3fdc/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java#L418-L424)?

thebauSoftwire (approved):
LGTM! Just one suggestion

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/service/IpvSessionService.java
undermad: Hi Thea, 
We probably should refactor it there. I'll raise separate PR for this task as part of this ticket.

PR #3598, opened 2025-11-26
PYIC-8776: update process-journey-event to handle direct routing to c…

thebauSoftwire:
…ross-browser-problem page when provided with cross-browser-problem event

❗ Note: this needs to be merged in _before_ the change to update `process-mobile-app-callback` to return the `problem-different-browser` event (done as part of this PR). This is so that, during canary deployment, when a "new" version of `process-mobile-app-callback` returns the new cross-browser event and core-front uses an "old" version of the `process-journey-event` lambda, the new cross-browser event is already supported and so won't break a user's journey.

### What changed

- updated `process-journey-event` lambda to route directly to `problem-different-browser` page when provided with new `problem-different-browser` event

### Why did it change

- A user doesn't have a valid ipv session (missing ipv session header) in a cross-browser scenario so they can't go through the rest of the lambda as it requires their session details

### Issue tracking

- [PYIC-8776](https://govukverify.atlassian.net/browse/PYIC-8776)

## Checklists

[PYIC-8776]: https://govukverify.atlassian.net/browse/PYIC-8776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blakeyp (dismissed):

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -76,6 +77,8 @@ public class ProcessJourneyEventHandler
     private static final String NEXT_EVENT = "next";
     private static final StepResponse BUILD_CLIENT_OAUTH_RESPONSE =
             new ProcessStepResponse(BUILD_CLIENT_OAUTH_RESPONSE_EVENT, null);
+    private static final StepResponse CROSS_BROWSER_PROBLEM_PAGE_RESPONSE =
+            new PageStepResponse(CROSS_BROWSER_PROBLEM_PAGE_EVENT, null, null);
```
hugh-emerson: You're using the `CROSS_BROWSER_PROBLEM_PAGE_EVENT` string for both the journey event and also the page id.

Not a big deal, but it's unnecessary for them to be equal and results in slightly odd looking code
- here, you create a page response containing an event id
- in the test you assert that the page id is equal to an event id

Would it make more sense to define the two strings separately (even if they happen to have the same value)?

hugh-emerson (dismissed):
Just a small suggestion for cleaner code

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/journeys/Pages.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.di.ipv.core.library.journeys;
+
+import lombok.Getter;
+import uk.gov.di.ipv.core.library.annotations.ExcludeFromGeneratedCoverageReport;
+
+@ExcludeFromGeneratedCoverageReport
+public enum Pages {
```
DanCorderIPV: What is this for? Do we need to decouple events from page names?

Maybe would be better to make this a separate PR and include all the pages

thebauSoftwire: yeah this came from Hugh's comment [above](https://github.com/govuk-one-login/ipv-core-back/pull/3598#discussion_r2577508917) to define the event and pageId separately. We already have enums for some [Events](https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/journeys/Events.java) so I added the Pages enum to match

DanCorderIPV: Ah, I was just looking at the latest changes. With this kind of thing I'd prefer to do it all together otherwise I suspect the next time we add a page it will get done the old way and not be added to your new enum.
I know Hugh said that the strings don't have to be the same, but is it helpful that they are? e.g. someone looking up what the event is related to.
Maybe it would be better to keep them the same constant but rename the constant?

thebauSoftwire: I've left the journey event enum in for now to match how the `build-client-oauth-response` is being treated. I'll leave it for a future tech debt ticket to clean up/refactor how those enums should be used

DanCorderIPV (approved):

PR #3604, opened 2025-12-01
PYIC-8824: update generateUserIdentity to only include successful and…

thebauSoftwire:
… M1C fraud VCs in the user identity vcs claim

### What changed

- generateUserIdentiy filters the vcs (used for the vc claim) to only use successful (non-zero scoring) vcs and the M1C fraud vcs

### Why did it change

- SPOT is rejecting users with no birthdate claim in their VCs. These VCs they get from the user identity bundle that core sends to orch. This appears on zero-scoring VCs e.g. failing the app and switching to the web route.

### Issue tracking

- [PYIC-8824](https://govukverify.atlassian.net/browse/PYIC-8824)

## Checklists

[PYIC-8824]: https://govukverify.atlassian.net/browse/PYIC-8824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CoreFeatureFlag.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/CoreFeatureFlag.java
@@ -7,7 +7,8 @@ public enum CoreFeatureFlag implements FeatureFlag {
     SQS_ASYNC("sqsAsync"),
     DL_AUTH_SOURCE_CHECK("drivingLicenceAuthCheck"),
     STORED_IDENTITY_SERVICE("storedIdentityServiceEnabled"),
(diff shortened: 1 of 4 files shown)


blakeyp (commented):
Looks good, we do want to put this change behind a feature flag so we can test in staging first:
> We should put this change behind a feature flag so that we can test it in staging fully e2e via the RP stub (which exercises real SPOT) before we go-live in prod.

thebauSoftwire:
ah I missed that 🤦‍♀️ thanks! config PR here to add the feature flag: https://github.com/govuk-one-login/ipv-core-common-infra/pull/1436

blakeyp (approved):

PR #3616, opened 2025-12-05
PYIC-8797: record high resolution identityReuse and identityProving metrics

thebauSoftwire:
### What changed

- record the existing `identityReuse` at high resolution*. This is recorded only for reuse journeys.
- add new `identityProving` high resolution metric. This is recorded for the following journeys:
  - new identity proving
  - separate session mitigations
  - repeat fraud check
  - update journeys
  - reprove identity
  - mfa reset (if they have an identity, as they require an identity to go through the identity proving part of mfa reset)

Notes:
*high resolution metrics are data with granularity of one second
- we don't record identityProving for DCMAW async recovery scenarios (ie an app user experiences cross-broswer issue) since then we'd be counting twice for a single proving journey

### Why did it change

- We want to report more granularly on journeys/second and be able to break this down by reuse and proving.

### Issue tracking

- [PYIC-8797](https://govukverify.atlassian.net/browse/PYIC-8797)

## Checklists

[PYIC-8797]: https://govukverify.atlassian.net/browse/PYIC-8797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
@@ -26,6 +26,7 @@ import uk.gov.di.ipv.core.library.domain.JourneyRequest;
 import uk.gov.di.ipv.core.library.domain.JourneyState;
 import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.IpvSessionNotFoundException;
+import uk.gov.di.ipv.core.library.helpers.EmbeddedMetricHelper;
 import uk.gov.di.ipv.core.library.helpers.LogHelper;
 import uk.gov.di.ipv.core.library.helpers.RequestHelper;
 import uk.gov.di.ipv.core.library.helpers.StepFunctionHelpers;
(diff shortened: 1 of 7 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/ProcessJourneyEventHandler.java
```diff
@@ -238,6 +245,17 @@ private StepResponse executeJourneyEvent(
...
                                         journeyChangeState.getInitialState()));
 
+                // We don't want to record another identityProving metric for a user
+                // starting an update journey if they require a repeat fraud check (RFC)
+                // as this would only count as a single identity proving journey (we
+                // already record this metric at the start of the RFC journey).
+                if (UPDATE_JOURNEY_TYPES.contains(journeyChangeState.getJourneyType())
+                        && !ipvSessionItem
+                                .getActiveJourneyContexts()
+                                .contains(REPEAT_FRAUD_CHECK_JOURNEY_CONTEXT)) {
```
thebauSoftwire, line 255: I'm not the biggest fan of this because there's nothing that ties `REPEAT_FRAUD_CHECK_JOURNEY_CONTEXT` to the actual context used so if that context changes, the metrics would be affected.

I think a more flexible way of recording metrics would be to add a new feature to the journey map to record a metric when a user reaches a given state, similar to how we record audit events from the journey map. So, this would look like:
```
  NAMES_ONLY_AFTER_REUSE_START:
    events:
      next:
        targetState: NAMES_ONLY_AFTER_REUSE
        embeddedMetrics:
          - IDENTITY_PROVING
```
But this is probably out of scope and to avoid bloating the ticket, I've gone for this simpler solution.

blakeyp (approved):

PR #2736, opened 2024-11-25
PYIC-7802 Tidy up dobExperianCheckEnabled feature flag

blakeyp:
### What changed

Remove dobExperianCheckEnabled feature flag 

### Why did it change

Now that we're live this should be enabled by default across all envs, and we no longer need it flagged, so can clean up the flag entirely

### Issue tracking
- [PYIC-7802](https://govukverify.atlassian.net/browse/PYIC-7802)

[PYIC-7802]: https://govukverify.atlassian.net/browse/PYIC-7802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3638, opened 2025-12-16
PYIC-8791: greater logging granularity on correlation failure types

thebauSoftwire:
### What changed

- We now have logs for when a user fails:
  - name correlation only
  - dob correlation only 
  - name AND dob correlation
- Updated unit test as it was checking testing for more than one thing (name failure and DOB failure) and was only passing because it had a name failure but in reality, we expected it to throw as the VC had a missing DOB despite being an identity credential
- added a reset function to the LoggerCollector to clear all log messages. This is so that a shared log collector can be re-used between tests.

### Why did it change

- KIWI are investigating a new feature and wanted to observe the number of users facing different types of correlation failures in order to inform their decision-making. Previously, the correlation checks were performed one after the other so only the first failure type was logged. They want visibility on the number of users who fail both name and DOB correlation so we need to check for both name and dob correlation and log the failure type appropriately.

### Issue tracking

- [PYIC-8791](https://govukverify.atlassian.net/browse/PYIC-8791)

## Checklists

[PYIC-8791]: https://govukverify.atlassian.net/browse/PYIC-8791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
+++ b/libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
@@ -169,16 +169,20 @@ public class UserIdentityService {
             throws HttpResponseExceptionWithErrorBody {
         var successfulVcs = getSuccessfulVcs(vcs);
 
-        if (!checkNameAndFamilyNameCorrelationInCredentials(successfulVcs)) {
-            LOGGER.info(LogHelper.buildErrorMessage(ErrorResponse.FAILED_NAME_CORRELATION));
-            return false;
+        var successfulNameCorrelation =
+                checkNameAndFamilyNameCorrelationInCredentials(successfulVcs);
+        var successfulDobCorrelation = checkBirthDateCorrelationInCredentials(successfulVcs);
+
+        if (!successfulDobCorrelation && !successfulNameCorrelation) {
+            LOGGER.error(
+                    LogHelper.buildErrorMessage(ErrorResponse.FAILED_NAME_AND_DOB_CORRELATION));
+        } else if (!successfulDobCorrelation) {
+            LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_BIRTHDATE_CORRELATION));
+        } else if (!successfulNameCorrelation) {
+            LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_NAME_CORRELATION));
         }
 
-        if (!checkBirthDateCorrelationInCredentials(successfulVcs)) {
-            LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_BIRTHDATE_CORRELATION));
-            return false;
-        }
-        return true;
+        return successfulDobCorrelation && successfulNameCorrelation;
     }
 
     public boolean areNamesAndDobCorrelated(List<VerifiableCredential> vcs)
--- a/libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
+++ b/libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
@@ -29,6 +29,7 @@ import uk.gov.di.ipv.core.library.exceptions.HttpResponseExceptionWithErrorBody;
 import uk.gov.di.ipv.core.library.exceptions.UnrecognisedCiException;
 import uk.gov.di.ipv.core.library.helpers.vocab.BirthDateGenerator;
 import uk.gov.di.ipv.core.library.service.ConfigService;
+import uk.gov.di.ipv.core.library.testhelpers.unit.LogCollector;
 import uk.gov.di.model.AddressAssertion;
 import uk.gov.di.model.CheckDetails;
 import uk.gov.di.model.ContraIndicator;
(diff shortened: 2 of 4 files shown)


libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
```diff
@@ -294,13 +294,34 @@ void areVCsCorrelatedReturnFalseWhenNameDifferentForBavCRI() throws Exception {
                                 USER_ID_1,
                                 BAV,
                                 createCredentialWithNameAndBirthDate(
-                                        "Jimmy", "Jones",
-                                        ""))); // BAV cri doesn't provide birthdate
+                                        "Jimmy", "Jones", "1000-01-01")));
```
thebauSoftwire: This test was passing before since it was checking name correlation first and returned false when it had failed. Is there a reason we want to prioritise name correlation failures over a failed dob check? I'd expect this to throw since the VC would be invalid since it's an identity check credential with a missing DOB

DanCorderIPV: Given we have a comment here saying that BAV doesn't provide DoBs should this be throwing an exception? Is the comment wrong, or are we OK with a missing DoB, or is it that the DoB field shouldn't exist at all in this case?

libs/user-identity-service/src/main/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityService.java
```diff
@@ -169,16 +169,20 @@ public boolean areVcsCorrelated(List<VerifiableCredential> vcs)
...
+        var successfulDobCorrelation = checkBirthDateCorrelationInCredentials(successfulVcs);
+
+        if (!successfulDobCorrelation && !successfulNameCorrelation) {
+            LOGGER.error(
+                    LogHelper.buildErrorMessage(ErrorResponse.FAILED_NAME_AND_DOB_CORRELATION));
+        } else if (!successfulDobCorrelation) {
             LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_BIRTHDATE_CORRELATION));
-            return false;
+        } else if (!successfulNameCorrelation) {
+            LOGGER.error(LogHelper.buildErrorMessage(ErrorResponse.FAILED_NAME_CORRELATION));
```
DanCorderIPV, line 182: Given that someone cared enough about to this to raise the ticket we should probably add tests that check the error logging (or add to existing tests)

DanCorderIPV (changes requested):

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
thebauSoftwire: yeah I did wonder about that comment. I found the original PR which [ignores BAV for DOB correlation](https://github.com/govuk-one-login/ipv-core-back/pull/2736/changes#diff-03e309f6bbe7b1e939e8d7a3f5ad31be61b375cf9c88a8a05048bed7970df470) but [this PR](https://github.com/govuk-one-login/ipv-core-back/pull/2736/changes) much later on that removes it from the exclusion list so I assumed it was an old test 🤔 @blakeyp it looks like you put up the latter. Do you know if we should be excluding the BAV VCs for DOB correlation?

blakeyp: Ah so the original HMRC-backed BAV CRI didn’t give us DOB, but the Experian-backed version does (which is what actually went live). So the comment/test case is out of date, we should be checking it

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
```diff
@@ -558,7 +598,13 @@ void areVCsCorrelatedShouldReturnFalseIfBavHasDifferentBirthDate() throws Except
                                         "Jimbo", "Jones", "2000-01-01")));
 
         // Act & Assert
+        var logCollector = LogCollector.getLogCollectorFor(UserIdentityService.class);
         assertFalse(userIdentityService.areVcsCorrelated(vcs));
+
+        var logMessages = logCollector.getLogMessages();
+        //        assertEquals(1, logMessages.size());
```
DanCorderIPV: uncomment?

libs/user-identity-service/src/test/java/uk/gov/di/ipv/core/library/useridentity/service/UserIdentityServiceTest.java
```diff
@@ -294,11 +302,64 @@ void areVCsCorrelatedReturnFalseWhenNameDifferentForBavCRI() throws Exception {
...
+                                USER_ID_1,
+                                PASSPORT,
+                                createCredentialWithNameAndBirthDate(
+                                        "Jimbo", "Jones", "1000-01-01")),
+                        generateVerifiableCredential(
+                                USER_ID_1,
+                                BAV,
+                                createCredentialWithNameAndBirthDate(
+                                        "Jimbo", "Jones",
+                                        ""))); // BAV cri doesn't provide birthdate
```
DanCorderIPV: I think we need to change/drop this comment

DanCorderIPV (changes requested):

DanCorderIPV (approved):
Thanks for tidying up the tests too

PR #3669, opened 2026-01-14
PYIC-8830: Add logic to determine if DCMAW VC has expired

AmritSidhu:
### What changed

- Add logic for expired DCMAW Driving Licence VCs
- If the DCMAW DL is expired and it's past the grace period, then we route them to [reprove-identity-start](https://identity.build.account.gov.uk/dev/template/reprove-identity-start/en) to reprove their identity

Note: this does not require a feature flag as Mobile isn't yet allowing people to prove their identity with an expired DL so this code shouldn't be touched

### Why did it change

We are introducing a validity period for driving licence VCs from the app. This is because a significant portion of DL scans fail because the DL has recently expired. This is a big issue for users who are attempting to renew their licence and blocks DVLA onboarding. Allowing time-limited expired-licence usage will help with improving these failure rates. 

Mobile are doing work to allow identity verification with DLs 90 days past their expiry (grace period) so we need to be able to support reuse scenarios where the user comes back with the expired driving licence; allowing them to reuse their identity if the current date is within a configured validity period from the VC's issue date.

A user's expired DCMAW DL VC is still valid if:
- the user's DL expired before the VC issue date
- there is a configurable validity period (in days)
- the time of reuse is within the VC's nbf + configurable validity period (days)
In which case, they can reuse their identity.

If not, they have to reprove their identity again.

### Issue tracking

- [PYIC-8830](https://govukverify.atlassian.net/browse/PYIC-8830)

## Checklists

[PYIC-8830]: https://govukverify.atlassian.net/browse/PYIC-8830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/account-intervention/journey-ending-interventions.feature
+++ b/api-tests/features/account-intervention/journey-ending-interventions.feature
@@ -109,10 +109,10 @@ Feature: Journey ending interventions
 
   Scenario: Blocked intervention at end of update identity journey
     Given the subject already has the following credentials
-      | CRI     | scenario                     |
-      | dcmaw   | kenneth-driving-permit-valid |
-      | address | kenneth-current              |
-      | fraud   | kenneth-score-2              |
+      | CRI     | scenario               |
+      | dcmaw   | kenneth-passport-valid |
+      | address | kenneth-current        |
+      | fraud   | kenneth-score-2        |
     And The AIS stub will return an 'AIS_NO_INTERVENTION' result
     When I start a new 'medium-confidence' journey
     Then I get a 'page-ipv-reuse' page response
@@ -127,7 +127,7 @@ Feature: Journey ending interventions
     When The AIS stub will return an 'AIS_ACCOUNT_BLOCKED' result
     When I submit 'kenneth-score-2' details with attributes to the CRI stub
       | Attribute          | Values                   |
-      | evidence_requested | {"identityFraudScore":2} |
+      | evidence_requested | {"identityFraudScore":1} |
     Then I get an OAuth response with error code 'session_invalidated'
 
   Scenario: Blocked intervention at end of initial F2F journey
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -63,10 +63,7 @@ import uk.gov.di.ipv.core.library.verifiablecredential.service.SessionCredential
 import uk.gov.di.model.ContraIndicator;
 import uk.gov.di.model.IdentityCheckCredential;
 
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.ZoneOffset;
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -407,10 +404,20 @@ public class CheckExistingIdentityHandler
                                                                         .getDrivingPermit()))
                                 .findFirst();
 
-                if (dcmawDlVc.isPresent()) {
-                    expiredDcmawDrivingPermit = checkForExpiredDcmawDrivingPermit(dcmawDlVc.get());
+                var dcmawExpiredDrivingPermitValidityPeriod =
+                        configService.getDcmawExpiredDlValidityPeriodDays();
+                if (dcmawDlVc.isPresent() && dcmawExpiredDrivingPermitValidityPeriod != null) {
+
+                    expiredDcmawDrivingPermit =
+                            VcHelper.hasExpiredDrivingPermitVc(
+                                    dcmawDlVc.get(),
+                                    dcmawExpiredDrivingPermitValidityPeriod,
+                                    Clock.systemUTC());
 
                     if (expiredDcmawDrivingPermit) {
+                        LOGGER.info(
+                                LogHelper.buildLogMessage(
+                                        "DCMAW Driving Permit VC is expired and past validity period."));
                         var vcsForUpdate = new ArrayList<>(List.of(dcmawDlVc.get()));
                         credentialBundle.credentials.stream()
                                 .filter(vc -> DRIVING_LICENCE.equals(vc.getCri()))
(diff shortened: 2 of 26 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -454,6 +498,41 @@ private JourneyResponse getJourneyResponse(
...
+                if (today.isAfter(cutoffDate)) {
+                    LOGGER.info(
+                            "User's driving licence is expired and outside of the grace period.");
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
```
blakeyp: I might be annoying and suggest instead of calling it "grace period" we call it "validity period"/"expiry validity period". Only because per the initiative terminology "grace period" refers to the period after DL expiry that a DL can be used for identity proving, not how long an expired DL DCMAW VC is valid for (see https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/5815795744/Expired+Driving+Licence+Requirements+Document)

I think we did spot this in refinement at the time but it looks like our ticket wasn't properly updated (cause it also refers to "90 days" which is the grace period rather than the 180 days for this validity period) 😞

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -454,6 +498,41 @@ private JourneyResponse getJourneyResponse(
...
+                ((IdentityCheckCredential) dcmawDlVc.getCredential())
+                        .getCredentialSubject()
+                        .getDrivingPermit()
+                        .getFirst()
+                        .getExpiryDate();
+
+        // Apply grace period check only when DL expiry is same day or earlier than VC
+        // issue date
+        LocalDateTime dlExpiryDate =
+                LocalDateTime.of(LocalDate.parse(dlExpiryDateString), LocalTime.MIDNIGHT);
```
blakeyp: Is it the VC issue date we should pin to midnight rather than the DL expiry date? 🤔

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -454,6 +498,41 @@ private JourneyResponse getJourneyResponse(
         }
     }
 
+    private boolean checkForExpiredDcmawDrivingPermit(VerifiableCredential dcmawDlVc) {
```
DanCorderIPV: Maybe move some of this logic into VcHelper so it can be more easily shared with the fraud expiry check

DanCorderIPV: Then we can also have more fine grained tests for e.g. what happens when crossing BST/GMT boundaries

thebauSoftwire: that's a good point! Thanks Dan, I've used your helper function and created one specific to driving permits

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -185,6 +189,35 @@ public static boolean isExpiredFraudVc(VerifiableCredential vc, ConfigService co
...
+    private static boolean hasExpired(
+            Instant vcIssueTime, int validityDurationInDays, Clock clock) {
+        var startOfIssueDay =
+                vcIssueTime.atZone(LONDON_TIMEZONE).toLocalDate().atStartOfDay(LONDON_TIMEZONE);
+
+        var endOfValidity = startOfIssueDay.plusDays(validityDurationInDays);
+        var now = ZonedDateTime.now(clock);
+
+        return endOfValidity.isBefore(now) || endOfValidity.isEqual(now);
+    }
```
thebauSoftwire, line 224: This has been copied over from @DanCorderIPV's PR [here](https://github.com/govuk-one-login/ipv-core-back/pull/3671/changes#diff-e69e112c340cbfd2a47535711f40476a7dafda5c8d40a640b60ff5f69939b004). Once his PR is merged in, I'll resolve any conflicts

blakeyp (commented):
@thebauSoftwire @AmritSidhu looks good, Dan's PR has been merged now so we can fix up conflicts. Also would you mind also updating the PR description to correct the "grace period"/"90 days" content?

api-tests/features/account-intervention/journey-ending-interventions.feature
```diff
@@ -109,10 +109,10 @@ Feature: Journey ending interventions
...
   Scenario: Blocked intervention at end of update identity journey
     Given the subject already has the following credentials
-      | CRI     | scenario                     |
-      | dcmaw   | kenneth-driving-permit-valid |
-      | address | kenneth-current              |
-      | fraud   | kenneth-score-2              |
+      | CRI     | scenario               |
+      | dcmaw   | kenneth-passport-valid |
+      | address | kenneth-current        |
+      | fraud   | kenneth-score-2        |
```
thebauSoftwire, line 115: making this passport here so we don't need to rely on the dcmaw VC being a driving permit (same reasoning for below where I've updated k`enneth-driving-permit-valid` -> `kenneth-passport-valid`

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -431,7 +482,7 @@ private JourneyResponse getJourneyResponse(
             }
 
             // No relevant async CRI
-            return getNewIdentityJourney(targetVot);
+            return getNewIdentityJourney(targetVot, expiredDcmawDrivingPermit);
```
blakeyp: as it's a bool maybe `hasExpiredDcmawDrivingPermit` ?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -199,6 +199,24 @@ private static boolean hasExpired(
         return endOfValidity.isBefore(now) || endOfValidity.isEqual(now);
     }
 
+    public static boolean hasExpiredDrivingPermitVc(
```
blakeyp: To align with the Fraud VC method -> `isExpiredDrivingPermitVc`? and if it's not a big faff could we align on passing in configService vs the config value itself?

Also maybe move this above the private method it uses?

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -188,19 +188,14 @@ public static boolean isExpiredFraudVc(
...
-
-        return endOfValidity.isBefore(now) || endOfValidity.isEqual(now);
-    }
+        if (validityDurationInDays == null) {
+            return false;
+        }
 
-    public static boolean hasExpiredDrivingPermitVc(
-            VerifiableCredential drivingPermitVc, int validityDurationInDays, Clock clock) {
         var vcIssueTime = drivingPermitVc.getClaimsSet().getNotBeforeTime().toInstant();
```
blakeyp: Sorry another thing, I noticed the Fraud VC check has a null check on nbf, I wonder if some very old VCs might not have the `nbf` field 🤔 

Though I think here we should actually allow through an old VC without nbf, since we expect all newly issued DCMAW VCs to have nbf (so any expired DL ones will once it goes live)

blakeyp (approved):

PR #3686, opened 2026-01-23
PYIC-8833: Emit audit events for expired Fraud and DCMAW DL VCs

undermad:
⚠️  DO NOT MERG UNTIL [EVENT CATALOGUE](https://github.com/govuk-one-login/event-catalogue/pull/524) PR IS MERGED ⚠️ 

### What changed

- Added helper logic to extract the nbf claim from Verifiable Credentials as an Instant.
- Implemented logic to emit the IPV_EXPIRED_FRAUD_VC_FOUND audit event.
- Implemented logic to emit the IPV_EXPIRED_DCMAW_DL_VC_FOUND audit event.

### Why did it change

When a user attempts to reuse their existing VC bundle and one or more credentials have expired, the user is required to either:

- repeat address and fraud checks
- reprove their full identity

Currently this outcome is not explicitly captured in telemetry, which makes it more difficult to:

- Understand how often VC expiry prevents reuse
- Identify expiry-driven transitions into repeat journeys
- Analyse the impact of expiry on downstream outcomes and abandonment
- Provide clear evidence for future policy adjustments to expiry lifetimes

### Issue tracking

Ticket is not fully update but please, see the comments.
- [PYIC-8833](https://govukverify.atlassian.net/browse/PYIC-8833)

## Checklists

[PYIC-8833]: https://govukverify.atlassian.net/browse/PYIC-8833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -669,6 +691,29 @@ private JourneyResponse buildReuseResponse(
                     false);
 
             EmbeddedMetricHelper.identityProving();
+
+            credentialBundle.credentials.stream()
```
undermad: On line 683 we check if fraud vcs are expired or from unavailable source. This code will only lunch it if the vcs are present and expired. Please advice if this will be accurate

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -669,6 +691,29 @@ private JourneyResponse buildReuseResponse(
...
+                    .filter(vc -> vc.getCri().equals(Cri.EXPERIAN_FRAUD))
+                    .findFirst()
```
undermad, line 703: Is there a chance that user going to have more than one fraud vcs. For example is HISTORICAL vc going to present here? If that is the case it should be changed to make sure we are using the current "valid" vc.

blakeyp: No, the bundle here won't include `HISTORIC` VCs so this is fine

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
blakeyp: This looks ok yeah, but hmm perhaps we could restructure things in VcHelper so we only have to call `isExpiredFraudVc` once and then fire the audit event? it makes me a little nervous that we call the check again separately for the audit event, especially considering it's time based

Could either split up the `allFraudVcsAreExpiredOrFromUnavailableSource` method, or pass in audit service and fire the events in VcHelper itself?

undermad: Hi Patrick,

I’ve amended the PR to address your concerns. Splitting the method while ensuring that the audit event is sent only for expired fraud VCs was a bit tricky. Passing the audit event logic into VcHelper would introduce side effects and go against a functional approach, so I avoided that.

To address the inconsistency in expired fraud checks, I created an overload of `allFraudVcsAreExpiredOrFromUnavailableSource` that accepts a Clock parameter, and I pass the previously created fixed Clock to ensure all checks are executed against the same timestamp.

Additionally, extracted fraud VCs from the credentials bundle to slightly improve performance.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -679,9 +682,15 @@ private JourneyResponse buildReuseResponse(
             String deviceInformation)
             throws VerifiableCredentialException {
         // check the result of 6MFC and return the appropriate journey
+        var fraudVcs =
+                credentialBundle.credentials.stream()
+                        .filter(vc -> vc.getCri().equals(EXPERIAN_FRAUD))
+                        .toList();
+        var clock = Clock.fixed(Instant.now(), LONDON_TIMEZONE);
```
blakeyp: `clock` -> `fixedClock` to make this a bit more descriptive?

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -152,6 +153,8 @@ public class CheckExistingIdentityHandler
     private static final String ACCOUNT_INTERVENTION_ERROR_DESCRIPTION =
             "Account intervention detected";
 
+    private static final ZoneId LONDON_TIMEZONE = ZoneId.of("Europe/London");
```
blakeyp: I think this is also defined elsewhere, could we pull it from a common single place? (not a huge deal)

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
blakeyp: Neat solution! just added a couple of minor suggestions

blakeyp (approved):

PR #3693, opened 2026-01-29
PYIC-8846: put the journey map behind google sso

thebauSoftwire:
### What changed

- places the journey map behind google sso via oidc authentication at the ALB 
- removes the API Gateway that sits in front of the ALB and makes the ALB public-facing
- removes the existing basic auth set-up

| Test | Screenshot |
|-|-|
| Google SSO - valid user | Screenshot 2026-02-02 at 10 07 32 |
| Google SSO - email outside of organisation | Screenshot 2026-01-30 at 15 02 12 |

### Why did it change

- Currently, the journey map is only protected by basic auth. We want it to be more secure by instead placing it behind Google SSO
- The ALB needs to be public-facing as Google's callback must point to the ALB's public URL (the current setup is an internal-facing ALB which sits behind an API Gateway). Without a public-facing ALB, the oauth flow breaks because Google can't reach the internal ALB. It should also be minimal risk to make the ALB public-facing since we're adding Google SSO.

### Issue tracking

- [PYIC-8846](https://govukverify.atlassian.net/browse/PYIC-8846)

## Checklists

[PYIC-8846]: https://govukverify.atlassian.net/browse/PYIC-8846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


journey-map/README.md
```diff
@@ -34,6 +34,24 @@ Use `npm start` to run the built code.
 
 In production, the journey map uses `../journey-map.Dockerfile` to run these steps.
 
+## Running the map in a dev environment
+
+To test the journey map deployed to AWS, it can be deployed to your dev environment.
+Note that a deployed journey-map requires authentication via Google SSO. To get this working,
+ask a member of the team with edit access to the `IPV Core Journey Map Link` Google Cloud Project to:
+
+- add your journey-map dev URI as a valid redirect URI. Your dev redirect URI will take a form like `https://dev-{username}-journey-map.02.core.dev.stubs.account.gov.uk/oauth2/idresponse`.
```
thebauSoftwire, line 43: I've manually added a journey-map redirect URI for everyones dev journey maps but we'll have to manually add the redirect URI of new devs as unfortunately, we can't use matchers and must be explicit when specifying the redirect URI

journey-map/README.md
```diff
@@ -34,6 +34,24 @@ Use `npm start` to run the built code.
...
+- (optionally) add you as a principal to give you access to the project
```
thebauSoftwire: I've manually added everyone on the team as principals to the project so everyone should have access to it. However, I had to add everyone's emails individually. Is there a group that the team is part of so I could use that instead? cc @blakeyp in case you know

journey-map/deploy/template.yaml
```diff
@@ -229,12 +187,17 @@ Resources:
           - Key: deletion_protection.enabled
             Value: false
           - !Ref AWS::NoValue
+        - Key: "access_logs.s3.enabled"
+          Value: "true"
+        - Key: "access_logs.s3.bucket"
+          Value: "ipv-core-dev02-access-logs"
```
DanCorderIPV: It's not super important, but would it make more sense to use build (or higher). The journey map is an internal tool, but it is "production" in a sense.

DanCorderIPV:
This looks sensible to me, but probably worth getting someone from SRE to look over the `template.yml` file if you haven't already

journey-map/README.md
blakeyp: There is a Google group `Identity-IPV-Core` is that what you need @thebauSoftwire ?

journey-map/README.md
```diff
@@ -34,6 +34,24 @@ Use `npm start` to run the built code.
 
 In production, the journey map uses `../journey-map.Dockerfile` to run these steps.
 
+## Running the map in a dev environment
+
+To test the journey map deployed to AWS, it can be deployed to your dev environment.
+Note that a deployed journey-map requires authentication via Google SSO. To get this working,
+ask a member of the team with edit access to the `IPV Core Journey Map Link` Google Cloud Project to:
```
blakeyp: Can we provide a link to the project?

journey-map/deploy/template.yaml
thebauSoftwire: ah, I forgot to update this 🤦‍♀️ thanks!!

journey-map/README.md
thebauSoftwire: yep that's what I needed, thank you :)

thebauSoftwire:
SREs have given a thumbs up :)

blakeyp (approved):

PR #3713, opened 2026-02-05
PYIC-8910: Add audit event queue environment variable to most lambdas

DanCorderIPV:
Some don't use it yet, but it's easy to overlook when adding the first audit event to a lambda and isn't caught by local testing.

### What changed

Add audit event queue environment variable to most lambdas

### Why did it change

The reset-session-identity lambda now raises an audit event and so fails without the config.
Some lambdas don't use the config yet, but it's easy to overlook when adding the first audit event to a lambda and isn't caught by local testing.

### Issue tracking

- [PYIC-8910](https://govukverify.atlassian.net/browse/PYIC-8910)

[PYIC-8910]: https://govukverify.atlassian.net/browse/PYIC-8910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -1101,6 +1101,7 @@ Resources:
         # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
         Variables:
           POWERTOOLS_SERVICE_NAME: !Sub issue-client-access-token-${Environment}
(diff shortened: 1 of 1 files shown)


thebauSoftwire (approved):
🕵️

PR #3722, opened 2026-02-10
[DCMAW-18492] Update DcmawCri consumer Pact tests

phad:
### What changed

This PR updates the expected VC JSON structures for VCs
asserted in the `pact/dcmawCri/ContractTest.java` Pacts.

- 

### Why did it change

In nearly 100% of test scenarios, the change is purely a JSON
structure update. We need to do this because Pact currently
does not support deep comparison of JWTs.   The DcmawCri
provider Pact tests fail because the VCs returned by DcmawCri
contain payloads serialised in a different order to those
hardcoded in the `ipv-core-back` DcmawCri contract test.

In a few scenarios there are some additional internal attributes
changes.  These arise because the DCMAW test VCs files which
are read by the Provider Pact tests have small differences from
the hard-coded VCs in `pact/dcmawCri/ContractTest.java`.

In one scenario - `DCMAW-410-AC1 with given name removed`
we need to discuss what is currently returned from the DCMAW
CRI, as it differs in one significant way from the asserted VC
in these contract tests.

- 

### Issue tracking

- [DCMAW-18492](https://govukverify.atlassian.net/browse/DCMAW-18492)

## Checklists

[DCMAW-18492]: https://govukverify.atlassian.net/browse/DCMAW-18492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
+++ b/lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
@@ -2767,7 +2767,7 @@ class ContractTest {
                       "documentNumber": "ZR8016200",
                       "expiryDate": "2024-02-25",
                       "icaoIssuerCode": "GBR",
-                      "documentType": "IR"
+                      "documentType": "CR"
                     }
                   ]
                 },
@@ -2798,7 +2798,7 @@ class ContractTest {
     // valid signature (using https://jwt.io works well) and record it here so the PACT file doesn't
     // change each time we run the tests.
     private static final String VALID_BRP_VC_SIGNATURE =
-            "dIXAu4yrIN0YGFisw8nhrJdS4mHrWR_BFAmveloHEwloM5nXQLk9cPfPPbWDtvd_ZwLIexnSrTdXNm1FgB_N5g"; // pragma: allowlist secret
+            "yGgpft2Kl3Xt9O8lC-bHrAgxQCX3ks8bTEQvD2IJWrrbBcrCnSZ4adRq4h1G9EYhaU5LD6VAR4El2b1Fbf3Qww"; // pragma: allowlist secret
 
     // 2099-01-01 00:00:00 is 4070908800 in epoch seconds
     // From DCMAW-5175-AC1
@@ -2815,6 +2815,10 @@ class ContractTest {
                 "@context": [
                   "https://www.w3.org/2018/credentials/v1",
                   "https://vocab.account.gov.uk/contexts/identity-v1.jsonld"
+                ],
+                "type": [
+                  "VerifiableCredential",
+                  "IdentityCheckCredential"
                 ],
                 "credentialSubject": {
                   "name": [
@@ -2836,24 +2840,20 @@ class ContractTest {
                       "value": "1980-01-01"
                     }
                   ],
-                  "residencePermit": [
-                    {
-                      "icaoIssuerCode": "GBR",
-                      "documentNumber": "ZR8016200",
-                      "expiryDate": "2024-02-25",
-                      "documentType": "IR"
-                    }
-                  ],
                   "deviceId": [
                     {
                       "value": "7d8f0d3c-6a0a-4298-afe5-0cf23633618c"
                     }
+                  ],
+                  "residencePermit": [
+                    {
+                      "documentNumber": "ZR8016200",
+                      "expiryDate": "2024-02-25",
+                      "icaoIssuerCode": "GBR",
+                      "documentType": "IR"
+                    }
                   ]
                 },
-                "type": [
-                  "VerifiableCredential",
-                  "IdentityCheckCredential"
-                ],
                 "evidence": [
                   {
                     "type": "IdentityCheck",
@@ -2884,7 +2884,7 @@ class ContractTest {
     // valid signature (using https://jwt.io works well) and record it here so the PACT file doesn't
     // change each time we run the tests.
     private static final String FAILED_BRP_VC_SIGNATURE =
-            "LWfukxi6ZCVz52LIVnNdUFg8Wcv1A5DqcRAQ4R5w3p3U3GNox-Kn6IcGgygt_nJFg4X4lgCqV2q-wSBdWDOGTg"; // pragma: allowlist secret
+            "2rdyetzKV4FPL9p6OZzjewd8nwqfF1MKfBBGmTN1xEcKTPgAR-hWgcwiHDN_KA3ZdMGrZxnYQBck0DImsixSIA"; // pragma: allowlist secret
 
     // 2099-01-01 00:00:00 is 4070908800 in epoch seconds
     // From BRC DCMAW-5176-AC1-valid-doc-chip-clone-detection-successful (there is also a BRP
@@ -2902,6 +2902,10 @@ class ContractTest {
                 "@context": [
                   "https://www.w3.org/2018/credentials/v1",
                   "https://vocab.account.gov.uk/contexts/identity-v1.jsonld"
+                ],
+                "type": [
+                  "VerifiableCredential",
+                  "IdentityCheckCredential"
                 ],
                 "credentialSubject": {
                   "name": [
@@ -2923,24 +2927,20 @@ class ContractTest {
                       "value": "1980-01-01"
                     }
                   ],
-                  "residencePermit": [
-                    {
-                      "icaoIssuerCode": "GBR",
-                      "documentNumber": "ZR8016200",
-                      "expiryDate": "2024-02-25",
-                      "documentType": "CR"
-                    }
-                  ],
                   "deviceId": [
                     {
                       "value": "7d8f0d3c-6a0a-4298-afe5-0cf23633618c"
                     }
+                  ],
+                  "residencePermit": [
+                    {
+                      "documentNumber": "ZR8016200",
+                      "expiryDate": "2024-02-25",
+                      "icaoIssuerCode": "GBR",
+                      "documentType": "CR"
+                    }
                   ]
                 },
-                "type": [
-                  "VerifiableCredential",
-                  "IdentityCheckCredential"
-                ],
                 "evidence": [
                   {
                     "type": "IdentityCheck",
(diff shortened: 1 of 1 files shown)


lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -1958,8 +1958,12 @@ private static OauthCriConfig getMockCredentialIssuerConfig(MockServer mockServe
                     {
                       "nameParts": [
                         {
-                          "value": "Doe The Ball",
-                          "type": "FamilyName"
+                          "type": "GivenName",
+                          "value": ""
```
phad, line 1901: Here the provider behaviour diverges from what was being asserted in the consumer Pact.
I'm not sure what was expected in this scenario in the ReadID responses that DCMAW
consumes in order to create a VC.

On the provider side, I cloned the existing `DCMAW-410-AC1-input.json`, which contains:
```
  "consolidatedIdentityData": {
    "primaryIdentifier": "Doe The Ball",
    "secondaryIdentifier": "Joe Shmoe",
    "selfieVerificationStatus": "SUCCEEDED",
    "visualVerification": "APPROVED_AUTOMATIC"
  }
```
and removed the content `Joe Shmoe` from `secondaryIdentifer`, leaving it empty.
This resulted in one `namePart` of type `GivenName` with the empty value seen here.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -2767,7 +2771,7 @@ private static OauthCriConfig getMockCredentialIssuerConfig(MockServer mockServe
                       "documentNumber": "ZR8016200",
                       "expiryDate": "2024-02-25",
                       "icaoIssuerCode": "GBR",
-                      "documentType": "IR"
+                      "documentType": "CR"
```
phad: I need to investigate the reason for this type change.
The equivalent input file on the DCMAW provider side has `"documentCode": "IR"` so I'm not clear yet why that translated to `CR` in the VC.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: @sandymay-dd  - we need to study this and figure out what the correct behaviour should be.
It could just be an edge case we've not considered before.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: @DanCorderIPV  the DcmawCRI and DcmawAsyncCRI no longer issue VCs for driving licences, if the driving licence issuer is missing.  This is a behaviour change made in Jan 2026:

https://github.com/govuk-one-login/mobile-id-check-biometric-credential/commit/8b3363c

Given this, do you think this Pact should be removed entirely?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -417,13 +417,15 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidDvlaVcWi
 
                                 assertEquals("CH1 1AQ", addressNode.get("postalCode").asText());
 
-                                assertEquals(1, nameParts.size());
-                                assertEquals("FamilyName", nameParts.get(0).get("type").asText());
+                                assertEquals(2, nameParts.size());
+                                assertEquals("GivenName", nameParts.get(0).get("type").asText());
+                                assertEquals("", nameParts.get(0).get("value").asText());
```
phad, line 422: @DanCorderIPV @sandymay-dd  we need to agree on what's expected / acceptable here.
Currently the mobile-id-check-biometric-credential library which formats VCs for dcmaw-cri and dcmaw-async-cri doesn't have specific handling of the vendor response having missing/empty `secondaryIdentifier` which is what is used to form the `GivenName` type of `nameParts`.
The change above reflects what our VCs currently contain in this situation.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -417,13 +417,15 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidDvlaVcWi
...
+                                assertEquals("FamilyName", nameParts.get(1).get("type").asText());
                                 assertEquals(
-                                        "Doe The Ball", nameParts.get(0).get("value").asText());
+                                        "Doe The Ball", nameParts.get(1).get("value").asText());
 
                                 assertEquals(
-                                        "2023-01-18", drivingPermitNode.get("expiryDate").asText());
+                                        "2046-11-07", drivingPermitNode.get("expiryDate").asText());
```
phad, line 428: All the `expiryDate` changes like this are because the input JSON files used in DcmawCri unit and pact tests have this specific date hardcoded.  So, this has diverged from the hardcoded VCs used in the IpvCoreBack consumer pacts.  It'd be good to just accept this, I don't think it reflects any change in the shape of the data so the pact isn't broken.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
```diff
@@ -1340,7 +1342,7 @@ void fetchVerifiableCredential_whenCalledAgainstDcmawCri_retrievesAValidBrpVc(
                                 assertEquals("2024-02-25", brp.get("expiryDate").asText());
                                 assertEquals("ZR8016200", brp.get("documentNumber").asText());
                                 assertEquals("GBR", brp.get("icaoIssuerCode").asText());
-                                assertEquals("IR", brp.get("documentType").asText());
+                                assertEquals("CR", brp.get("documentType").asText());
```
phad: I need to investigate this.  I don't think it's caused by the DcmawCri's own test data files containing CR instead of IR.  @sandymay-dd  does this ring any bell for you?

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
DanCorderIPV: That sounds sensible to me

phad: Thanks - I've removed it now.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: Done some digging ... lack of `secondaryIdentifier` is covered by a different DCMAW scenario: https://govukverify.atlassian.net/browse/DCMAW-4543

@sandymay-dd in our biometric-credential repo we have tests for this ticket but only for the error cases (missing `primaryIdentifier`) - we don't test the happy path (`primaryIdentifier` present, but `secondaryIdentifier` missing).

This gap in our test coverage probably means that we never defined correct behaviour for the GivenName fields in resulting VCs.

I think we should address that, and then have the IPVCoreBack consumer Pact reference DCMAW-4543-AC1 instead of using DCMAW-410-given-names-removed which doesn't correspond to any test case we have on our side.

phad: Ah, it looks like a later issue did address this: https://govukverify.atlassian.net/browse/DCMAW-13259
In which missing `secondaryIdentifier` (which once caused an error) leads to the current behaviour of a VC with `givenName: null`.  We need to consider that case too.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: This seems likely to be due to a mistake in our provider-side Pact test.  I'll investigate.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
sandymay-dd: Hi @DanCorderIPV - While DCMAW no longer issues a VC in this scenario, we _do_ return an error. Rather than remove this pact I think we should consider updating the expected result - what do you think?

The expected result would be to return a `404` response (option 3 in this design): https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3618078730/Userinfo+errors#Option-Proposed.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: It was a copy-paste error by me, I pasted the updated BRC VC over the BRP one :(  now fixed.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: I've restored the test, but converted it to one that checks for a 404 response from the CRI.

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: Resolving this conversation as this issue is also discussed further down...

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: Resolving this, it's discussed further up (caused by copy&paste error).

DanCorderIPV (approved):

lambdas/process-cri-callback/src/test/java/uk/gov/di/ipv/core/processcricallback/pact/dcmawCri/ContractTest.java
phad: We agreed to keep the empty `GivenName` field to start with, and to meet with BAs to agree if this is fine, or if something different (e.g. no `GivenName` in this circumstance) should be required - if so, we'll come back and amend it.

PR #3733, opened 2026-02-17
BAU: Fix expanding doubly-nested journeys with event overrides

Joe-Edwards-GDS:
Currently the 'expand nested journeys' option in the journey map doesn't work when there are doubly-nested journeys with `targetEntryEvent` overrides.

### DIFF
--- a/journey-map/src/helpers/expand-nested.test.ts
+++ b/journey-map/src/helpers/expand-nested.test.ts
@@ -300,6 +300,88 @@ describe("expandNested", () => {
     expect(states).toEqual(expected);
   });
 
+  it("should expand doubly-nested journeys with event overrides", () => {
+    // Arrange
+    const states: Record<string, JourneyState> = yaml.parse(`
+      ENTRY_STATE:
+        events:
+          next:
+            targetState: NESTED_STATE
+            targetEntryEvent: entryEventOverride
+
+      NESTED_STATE:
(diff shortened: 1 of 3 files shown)


journey-map/src/helpers/expand-nested.ts
```diff
@@ -81,8 +81,8 @@ export const expandNestedJourneys = (
                   (!t.targetEntryEvent && implicitEntryEvent === entryEvent),
               )
               .forEach((t) => {
+                delete t.targetEntryEvent; // Clear any resolved target entry event
                 Object.assign(t, entryEventDef);
-                delete t.targetEntryEvent;
```
Joe-Edwards-GDS, line 85: Previously if `entryEventDef` also contains a `targetEntryEvent` it got wiped (we only want to wipe the original one, which has already been resolved)

journey-map/src/helpers/expand-nested.test.ts
```diff
@@ -300,6 +300,88 @@ describe("expandNested", () => {
     expect(states).toEqual(expected);
   });
 
+  it("should expand doubly-nested journeys with event overrides", () => {
```
thebauSoftwire, line 303: 🤩

thebauSoftwire (approved):
thanks Joe! 🕵️ 🎉

PR #3754, opened 2026-02-24
PYIC-8933: Optimise api-tests

AmritSidhu:
### What changed

- Reduced polling interval for DCMAW credential receipt from 1s to 200ms with immediate return on success.
- Refactored credential generation to use Promise.all for concurrent stub requests across all table rows.
- Parallelised CIMIT detection calls to remove sequential I/O bottlenecks.
- Significantly reduces idle wait time in integration test suites.

### Why did it change

- Recently the API tests are taking around 15 minutes to complete and have a 15 minute timeout configured which means that sometimes the test hit the timeout and show as “cancelled”.

### Issue tracking

- [PYIC-8933](https://govukverify.atlassian.net/browse/PYIC-8933)

## Checklists

[PYIC-8933]: https://govukverify.atlassian.net/browse/PYIC-8933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/src/steps/cri-steps.ts
+++ b/api-tests/src/steps/cri-steps.ts
@@ -487,14 +487,15 @@ When(
     table: DataTable,
   ): Promise<void> {
     this.userId = this.userId ?? getRandomString(16);
-    const credentials: string[] = [];
-    for (const row of table.hashes()) {
+
+    const allGenerationPromises = table.hashes().flatMap((row) => {
       const numCredentials = row.numCredentials
         ? parseInt(row.numCredentials)
         : 1;
-      if (row.CRI === "dcmawAsync") {
-        const asyncCreds = Array.from(Array(numCredentials)).map(async () => {
-          return await dcmawAsyncStubClient.generateVc(
+
+      return Array.from(Array(numCredentials)).map(async () => {
+        if (row.CRI === "dcmawAsync") {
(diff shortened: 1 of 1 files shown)


api-tests/src/steps/cri-steps.ts
```diff
@@ -625,18 +625,27 @@ When(
...
-      await new Promise((resolve) => setTimeout(resolve, 1000));
+      if (this.strategicAppPollResult?.journey) {
+        return;
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, intervalMs));
+    }
+
+    if (!this.strategicAppPollResult) {
+      console.warn("Polling timed out after 10 seconds");
```
blakeyp: Do we want to throw an error to actually fail the test in this case?

AmritSidhu: In this case, we actually don't want to throw an error inside the polling step.
This is because some of our scenarios are specifically expect a 404 result. If we throw an error in the polling step, the test crashes before it can reach the validation step, causing those tests to fail incorrectly

blakeyp (approved):

PR #3821, opened 2026-03-23
PYIC-8718: add page context object to page step response

thebauSoftwire:
### What changed

- update PageStepResponse to return a `pageContext` object to core-front (currently not being used by core-front)
- updated the journey map yaml files to use the new `pageContext` where we're setting a context for a page
- added some validation for the `pageContext` so we know what values we can use
- updated the api tests to also check for the `pageContext` along with the string `context`

**Note on deployment**: it should be fine to make this change in one go (add the `pageContext` object AND update the journey map yaml files) since we don't really do anything like persisting the page `context` between lambdas (we only store the state name and pull from the yaml files when a new event comes in) and core-front doesn't use the new `pageContext` yet. 

I'll create a separate PR to clean up the old page `context` once core-front is using the new one.

**Note on documentation**: I'll update the journey map docs in the clean-up PR to remove the existing `context` on page states so I can get feedback on this PR first.

### Why did it change

We currently use a string `context` for routing to variants of a page. However, this has limitations because it is only a string. For example, [pyi-triage-desktop-download-app](https://identity.build.account.gov.uk/dev/template/pyi-triage-desktop-download-app/en?context=android) has four different context values: android, iphone, android-appOnly and iphone-appOnly which could make the logic in core-front messy/confusing. Instead, we can turn it into an object:
```
{
  isAppOnly: true/false,
  smartphone: iphone/android
}
```

### Issue tracking

- [PYIC-8718](https://govukverify.atlassian.net/browse/PYIC-8718)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-8718]: https://govukverify.atlassian.net/browse/PYIC-8718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-address.yaml
```diff
@@ -101,6 +101,8 @@ states:
       type: page
       pageId: page-ipv-success
       context: updateIdentity
+      pageContext:
+        journeyType: coi
```
thebauSoftwire, line 105: The `repeatFraudCheck` and `updateIdentity` contexts for this page have the same content so I've combined them into a single context `coi`

lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/validators/PageContextValidator.java
```diff
@@ -0,0 +1,39 @@
...
+                    Map.entry("pyi-triage-mobile-download-app", Set.of("smartphone", "isAppOnly")),
+                    Map.entry("pyi-triage-select-smartphone", Set.of("deviceType")),
+                    Map.entry("sorry-could-not-confirm-details", Set.of("isExistingIdentityValid")),
+                    Map.entry("sorry-technical-problem", Set.of("reason")),
+                    Map.entry(
+                            "uk-driving-licence-details-not-correct", Set.of("isFromStrategicApp")),
+                    Map.entry("update-details-failed", Set.of("isExistingIdentityInvalid")),
+                    Map.entry(
+                            "update-name-date-birth",
+                            Set.of("journeyType", "allowAccountDeletion")));
```
DanCorderIPV, line 33: There may not be a sensible way of doing this, but it would be great if we could link this to the list of pages and contexts in core-front

DanCorderIPV (approved):

PR #3822, opened 2026-03-23
PYIC-8969: remove legacy config key fallback for OAuth signing

AmritSidhu:
### What changed

- Remove fallback to hardcoded getPublicKeyMaterialForCoreToVerify in OAuthKeyService. 
- Enforce JWKS URL as the source of truth for client signing keys. 
- Return existing cached JWKS (even if expired) on fetch failure, otherwise fail outright.
- Update unit tests to assert ConfigException when JWKS keys or URLs are missing. 

### Why did it change

- A bit we seem to have missed in PYIC-7623, cleaning up the hardcoded signing keys for our OAuth clients Orch and Auth (for MFA Reset). They are both hosting JWKS endpoints we are consuming and the keys in staging have been rotated so the fallback config is now out of date anyway.

### Issue tracking

- [PYIC-8969](https://govukverify.atlassian.net/browse/PYIC-8969)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-8969]: https://govukverify.atlassian.net/browse/PYIC-8969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
@@ -11,7 +11,8 @@ import lombok.extern.jackson.Jacksonized;
 public class ClientConfig {
     @NonNull String id;
     @NonNull String issuer;
+    String publicKeyMaterialForCoreToVerify; // to be removed once PYIC-8969 is merged
     @NonNull String validRedirectUrls;
     @NonNull String validScopes;
-    String jwksUrl; // Null for API test client configs
+    String jwksUrl; // to be made @NonNull once PYIC-8969 is merged
 }
--- a/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
+++ b/libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
@@ -11,7 +11,6 @@ import lombok.extern.jackson.Jacksonized;
 public class ClientConfig {
     @NonNull String id;
     @NonNull String issuer;
-    @NonNull String publicKeyMaterialForCoreToVerify;
     @NonNull String validRedirectUrls;
     @NonNull String validScopes;
     String jwksUrl; // Null for API test client configs
--- a/libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
+++ b/libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
@@ -70,7 +70,7 @@ public class OAuthKeyService {
                 .orElseThrow();
     }
 
-    public ECKey getClientSigningKey(String clientId, JWSHeader jwsHeader) throws ParseException {
+    public ECKey getClientSigningKey(String clientId, JWSHeader jwsHeader) {
         var keyId = jwsHeader.getKeyID();
         if (keyId == null) {
             LOGGER.error(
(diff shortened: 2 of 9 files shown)


libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
```diff
@@ -11,7 +11,6 @@
 public class ClientConfig {
     @NonNull String id;
     @NonNull String issuer;
-    @NonNull String publicKeyMaterialForCoreToVerify;
```
blakeyp, line 14: I guess this will fail AppConfig validation until https://github.com/govuk-one-login/ipv-core-common-infra/pull/1504 is merged (which should wait until core-back's changes are through to prod)? For safety in the meantime we could make this nullable for now, and then remove after? We probably also then want to make jwksUrl not nullable?

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -142,6 +132,7 @@ private JWKSet getCachedJWKSet(URI jwksUrl) {
                                             LogHelper.buildErrorMessage(
                                                             "Error creating cached JWK set", e)
                                                     .with(LOG_JWKS_URL.getFieldName(), jwksUrl));
+                                    return existingCachedJWKSet;
```
blakeyp: This looks redundant as there's already a `return existingCachedJWKSet;` below which the catch block will fall to anyway?

libs/common-services/src/main/java/uk/gov/di/ipv/core/library/config/domain/ClientConfig.java
```diff
@@ -11,7 +11,8 @@
 public class ClientConfig {
     @NonNull String id;
     @NonNull String issuer;
+    String publicKeyMaterialForCoreToVerify; // to be removed once PYIC-8969 is merged
     @NonNull String validRedirectUrls;
     @NonNull String validScopes;
-    String jwksUrl; // Null for API test client configs
+    @NonNull String jwksUrl;
```
blakeyp: We probably want to keep this nullable for now and make it not null after we've merged the config PR? else any config deployments will fail in the meantime

libs/oauth-key-service/src/main/java/uk/gov/di/ipv/core/library/oauthkeyservice/OAuthKeyService.java
```diff
@@ -109,10 +109,11 @@ public ECKey getClientSigningKey(String clientId, JWSHeader jwsHeader) {
...
     private URI getClientJwksUrl(String clientId) {
         try {
-            var jwksUrl = configService.getConfiguration().getClientConfig(clientId).getJwksUrl();
-            if (jwksUrl == null) {
+            var clientConfig = configService.getConfiguration().getClientConfig(clientId);
+            if (clientConfig == null) {
                 return null;
             }
+            String jwksUrl = clientConfig.getJwksUrl();
             return URI.create(jwksUrl);
```
blakeyp: It shouldn't be but what if jwksUrl is null?

AmritSidhu: I've updated getClientJwksUrl to be more robust before I mark jwksUrl as @NonNull. Once I add that annotation, a manual if (jwksUrl == null) check becomes redundant (and the IDE will say it's always false) because the Lombok builder will throw an exception the moment it tries to create the object with missing data. So I can expand the catch block to include NullPointerException, that we somehow if it is null we can safely log the error and return null if the config is broken

blakeyp (approved):

PR #3830, opened 2026-03-24
PYIC-8718: refine page context object

thebauSoftwire:
### What changed/Why did it change

- remove unnecessary context for `update-name-date-birth` (`repeatFraudCheck` isn't used so there's really only two valid variants: rfc/reuse)
- add validation to make sure that, if a page defines a `pageContext`, it has a value for all the keys. This is so that we know the exact shape of the pageContext when using it in core-front
- remove accidentally committed empty pages

This is a follow-up PR of: https://github.com/govuk-one-login/ipv-core-back/pull/3821

Context:
We currently use a string `context` for routing to variants of a page. However, this has limitations because it is only a string. For example, [pyi-triage-desktop-download-app](https://identity.build.account.gov.uk/dev/template/pyi-triage-desktop-download-app/en?context=android) has four different context values: android, iphone, android-appOnly and iphone-appOnly which could make the logic in core-front messy/confusing. Instead, we can turn it into an object:
```
{
  isAppOnly: true/false,
  smartphone: iphone/android
}
```

### Issue tracking

- [PYIC-8718](https://govukverify.atlassian.net/browse/PYIC-8718)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-8718]: https://govukverify.atlassian.net/browse/PYIC-8718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/validators/AbstractPageContextValidator.java
+++ b/lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/validators/AbstractPageContextValidator.java
@@ -21,12 +21,21 @@ public abstract class AbstractPageContextValidator implements IPageContextValida
                             .formatted(pageId));
         }
 
-        for (var key : pageContext.keySet()) {
-            if (!allowedContexts.contains(key)) {
-                throw new StepResponseException(
-                        "Context '%s' is not allowed for page '%s'. If required, please add to ALLOWED_CONTEXTS_BY_PAGE."
-                                .formatted(key, pageId));
-            }
+        var contextKeys = pageContext.keySet();
+        var disallowedKeys =
+                contextKeys.stream().filter(key -> !allowedContexts.contains(key)).findFirst();
+        if (disallowedKeys.isPresent()) {
+            throw new StepResponseException(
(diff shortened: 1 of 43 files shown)


lambdas/process-journey-event/src/main/java/uk/gov/di/ipv/core/processjourneyevent/statemachine/validators/AbstractPageContextValidator.java
```diff
@@ -21,12 +21,21 @@ public void validate(String pageId, Map<String, Object> pageContext) {
                             .formatted(pageId));
         }
 
-        for (var key : pageContext.keySet()) {
-            if (!allowedContexts.contains(key)) {
-                throw new StepResponseException(
-                        "Context '%s' is not allowed for page '%s'. If required, please add to ALLOWED_CONTEXTS_BY_PAGE."
-                                .formatted(key, pageId));
-            }
+        var contextKeys = pageContext.keySet();
```
DanCorderIPV, line 24: Are there any unit tests to cover these cases?

DanCorderIPV (changes requested):

DanCorderIPV (dismissed):

DanCorderIPV (approved):

PR #3833, opened 2026-03-24
BAU: log request size for TICF calls

thebauSoftwire:
### What changed

- log request byte size for TICF calls

### Why did it change

- TICF are looking to update their WAF rule to increase their payload size limit so we get less 403s. To help them get a better idea of how much to increase it by, we can temporarily log the request size

## Checklists

    <details>
        <ul>
        </ul>
    </details>

### DIFF
--- a/libs/ticf-cri-service/src/main/java/uk/gov/di/ipv/core/library/ticf/TicfCriService.java
+++ b/libs/ticf-cri-service/src/main/java/uk/gov/di/ipv/core/library/ticf/TicfCriService.java
@@ -98,13 +98,19 @@ public class TicfCriService {
                                     .map(VerifiableCredential::getVcString)
                                     .toList());
 
+            var requestBody = OBJECT_MAPPER.writeValueAsString(ticfCriRequest);
+            LOGGER.info(
+                    LogHelper.buildLogMessage("TICF CRI request body size")
+                            .with(
+                                    "requestBodyBytes",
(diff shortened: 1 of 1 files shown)


libs/ticf-cri-service/src/main/java/uk/gov/di/ipv/core/library/ticf/TicfCriService.java
```diff
@@ -98,13 +98,19 @@ public List<VerifiableCredential> getTicfVc(
...
                                     .toList());
 
+            var requestBody = OBJECT_MAPPER.writeValueAsString(ticfCriRequest);
+            LOGGER.info(
+                    LogHelper.buildLogMessage("TICF CRI request body size")
+                            .with(
+                                    "requestBodyBytes",
+                                    requestBody.getBytes(java.nio.charset.StandardCharsets.UTF_8)
+                                            .length));
+
```
blakeyp, line 103: Is this failsafe if something blows up here?
I think best we make this temporary and revert after a couple of days as this calculation _could_ have a slight perf impact

thebauSoftwire: good point!! I've wrapped it in a try-catch so it doesn't block any users if it fails. I'll set a reminder to revert this  PR after a few days (by eow)

blakeyp (approved):

PR #3883, opened 2026-04-10
PYIC-9047 Fix DCMAW-DL VC treated as expired when proved on day of expiry

blakeyp:
### What changed

- Updated DL expiry check to exclude DCMAW-DL VCs with issue date same as DL expiry date. Also corrected test case that failed to catch this scenario

### Why did it change

-  DL expiry lasts until the end of the day of expiry

### Issue tracking

- [PYIC-9047](https://govukverify.atlassian.net/browse/PYIC-9047)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9047]: https://govukverify.atlassian.net/browse/PYIC-9047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
+++ b/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
@@ -202,6 +202,9 @@ public class VcHelper {
         var nbf = drivingPermitVc.getClaimsSet().getNotBeforeTime();
 
         if (validityDurationInDays != null && nbf != null) {
+            // We pin both the VC issue time and DL expiry to the start of the day, for consistency
+            // with the hasExpired check and to ensure a licence is treated as valid until the very
+            // end of its expiry day.
             var vcIssueTime = nbf.toInstant();
             var startOfIssueDay = getLocalStartOfDay(vcIssueTime);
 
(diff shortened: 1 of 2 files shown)


libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
```diff
@@ -203,34 +203,37 @@ public static boolean isExpiredDrivingPermitVc(
...
 
             var dlExpiryDateString =
                     ((IdentityCheckCredential) drivingPermitVc.getCredential())
                             .getCredentialSubject()
                             .getDrivingPermit()
                             .getFirst()
                             .getExpiryDate();
-
-            var dlExpiryDate =
+            var startOfDlExpiryDay =
```
undermad, line 214: Hmm, this method serve the logic but maybe instead of pushing both vc issued time and dl expiry time to the beginning of the day we should push dl expiry time to the end of the day.
This would be more accurate with DL expiry data considered as valid on the last day.

undermad (approved):

libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/helpers/VcHelper.java
blakeyp: I see what you mean yeah as that would read clearer, but I wanted to maintain the 'pinning issue date to start of day' requirement (from the original ticket) for consistency. I've added in a few comments to make this explicit

undermad (approved):

PR #3901, opened 2026-04-17
PYIC-8459: Update call to EVCS POST /identity endpoint to accept VCs.

undermad:
### What changed

- Added logic to store the SI object and VCs in a single request to the EVCS POST /identity endpoint according to the API changes. The updated endpoint is fail-closed and redirects the user to an error page if the request fails.
Hidden behind the EVS_API_UPDATES feature flag.

### Why did it change

Currently, we use the POST /identity endpoint for storing Stored Identity objects. The T&R team have updated this endpoint so that:

- the request body takes a [list of VCs](https://github.com/govuk-one-login/ipv-identity-reuse-storage/blob/7b551a6d0a0c3735988af5cc63045aa459324ca5/src/specs/api.yaml#L831-L832) for storage
- the request body also takes a [govuk_signin_journey_id](https://github.com/govuk-one-login/ipv-identity-reuse-storage/blob/7b551a6d0a0c3735988af5cc63045aa459324ca5/src/specs/api.yaml#L779-L786)
- an existing SI will be overridden by the si in the request
- VCs which do not already exist in EVCS will be created.
- VCs which already exist in EVCS will have their state and metadata updated. 
- VCs not referenced in the request will not be updated. To change the state or metadata of a stored VC the caller must include it in the request and EVCS will not automatically change the state of Verifiable Credentials.

With this addition, we can store both the user’s Stored Identity (SI) and their VCs in a single API call.

### Issue tracking

- [PYIC-8459](https://govukverify.atlassian.net/browse/PYIC-8459)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-8459]: https://govukverify.atlassian.net/browse/PYIC-8459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
+++ b/lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
@@ -35,6 +35,7 @@ import uk.gov.di.ipv.core.library.enums.TicfCode;
 import uk.gov.di.ipv.core.library.enums.Vot;
 import uk.gov.di.ipv.core.library.evcs.dto.EvcsGetUserVCDto;
 import uk.gov.di.ipv.core.library.evcs.exception.EvcsServiceException;
+import uk.gov.di.ipv.core.library.evcs.exception.FailedToCreateStoredIdentityForEvcsException;
 import uk.gov.di.ipv.core.library.evcs.service.EvcsService;
 import uk.gov.di.ipv.core.library.exceptions.CiExtractionException;
 import uk.gov.di.ipv.core.library.exceptions.CredentialParseException;
@@ -82,6 +83,7 @@ import java.util.stream.Stream;
 import static java.lang.Boolean.TRUE;
 import static uk.gov.di.ipv.core.library.domain.Cri.TICF;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.ERROR_PROCESSING_TICF_CRI_RESPONSE;
+import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_CREATE_STORED_IDENTITY_FOR_EVCS;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_EXTRACT_CIS_FROM_VC;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.FAILED_TO_PARSE_ISSUED_CREDENTIALS;
 import static uk.gov.di.ipv.core.library.domain.ErrorResponse.IPV_SESSION_NOT_FOUND;
@@ -301,6 +303,14 @@ public class ProcessCandidateIdentityHandler
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
+        } catch (FailedToCreateStoredIdentityForEvcsException e) {
+            LOGGER.error(
+                    LogHelper.buildErrorMessage("Failed to construct stored identity record", e));
+            return new JourneyErrorResponse(
+                            JOURNEY_ERROR_PATH,
+                            HttpStatusCode.INTERNAL_SERVER_ERROR,
+                            FAILED_TO_CREATE_STORED_IDENTITY_FOR_EVCS)
+                    .toObjectMap();
         } catch (CredentialParseException e) {
             LOGGER.error(LogHelper.buildErrorMessage("Failed to parse credentials", e));
             return new JourneyErrorResponse(
@@ -351,7 +361,8 @@ public class ProcessCandidateIdentityHandler
                     HttpResponseExceptionWithErrorBody,
                     CredentialParseException,
                     CiExtractionException,
-                    AccountInterventionException {
+                    AccountInterventionException,
+                    FailedToCreateStoredIdentityForEvcsException {
         List<EvcsGetUserVCDto> evcsUserVcs = null;
         var userId = clientOAuthSessionItem.getUserId();
         var auditEventParameters =
@@ -463,7 +474,7 @@ public class ProcessCandidateIdentityHandler
             List<EvcsGetUserVCDto> evcsUserVcs,
             CandidateIdentityType processIdentityType,
             SharedAuditEventParameters auditEventParameters)
-            throws EvcsServiceException {
+            throws EvcsServiceException, FailedToCreateStoredIdentityForEvcsException {
         var achievedVot = ipvSessionItem.getVot();
         VotMatchingResult.VotAndProfile strongestMatchedVot =
                 Objects.isNull(votMatchingResult)
--- a/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
+++ b/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
@@ -172,10 +172,10 @@ public class EvcsService {
             boolean isPendingIdentity)
             throws EvcsServiceException {
 
-        var newUserVcs = findNewUserVcs(credentials, existingEvcsUserVCs);
         var evcsCreateUserVCsDtos =
                 mapNewVcsToEvcsCreateUserVCsDto(
-                        newUserVcs, isPendingIdentity ? PENDING_RETURN : CURRENT);
+                        findNewUserVcs(credentials, existingEvcsUserVCs),
+                        isPendingIdentity ? PENDING_RETURN : CURRENT);
 
         if (!CollectionUtils.isEmpty(existingEvcsUserVCs)) {
             var existingVcsToUpdate =
@@ -236,7 +236,10 @@ public class EvcsService {
                         .toList();
         var evcsStoreIdentityDto =
                 new EvcsPostIdentityDto(
-                        userId, govukSigninJourneyId, vcToCreateAndUpdate, storedIdentityJwt);
+                        userId,
+                        vcToCreateAndUpdate.isEmpty() ? null : govukSigninJourneyId,
+                        vcToCreateAndUpdate.isEmpty() ? null : vcToCreateAndUpdate,
+                        storedIdentityJwt);
 
         return evcsClient.storeUserIdentity(evcsStoreIdentityDto);
     }
@@ -275,7 +278,7 @@ public class EvcsService {
             List<VerifiableCredential> credentials,
             List<EvcsGetUserVCDto> existingEvcsUserVCs,
             BiFunction<List<EvcsGetUserVCDto>, EvcsVCState, List<T>> mapper,
-            Boolean isPendingIdentity) {
+            boolean isPendingIdentity) {
         var existingPendingReturnUserVcsNotInSessionToUpdate =
                 mapper.apply(
                         findExistingVcsNotInSessionCredentials(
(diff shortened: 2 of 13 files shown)


libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
```diff
@@ -252,11 +203,112 @@ public HttpResponse<String> storeStoredIdentityRecord(
...
+            Vot achievedVot)
+            throws FailedToCreateStoredIdentityForEvcsException, EvcsServiceException {
+        var storedIdentityJwt =
+                storedIdentityService.getStoredIdentityForEvcs(
+                        userId, credentials, strongestAchievedVot, achievedVot);
+
+        var newUserVcsToStore =
+                mapNewUsersVcsToEvcsCreateUserVcsDto(credentials, existingEvcsUserVcs);
+        var existingVcsToUpdate =
+                mapAndUpdateStateOfExistingUserVcs(credentials, existingEvcsUserVcs);
```
undermad, line 226: I'll appreciate feedback on those methods and if I correctly set the state, metadata and provenance on DTO objects.
I tried to replicate it according to this [updateExistingUserVcs](https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java#L264)

local-running/core.local.params.yaml
```diff
@@ -358,7 +358,7 @@ core:
     kidJarHeaderEnabled: true
     drivingLicenceAuthCheck: true
     sisVerificationEnabled: true
-    evcsApiUpdatesEnabled: false
```
undermad, line 361: Let me know if this is to early to enable it in API tests.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -301,6 +303,14 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
             return new JourneyErrorResponse(
                             JOURNEY_ERROR_PATH, e.getResponseCode(), e.getErrorResponse())
                     .toObjectMap();
+        } catch (FailedToCreateStoredIdentityForEvcsException e) {
```
DanCorderIPV, line 306: Should this be a different exception? Presumably it could be thrown even if the SI is fine, but one of the VCs being saved has an issue?
Same for the log message below

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
```diff
@@ -301,6 +303,14 @@ public Map<String, Object> handleRequest(ProcessRequest request, Context context
...
+            LOGGER.error(
+                    LogHelper.buildErrorMessage("Failed to construct stored identity record", e));
+            return new JourneyErrorResponse(
+                            JOURNEY_ERROR_PATH,
+                            HttpStatusCode.INTERNAL_SERVER_ERROR,
+                            FAILED_TO_CREATE_STORED_IDENTITY_FOR_EVCS)
```
DanCorderIPV, line 312: Similarly should this be a new error code?

local-running/core.local.params.yaml
DanCorderIPV: I think it's fine to enable the flag in local running, but we should probably add a feature set to disable `evcsApiUpdatesEnabled` and then add API tests that use that feature set so that we have tests running the old and the new code until we turn the flag on for all environments.
The new tests can be labelled with the clean-up ticket number and deleted as part of the clean-up once this is live.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClientTest.java
```diff
@@ -110,7 +110,10 @@ class EvcsClientTest {
                             "VC_Signature2", EvcsVCState.ABANDONED, TEST_METADATA));
     private static final EvcsPostIdentityDto EVCS_POST_IDENTITY_DTO_SI_ONLY =
             new EvcsPostIdentityDto(
-                    TEST_USER_ID, null, new EvcsStoredIdentityDto("storedIdentityJwt", Vot.P2));
+                    TEST_USER_ID,
+                    null,
```
DanCorderIPV, line 115: We should probably have an equivalent `EVCS_POST_IDENTITY_DTO_SI_AND_VCS` DTO here and a matching test

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/evcs/pact/ContractTest.java
```diff
@@ -514,7 +514,7 @@ public RequestResponsePact postIdentityReturns202(PactDslWithProvider builder) {
     void testPostIdentityReturns202(MockServer mockServer) throws EvcsServiceException {
         // Arrange
         var evcsPostIdentityDto =
-                new EvcsPostIdentityDto(TEST_USER_ID, null, EVCS_STORED_IDENTITY_DTO);
+                new EvcsPostIdentityDto(TEST_USER_ID, null, null, EVCS_STORED_IDENTITY_DTO);
```
DanCorderIPV, line 517: We should probably add some more contract tests that include VCs and the journey ID - we'll need to check with the EVCS team that the new tests won't break their pipeline.
And we can mark these tests for deletion once we enable the flag in prod?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
```diff
@@ -160,65 +161,13 @@ public List<EvcsGetUserVCDto> getUserVCs(
         return evcsClient.getUserVcs(userId, evcsAccessToken, List.of(states)).vcs();
     }
 
-    public void storePendingIdentityWithPostIdentity(
```
DanCorderIPV, line 163: Looks like these methods were already obsolete and you're just deleting them?

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
```diff
@@ -252,11 +203,112 @@ public HttpResponse<String> storeStoredIdentityRecord(
...
+                Stream.concat(newUserVcsToStore.stream(), existingVcsToUpdate.stream()).toList();
+        var evcsStoreIdentityDto =
+                new EvcsPostIdentityDto(
+                        userId, govukSigninJourneyId, vcToCreateAndUpdate, storedIdentityJwt);
 
         return evcsClient.storeUserIdentity(evcsStoreIdentityDto);
     }
 
+    private List<EvcsCreateUserVCsDto> mapNewUsersVcsToEvcsCreateUserVcsDto(
+            List<VerifiableCredential> credentials, List<EvcsGetUserVCDto> existingEvcsUserVcs) {
```
DanCorderIPV: Can you re-use this in `storeCompletedOrPendingIdentityWithPostVcs()`? The only difference appears to be that `CURRENT` is hardcoded here and might be `PENDING_RETURN` in `storeCompletedOrPendingIdentityWithPostVcs()`

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
DanCorderIPV: The methods look the same except for the different DTOs produced - can the filtering be extracted into common methods?
Maybe return a `Dictionary<EvcsVCState, List<EvcsGetUserVCDto>>` for the updates?

DanCorderIPV (changes requested):

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
undermad: Yes, those methods was created a head of time. As requirements changed it wasn't suitable to use them.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
undermad: This exception is thrown only when StoredIdentityService fail to create SI object.
FAILED_TO_CREATE_STORED_IDENTITY_FOR_EVCS seems to be unused and match the description of the error, therefore I decided to reuse it instead of creating new. 
All occurrences where this error is thrown are [here](https://github.com/govuk-one-login/ipv-core-back/blob/main/libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/StoredIdentityService.java#L89-L117).
This endpoint become fail close if we enable feature flag, so we should respond with 500 if we cannot create SI object. 
In other scenarios we throw EvcsServiceException with log message "Failed to store identity" which seems to be accurate.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
undermad: Same as above, this error code and message is unused.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/evcs/client/EvcsClientTest.java
undermad: Agree, added additional test and test DTO object.

libs/evcs-service/src/test/java/uk/gov/di/ipv/core/library/evcs/pact/ContractTest.java
undermad: I posted message on Slack to arrange some work around contract tests, I assume this will be required for other endpoints as well. From the past experience I remember we usually have separate ticket for Contract Tests.

libs/evcs-service/src/main/java/uk/gov/di/ipv/core/library/evcs/service/EvcsService.java
undermad: It was quite tricky to refactor, but I leveraged usage of Java generics and BiFunction to pass mapper to the method. It become highly reusable and perfectly fit this example.

local-running/core.local.params.yaml
undermad: Sorry, I think it would be better if we keep it disabled so all our tests run against same config as it is in production environment. I crated few scenarios to cover the evcsApiUpdates feature flag, but I wasn't sure which journeys to choose and how many of them to add.

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
DanCorderIPV: Sorry, I misunderstood what this was for, fine to leave it

lambdas/process-candidate-identity/src/main/java/uk/gov/di/ipv/core/processcandidateidentity/ProcessCandidateIdentityHandler.java
DanCorderIPV: Sorry, I misunderstood what this was for, fine to leave it

DanCorderIPV (approved):

PR #3939, opened 2026-05-06
PYIC-8711: Routing to avoid a dead end for COI/RFC users

AmritSidhu:
### What changed

- Add new journeyMap update-details-higher-strength. This follows same pattern as intervention-reprove-identity. Due to specific scenarios such as the link on page firing the preferNoApp event we need to dupliate the strategicAppTriage steps manually. 
 
- Scenarios covered:
  - COI / 6MFC / user route due to GPG45 evaluation failures
  - COI / 6MFC / User drop-off in DCMAW eligibility screens/ ineligible for app
  - COI / 6MFC / User chooses “Confirm your identity another way” in the app - abandoned event
  - COI / 6MFC / System error with option to exit
  - COI / 6MFC / Fail app twice
  - COI / 6MFC / Fail final fraud

### Why did it change

- In One Login, users are required to maintain a level of confidence in their identity to access services. Its possible that during COI journey while updating their details , some users may lose their medium confidence status.

- Currently, these users can encounter a dead end — either being shown a failure page or being forced to return to the relying party without a valid route to recover their status.

- To avoid this, we should provide users in both journeys with a clear option to use a valid passport via the GOV.UK One login app to re-establish their medium confidence level.

### Issue tracking

- [PYIC-8711](https://govukverify.atlassian.net/browse/PYIC-8711)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-8711]: https://govukverify.atlassian.net/browse/PYIC-8711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/update-details-higher-strengh/update-details-higher-strengh.feature
+++ b/api-tests/features/update-details-higher-strengh/update-details-higher-strengh.feature
@@ -1,0 +1,380 @@
+@Build @QualityGateIntegrationTest @QualityGateRegressionTest
+Feature: Update details higher strength
+  Rule: Repeat Fraud Check
+    Background:
+      Given the subject already has the following credentials
+        | CRI     | scenario               |
+        | dcmaw   | kenneth-passport-valid |
+        | address | kenneth-current        |
+      And the subject already has the following expired credentials
+        | CRI   | scenario        |
+        | fraud | kenneth-score-2 |
+      And I have an existing stored identity record with a 'P3' vot
+      And I activate the 'coi6mfcDeadEndRouting' feature set
+      When I start a new 'medium-confidence' journey
+      Then I get a 'confirm-your-details' page response
+
+    Scenario: Successful 6MFC identity recovery via App for Name and Address change
+      When I submit a 'family-name-and-address' event
+      Then I get a 'page-update-name' page response and pageContext
+        | Context     | Value            |
+        | journeyType | repeatFraudCheck |
+      When I submit an 'update-name' event
+      Then I get an 'identify-device' page response
+      When I submit an 'appTriage' event
+      Then I get a 'pyi-triage-select-device' page response
+      When I submit a 'computer-or-tablet' event
+      Then I get a 'pyi-triage-select-smartphone' page response and pageContext
+        | Context    | Value |
+        | deviceType | dad   |
+      When I submit an 'android' event
+      Then I get a 'pyi-triage-desktop-download-app' page response and pageContext
+        | Context    | Value   |
+        | smartphone | android |
+        | isAppOnly  | true    |
+      When the async DCMAW CRI produces a 'kenneth-passport-fail-no-ci' VC
+      And I pass on the DCMAW callback
+      When I poll for async DCMAW credential receipt
+      Then the poll returns a '201'
+      When I submit the returned journey event
+      Then I get an 'need-more-information-confirm-change-details' page response and pageContext
+        | Context              | Value             |
+        | journeyType          | repeatFraudCheck  |  
+      When I submit an 'passport' event  
+      Then I get an 'identify-device' page response
+      When I submit an 'appTriage' event
+      Then I get a 'pyi-triage-select-device' page response
+      When I submit a 'computer-or-tablet' event
+      Then I get a 'pyi-triage-select-smartphone' page response and pageContext
+        | Context    | Value |
+        | deviceType | dad   |
+      When I submit an 'android' event
+      Then I get a 'pyi-triage-desktop-download-app' page response and pageContext
+        | Context    | Value   |
+        | smartphone | android |
+        | isAppOnly  | true    |
+      When the async DCMAW CRI produces a 'kenneth-changed-given-name-passport-valid' VC
+      And I poll for async DCMAW credential receipt
+      Then the poll returns a '201'
+      When I submit the returned journey event
+      Then I get a 'page-dcmaw-success' page response and pageContext
+        | Context   | Value |
+        | noAddress | true  |
+      When I submit a 'next' event
+      Then I get a 'address' CRI response
+      When I submit 'kenneth-changed' details to the CRI stub
+      Then I get a 'fraud' CRI response
+      When I submit 'kenneth-changed-given-name-score-2' details with attributes to the CRI stub
+        | Attribute          | Values                   |
+        | evidence_requested | {"identityFraudScore":1} |
+      Then I get a 'page-ipv-success' page response and pageContext
+        | Context     | Value |
+        | journeyType | coi   | 
+
+    Scenario: Successful 6MFC identity recovery via App for Name change only
+      When I submit a 'given-names-only' event
+      Then I get a 'page-update-name' page response and pageContext
+        | Context     | Value            |
+        | journeyType | repeatFraudCheck |
+      When I submit an 'update-name' event
+      Then I get an 'identify-device' page response
+      When I submit an 'appTriage' event
+      Then I get a 'pyi-triage-select-device' page response
+      When I submit a 'computer-or-tablet' event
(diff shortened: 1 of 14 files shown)


api-tests/features/update-details-higher-strengh/update-details-higher-strengh.feature
```diff
@@ -0,0 +1,380 @@
...
+        | Context    | Value   |
+        | smartphone | android |
+        | isAppOnly  | true    |
+      When the async DCMAW CRI produces a 'kenneth-changed-given-name-passport-valid' VC
+      And I poll for async DCMAW credential receipt
+      Then the poll returns a '201'
+      When I submit the returned journey event
+      Then I get a 'page-dcmaw-success' page response and pageContext
+        | Context   | Value |
+        | noAddress | true  |
```
blakeyp: As this is a name+address change shouldn't this be the no-context variant?

api-tests/features/update-details-higher-strengh/update-details-higher-strengh.feature
```diff
@@ -0,0 +1,380 @@
...
+      Then I get a 'pyi-triage-select-smartphone' page response and pageContext
+        | Context    | Value |
+        | deviceType | dad   |
+      When I submit an 'android' event
+      Then I get a 'pyi-triage-desktop-download-app' page response and pageContext
+        | Context    | Value   |
+        | smartphone | android |
+        | isAppOnly  | true    |
+      When I submit a 'preferNoApp' event
+      Then I get an 'update-details-failed' page response
```
blakeyp, line 444: Nice coverage of the App scenarios 👍 do we also need to cover failing GPG45 evaluation after the (first) Fraud check e.g. use/used a DL in the app and then only get a Fraud score of 1 -> pivot to using passport in the app instead

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
```diff
@@ -74,6 +74,16 @@ states:
       next:
         targetState: PROCESS_INCOMPLETE_IDENTITY_BEFORE_NO_MATCH_NINO
 
+  FAILED_RFC:
+    events:
+      next:
+        targetState: PROCESS_INCOMPLETE_IDENTITY_BEFORE_NO_MATCH_REPEAT_FRAUD_CHECK
+
+  FAILED_COI:
```
blakeyp: Wonder if we could name these "FAILED_UPDATE_DETAILS_HIGHER_STRENGTH_X" to give a bit more context? Also could possibly use the `rfc` journey context to distinguish COI vs RFC like the `FAILED_CONFIRM_DETAILS` state above?

blakeyp: Also I see `FAILED_CONFIRM_DETAILS` calls an audit event `IPV_USER_DETAILS_UPDATE_END` which we probably need here too - happy for this to be a follow-up ticket since it's not called out in the requirements

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
blakeyp, line 352: I think we need to handle here too to cover the case of insufficient Fraud scores - e.g. DL+fraud score of 1 would be caught here as no GPG45 profile met

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-name.yaml
```diff
@@ -299,9 +319,13 @@ states:
         targetState: PROCESS_UPDATE_IDENTITY
       fraud-fail-with-no-ci:
         targetState: PROCESS_UPDATE_IDENTITY
+        checkFeatureFlag:
+          coi6mfcDeadEndRoutingEnabled:
+            targetJourney: UPDATE_DETAILS_HIGHER_STRENGTH
+            targetState: START_HIGHER_STRENGTH
+            journeyContextToSet: nameAndAddress
```
blakeyp, line 327: We ideally need to unset this somewhere (in case a user somehow winds up on the journey again in the same session) - probably as soon as it's used?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-details-higher-strength.yaml
```diff
@@ -0,0 +1,523 @@
...
+      fraud-fail-with-no-ci:
+        targetState: PROCESS_UPDATE_IDENTITY
+
+  # Remove any DCMAW Async VC from previous attempts
+  RESET_SESSION_BEFORE_NEW_IDENTITY:
+    response:
+      type: process
+      lambda: reset-session-identity
+      lambdaInput:
+        resetType: PENDING_DCMAW_ASYNC_ALL
```
blakeyp: I'd thought this would actually clear all session VCs (+DCMAW pending record) but it's actually only the DCMAW VC: https://github.com/govuk-one-login/ipv-core-back/blob/28e1dc7462b6bf093688c6d4f59348f0a7d16204/libs/verifiable-credentials/src/main/java/uk/gov/di/ipv/core/library/verifiablecredential/service/SessionCredentialsService.java#L123 🤔 
I think we may need to clear the Fraud VC too, since the user could use a doc with slightly different name?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-details-higher-strength.yaml
```diff
@@ -0,0 +1,523 @@
...
+        targetState: PROCESS_INCOMPLETE_IDENTITY
+
+  DELETE_LOGIN:
+    response:
+      type: page
+      pageId: delete-handover
+      pageContext:
+        journeyType: reprove
+
+  PROCESS_INCOMPLETE_IDENTITY:
```
blakeyp: I think these aren't used in this journey?

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-details-higher-strength.yaml
```diff
@@ -0,0 +1,523 @@
...
+    events:
+      next:
+        targetState: FRAUD_CHECK_NAMES_ONLY
+    
+  POST_APP_DOC_CHECK_NAMES_WITH_ADDRESS:
+    response:
+      type: page
+      pageId: page-dcmaw-success
+      pageContext:
+        noAddress: true
```
blakeyp: For with address should be no context?

AmritSidhu: Ah this was copied over from update-name journeyMap. So looks like we need to re-visit that to do a clean up. I'll make a note of that

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-details-higher-strength.yaml
AmritSidhu: So this is used for when the user selects they do not have a device are part of the final attempt to use the app. I've followed what the intervention-reprove journey map does

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/failed.yaml
AmritSidhu: Other than just adding the audit event to the journey event, is there anything else that would need to be updated?

blakeyp (approved):

PR #4001, opened 2026-06-03
PYIC-9095: Tidy up repeatFraudCheckEnabled feature flag

AmritSidhu:
### What changed

- Updated the CheckExistingIdentityHandlerTest suite to align with the decommissioning of the REPEAT_FRAUD_CHECK feature flag. 
Because the fraud check expiration logic is now permanent and executes on every request, standard identity reuse tests were incorrectly routing to the repeat fraud check path due to missing/empty mock fraud data. 

- Added MockedStatic<VcHelper> blocks to the failing standard reuse tests.
- Stubbed VcHelper.allFraudVcsAreExpiredOrFromUnavailableSource to return false for reuse scenarios, forcing the code to bypass the fraud check rerouting and cleanly hit the expected identity reuse assertions.

### Why did it change

- The repeat fraud check feature has long been live and there are no plans to turn it off, in fact it will be built into SIS when Phase 3 goes live. We no longer need it behind a feature flag in core-back.

### Issue tracking

- [PYIC-9095](https://govukverify.atlassian.net/browse/PYIC-9095)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9095]: https://govukverify.atlassian.net/browse/PYIC-9095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
+++ b/lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
@@ -224,6 +224,7 @@ class CheckExistingIdentityHandlerTest {
     @Captor private ArgumentCaptor<AuditEvent> auditEventArgumentCaptor;
     private JourneyRequest event;
     private ClientOAuthSessionItem clientOAuthSessionItem;
+    private MockedStatic<VcHelper> mockVcHelper;
 
     @BeforeEach
     void setUpEach() {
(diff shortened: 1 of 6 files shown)


libs/common-services/src/test/java/uk/gov/di/ipv/core/library/service/AppConfigServiceTest.java
```diff
@@ -290,11 +290,6 @@ void getParameterReturnsParametersWithFeatureOverride() {
         assertEquals("alternate-component-id", param);
     }
 
-    @Test
```
DanCorderIPV, line 293: This is testing the config code so we can't just delete it, we need to change to a different feature flag that is enabled (or make a test one that can be used here and permanently enabled)

lambdas/check-existing-identity/src/test/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandlerTest.java
DanCorderIPV: This is quite verbose and repeated a lot, is there anyway we can pull it out into a helper method or similar?

DanCorderIPV (changes requested):

DanCorderIPV (approved):

PR #4050, opened 2026-06-19
PYIC-9170: Rename KBV find-another-way pages.

undermad:
### What changed

P2 Journey Updates
Introduced four new states for the P2 journey:
  - MITIGATION_NO_PHOTO_ID_WEB_FIND_ANOTHER_WAY
  - MITIGATION_PHOTO_ID_WEB_FIND_ANOTHER_WAY
  - DROPOUT_NO_PHOTO_ID_WEB_FIND_ANOTHER_WAY
  - DROPOUT_PHOTO_ID_WEB_FIND_ANOTHER_WAY
  
Additionally: 
  - Pinned the journey flow to these new states.
  - The previous states have been retained to avoid impacting users currently progressing through journeys.

P1 Journey Updates
  - Renamed state names, response page ids, and pinned journey map states 
  - Unlike the P2 changes, no state duplication or staging was required for P1, as there is currently no live traffic on these journeys.
  
API Tests
  - Updated all occurrences of the above pages to match new name

### Why did it change

The URL paths associated with the "KBV find another way" screens have been updated to support new dynamic Post Open Banking frontend pages.

Previously, these pages were dedicated to users who reached the route following KBV outcomes. The new implementation allows the same frontend pages to be reused dynamically for both KBV and Open Banking user journeys.

### Issue tracking

- [PYIC-9170](https://govukverify.atlassian.net/browse/PYIC-9170)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9170]: https://govukverify.atlassian.net/browse/PYIC-9170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
+++ b/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
@@ -314,10 +314,10 @@ states:
         targetJourney: INELIGIBLE
         targetState: INELIGIBLE_SKIP_MESSAGE
 
-  PYI_KBV_DROPOUT_PHOTO_ID:
+  DROPOUT_PHOTO_ID_WEB_FIND_ANOTHER_WAY:
     response:
       type: page
-      pageId: photo-id-security-questions-find-another-way
+      pageId: photo-id-web-find-another-way
       pageContext:
         reason: dropout
     events:
@@ -340,7 +340,7 @@ states:
     nestedJourney: KBVS
     exitEvents:
       fail-with-no-ci:
-        targetState: PYI_KBV_DROPOUT_PHOTO_ID
+        targetState: DROPOUT_PHOTO_ID_WEB_FIND_ANOTHER_WAY
         checkMitigation:
           alternate-doc-invalid-passport:
             targetJourney: INELIGIBLE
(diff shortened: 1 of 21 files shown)


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p1-identity.yaml
```diff
@@ -335,12 +335,33 @@ states:
             targetJourney: TECHNICAL_ERROR
             targetState: ERROR
 
+  PYI_KBV_DROPOUT_PHOTO_ID_FIND_ANOTHER_WAY:
```
blakeyp: Do we want to rename it `PHOTO_ID_WEB_FIND_ANOTHER_WAY` to align with the new page URL and so it can be reused?

undermad: Sure, is those include the following states in P1 and P2?
- PYI_KBV_DROPOUT_PHOTO_ID_FIND_ANOTHER_WAY
- PYI_KBV_DROPOUT_NO_PHOTO_ID_FIND_ANOTHER_WAY

I assume mitigation states should remain related to KBV?

blakeyp: Ah sorry I forgot there are 2 variants of this, in that case I think something like:
`DROPOUT_PHOTO_ID_WEB_FIND_ANOTHER_WAY`
`MITIGATION_PHOTO_ID_WEB_FIND_ANOTHER_WAY`
(and yes across all the relevant states in P1 and P2)

This mitigation route may end up being used by Open Banking in future too so worth keeping it generic I think

blakeyp (approved):

PR #4082, opened 2026-07-02
PYIC-9066: Add routing for explicit mitigation journey instead of mitigating all CIs

DanCorderIPV:
### What changed

Added explicit mitigation events

### Why did it change

So mitigation can be controlled by config

### Issue tracking

- [PYIC-9066](https://govukverify.atlassian.net/browse/PYIC-9066)

## Checklists

This is the first stage - adding new routes that won't be used until we update the config

[PYIC-9066]: https://govukverify.atlassian.net/browse/PYIC-9066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


local-running/core.local.params.yaml
```diff
@@ -36,10 +36,6 @@ core:
         detectedScore: 11
         checkedScore: -11
         returnCode: "breaching-p2-when-combined-with-non-breaching"
-      - ci: "BREACHING-P3-WHEN-COMBINED-WITH-NON-BREACHING"
```
DanCorderIPV, line 39: This isn't used and doesn't make sense any more

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/nested-journeys/address-and-fraud.yaml
```diff
@@ -46,3 +46,5 @@ nestedJourneyStates:
         exitEventToEmit: fraud-fail-with-no-ci-fatal
       fail-with-ci:
         exitEventToEmit: fraud-fail-with-ci
+      liveness-likeness:
```
blakeyp: Do we need to be using the `checkMitigation` predicate here, or is that still to come?

Also the exit event name could be more descriptive something like `fraud-fail-needs-liveness-likeness` if that's possible?

DanCorderIPV: Sorry, I realised on the way home I've misunderstood how this works - it should use `checkMitigation` and the event will remain as `fail-with-ci`
Relatedly maybe we should update the config so the mitigations don't look like journey events. Currently they look like `/journey/enhanced-verification` and then during journey map processing the `/journey/` is chopped off.

blakeyp (approved):

PR #4132, opened 2026-07-27
PYIC-9207: Add evcsStubVerifyKey secret to api tests.

undermad:
⚠️  This [PR](https://github.com/govuk-one-login/ipv-core-common-infra/pull/1580) need to be merged first ⚠️  

### What changed

Created EVCS_STUB_VERIFY_KEY secret and added to api tests:
- added to .env.template
- added to pipeline running instance
- changed evcs access token generation to use new secret
- updated readme

### Why did it change

Currently, the same key is used both to sign JAR requests and to generate the EVCS access token in API tests. To support key rotation, separate keys will be used for JAR signing and EVCS access token, same as it's used in orchestrator stub.

⚠️  I going to notify developers about new mandatory secret that need to be added to .env file.

### Issue tracking

- [PYIC-9207](https://govukverify.atlassian.net/browse/PYIC-9207)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9207]: https://govukverify.atlassian.net/browse/PYIC-9207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-tests/src/config/config.ts
```diff
@@ -32,6 +32,7 @@ const config = {
   orch: {
     redirectUrl: getMandatoryConfig("ORCHESTRATOR_REDIRECT_URL"),
     signingKey: getMandatoryConfig("JAR_SIGNING_KEY"),
+    evcsStubVerifyKey: getMandatoryConfig("EVCS_STUB_VERIFY_KEY"),
```
undermad, line 35: I put it in orch section as our orchestrator is signing the token and passing it in jar request.
In API tests we mimic the orchestrator behaviour therefore it should be correct place to define this secret.

blakeyp (approved):

PR #4135, opened 2026-07-28
PYIC-9223 Stop calling invalidate at journey start

blakeyp:
### What changed

- Removed the unconditional stored-identity invalidate call from CheckExistingIdentity, which ran at the start of every journey. Invalidation still happens in reset-identity when a user commits to an update.
- Updated API test scenarios to expect invalidation after the update is committed rather than at journey start.

### Why did it change

EVCS hit DynamoDB conflict errors seemingly caused by users starting parallel reuse journeys, where invalidating at journey start produced a rapid /invalidate→/identity sequence. We only need to invalidate when a user actually opts to update their details, which reset-identity already does - so the blanket call is unnecessary and removing it avoids the conflicting sequence. The up-front call was groundwork for SIS Phase 3, which will instead trigger invalidation from a new JAR claim.

### Issue tracking

- [PYIC-9223](https://govukverify.atlassian.net/browse/PYIC-9223)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9223]: https://govukverify.atlassian.net/browse/PYIC-9223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/stored-identity.feature
+++ b/api-tests/features/stored-identity.feature
@@ -37,6 +37,8 @@ Feature: Stored Identity
       Then I get a 'page-update-name' page response and pageContext
         | Context     | Value            |
         | journeyType | repeatFraudCheck |
+      When I submit a 'update-name' event
+      Then I get an 'identify-device' page response
       And I have a stored identity record with a 'P2' max vot that is 'invalid'
 
     Scenario: Update address invalidates stored identity
@@ -53,4 +55,6 @@ Feature: Stored Identity
       Then I get a 'page-update-name' page response and pageContext
         | Context     | Value            |
         | journeyType | repeatFraudCheck |
+      When I submit a 'update-name' event
+      Then I get an 'identify-device' page response
       And I have a stored identity record with a 'P2' max vot that is 'invalid'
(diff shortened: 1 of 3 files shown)


api-tests/features/stored-identity.feature
```diff
@@ -37,6 +37,8 @@ Feature: Stored Identity
       Then I get a 'page-update-name' page response and pageContext
         | Context     | Value            |
         | journeyType | repeatFraudCheck |
+      When I submit a 'update-name' event
+      Then I get an 'identify-device' page response
```
blakeyp, line 41: The reset-identity invalidate call only happens once the user commits to starting an update journey (i.e. goes to the App)

blakeyp:
Also updated the EVCS touchpoints Mural

blakeyp:
Build tests against my dev env ✅ 
https://github.com/govuk-one-login/ipv-core-tests/actions/runs/30619184025

DanCorderIPV (approved):

PR #4189, opened 2026-08-12
PYIC-9068: Fix new CI mitigation journeys - back button and cross-browser scenarios.

undermad:
### What changed

- Fix back button on select device page
- Fix cross-browser scenarios during fraud mitigation journey

### Why did it change

- State before select device page was lambda process - rewired this lambda to happen after address and fraud nested journey
- Cross-browser scenarios required some changes in check-existing-identity lambda. As fraud mitigation via app is last step in the journey, returning user in cross-browser scenario, when proceeded to process-candidate-lambda need to have access to address and fraud VCs from previous session to build GPG45 profile.
During fraud mitigation we reset only pending dcmaw async VC which means we can fetch them using previousIpvSession id.

### Issue tracking

- [PYIC-9068](https://govukverify.atlassian.net/browse/PYIC-9068)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9068]: https://govukverify.atlassian.net/browse/PYIC-9068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -608,6 +608,7 @@ public class CheckExistingIdentityHandler
         return JOURNEY_F2F_FAIL;
     }
 
+    @SuppressWarnings("java:S107") // Methods should not have too many parameters
     private JourneyResponse buildDCMAWContinuationResponse(
             VerifiableCredentialBundle credentialBundle,
             List<ContraIndicator> contraIndicators,
--- a/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
+++ b/lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
@@ -72,6 +72,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import static com.nimbusds.oauth2.sdk.http.HTTPResponse.SC_NOT_FOUND;
 import static software.amazon.awssdk.utils.CollectionUtils.isNullOrEmpty;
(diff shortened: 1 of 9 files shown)


lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -641,8 +642,33 @@ private JourneyResponse buildDCMAWContinuationResponse(
...
+            // from the previous session, as it may not have been updated.
+            var newSessionVcs =
+                    Stream.concat(
+                                    credentialBundle.credentials.stream()
+                                            .filter(vc -> vc.getCri().equals(DCMAW_ASYNC)),
+                                    previousSessionVcs.stream()
+                                            .filter(vc -> !vc.getCri().equals(DCMAW_ASYNC)))
+                            .toList();
+            sessionCredentialsService.persistCredentials(
+                    newSessionVcs, auditEventUser.getSessionId(), true);
```
blakeyp, line 671: I like the idea, but I'm a little nervous about the side effects this may have and the impact on different types of journeys (e.g. update details) which we'd also need to test (adding a further maintenance burden). The cross-browser recovery fix wasn't intended to be a save-and-resume style feature, which this strays into - I think any broad solution probably warrants a bigger piece of work more fully thought through?

I think we could keep this as a tactical fix for now if we can scope it only to this specific Fraud mitigation scenario, but I'm not sure if this is possible?

undermad: I agree, to approach this safe we propably should limit this feature just to new Fraud mitigation journey.
I think it should be possible to check in runtime, we are doing those checks constantly in journey engine. 
I'll see what we can do.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -676,6 +706,13 @@ private JourneyResponse buildDCMAWContinuationResponse(
         };
     }
 
+    private boolean isLivenessLikenessMitigation(List<ContraIndicator> contraIndicators, Vot vot) {
+        var mitigationEvent =
+                cimitUtilityService.getMitigationEventIfBreachingOrActive(contraIndicators, vot);
+        return mitigationEvent.isPresent()
+                && mitigationEvent.get().equals(Events.LIVENESS_LIKENESS);
+    }
```
undermad, line 714: I thought about reusability of this method, but even IDE suggested to inline this instead of passing event.
Maybe if in the future we decide do some work around those events we may move it to cimit utility class.
For now its just scoped for this specific scenario.

lambdas/check-existing-identity/src/main/java/uk/gov/di/ipv/core/checkexistingidentity/CheckExistingIdentityHandler.java
```diff
@@ -641,8 +646,33 @@ private JourneyResponse buildDCMAWContinuationResponse(
                 deviceInformation,
                 new AuditExtensionPreviousIpvSessionId(previousIpvSessionItem.getIpvSessionId()));
 
-        sessionCredentialsService.persistCredentials(
-                credentialBundle.credentials, auditEventUser.getSessionId(), true);
+        if (isLivenessLikenessMitigation(contraIndicators, lowestGpg45ConfidenceRequested)) {
```
undermad, line 649: One more question. I believe we don't need to hide it behind feature flag as it's guarded in the journey map?

blakeyp (approved):

PR #4204, opened 2026-08-18
PYIC-9125: Fraud mitigation journeys - COI scenarios.

undermad:
### What changed

- Wire update-details-higher-strength journey after COI/6MFC fraud CRI CIs scenarios.

### Why did it change

- To support Fraud CRI mitigation in COI and 6MFC journeys.

### Issue tracking

- [PYIC-9125](https://govukverify.atlassian.net/browse/PYIC-9125)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9125]: https://govukverify.atlassian.net/browse/PYIC-9125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-tests/features/fraud-mitigation/coi-fraud-mitigation.feature
+++ b/api-tests/features/fraud-mitigation/coi-fraud-mitigation.feature
@@ -47,7 +47,7 @@ Feature: Fraud mitigation - COI
         | Attribute          | Values                   |
         | evidence_requested | {"identityFraudScore":2} |
       Then I get an 'need-more-information-confirm-change-details' page response and pageContext
-        | Context     | Value         |
+        | Context     | Value            |
         | journeyType | repeatFraudCheck |
       When I submit an 'passport' event
       Then I get an 'identify-device' page response
@@ -89,7 +89,7 @@ Feature: Fraud mitigation - COI
         | Attribute          | Values                   |
         | evidence_requested | {"identityFraudScore":2} |
       Then I get an 'need-more-information-confirm-change-details' page response and pageContext
-        | Context     | Value         |
+        | Context     | Value            |
         | journeyType | repeatFraudCheck |
       When I submit a 'returnToRp' event
       Then I get an OAuth response
@@ -102,7 +102,7 @@ Feature: Fraud mitigation - COI
         | Attribute          | Values                   |
         | evidence_requested | {"identityFraudScore":2} |
       Then I get an 'need-more-information-confirm-change-details' page response and pageContext
-        | Context     | Value         |
+        | Context     | Value            |
         | journeyType | repeatFraudCheck |
       When I submit an 'passport' event
       Then I get an 'identify-device' page response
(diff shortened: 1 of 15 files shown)


undermad:
I'm not sure if those scenarios should use auto mitigation feature?
Also, I assume unhappy paths are covered by update-details-higher-strength API tests?

api-tests/features/fraud-mitigation/coi-fraud-mitigation.feature
```diff
@@ -0,0 +1,454 @@
...
+      When I submit the returned journey event
+      # TODO: This should not be a updateDetails reason as user is not allowed to continue
+      Then I get an 'pyi-no-match' page response and pageContext
+        | Context | Value         |
+        | reason  | updateDetails |
+      When I submit a 'next' event
+      Then I get an OAuth response
+      When I use the OAuth response to get my identity
+      Then I am issued a 'P0' identity
+      And I have a stored identity record with a 'P3' max vot that is 'invalid'
```
undermad: I left this one scenario here as an example of improvements/follow ups we need to apply to the update-details-higher-strenght journey. 
When user get CI from app or/and VC are not correlated then user may end up in the pyi-no-match page with page context updateDetails which indicate that users identity is valid where this is not true.
Ideally we would cover this as follow up ticket as it is going outside of the scope of this ticket.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/update-details-higher-strength.yaml
```diff
@@ -14,6 +14,16 @@ states:
...
 
+  MITIGATION_START_HIGHER_STRENGTH:
+    events:
+      next:
+        # Temporary - User with CI cannot continue journey without updating their details
+        # need-more-information-confirm-change-details needs new context which doesn't indicate 6MFC
+        # Curren wiring from this journey is to page with updateDetails context which indicate that user's identity is valid
+        # as user won't be able to continue due to the existing CI and profile unmet evaluation in process-candidate-identity lambda
+        # we want to redirect user to page which will display accurate information
+        targetState: NEED_MORE_INFORMATION_CONFIRM_CHANGE_DETAILS_REPEAT_FRAUD_CHECK
```
blakeyp: An idea but not sure if it will work without side effects: instead of doing this, just set the journeyContext="rfc" which should fix the pyi-no-match page too? it's a bit of a hack but may work for now

undermad: This approach will create another tech debt to fix in the future. Instead, I changed fail-with-ci events across the whole update-details-higher-strength journey to new entry point in the failed journey.
As identity with breaching CI is invalid, this will be more accurate to point it to the pyi-no-match page with no context. There is also branch for 'rfc' which redirect user to pyi-no-match with repeatFraudCheck context. Couldn't reuse basic entry to Failed journey as coi/6mfc lunch the audit event in journey engine.

For entry to update-details-higher-strength I added checkMitigation to redirect user to same page as rfc and I going to raise ticket (once tested by QA and there won't be any mandatory changes) to rename this context to more reusable name. This is the only tech debt we'll need to cover.

blakeyp (approved):

PR #4222, opened 2026-08-27
BAU Clear CRI OAuth session to try and reduce duplicate CRI token requests

blakeyp:
### What changed

Clear criOAuthSessionId after callback validation and before token exchange, rather than waiting for the subsequent journey step. This shrinks the window in which a repeat CRI callback can pass OAuth state validation and trigger a second token request to the CRI. A second callback should now see the session is already cleared and will be routed to pyi-attempt-recovery where the user can get back on track.

### Why did it change

Trying this because we have seen a recent increase in repeat callbacks from CRIs, seemingly due to a possible iOS Safari bug causing duplicate requests. Note this is a best-effort mitigation - the clear is a non-atomic write and the session read is eventually consistent, so truly concurrent callbacks may still race.


blakeyp:
Build regression against my dev env: https://github.com/govuk-one-login/ipv-core-tests/actions/runs/33083998839 🟢

AmritSidhu (approved):

PR #4228, opened 2026-08-28
PYIC-9263: Add cross browser routes for Fraud CI mitigations.

undermad:
### What changed

- Added routes for cross browser / separate session Fraud CI mitigation
- Added API tests coverage

### Why did it change

- To support cross-browser and separate session Fraud CI mitigation

### Issue tracking

- [PYIC-9263](https://govukverify.atlassian.net/browse/PYIC-9263)

## Checklists

    <details>
        <ul>
        </ul>
    </details>

[PYIC-9263]: https://govukverify.atlassian.net/browse/PYIC-9263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
```diff
@@ -80,7 +80,7 @@ states:
         targetState: POST_APP_DOC_CHECK_INTERNATIONAL_SUCCESS_PAGE
         checkMitigation:
           liveness-likeness:
-            targetState: PROCESS_NEW_IDENTITY
+            targetState: MITIGATION_ADDRESS_AND_FRAUD_SEPARATE_SESSION
```
DanCorderIPV, line 83: Is this definitely correct? Isn't this the same-session mitigation where the user has already done address and fraud? (same for P1)

api-tests/features/app-cross-browser/p2-cross-browser.feature
DanCorderIPV: Do we actually need to disable Open Banking here? If so we should have a parallel test with Open Banking enabled.

DanCorderIPV: Given this is all in the app it will probably work fine with Open Banking enabled

DanCorderIPV (changes requested):

api-tests/features/app-cross-browser/p2-cross-browser.feature
undermad: I can change it to be with OpenBanking enabled, it will be easier to clear it once we remove the feature flag for open banking. I think it shouldn't impact the end of the journey whether it is with FF enabled or not as we getting the DCMAW async with Fraud and Address VCs. Correct me if I'm wrong.

DanCorderIPV: Ideally, test it with Open Banking enabled to make sure, but then delete the flag completely if it works both ways.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
undermad: This entry event coming from initial journey selection [here](https://github.com/govuk-one-login/ipv-core-back/blob/main/lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/initial-journey-selection.yaml#L38-L43). This means user went through the initialise IPV session lambda and are in the separate session.
There is not other places pointing into this state in journey map.

Same session shouldn't reach this state anyway as they should be redirected straight from the nested journey to accurate place.
As far as I checked there is only 3 states (for each p1 and p2) pointing to new P2/P1 journeys and all lives in f2f-hand-off journey.

undermad: Actually, we should double check the original target state for this event as it redirect user to international address and fraud . This mean each user that drop off the journey after application success check is redirected to this state.

api-tests/features/app-cross-browser/p2-cross-browser.feature
undermad: In P2 I enabled the feature set and it changed the route but in reuse scenario it didn't change anything so I left it without (there is one test on the very bottom of the file). P1 seems to not have open banking routes.

lambdas/process-journey-event/src/main/resources/statemachine/journey-maps/new-p2-identity.yaml
DanCorderIPV: Sorry, yes, you are right, I must have misread the state name

api-tests/features/app-cross-browser/p2-cross-browser.feature
DanCorderIPV: For other tests I've created duplicate tests for open banking enabled and disabled, but as here it only affects the set-up step to get to the actual mitigation that is being tested I guess we can get away without duplication.

DanCorderIPV (approved):

